PR #6, opened 2026-06-30
feat(skill-distillation): add config and Rust contracts

elyasmnvidian:
## Summary

This PR establishes the first skill-distillation boundary in Switchyard:

- `switchyard configure --skill-distillation NAME` saves the namespace that later launcher and runtime work will use.
- `switchyard configure --disable-skill-distillation` removes that saved namespace.
- The new `switchyard-skill-distillation` Rust crate defines the source-neutral records, validation rules, and async extension points shared by future trajectory sources, distillers, validators, and skill stores.

Existing provider and launcher configuration remains unchanged. A skill-only config update writes `~/.config/switchyard/config.json` and does not require provider credentials.

`UserConfig` owns the saved namespace. The Rust crate owns the cross-implementation API contracts. It does not choose a provider, store, model, agent runtime, or activation policy.

This PR does not yet save sessions, run a distiller, write or activate skills, import external runs, or mount skills into launched agents.

## Public Contract

The Rust crate exports:

- safe `SkillNamespace`, `SkillEvidenceId`, and `SkillVersionId` identifiers;
- versioned `Trajectory`, `DistillationRequest`, and `SkillCandidate` records;
- task, execution, source, event, outcome, provenance, validation, and activation records; and
- async `TrajectorySource`, `SkillDistiller`, `SkillValidator`, and `SkillStore` traits.

`DistillationRequest::validate_candidate()` checks that a candidate uses the request namespace, references only input evidence, and names the request's base skill as its parent. Candidate validation also requires source-evidence provenance and rejects invalid validation metrics.

The contract crate has no provider, filesystem, Python binding, benchmark, or model implementation dependency. Provider-specific event data stays inside JSON payload and metadata fields.

## Feature Proof

```text
$ SWITCHYARD_CONFIG_DIR=/private/tmp/switchyard-skill-pr-proof .venv/bin/switchyard configure --skill-distillation tooluniverse-trialqa
Saved Switchyard config to /private/tmp/switchyard-skill-pr-proof/config.json

$ SWITCHYARD_CONFIG_DIR=/private/tmp/switchyard-skill-pr-proof .venv/bin/switchyard configure --show
Skill distillation
  configured: True
  namespace: tooluniverse-trialqa
  session learning: namespace saved

skill distillation: configured; namespace: tooluniverse-trialqa; session learning: namespace saved
```

Saved config excerpt:

```json
{
  "skill_distillation": {
    "namespace": "tooluniverse-trialqa"
  }
}
```

The Rust API can be implemented by independent adapters without depending on Switchyard's Python CLI:

```rust
use switchyard_skill_distillation::{
    DistillationRequest, SkillDistiller, SkillStore,
    SkillValidator, TrajectorySource,
};
```

Invalid namespaces, path traversal in contract identifiers, non-string namespaces, unsupported manually edited `skill_distillation` keys, missing namespace values, empty CLI namespaces, conflicting configure modes, duplicate skill-evidence IDs, missing candidate provenance, and mismatched candidate/request evidence fail closed.

## Design Decisions

The namespace is saved user configuration, not a per-request header. Future requests may be recorded inside saved sessions, but one HTTP request must not reconfigure which skill is learned or used.

Skill generation should happen after sessions finish, not during normal model request handling. Later implementations can use the Rust traits while keeping provider, storage, validation policy, and agent-specific mounting behavior outside the contract crate.

Generated output should remain reviewable text such as `SKILL.md` plus support reports. Future updates should retain source-session provenance and previous active versions so people can inspect and roll back changes.

### DIFF
--- a/docs/skill_distillation.md
+++ b/docs/skill_distillation.md
@@ -1,0 +1,102 @@
+# Skill Distillation
+
+Skill distillation turns selected past agent sessions into reusable guidance,
+often a portable skill centered on `SKILL.md`. The model is not retrained.
+Switchyard's role is the trace substrate: name the workflow, collect structured
+session traces when that support lands, and keep enough provenance for a
+separate distillation tool or harness-specific adapter to consume later.
+
+The current implementation defines the product shape and saved configuration
+only. It does not implement session capture, drafting, adoption, rollback,
+external import, validation runs, or agent activation.
+
+## V1 Workflow
+
+The intended v1 flow is:
+
+```text
+run agents -> save session traces -> distill outside the proxy -> review/test -> adopt in the target harness
+```
+
+The user-facing command shape is:
+
+```bash
+switchyard configure --skill-distillation tooluniverse-trialqa
+switchyard launch codex
+switchyard configure --show
+```
+
+For now, `configure` stores only the namespace so future trace capture has a
+stable grouping key. Distillation, review, adoption, and skill activation remain
+future work and are expected to be harness-specific.
+
+## Config
+
+Skill distillation config is stored in `~/.config/switchyard/config.json` under
+the top-level `skill_distillation` key:
+
+```json
+{
+  "skill_distillation": {
+    "namespace": "tooluniverse-trialqa"
+  }
+}
+```
+
+The config is intentionally small:
+
+| Field | Default | Meaning |
+|---|---|---|
+| `namespace` | unset | Required when configuring the future trace collection namespace. |
+
+Namespaces must be safe local path components: letters, numbers, dot,
+underscore, and hyphen only. They cannot be `.` or `..`.
+
+The top-level `skill_distillation` key is omitted when the workflow is not
+configured. Use `switchyard configure --disable-skill-distillation` to remove
+it without touching provider credentials. `namespace` is the only supported key
+today; any extra manually edited keys are rejected instead of being treated as
+dormant product behavior.
+
+## Decisions
+
+Generated outputs should stay portable and reviewable. A future candidate may
+contain `SKILL.md` and human-readable support reports, but generated Python or
+other executable artifacts are out of scope for the initial workflow.
+
+Switchyard should not become a request-path memory system, and skill generation
+should not run inside normal request handling. Distillation and activation are
+closer to the target agent than to the proxy, so future implementations should
+use harness-specific adapters rather than assuming one generated artifact can
+be mounted everywhere.
+
+The namespace is saved user configuration, not a per-request header. Future
+request metadata may be captured as trace evidence, but individual requests
+should not reconfigure skill distillation behavior.
+
+Cheating prevention is explicit rather than automatic adoption. Drafts stay
+staged by default, humans adopt them, every final rule should link back to
+supporting sessions, and later review checks should flag answer leakage, source
+IDs, URLs, benchmark shortcuts, and task-specific details before a skill is used.
+
+## Planned Store Layout
+
+Future work should use inspectable local files:
+
+```text
+<project>/.switchyard/skill-distillation/<namespace>/
+  sessions/<session-id>/
+    session.json
+    turns.jsonl
+    stats.json
+  distillation-ledger.jsonl
+  candidates/<version-id>/
+    skill/SKILL.md
+    report.md
+  active/SKILL.md
+  history.jsonl
+```
+
+The ledger should track which captured sessions have already been consumed by a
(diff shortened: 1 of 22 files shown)


docs/skill_distillation.md
```diff
@@ -0,0 +1,134 @@
...
+
+```bash
+switchyard configure --disable-skill-distillation
+```
+
+## Intended Workflow
+
+Once launcher support lands, the flow should be automatic:
+
+```text
```
sabhatinas, line 39: Its not clear who owns what in this flow. Whos is saving the session and how? Is it all switchyard internals?

docs/skill_distillation.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Config
+
+Skill distillation config is stored in `~/.config/switchyard/config.json` under
+the top-level `skill_distillation` key:
+
+```json
+{
+  "skill_distillation": {
+    "namespace": "tooluniverse-trialqa"
```
sabhatinas, line 79: can you clarify what a namespace means? Is it going to be used as session-id to track trajectories?

docs/skill_distillation.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Rust Contracts
+
+The `switchyard-skill-distillation` crate defines the source-neutral records
+and extension points shared by future implementations and adapters. It does not
+choose a provider, storage format, agent runtime, or model implementation.
+
+The public contract includes:
+
+- safe `SkillNamespace`, `TrajectoryId`, and `SkillVersionId` identifiers;
```
sabhatinas: Im not sure I understand what TrajectoryId is and if its different from the session awareness we have implemented already

sabhatinas:
@elyasmnvidian Can you share how the rust traits are going to be used within switchyard ? Im trying to follow the API but its not clear to me how everything ties in.

docs/skill_distillation.md
elyasmnvidian: Thanks—yes, for the initial implementation this is entirely owned by Switchyard internals. The user’s only action is to configure a skill-distillation namespace.

For each switchyard launch, Switchyard creates the session ID and owns the capture lifecycle. It attaches internal request/response processors to the same proxy used by the launched agent, records completed model turns and routing/usage metadata, and finalizes the project-local session files, stats, and distillation ledger when the agent exits. The user does not configure a session store, provide session IDs, or run a separate recorder.

This PR establishes the configuration and contracts; the launcher implementation supplies the automatic capture path. We’ll update the documentation to make that ownership and current-versus-future boundary explicit.

Longer term, NeMo Relay could provide richer agent, tool, and LLM trajectory capture behind the same internal boundary. That should remain an implementation detail—the user-facing “enable it once” workflow would not change.

I updated this md file

docs/skill_distillation.md
elyasmnvidian: One namespace = One skill
We can have one namespace over many sessions or many trajectories.
The idea is we have one growing "skill" and if we ever want a new skill we change the namespace.

A namespace is the stable, user-chosen name for one kind of work and its learned skill—for example, tooluniverse-trialqa. It is not a session ID. Each switchyard launch creates its own internal session ID, and many sessions or trajectories can contribute to the same namespace over time. Switchyard uses the namespace to determine which skill history to update and which active skill to load for later launches. We’ll make that distinction explicit in the documentation.

elyasmnvidian: I updated the md file to make this more clear

elyasmnvidian:
> @elyasmnvidian Can you share how the rust traits are going to be used within switchyard ? Im trying to follow the API but its not clear to me how everything ties in.

You’re right—the Switchyard code that calls these traits is not in this PR. This PR defines four expected steps: TrajectorySource loads completed runs, SkillDistiller creates a candidate skill, SkillValidator checks it, and SkillStore saves it and can make it active.
Switchyard will own the code that calls these steps, but we have not settled the exact Rust/Python split, when distillation runs, or when a candidate becomes active. We will show the complete working flow in the next draft PR so the full call path is visible. After we validate it end to end, we will split it into smaller PRs.
I also updated the Rust docs to state that no production Switchyard code uses these traits yet.

docs/skill_distillation.md
elyasmnvidian: Renaming to "SkillEvidenceId"

basically this will be the thing fed into an LLM that will distill into a skill.

It can be a piece of a session, or perhaps multiple sessions.

sabhatinas (approved):

PR #8, opened 2026-06-30
feat: add Relay routing Decision API

bbednarski9:
### Summary

Adds the first, independently functional layer of the [NeMo Relay](https://github.com/NVIDIA/NeMo-Relay) integration:

- a versioned JSON routing-request and routing-decision contract;
- decision-only Random and LLM routing through the existing profile runtimes;
- `POST /v1/routing/decision` with distinct malformed, unknown-profile, and unsupported-profile errors;
- Relay-compatible route metadata, including the configured Switchyard target ID as `backend_id`.

This is the Decision API slice of a previous end-to-end implementation, split into a focused public-repository change.

## Why

Relay needs Switchyard to select a configured target without dispatching that target itself. Keeping this contract and the two existing v2 routing profiles in a focused first PR makes the wire API, profile capability boundary, and server error behavior reviewable before ATOF state or Cascade routing are introduced.

### Testing summary

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p switchyard-components-v2` (46 unit tests plus profile integration suites)
- [x] `cargo test -p switchyard-server` (21 server tests plus migration coverage)
- [x] `cargo test --workspace --exclude switchyard-py`
- [x] `cargo check -p switchyard-py`
- [x] Loopback classifier integration verifies LLM classification without calling the selected target.
- [ ] `uv run ruff check .` — not applicable; no Python source changed.
- [ ] `uv run mypy switchyard` — not applicable; no Python source changed.
- [ ] `uv run pytest tests/` — not applicable; the affected API is covered by Rust component/server tests.

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. (N/A: Rust API change.)
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (N/A: symbols are exported from the Rust crate root.)
- [x] Unit tests added for new components / bug fixes.
- [x] README / `--help` updated if customer-facing surface changed. (`DESIGN.md` documents the runtime contract.)
- [x] Commits signed off (`Signed-off-by: Your Name <email>`) per the DCO.

### Review guidelines

Suggested review order:

1. `decision.rs` for the wire contract, protocol/materialization validation, and response conversion.
2. `profile.rs` and `registry.rs` for the single-runtime capability path.
3. Random/LLM profile adapters to verify that decision-only routing stops before selected-target dispatch.
4. Server endpoint/error tests for Relay compatibility and status-code precedence.

[...]

PR #9, opened 2026-07-01
feat: add bounded Relay ATOF snapshots

bbednarski9:
### Summary

Adds the second independently functional [NeMo Relay](https://github.com/NVIDIA/NeMo-Relay) layer on top of [#8](https://github.com/NVIDIA-NeMo/Switchyard/pull/8):

- a router-neutral `RelaySnapshotAccumulator` in `switchyard-components-v2::features`;
- atomic finite-NDJSON ingestion at `POST /v1/atof/events`;
- exact `(session_id, owner_id)` snapshots with Relay-compatible identity fallbacks;
- deterministic identity, history, dedupe, request-size, and 64 MiB total retained-state bounds;
- optional bearer authentication through the Rust CLI, PyO3 binding, Python wrapper, and Python CLI;
- a checked-in Relay `http_post` / `field_name_policy = "preserve"` fixture.

This is the ATOF subscription and accumulation slice of a previous end-to-end implementation, split into a focused public-repository change.

## Why

History-aware routers need one bounded, reusable snapshot source rather than per-router event state. This PR deliberately stops at ingestion and router-neutral reconstruction: Cascade projection and decision behavior remain in PR 3.

### Testing summary

- [x] `cargo test --workspace --exclude switchyard-py`
- [x] `cargo check -p switchyard-py`
- [x] `uv run ruff check` on all changed Python/test files
- [x] `uv run mypy switchyard/cli/switchyard_cli.py switchyard_rust/core.py switchyard_rust/server.py`
- [x] `uv run maturin develop`
- [x] `uv run pytest tests/test_serve_profile_config.py tests/test_route_bundle.py tests/test_profile_migration.py -q -o addopts=` (97 passed)
- [x] One-record Relay-style POST, multiline atomic rejection, auth, media type, event size, batch size, count pruning, total-byte pruning/accounting, global dedupe, canonical-event validation, and every identity fallback are covered.

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. (N/A: Rust/Python functional API change.)
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (N/A: Rust symbols are exported from the crate root; Python constants are exported from `switchyard_rust.server`.)
- [x] README / `--help` updated if customer-facing surface changed.

### Dependency and delta

This PR depends on [#8](https://github.com/NVIDIA-NeMo/Switchyard/pull/8). Because the author has read-only access to the upstream repository, this draft must target upstream `main` and temporarily contains the #8 commit as its prerequisite.

Review the #9 delta commits `edd342d0` (snapshot memory/reporting fixes), `3857a695` (envelope-validation coverage), and `e81cd9ec` (shared Relay auth documentation), or use the [exact fork delta](https://github.com/bbednarski9/Switchyard/compare/feature/relay-decision-api...feature/relay-atof-snapshots). The earlier commits are prerequisites from #8 and the original ATOF implementation; after #8 merges, this branch will be rebased onto updated `main`.

### Review guidelines

Suggested review order:

1. `features.rs` for atomic preparation, canonical event validation, identity resolution, global dedupe, and retained-byte accounting.
2. `switchyard-server` for optional auth, bounded body parsing, and the single server-owned accumulator.
3. Rust/Python CLI and binding propagation for every limit and the bearer token.
4. `examples/relay_atof_switchyard.toml` and CLI docs for the supported Relay transport contract.

The endpoint intentionally supports finite Relay `transport = "http_post"` requests (one event per POST or a finite multiline batch). Relay's long-lived `transport = "ndjson"` upload is out of scope because atomic batch application waits for request completion.

### Previous feedback addressed

- Replaced the Cascade-specific accumulator with a reusable router-neutral snapshot contract while honoring the selected source location in `components-v2::features`.
- Bounded identity count, per-identity history, global dedupe entries, event bytes, batch bytes, and total retained-memory accounting bytes (64 MiB default).
- Parses and semantically validates the entire batch before taking the mutation lock; malformed recognized scopes cannot partially warm state.
- Restored global UUID/phase idempotency so changed identity metadata cannot bypass dedupe.
- Preserved the previous implementation's full compatibility fallback order for four session and seven owner sources, with precedence tests.
- Threads optional bearer auth through both Rust and Python startup paths; explicitly blank tokens fail configuration instead of silently disabling auth.
- Removes fallible test `.expect()` usage and documents the auth/body-validation contract.
- The Relay fixture pins `http_post`, `field_name_policy = "preserve"`, and a replaceable bearer header without assuming unsupported header environment interpolation.

Relay-aware Cascade consumption, typed freshness, and unified request session metadata are intentionally deferred to [#11](https://github.com/NVIDIA-NeMo/Switchyard/pull/11).

### Follow-up fixes

- Uses a read/write lock so concurrent snapshot readers do not serialize behind one global mutex.
- Makes the retained-state cap a conservative heap-accounting budget (serialized payload plus value/container overhead), rather than a wire-byte-only limit.
- Defines accepted_events as state-mutating ingests and exposes parsed received_events separately.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -99,10 +99,6 @@
       "filename": ".secrets.baseline"
     },
     {
-      "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
-      "min_level": 2
-    },
-    {
       "path": "detect_secrets.filters.heuristic.is_indirect_reference"
     },
     {
@@ -308,7 +304,7 @@
         "filename": "docs/cli_reference.md",
         "hashed_secret": "414cb2d928610e260b1b00de84c809d2adef2480",
         "is_verified": false,
-        "line_number": 561
+        "line_number": 605
       }
     ],
     "examples/minimal.py": [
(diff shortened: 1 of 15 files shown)


tim-mcdonnell:
Completed on main in 7df0101f, 6e2eb885, and 203ed53e. The committed-tree carry-patch inventory check passes with no findings.

PR #10, opened 2026-07-01
feat(routing): add per-request routing traces

ayushag-nv:
## Why

[SWITCH-842](https://linear.app/nvidia/issue/SWITCH-842/enable-collecting-routing-decision-data) needs per-request routing evidence that can be compared with whether a benchmark agent solved its task. Switchyard must remain benchmark-agnostic, so task identity is correlated outside the proxy through an opaque request ID.

## What

- Add a minimal routing event envelope: `request_id`, `sequence`, `timestamp_ms`, `name`, and an algorithm-owned JSON `payload`.
- Add opt-in, fail-open JSONL export through `SWITCHYARD_ROUTING_TRACE_JSONL`; raw content additionally requires `SWITCHYARD_ROUTING_TRACE_CAPTURE_CONTENT=1`.
- Emit one complete `llm_classifier.route` event containing classifier input, output, parsed signals, final tier decision, and selected model.
- Propagate a generic `x-request-id` and write a minimal per-trial `request_map.jsonl`; no task-specific header enters Switchyard.
- Keep shared raw `routing-traces/events.jsonl` at run level and write joined `artifacts/routing_trace.jsonl` beside each Harbor trial result.

## How

Routing algorithms choose their own event names, payload shape, and event granularity. The framework only assigns correlation and ordering metadata and appends JSONL without affecting routing behavior.

During a benchmark run, each trial proxy assigns UUIDs to its LLM requests while the shared Switchyard process writes interleaved events. After Harbor finishes, the benchmark joins events to trials through each trial's request map and attaches task name, trial name, and reward.

A later route-changing action, such as a context fallback, is a separate event rather than a mutation of the original classifier decision.

## Flow

1. The trial proxy intercepts an LLM request, assigns an opaque `x-request-id`, and records only that ID in the trial-local `request_map.jsonl`.
2. Switchyard copies the generic request ID into the request context. The classifier captures its input, output, parsed signals, and any failure metadata; the deterministic selector adds the tier decision, reason, and confidence.
3. The deterministic-routing backend resolves the final target and model, then appends one `llm_classifier.route` event to the shared run-level `routing-traces/events.jsonl`. The framework adds `request_id`, sequence, and timestamp; the routing algorithm owns the payload.
4. The shared event file may contain interleaved requests from concurrent trials. It remains separate from benchmark task identity, and trace-export failures never affect routing.
5. After Harbor completes, the joiner uses each trial's request map to associate its events with `result.json`, producing `artifacts/routing_trace.jsonl` containing the task name, trial name, reward, request ID, and routing events.

Raw prompt and classifier-output content is captured only when `SWITCHYARD_ROUTING_TRACE_CAPTURE_CONTENT=1`; otherwise those fields carry SHA-256 and size metadata. Without `SWITCHYARD_ROUTING_TRACE_JSONL`, tracing is a no-op.

## Limitations

- Traces are correlated to a Harbor trial and written beside that trial's result, but this PR does not materialize a reference to `agent/trajectory.json` or map an individual routing request to a trajectory event or turn. Consumers that need per-turn attribution must perform that correlation themselves.
- The shared run-level event file is intentionally interleaved across concurrent trials; `request_map.jsonl` is the supported key for assigning events back to a trial.
- The trace envelope is generic, but this PR currently emits rich routing-decision traces only for the LLM-classifier path. Other routing algorithms and decision stages do not yet produce equivalent trace events.

## Where to Start Review

1. `crates/switchyard-core/src/routing_trace.rs`
2. `switchyard/lib/routing_trace.py`
3. `switchyard/lib/processors/llm_classifier/request_processor.py`
4. `switchyard/lib/backends/deterministic_routing_llm_backend.py`
5. `benchmark/closed_book_proxy/proxy/rewriter.py`
6. `benchmark/routing_trace_joiner.py`
7. `benchmark/run-baseline.sh`

## Test Plan

- `cargo fmt --all --check`
- `cargo test --workspace`
- `uv run ruff check .`
- `uv run mypy switchyard`
- Hermetic Python suite: 1,875 passed, 12 skipped, 43 integration tests deselected
- Live TB-lite Harbor run: 2 tasks at concurrency 2 using GPT-5.5, Nemotron 3 Ultra, and Gemini 3.5 Flash
  - Harbor exit 0
  - 12 routed requests across 2 trials
  - 12/12 proxy request IDs matched routing traces
  - 0 missing traces, 0 orphan traces, 0 sequence errors


ryan-lempka:
@ayushag-nv let's add cost_usd and capture routing cost similar to existing per-request telemetry.

ayushag-nv:
> @ayushag-nv let's add cost_usd and capture routing cost similar to existing per-request telemetry.

will add as a follow on. Lets get the bare bones merged first

ayushag-nv:
not required

PR #12, opened 2026-07-01
ci(release): add Rust crate publishing path

nachiketb-nvidia:
## What

Adds the release-infrastructure needed to publish the first public Rust crates for Switchyard.

- Makes public workspace crates publish-ready by adding shared Cargo package metadata and versioned internal path dependencies.
- Keeps `switchyard-py` private with `publish = false` because it is distributed through the `nemo-switchyard` Python wheel.
- Adds `scripts/release/publish_crates.py` as the single source of truth for Rust crate publish order.
- Adds GitHub Actions support for:
  - manual Rust crate package dry-runs that upload `.crate` artifacts;
  - manual first crate publish from `main` behind the protected `crates-io` environment;
  - tag-gated crates.io publishing before the Python/PyPI publish job.
- Updates release docs and the `publish-package` skill so future agents follow the same path.
- Adds a focused test that guards publish order and manifest metadata.

## Why

SWITCH-456 requires the first Rust crate versions to be published in dependency order and verified on crates.io/docs.rs. Cargo publishing is stricter than Python wheel publishing: versions are immutable, dependent crates must resolve already-published versions, and path-only internal dependencies are not publishable. This PR prepares the safe workflow before any actual upload happens.

## How

The public crate order is:

1. `switchyard-core`
2. `switchyard-translation`
3. `switchyard-components`
4. `switchyard-components-v2-macros`
5. `switchyard-components-v2`
6. `switchyard-server`

The new helper runs `cargo publish --dry-run` by default. During dry-runs only, it patches unpublished workspace dependencies back to local paths so dependent crates can be validated before the first dependency versions exist on crates.io. Real publishes do not use those patches and still require `CARGO_REGISTRY_TOKEN` from the protected `crates-io` GitHub environment.

## What To Review

- Whether the selected public crate list is correct and `switchyard-py` should remain wheel-only.
- Whether the crate publish order matches dependency order.
- Whether tag-gated publishing should include crates.io before PyPI in the official release path.
- Whether the manual `publish_rust_crates` path being restricted to `main` is enough guardrail for the first publish.
- Whether docs clearly separate dry-run artifact generation from real crates.io publishing.

## Validation

- `python scripts/release/publish_crates.py --allow-dirty`
- `uv run pytest tests/test_rust_crate_publishing.py -v`
- `uv run ruff check scripts/release/publish_crates.py tests/test_rust_crate_publishing.py`
- `cargo metadata --format-version 1 --no-deps`
- `git diff --check`
- `cargo search` for the planned crate names returned no existing crates.

Pre-commit also ran and passed strict secret detection, SPDX, ruff, cargo fmt, cargo clippy, mypy, and commitlint. Pyright remains non-blocking and reported pre-existing unrelated issues.

Linear: https://linear.app/nvidia/issue/SWITCH-456/publish-first-rust-crate-versions


nv-msvoboda:
@sir-merge-a-lot groom

PR #14, opened 2026-07-01
chore: support deterministic routing tiers

erichards-nvidia:
Adds configurable deterministic routing tiers beyond weak/strong while preserving the existing weak/strong configuration path. This lets downstream demos map classifier recommendations onto multiple backend tiers such as local GPU, Spark, NIM, and Frontier.


sabhatinas:
Can you make this change apply across routing algorithms implemented in python lib as well as components-v2 rust crates?

sabhatinas:
I believe routing profiles are getting deprecated and we will only have --config. You can isolate your changes to --config

elyasmnvidian:
Closing, feel free to open back

PR #19, opened 2026-07-03
docs: clarify type-name conventions between route bundles and v2 profile configs

himorishige:
### What

Adds a short note to the getting-started guide clarifying that legacy route bundles (`--routing-profiles routes.yaml`) use snake_case route types (`random_routing`) while v2 profile configs (`switchyard serve --config profiles.yaml`) use hyphenated profile types (`random-routing`, `llm-routing`).

### Why

Following the quickstart, we copied `type: random_routing` into a v2 profile config and hit `unknown profile type 'random_routing'`. The two formats are easy to conflate for new users, since the quickstart teaches the bundle format while `examples/profiles.yaml` and the serve docs use the v2 format. A one-paragraph note at the point of first contact avoids the confusion.

The note follows the existing blockquote style used by the adjacent "Format default and caching" note in the same file.

### Notes

- Docs-only change; no code affected.


ayushag-nv:
@himorishige Thanks for the contribution. This deserves a fix in the code to unify rather than docs. So I will prefer that

himorishige:
Closing in favor of #22 — per your suggestion, this unifies the type-name spellings in code (the v2 `--config` path now treats `-` and `_` as equivalent, so `type: random_routing` resolves to `random-routing` and `stage-router` to `stage_router`) instead of documenting the difference. Thanks for the steer!

PR #20, opened 2026-07-06
refactor: rename cascade router to stage-router

sabhatinas:
## Summary

Two related renames on this branch, both applied across **Python, Rust, docs, and tests**:

1. **`cascade` → `stage-router`** — the router itself.
2. **`strong`/`weak` tiers → `capable`/`efficient`** — capability-specifying tier vocabulary for the stage-router.

---

## 1. cascade → stage-router

Case-aware mapping:
- `cascade` → `stage_router`, `Cascade` → `StageRouter`, `CASCADE` → `STAGE_ROUTER`
- Profile type value `type: cascade` → `type: stage_router`
- Example profile keys `smart-cascade`/`my-cascade` → `smart-stage-router`/`my-stage-router`
- Prose, headings, nav titles → "stage-router" / "Stage-Router Routing"

Deliberately **not** renamed (the English word, unrelated to the router): `defaults cascade` / `credentials cascade` (`route_bundle.py`), "a cascade of failures" (`test_verify.py`), and "cascading 504s" throughout.

22 files renamed via `git mv` (e.g. `processors/cascade/` → `processors/stage_router/`, `profiles/cascade{,_config}.py`, `crates/.../profiles/cascade.rs`, `docs/.../cascade_routing.md` → `stage_router_routing.md`, all `test_cascade_*` → `test_stage_router_*`). `mkdocs.yml`, `pyproject.toml` package-data, and `.secrets.baseline` updated to match.

## 2. strong/weak tiers → capable/efficient

Scoped to the **stage-router feature only**, applied consistently in **both** the Rust `switchyard-components-v2` profile and the Python `switchyard.lib` implementation (kept in sync):

- Config fields `strong:`/`weak:` → `capable:`/`efficient:` (`StageRouterConfig` / Rust `StageRouterProfileConfig`, serde field names).
- Picker modes `stage_router_strong_default`/`stage_router_weak_default` → `capable_first`/`efficient_first` (Python `pick_*` fns + `StageRouterPickerMode` enum variants).
- Tier enum/labels: `StageRouterTier::{Strong,Weak}` → `{Capable,Efficient}` (stats now bucket stage-router tiers as `capable`/`efficient`).
- Classifier verdict + prompts: `{"tier": "strong"|"weak"}` → `{"tier": "capable"|"efficient"}`.

**Left as strong/weak (out of scope / shared vocabulary):**
- `random_routing` (incl. `strong_probability`), `plan_execute`, `deterministic`, `routellm` — they own their own strong/weak config.
[...]

PR #21, opened 2026-07-07
fix: record streaming token usage in v2 profile stats

### LINKED ISSUE #18: `/v1/routing/stats` records zero token usage for streaming responses (v2 profile server)
### Summary

When a client sends `"stream": true` to a `switchyard serve --config <profiles.yaml>` proxy, the request is counted in `GET /v1/routing/stats` but its token usage is recorded as zero. Non-streaming requests on the same profile record usage correctly. This makes `cost_estimate`, `token_pct`, and per-tier token totals unusable for streaming clients — which includes most agent frameworks.

### Reproduction (v0.1.0)

`profiles.yaml` with a single passthrough profile `fast` pointing at an OpenAI-compatible endpoint (OpenRouter in our case), then three requests:

```bash
# 1) non-streaming — usage recorded (+49 tokens)
curl -s localhost:4000/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"fast","max_tokens":32,"messages":[{"role":"user","content":"Say OK"}]}'

# 2) streaming — counted as a call, usage +0
curl -s localhost:4000/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"fast","max_tokens":32,"stream":true,"messages":[{"role":"user","content":"Say OK"}]}'

# 3) streaming with an explicit usage frame — usage still +0
curl -s localhost:4000/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"fast","max_tokens":32,"stream":true,"stream_options":{"include_usage":true},"messages":[{"role":"user","content":"Say OK"}]}'
```

After the three requests, `GET /v1/routing/stats` shows `calls: 3, total_tokens: 49` — only the buffered response contributed usage. In case 3 the upstream demonstrably sent a `usage` frame in the SSE stream (visible in the raw response), and it is still not recorded.

### Real-world impact

We routed a personal agent's daily traffic (streaming Chat Completions) through an `llm-routing` profile for half a day. The provider's own analytics recorded ~5.17M tokens for the weak tier, while `/v1/routing/stats` reported 0 tokens across 108 calls. The stats then report `token_pct: 100%` for the strong tier, which only had 2 buffered test calls — the exact opposite of the real traffic distribution.

### Likely root cause

The v2 profiles record usage from the buffered body only, e.g. in the `record_success` path:

```rust
let usage = response.body().map(usage_from_body).unwrap_or_default();
```

- `crates/switchyard-components-v2/src/profiles/llm_routing.rs`
- same pattern in `passthrough.rs`, `random_routing.rs`, and `cascade.rs`

For streaming responses `response.body()` is `None`, so usage silently defaults to zero. The stream-tap machinery already exists on the v1 components side (`switchyard-components/src/stats/usage.rs` — `openai_chat_usage_from_stream_event`, and `response_processors/stats.rs`, which wraps streams so usage is captured on completion), but it does not appear to be wired into the v2 profile chain.

### Environment

- nemo-switchyard 0.1.0 (PyPI wheel), macOS arm64 / Python 3.13, and reproduced on Linux aarch64 (DGX Spark)
- Upstream: OpenRouter (OpenAI-compatible)

Happy to contribute a fix if wiring the existing stream-usage tap into the v2 chain is the direction you'd take — or to test a patch against our streaming workload.

himorishige:
### What

Adds a native v2 stream tap (`switchyard-components-v2/src/stats/stream_tap.rs`) and wires it into all five v2 profiles (passthrough, random-routing, llm-routing, cascade, latency-service), so `/v1/routing/stats` records token usage for streaming responses. Fixes the `108 calls / 0 tokens` shape reported in #18.

### Root cause

All v2 profiles read usage from the buffered body only:

```rust
let usage = response.body().map(usage_from_body).unwrap_or_default();
```

`ChatResponse::body()` returns `None` for the three stream variants, so streaming usage silently defaulted to zero while `record_success` still counted the call.

### Approach

Per the guidance in #18, this does **not** wire in the v1 `StatsResponseProcessor` (it depends on `ProxyContext` state that v2 doesn't have). Instead:

- `UsageAttribution` captures what v2 already holds locally before the stream is consumed: the stats accumulator, model, tier, request-start `Instant`, and backend latency.
- `record_usage_or_tap_stream(response, attribution)` records buffered bodies immediately and wraps the three stream variants so usage is committed once when the stream completes.
- Usage parsing reuses the existing `switchyard-components::stats` parsers (`openai_chat_usage_from_stream_event`, `openai_responses_usage_from_stream_event`, `AnthropicStreamUsage`).
- Usage is recorded via `record_usage_after_success_attribution`, so the tier call counted by `record_success` is not double-counted.

Profile wiring stays local: passthrough / random-routing / latency-service replace the inline usage read; llm-routing / cascade change `record_success` to take the response by value and return the (possibly wrapped) response. The cascade `enable_stats=false` guard is preserved (no tap when stats are disabled).

### Semantics notes (for review)

- **Last usage frame wins, committed at stream completion.** Spec-compliant upstreams send usage only on the final chunk; cumulative emitters (e.g. vLLM `stream_options.continuous_usage_stats`) update every chunk, so committing the first frame would under-count.
- **Usage-less streams still record latency.** Streams without a usage frame commit zero usage at completion, so total-latency samples aren't lost (matches the previous v2 behavior of recording zero usage).
- **Total latency spans until the stream completes**, matching the v1 stream taps. This includes client read time for streams — happy to change if you'd prefer header-time latency.
- **Client-disconnected streams commit nothing** — same limitation as the v1 taps.
- **Classifier-path usage is untouched** (llm-routing / cascade classifiers are always non-streaming, so `record_classifier_usage` still reads the buffered body).

### Testing

- Unit tests for the tap: buffered, OpenAI Chat / Responses / Anthropic streams, cumulative usage frames, usage-less streams (latency-sample preservation).
- Per-profile streaming tests for all five profiles, asserting usage is zero before the stream is consumed and recorded after.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` all pass (toolchain 1.94.0).
- Live re-run of the #18 reproduction (passthrough profile → OpenRouter, Linux aarch64):

| Request | before (v0.1.0) | after this fix |
| --- | --- | --- |
| non-streaming | +49 | +43 |
| `"stream": true` | **+0** | **+50** |
| `"stream": true` + `stream_options.include_usage` | **+0** | **+50** |

(deltas are `total_tokens` in `/v1/routing/stats` per request; absolute counts vary with model output length)

Closes #18

### DIFF
--- a/crates/switchyard-components-v2/Cargo.toml
+++ b/crates/switchyard-components-v2/Cargo.toml
@@ -12,7 +12,9 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
+async-stream = "0.3"
 async-trait = "0.1"
+futures-util = "0.3"
 parking_lot = "0.12"
 rand = "0.8"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots"] }
--- a/crates/switchyard-components-v2/src/profiles/latency_service/mod.rs
+++ b/crates/switchyard-components-v2/src/profiles/latency_service/mod.rs
@@ -16,7 +16,6 @@ use std::time::Instant;
 
 use async_trait::async_trait;
 use parking_lot::RwLock;
-use switchyard_components::stats::usage_from_body;
 use switchyard_components::StatsAccumulator;
 use switchyard_core::{ChatResponse, LlmTarget, LlmTargetId, Result, SwitchyardError};
 
@@ -26,6 +25,7 @@ use self::selection::select_target;
 pub use self::selection::SelectedTarget;
 use crate::backend::{native_target_backend, TargetBackend};
 use crate::profile_stats_accumulator;
+use crate::stats::{record_usage_or_tap_stream, UsageAttribution};
 use crate::{
     profile_config, Profile, ProfileConfig, ProfileHooks, ProfileInput, ProfileResponse,
     RoutingMetadata,
@@ -329,16 +329,16 @@ impl Profile for LatencyServiceProfile {
                         Some(backend_latency_ms),
                         None,
                     )?;
-                    let total_latency_ms = profile_started_at.elapsed().as_secs_f64() * 1000.0;
-                    let routing_overhead_ms = (total_latency_ms - backend_latency_ms).max(0.0);
-                    let usage = response.body().map(usage_from_body).unwrap_or_default();
-                    self.stats.record_usage_after_success_attribution(
-                        stats_model,
-                        usage,
-                        Some(total_latency_ms),
-                        Some(routing_overhead_ms),
-                        None,
-                    )?;
+                    let response = record_usage_or_tap_stream(
+                        response,
+                        UsageAttribution::new(
+                            self.stats.clone(),
+                            stats_model,
+                            None,
+                            profile_started_at,
+                            backend_latency_ms,
+                        ),
+                    );
                     let processed = LatencyServiceProcessedRequest {
                         profile_input: input,
                         selected,
--- a/crates/switchyard-components-v2/src/profiles/latency_service/tests.rs
+++ b/crates/switchyard-components-v2/src/profiles/latency_service/tests.rs
@@ -15,6 +15,7 @@ use std::time::Duration;
 
 use async_trait::async_trait;
 use futures_core::Stream;
+use futures_util::StreamExt;
 use parking_lot::RwLock;
 use serde_json::{json, Value};
 use switchyard_core::{BackendFormat, ChatRequest, LlmTargetId, ModelId, StreamEvent};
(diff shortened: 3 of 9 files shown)


ayushag-nv:
@himorishige Thanks for putting up the PR. Can you please rebase and resolve the merge conflicts once

himorishige:
Rebased onto `main` to resolve conflicts. The stats wiring and the new streaming-usage test moved from `cascade.rs` to `stage_router.rs` (following the #20 rename and the strong/weak → capable/efficient tier renames); no behavior change from the previous revision. `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are green on the rebased branch.

grahamking:
Replaced by https://github.com/NVIDIA-NeMo/Switchyard/pull/64

PR #28, opened 2026-07-07
feat(latency-service): credential policy, request-type routing, metric attribution, response fidelity, and shared session-affinity store

eric-liu-nvidia:
## Summary

Brings the latency-aware router up to date with features developed prior to this repository's migration, one commit per feature for reviewability. All of this code shipped previously with green CI and review; this PR re-lands it here.

| Commit | What it does |
|---|---|
| caller_required credential policy | Opt-in mode where upstream calls always use the caller's API key and never fall back to the service key (401 when absent) — for multi-tenant deployments with per-user billing. Adds the `switchyard_latency_upstream_attempts_total{requested_model,upstream_model,outcome,code}` metric. |
| request_type on route-bundle endpoints | The routing YAML now accepts `request_type: openai_responses` per latency endpoint, so Responses-only upstream models are configurable via `--routing-profiles` (previously the loader rejected the key at startup). |
| route-model metric attribution | `requested_model` is bounded to config-derived ids (route id + endpoint ids), so route-key traffic is attributed by name instead of collapsing to the `other` sentinel. |
| error-source annotation + Responses fidelity | Failed responses carry `x-switchyard-error-source` (`switchyard` \| `provider`) and `x-switchyard-upstream-model` headers (plus span tags and structured-log fields). Responses API calls through `request_type: openai_responses` endpoints return the upstream payload verbatim — exact JSON body non-streaming, verbatim SSE frames streaming — while usage accounting still works from raw frames. |
| shared session-affinity pin store (Redis L2) | Optional `affinity_store: redis` shares conversation→endpoint pins across workers/pods and persists them across pod churn. Fail-open with a 0.1s socket timeout and a 3-failure/10s-cooldown circuit breaker (`switchyard_affinity_l2_breaker_open` gauge). New optional extra: `nemo-switchyard[affinity-redis]`. |

Support commits: an operations-docs import (`docs/internal/latency_service_routing.md` including the deployment contract for the Redis store, `docs/internal/metrics_reference.md`, and `examples/prometheus/` scrape config + alert rules — droppable if maintainers prefer to keep latency-router deployment docs out of this repo), a comment/docs cleanup, a secrets-baseline chore, and a strict-mypy compatibility tweak for the optional redis import.

## Adaptations to this repository

- `all` extra uses the `nemo-switchyard` package name and now includes `affinity-redis`.
- `uv.lock` regenerated (adds `redis==5.3.1` as an optional extra only — the default install never pulls it in).
- Backend `shutdown()` becomes async; this repo's `_run_lifecycle_method` awaits awaitable shutdowns, so teardown semantics are unchanged.

## Validation (this repo's toolchain, Rust 1.96.1)

- Full hermetic suite: **1974 passed, 12 skipped** — includes all ported tests: credential-policy wire tests, request_type regression, route-model metric labels, error-source headers/spans/logs, byte-exact streaming for OpenAI- and Azure-shaped samples, and the Redis pin-store/circuit-breaker suite
- `ruff check .` clean; `mypy switchyard` (strict) clean
- `cargo fmt` / `clippy --workspace -D warnings` clean; `cargo test --workspace`: all 43 suites ok
- detect-secrets scan clean

## Summary by CodeRabbit

* **New Features**
  * Added latency-aware routing with optional shared session affinity, including Redis-backed pin storage and a new caller-required authentication mode.
  * Responses streaming now preserves raw SSE frames while still capturing usage metrics.
  * Added richer operational docs, Prometheus example configs, and alerting rules.

* **Bug Fixes**
  * Improved error responses with clearer source and upstream-model details.
  * Enhanced request metadata handling to prefer and redact caller credentials safely.
  * Fixed streaming and passthrough behavior so upstream payloads are returned more faithfully.


linj-glitch (approved):

ayushag-nv (changes requested):
Please hold on from merging. @eric-liu-nvidia @ryan-lempka are we aligned to bring back these docs in public repo ??

ayushag-nv:
@eric-liu-nvidia Can u address the code rabbit comments ?

eric-liu-nvidia:
> Please hold on from merging. [@eric-liu-nvidia](https://github.com/eric-liu-nvidia) [@ryan-lempka](https://github.com/ryan-lempka) are we aligned to bring back these docs in public repo ??

@ayushag-nv all the docs doesn't have internal information and were meant for customer facing at first place.

latency_service_routing.md: how the latency service get designed.
metrics_reference.md: list of metrics emitted for user using the latency aware router

what are the exact concerns of bringing back these docs?

eric-liu-nvidia:
Review findings addressed in 8a7dce6 — resolution per finding:

**`docs/internal/metrics_reference.md` (terminology)** — Fixed. The routing-overhead formula now reads `total_latency_ms − backend_call_latency_ms`, matching the instrumentation field and the rest of the docs.

**`examples/prometheus/prometheus.yml` (docs path)** — Fixed. Comment now points at `docs/internal/metrics_reference.md`; the alert-rules reference is unchanged.

**`examples/prometheus/README.md` (broken links)** — Fixed. All three occurrences (the markdown link and both prose mentions) now resolve to `docs/internal/metrics_reference.md`. Also fixed the same stale path in `switchyard.rules.yaml`'s header comment, which the sweep surfaced.

**`switchyard.rules.yaml` (LatencyServicePollStale timing)** — Fixed by removing the `for: 60s` clause rather than lowering the threshold: `poll_age_seconds` is a monotonically growing gauge that only climbs while polls keep failing, so the `> 60` threshold is already self-debounced (healthy age oscillates 0–10 s) and the `for:` added pure delay past the documented 60 s staleness budget. The alert now fires when the description says it does. A comment in the rule records the reasoning.

**`latency_service_llm_backend.py` (caller_required ordering)** — Partially applied. The credential-policy check now runs at the top of `call()`, **before** the affinity lookup, so a doomed request never pays a pin resolution (which can be a shared-store round-trip). However, the rejection intentionally remains **after** `CTX_UPSTREAM_ATTEMPTS_RECORDED` is claimed: `switchyard_upstream_attempts_total` counts *individual upstream call attempts*, and a caller_required 401 makes none — counting a synthetic 401 there would inflate the direct-to-endpoint baseline error rate that the router-vs-baseline ratio depends on. The 401 stays observable via `switchyard_client_responses_total` and the stashed 401 envelope. A code comment now documents this.

**`upstream_error.py` (context_exhausted_response envelope)** — Fixed. Delegates to `error_response()`; identical status, payload shape, and `switchyard` source header, with the hand-built envelope (and the now-orphaned `ERROR_SOURCE_HEADER` import) removed.

**Tests (attempt-tracking assertions)** — Added. Both caller_required rejection tests now assert `CTX_UPSTREAM_ATTEMPTS_RECORDED is True` (and the missing-key test additionally asserts the per-model attempts metric stays absent), pinning that rejections stay invisible to upstream-attempt accounting. A new `test_caller_required_rejection_skips_affinity_lookup` pins the ordering fix.

Validation: full hermetic suite 1975 passed / 12 skipped, ruff clean, strict mypy clean.

ayushag-nv (approved):

PR #40, opened 2026-07-09
docs: add external contribution quick guide

ryan-lempka:
## Summary

Adds an External Contributions section at the top of CONTRIBUTING.md so first-time contributors see the end-to-end flow before the internal development details. Modeled on the ai-dynamo/dynamo contribution guide, trimmed to the essentials:

- Fork-and-clone workflow (the existing setup instructions assume push access to this repo, which external contributors do not have)
- Size-based process: small fixes go straight to a PR, larger changes (~100+ lines or new features) start with an issue
- DCO sign-off step, pointing at the existing Signing Your Work section
- Note that reviewers are requested automatically via CODEOWNERS
- Expectation to understand and be able to explain AI-assisted code

Also includes a few small related fixes:

- Internal workflow now notes the DCO sign-off requirement (the DCO check is required on every PR as of today), and the stale note about which status checks should be required now reflects reality (CI Success + DCO)
- Issue-template contact link for questions pointed at the wrong org (NVIDIA/switchyard, a 404); now points at this repo's Discussions, which is enabled
- detect-secrets baseline refreshed (line numbers shifted by the CONTRIBUTING.md insertion)

PR #42, opened 2026-07-09
fix: point question contact link at this repo's Discussions

ryan-lempka:
The issue-template contact link for questions pointed at https://github.com/NVIDIA/switchyard/discussions, which is the wrong org (404). Discussions is now enabled on this repo; this points the link at the right place.

With blank issues disabled, this link is the only path the new-issue chooser offers for questions, so it was a dead end for anyone trying to ask one.

## Summary by CodeRabbit

* **Chores**
  * Updated the issue template’s external discussion link to point to the current community support location.


ryan-lempka:
Folded into #40.

PR #43, opened 2026-07-09
chore: remove detect-secrets in favor of GitHub secret scanning

ryan-lempka:
## Summary

Removes the detect-secrets pre-commit hook, its `.secrets.baseline`, the custom NVIDIA-key plugin, and the Secret scan CI job. Secret protection moves to GitHub's platform-level scanning: secret scanning alerts and push protection are both enabled on this repo (Settings > Advanced Security).

## Why

**Baseline maintenance scales with repo churn.** `.secrets.baseline` records the line number of each allowlisted example secret (~130 entries across `tests/`, docs, workflows, and Rust source). Any commit that shifts lines in one of those files requires a companion baseline-refresh commit to keep the Secret scan job green; #40 needed two of them. As the repo grows, more files carry allowlisted examples and more edits require a refresh.

**The entries being tracked are placeholder keys.** The baseline's role today is allowlisting the fake credentials used in tests and docs, so the ongoing maintenance mostly serves examples rather than live-secret detection.

**Platform scanning intervenes earlier for real tokens.** GitHub secret scanning detects real token formats for its supported providers, and push protection rejects a push containing one before it reaches the repository. A CI job necessarily runs after the commit is already pushed, when rotation is required regardless of the check's result. Detections and bypass events are auditable under the repo's Security tab.

**Coverage gaps, stated plainly:**

- detect-secrets' keyword and entropy heuristics can flag generic secrets (e.g. a hard-coded password matching no known token format), which platform scanning does not cover by default. GitHub's non-provider pattern detection is a follow-up option.
- NVIDIA API keys (`nvapi-`) do not appear on GitHub's [supported patterns list](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns), so removing `.hooks/detect_nvidia_api_key.py` removes the only NVIDIA-specific detection. If we want that back, a baseline-free pattern check (a strict `nvapi-` regex that placeholder keys don't match) would restore it without the churn above.

## Changes

- `.pre-commit-config.yaml`: drop the detect-secrets hook
- `.github/workflows/ci.yml`: drop the Secret scan job and its `ci-success` gate entry
- Delete `.secrets.baseline` and `.hooks/detect_nvidia_api_key.py` (the plugin existed only to feed detect-secrets)


ayushag-nv (approved):
Looks Good. Thanks for this. It has been causing troubles.

PR #49, opened 2026-07-10
feat(advisor): executor+advisor strategies on any Chat wire (tool_call default, review_gate)

eric-liu-nvidia:
## What

Ports the **advisor strategy** from the internal dev repo (`advisor-strategy` branch, including its final in-flight state) into this repo's profiles architecture, and generalizes the shipping `tool_call` strategy into a **wire-format-agnostic performance gear** — it works with any Chat-shaped executor/advisor pair, not just Anthropic:

- **`type: advisor` route / `AdvisorProfileConfig`** — pairs a faster **executor** with a stronger **advisor** model. Each tier's `format` selects its wire independently and tiers mix freely:
  - `anthropic` — served native Anthropic Messages (`/v1/messages`); the executor call is delegated **verbatim** to `AnthropicNativeBackend`, so the client's `cache_control` breakpoints survive and prompt caching is honored.
  - `openai` — served OpenAI Chat Completions (`/chat/completions`) via `OpenAiNativeBackend`, covering OSS models (Qwen, DeepSeek on vLLM/NIM) and OpenAI. The loop runs natively on OpenAI wire: function-shaped advisor tool, `tool_calls` interception, `role:"tool"` feedback (1:1 with tool_call ids), delta reassembly + final-usage chunk, verbatim stream replay.
  - `responses` targets are rejected at config validation (the loop is Chat-shaped); `format: auto` resolves at build time.
- `AdvisorConfig.strategy` selects the design:
  - **`tool_call` (shipping default, any wire)** — `AdvisorToolCallBackend`: a proxy-side re-creation of Anthropic's `advisor_20260301` server tool for gateways/models that cannot run it server-side. A real, parameterless `advisor` tool is injected; each advisor call is intercepted before it reaches the client, the advisor model is consulted on the transcript, and the advice loops back as the tool result (up to `max_uses`, hard cap 8 executor turns). Wire specifics live in two small private dialect objects (`_AnthropicDialect` / `_OpenAiDialect`); the loop itself is format-independent.
  - **`review_gate` (any wire, like tool_call)** — `AdvisorLoopBackend`: no tool injected; at the executor's first no-tool-call turn (a plan, or a claim of "done") the advisor reviews **once per session** → APPROVE (return as-is) or REDO (feed the optimized plan back and re-invoke). Preserved for A/B comparison.
- The **advisor caller** likewise dispatches on `advisor.format`: Anthropic Messages (httpx, Bearer) or OpenAI Chat Completions (`OpenAILLMClient`, `max_retries=0` so failures fall through to the loop's own `fail_open`); target `extra_headers`/`extra_body` are forwarded so gateway auth and vLLM chat-template hints work.
- **Rust: native-Anthropic verbatim passthrough** (`AnthropicNativeBackend::outbound_body`) — a native Anthropic request is now forwarded byte-faithful (only the model id is rewritten + operator `extra_body` merged). The strip/normalize/sanitize steps now apply only to the translated (OpenAI/Responses→Anthropic) path. This preserves `context_management`, signed thinking blocks, mid-conversation system turns, client tool ids — and, critically, `cache_control` breakpoints.
- Claude Opus 4.8 pricing in `cost_estimator.py` (3 id variants), public exports (`AdvisorConfig`, `AdvisorPresets`, `AdvisorProfileConfig`), an advisor row in `switchyard-lib-core`'s Quick Reference, and AGENTS.md structure sync.

Port mapping (dev repo → here): `lib/factories/advisor/{config,factory,advisor_presets,prompts}.py` + `SwitchyardRecipes.advisor_recipe` → `lib/profiles/advisor{_config,,_presets,_prompts}.py`, per this repo's profile-owned-construction contract (no factories/recipes/registries). The internal benchmark runbook skill intentionally stays in the internal repo.

## Why

This migrates the complete advisor line of work — code and the design rationale behind it — so it ships from this repo, and makes it usable across every endpoint family Switchyard fronts. Condensed design history:

**v1 — executor-triggered tool loop**: a mostly spec-faithful re-creation of Anthropic's advisor tool. Trace analysis showed front-loaded advice **suppressed the executor's own test-and-iterate loop** — it trusted the plan, one-shot it, and declared done prematurely, losing tasks the unadvised baseline solved by iterating. Retired in favor of:

**v2 — once-per-session review gate**: the executor works with its own tools; its first no-tool-call turn is reviewed once → APPROVE/REDO. A near-superset of solo behavior (identical until "done", plus one quality gate), so it is downside-protected while catching premature convergence.

**v3 — native Anthropic everywhere**: the cost lever. Serving the executor through OpenAI-Chat translation strips `cache_control` (the OpenAI schema has no such field), so upstream prompt caching never engages and all input bills at the full rate — on agentic workloads (~95% input tokens) that was roughly a **7× cost difference** vs the cached path. Both tiers therefore run native `/v1/messages`, and the executor call is delegated verbatim to `AnthropicNativeBackend` (this PR's Rust change) so the client's breakpoints reach the upstream. For gateways that authenticate `/v1/messages` with `Authorization: Bearer` rather than Anthropic's `x-api-key`, the preset sets `api_key: ""` (suppresses x-api-key) and carries the key in `extra_headers`.

**Benchmark outcome (internal coding-agent benchmark, review-gate arm vs executor-solo, both native-Anthropic on a healthy cluster):** the gate **edged the solo baseline directionally but not statistically significantly** — it rescued more tasks than it regressed (premature-convergence losses recovered via REDO), at a small cost and latency premium. Earlier runs suggesting the advisor lost were traced to an infrastructure incident and discarded.

**v4 — advisor-as-tool-call restored as the shipping default**: ship the strategy that matches the public product surface (Anthropic's `advisor_20260301`; OpenRouter's analogous `openrouter:advisor`), keeping the gate selectable for A/B. The rebuilt `AdvisorToolCallBackend` fixes the v1 deviations found by auditing against Anthropic's advisor-tool doc:

1. The advisor now sees **the executor's current-turn text** (the doc forwards "the text the executor has produced so far in this turn"; v1 consulted before appending it), plus a compact summary of the executor's tools.
2. Transcript truncation is **tail-keep** — the oldest turns drop first (v1 kept the head, discarding the consult-relevant recent turns).
3. `max_uses` follows the spec's semantics: over-cap calls receive a `max_uses exceeded` **tool result** and the executor continues (v1 withdrew the tool, which the Anthropic wire format rejects once history contains `tool_use` blocks).
4. Steering is injected **cache-stably**: system-prompt prepend + the doc's length line on the **first** user message. The doc places the length line on the *latest* user message, but the client never sees this injection — re-injecting per turn would shift the upstream cache prefix on every turn. The same placement holds on the OpenAI wire, where vLLM/OpenAI automatic prefix caching also wants a stable prefix.

One v1 behavior kept deliberately: a turn mixing advisor + client tool calls is regenerated with the sibling client calls dropped (they are re-issued advice-informed on the next iteration). The native API pauses with the client calls pending; a proxy cannot — it can neither execute the client's tools nor return a turn containing a tool the client was never offered.

**v5 — generic wire formats (this PR's final shape)**: Switchyard users front OSS models (Qwen, DeepSeek) on OpenAI-compatible endpoints alongside commercial Claude/OpenAI, and want the advisor as a generic model-performance gear rather than an Anthropic-only feature. This is also where the feature most plausibly pays off: Anthropic's published guidance puts tool-call lifts on **weaker executors** (strong executors tend to over-call), which is exactly the OSS-executor + strong-advisor pairing. The loop was refactored behind a private wire-dialect seam — the Anthropic dialect delegates to the pre-existing helpers (the original Anthropic test suite passes **unmodified**, which is the byte-for-byte regression proof), and the OpenAI dialect adds Chat-wire handling with OSS-endpoint hardening: advisor detection by tool-call presence (never `finish_reason`, which some servers mislabel), synthesized ids for servers that omit them in stream deltas, `"{}"` for empty arguments, vendor fields (`reasoning_content`) whitelisted out of rebuilt turns so strict endpoints accept the replayed history, and missing-usage tolerance.

Cost accounting: advisor consults **and** the intercepted executor turns (which the client never sees, so the outer stats processor cannot count them) are priced into the stats planner bucket (OpenAI cached tokens read from `prompt_tokens_details`); each consult emits an `advisor_call={…}` audit line (`advisor_review={…}` on the gate arm).

Open follow-ups carried from the dev discussion: (a) verify end-to-end that planner-bucket advisor cost lands in the deployed cost output; (b) request-declared advisor (parsing/stripping a client-sent advisor tools entry, with an opt-in route mode) was scoped but not built — today the advisor is route-configured, always-on, and invisible to the API contract.

## How tested

- [x] `uv run mypy switchyard` clean (strict)
- [x] `uv run pytest tests/` — **2070 passed** (100 advisor tests: the Anthropic tool-call suite — unmodified across the format generalization — plus its OpenAI mirror, review-gate contract, both advisor callers via respx, profile wiring + format validation, route-bundle `type: advisor` incl. OSS-tier and rejected-format cases, presets, exports)
- [x] `cargo test -p switchyard-components` (246 tests, incl. the rewritten native-passthrough adversarial suite), `cargo clippy` clean, `cargo fmt --check` clean
- [ ] Live e2e not run from this host (no credentials); the Anthropic chain was exercised live during the internal benchmark runs summarized above

## Checklist

- [x] New public symbols exported from `switchyard/__init__.py.__all__` (`AdvisorConfig`, `AdvisorPresets`, `AdvisorProfileConfig`).
- [x] Unit tests added for new components.
- [x] Skills synced (`switchyard-lib-core` advisor row); AGENTS.md backends list updated.

## Notes for reviewers

- **The Rust passthrough change is a behavior change** for every native-Anthropic route, not just advisor ones: fields the old path stripped from real Anthropic clients (`reasoning_effort`, `context_management`, unsigned thinking blocks, message-level system turns, unsanitized tool ids) now forward verbatim, and the upstream API decides what to accept. The adversarial tests were rewritten to pin the new contract; translated-path (OpenAI/Responses→Anthropic) stripping is unchanged.
- Route YAMLs should **pin `strategy:` explicitly** — the config default is `tool_call`, so an unpinned `type: advisor` route means the tool-call arm. `review_gate` is wire-generic too: it dispatches on `executor.format`, its REDO feedback prefix is configurable (`redo_feedback_prefix`), and its proxy-side trigger makes it the recommended strategy for weak executors that rarely call tools.
- `AdvisorToolCallBackend.supported_request_types` follows the **executor's** wire; the chain's TranslationEngine normalizes any inbound client format (OpenAI Chat, Anthropic Messages, Responses) to it once, so e.g. Claude Code can front a Qwen executor and an OpenAI SDK client can front a Claude executor.
- Both advisor backends are Python multi-call backends that do their own stats accounting (like `LatencyServiceLLMBackend`): do **not** wrap them in `StatsLlmBackend`; `with_runtime_components` attaches the shared accumulator via the `_stats` hook (covered by a test).
- Example OSS route (Qwen executor + DeepSeek advisor):
  ```yaml
  routes:
    my-model:
      type: advisor
      strategy: tool_call
      executor: {model: qwen/qwen3-max, base_url: https://exec.example/v1, api_key: "${KEY}", format: openai}
      advisor:  {model: deepseek/deepseek-r2, base_url: https://adv.example/v1, api_key: "${KEY}", format: openai}
  ```

## Update 2026-07-22 — judge-side controls and session seeding

Two commits add YAML-tunable advisor controls, all default-off and backward compatible:

- `seed_plan_advice` + `seed_advice_prefix` (both strategies): consult the advisor once at each session-opening request and re-inject its plan into the first user message on every turn (stable cache prefix; fail-open; `advisor_seed=` audit line).
- `max_reviews` (review_gate): per-session review budget — re-declared completions after a REDO are re-verified (sequential best-of-N with the advisor as judge).
- `gate_stall_turns` (review_gate): once-per-session mid-task checkpoint when the conversation exceeds N assistant turns without the completion trigger firing.
- `gate_min_tool_results` (review_gate, `no_tool_call` trigger): only review no-tool-call turns after real work exists — skips early commentary turns on chatty harnesses.
- benchmark runner: env-gated `SWITCHYARD_DOCKER_NETWORK_MODE=host` for hosts whose Docker bridge cannot route to the upstream.

In our evaluation, the review-budget configuration produced the strongest results of the strategies tested; upfront seeding measured neutral-to-slightly-negative and ships as ablation surface. Validation: full test suite, ruff, and mypy green (new failing-first unit tests cover seeding, budgets, stall checkpoint, and the trigger guard).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,6 +158,23 @@ Direct Rust bindings for migrated concrete processors/backends are exposed from
 | Response component | Plain Python/Rust object | `async process(ctx, response) -> ChatResponse` | Post-process (logging, stats) |
 | `TranslationEngine` | `switchyard_rust.translation` | `async translate(ctx, request, response) -> Any` | Convert to client's wire format |
 
+### Multi-call backends
+
+`AdvisorProfileConfig` (`switchyard/lib/profiles/advisor.py`) dispatches on `AdvisorConfig.strategy`
+to one of two backends. `AdvisorToolCallBackend` (`tool_call`, default) offers the executor a
+proxy-intercepted `advisor` tool; `AdvisorLoopBackend` (`review_gate`) injects no tool and instead
+reviews the executor's first no-tool-call turn, because front-loaded advice suppressed the
+executor's own test-and-iterate loop — its trigger is proxy-side, so prefer it for executors that
+rarely call tools. Advisor text goes into the system prompt and the first user message, never the
+newest turn, so the upstream cache prefix stays stable across a session. Both are multi-call — one
+`call(...)` issues several upstream requests before returning one `ChatResponse`, so "exactly one
+per chain" holds at the chain level only — and both do their own stats accounting into the
+classifier bucket, so they must not be wrapped in `StatsLlmBackend`, which rejects Python-only
+backends; `with_runtime_components` attaches the accumulator instead. Executor and advisor targets
+dispatch independently on `LlmTarget.format`; `responses` is rejected at `AdvisorConfig` validation.
+Compose with a `type: advisor` route (`switchyard/cli/route_bundle.py`) or an `AdvisorPresets`
+helper.
+
 ## Project Structure
 
 ```
@@ -183,6 +200,8 @@ switchyard/
 │   │   ├── openai_llm_backend.py           # OpenAiPassthroughBackend
 │   │   ├── openai_native_backend.py        # OpenAiNativeBackend
 │   │   ├── anthropic_native_llm_backend.py # AnthropicNativeBackend
+│   │   ├── advisor_tool_call_backend.py    # AdvisorToolCallBackend (advisor tool_call strategy)
+│   │   ├── advisor_loop_backend.py         # AdvisorLoopBackend (advisor review_gate strategy)
 │   │   ├── llm_target.py                   # LlmTarget, BackendFormat
 │   │   ├── multi_llm_backend.py            # MultiLlmBackend helpers
 │   │   ├── stats_llm_backend.py            # StatsLlmBackend
--- a/benchmark/run-baseline.sh
+++ b/benchmark/run-baseline.sh
@@ -881,9 +881,21 @@ if [[ "\${SERVER_ENABLED}" == "1" ]]; then
     DOCKER_RUN_ARGS=(
         -d --rm
         --name "\${SWITCHYARD_DOCKER_CONTAINER}"
-        --network "\${SWITCHYARD_DOCKER_NETWORK}"
-        --network-alias "\${SWITCHYARD_DOCKER_SERVICE_NAME}"
-        -p "127.0.0.1:$(q "${PORT}"):$(q "${PORT}")"
+    )
+    if [[ "\${SWITCHYARD_DOCKER_NETWORK_MODE:-bridge}" == "host" ]]; then
+        # Host networking: for upstreams only routable from the host (VPN /
+        # corp-internal gateways that Docker bridge networks cannot reach).
+        # Pair with --harbor-server-url http://<host-ip>:<port> so task
+        # containers reach the server at the host address.
+        DOCKER_RUN_ARGS+=(--network host)
+    else
+        DOCKER_RUN_ARGS+=(
+            --network "\${SWITCHYARD_DOCKER_NETWORK}"
+            --network-alias "\${SWITCHYARD_DOCKER_SERVICE_NAME}"
+            -p "127.0.0.1:$(q "${PORT}"):$(q "${PORT}")"
+        )
+    fi
+    DOCKER_RUN_ARGS+=(
         -v "\${REPO_ROOT}:\${REPO_ROOT}:ro"
         -v "\${RUN_DIR}:\${RUN_DIR}"
     )
--- a/crates/switchyard-components/src/backends/anthropic.rs
+++ b/crates/switchyard-components/src/backends/anthropic.rs
@@ -80,24 +80,36 @@ impl AnthropicNativeBackend {
     }
 
     fn outbound_body(&self, request: &ChatRequest) -> Result<Value> {
-        let mut body = match request.request_type() {
-            ChatRequestType::Anthropic => request.body().clone(),
-            source => {
-                self.translation
-                    .translate_request(
-                        request_wire_format(source),
-                        WireFormat::AnthropicMessages,
-                        request.body(),
-                        &self.translation_policy,
-                    )
-                    .map_err(|error| {
-                        SwitchyardError::Backend(format!(
-                            "failed to translate {source:?} request to Anthropic Messages: {error}"
-                        ))
-                    })?
-                    .body
-            }
-        };
+        // Native Anthropic requests are a TRUE passthrough: forward the client's
+        // body verbatim, rewriting only the model id (routing) and merging any
+        // operator-configured `extra_body`. The strips/normalization in the
+        // translated branch below exist for OpenAI/Responses -> Anthropic bodies,
+        // which can carry fields/shapes the Anthropic API rejects. Applying them
+        // to a real Anthropic client (e.g. Claude Code) silently drops valid
+        // Anthropic features — `context_management` (context auto-compaction),
+        // signed thinking blocks, mid-conversation system turns, client tool ids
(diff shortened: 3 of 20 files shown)


nv-msvoboda:
@sir-merge-a-lot groom

ayushag-nv:
closing this as we have equivalent of this being done in rust.

PR #17, opened 2026-07-02
feat(libsy): core interfaces and examples

messiaen:
## Summary

Adds **`libsy`** — a lightweight, embeddable Rust SDK for multi-LLM agent
optimization, with **routing** as the first use case. An algorithm decides —
statefully, using more than the request — which model(s) to call and how, and
never performs the call itself ("ask, don't call"), keeping the core provider-
and transport-agnostic. Ships with reference routers, runnable examples, and a
demo HTTP proxy that wires libsy routing into Switchyard's existing server /
translation crates.

## What's included

- **Core model** (`crates/libsy/src/lib.rs`)
  - `OrchAlgo` (the algorithm) + `MultiLlmOrchestrator` (the driver).
  - `LlmTarget` targets that either **serve** a call (carry an
    `LlmClient`) or **offload** it as a promise the host fulfills.
  - `DecisionTrace` trait-object decisions (uniform to read, downcast when the
    algorithm is known).
  - Two run modes: `orchestrate_direct` (all targets self-serve → `(trace,
    response)`) and streaming `orchestrate` (offloaded calls surface as
    `CallLlm` promises; `set_response(Ok/Err)` propagates results and errors).
- **Concurrency** — one shared `Arc<dyn OrchAlgo>` with no lock; `process_request(&self)`
  runs requests in parallel; each `orchestrate` run gets its own promise channel
  via a task-local, so concurrent requests' offloaded calls never cross.
- **Reference algorithms**
  - `RandomOrchAlgo` (`rand.rs`) — uniform random, single call.
  - `LlmClassifierOrchAlgo` (`llm_class.rs`) — classify → strong/weak, fail-open.
  - `EnsembleOrchAlgo` (`ensemble.rs`) — stateful fan-out → judge → commit.
- **Examples** — `research_agent.rs` (self-serving / `orchestrate_direct`) and
  `research_agent_core.rs` (offloaded / streaming).
- **Docs** — `README.md` (code-first scope) and `DESIGN.md` (narrative design).
- **Demo proxy** (`demo/libsy-proxy`) — a real HTTP proxy: Switchyard's server +
  translation crates handle the OpenAI / Anthropic / Responses APIs and format
  translation, while **all routing** is libsy's classifier algorithm. Shows
  libsy embedded in a production-shaped I/O stack without depending on
  Switchyard's own routing.

## Architecture

request → OrchAlgo.process_request (N × target.call) → (DecisionTrace, response)
                       │
[...]

PR #58, opened 2026-07-13
refactor(libsy): use conversation IR request and response types

nachiketb-nvidia:
Linear: [SWITCH-620 - Run a 24 hr stress test on Switchyard server](https://linear.app/nvidia/issue/SWITCH-620/run-a-24-hr-stress-test-on-switchyard-server)

## Why

PR #17 introduced libsy with bespoke `LlmRequest` / `LlmResponse` structs that only carried a text prompt and completion. That makes libsy look like it owns a new LLM wire abstraction, when Switchyard already has a shared provider-neutral conversation IR in `switchyard-translation`.

This change makes libsy align with that direction now: `LlmRequest` and `LlmResponse` are semantic aliases over `ConversationRequest` and `ConversationResponse`, while provider-specific details stay in clients/hosts rather than in router algorithms.

## What

- Replaces libsy-owned `LlmRequest` and `LlmResponse` structs with aliases to `ConversationRequest` and `ConversationResponse`.
- Adds small text helpers for text-only router logic and examples: `text_request`, `text_response`, `request_text`, and `response_text`.
- Updates `llm_class`, `ensemble`, and `rand` to inspect conversation IR via the helpers instead of old `prompt` / `completion` fields.
- Boxes `Step::ReturnToAgent` so the larger conversation response does not inflate the stream enum.
- Updates libsy examples and README to teach the conversation IR contract.
- Updates the libsy proxy demo to the current `Algorithm` / `Request` / `Response` API and compiles it against the shared router builder.

## How

The algorithms still operate on plain text where their current logic requires text, but that text is now extracted from the shared IR. Generated subcalls are built as text-only `ConversationRequest`s with the inbound model preserved in `model`; routed target selection remains on `Decision::selected_model()`.

The proxy demo still lets libsy make routing decisions and Switchyard own HTTP/profile serving. The demo now creates a text-only libsy request from inbound prompt text, runs the classifier algorithm directly, and returns a chat-completion response for Switchyard to translate back to the inbound format.

## What to Review

- Whether aliasing `LlmRequest` / `LlmResponse` to the translation IR is the right libsy contract for this PR.
- Whether the helper functions are the right minimum bridge for text-based router algorithms without reintroducing a second IR.
- Whether boxing `Step::ReturnToAgent` is acceptable for the stream API after the response type becomes larger.
- Whether the demo proxy update is appropriately scoped to compiling against the current libsy/server APIs.

## Validation

- `cargo test -p libsy`
- `cargo check -p libsy --examples`
- `cargo check -p libsy-proxy`
- `cargo test -p libsy-proxy`
- pre-commit hook: `cargo fmt`, `cargo clippy`

## Summary by CodeRabbit

* **New Features**
  * Added a shared, provider-neutral protocol for structured LLM conversations, including messages, roles, content blocks, and responses.
  * Updated the library API to support model selection and multi-part response content.
  * Improved routing and orchestration examples to use structured requests and responses.

* **Documentation**
  * Refreshed usage examples and API documentation to demonstrate the new conversation format.

* **Refactor**
  * Unified protocol types across the library and translation components.


crates/libsy/src/types.rs
messiaen, line 1: Let's move these types into a `protocol` or similar crate.

messiaen:
#17 is merged can you rebase? @nachiketb-nvidia

messiaen (changes requested):

crates/libsy/src/lib.rs
```diff
@@ -127,23 +117,12 @@ pub struct Request {
...
-}
-
-/// A response leaving the orchestrator: the neutral [`LlmResponse`] plus optional
+/// A response leaving the orchestrator: the neutral [`ConversationResponse`] plus optional
 /// correlation [`Metadata`].
 #[derive(Clone)]
 pub struct Response {
     /// The neutral model response.
-    pub llm_response: LlmResponse,
+    pub llm_response: ConversationResponse,
```
messiaen: @nachiketb-nvidia thoughts on renaming `ConversationResponse` to `LlmResponse`? ans similarly for Request

crates/switchyard-protocol/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "switchyard-protocol"
```
messiaen: wdyt of naming this `libsy-protocol`?

messiaen (commented):
Couple of minor naming comments otherwise LGTM

crates/libsy/src/lib.rs
nachiketb-nvidia: sounds great!

crates/switchyard-protocol/Cargo.toml
nachiketb-nvidia: perfect, we should make everything libsy eventually

crates/switchyard-translation/src/codecs/stream.rs
messiaen: Should this type also be move to `libsy-protocol`?
Can be separate PR.

messiaen (approved):

messiaen (approved):

PR #62, opened 2026-07-13
feat(router): add escalation_router profile with judge-latched strong/weak routing

linj-glitch:
## Summary

Ports the escalation-router work from the private fork (branch `linj_router_exp`) to the public repo.

- Adds an `escalation_router` profile with judge-latched strong/weak routing, including the escalation-judge request processor, profile configs, and `route_bundle` CLI support.
- ESC8 escalation-judge campaign through ESC7 plus a prompt generalization audit: judge prompt refinements, cost-estimator updates (`switchyard/lib/cost_estimator.py`, `crates/switchyard-components/src/stats/cost.rs`), and expanded tests.
- Adds routing-algorithm docs (`docs/routing_algorithms/escalation_router_routing.md`) and registers them in `mkdocs.yml`.

Note: changes under the `benchmark/` folder from the original branch are intentionally excluded from this PR.

## Test plan

- `tests/test_escalation_judge_request_processor.py`
- `tests/test_escalation_router_profile.py`
- `tests/test_route_bundle.py`

## Summary by CodeRabbit

* **New Features**
  * Added escalation-router routing, which starts requests on a weaker tier and can permanently promote them to a stronger tier when a judge detects difficulty.
  * Added YAML configuration support and public configuration exports for escalation routing.
  * Added session-affinity latching, configurable confirmation rules, fallbacks, timeouts, and judge controls.
  * Added pricing support for Nemotron 3 Ultra and additional Claude model variants.

* **Documentation**
  * Added comprehensive escalation-router configuration and behavior documentation.
  * Updated routing strategy navigation and overview guidance.


switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,815 @@
...
+
+    first_user_seen = False
+    tail_taken = 0
+    for m in messages:
+        if not isinstance(m, dict):
+            continue
+        role = m.get("role")
+        if role in ("system", "developer"):
+            hasher.update(_message_text(m).encode("utf-8"))
+            hasher.update(b"\x00")
```
ayushag-nv: The session_key is already derived in SesssionAffinity classes, please try to re-use that

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,815 @@
...
+steering the session (a new instruction, a correction, an answer), and
+is the most current statement of what the agent should be doing. Judge
+the *trajectory* — is the agent making real progress toward what the
+user currently wants — not the difficulty of the task itself. Return
+exactly one JSON object:
+
+{"escalate": boolean, "reason": "one short sentence naming the pattern"}
+
+Escalation is one-way for the rest of the session and expensive.
+Escalate only on a clear PATTERN of trouble, never on a single failed
```
ayushag-nv: Prompt should not be hardcoded. This is a tech debt to maintain and gives user less flexibility to change on runtime . Please make this configurable and move this to a prompt file. hardocded prompts should not be in the code files.

switchyard/cli/route_bundle.py
```diff
@@ -1008,6 +1051,147 @@ def _deterministic_switchyard(
...
+        strong:
+          model: anthropic/claude-opus-4.7
+          api_key: ${OPENROUTER_API_KEY}
+          base_url: https://openrouter.ai/api/v1
+        weak:
+          model: deepseek/deepseek-v4-pro
+          api_key: ${OPENROUTER_API_KEY}
+          base_url: https://openrouter.ai/api/v1
+        fallback_target_on_evict: strong
+        session_key_depth: 0        # >0 only for k>1 benchmark trials
```
ayushag-nv: These kind of doc strings just increases lines of code. Please keep this concise.

switchyard/lib/profiles/escalation_router_profile_config.py
```diff
@@ -0,0 +1,133 @@
...
+            EscalationJudgeRequestProcessor,
+        )
+        from switchyard.lib.processors.reasoning_effort_normalizer import (
+            ReasoningEffortNormalizer,
+        )
+        from switchyard.lib.profiles.deterministic_routing_profile_config import (
+            _apply_deepseek_overrides,
+            _apply_default_tier_timeout,
+        )
+        from switchyard.lib.session_affinity import SessionAffinity
```
ayushag-nv: Please move these imports to top level. We don't do imports under methods unless it is absolute necessary

ayushag-nv (dismissed):
This looks like a extension of LLM classifier or existing stage router. I suggest we should consider this as a design to extend our existing routers to accomodate this like we did for Session Pinning in the existing routers.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Good catch, thanks! Done in 1fc31a0 — the memoized derivation that lived in `SessionAffinity._session_key` is now a shared `resolve_session_key()` in `session_affinity.py`, and the judge's streak bookkeeping reuses it instead of reimplementing the lookup.

One part is intentionally kept separate: `_deep_session_key` is the `session_key_depth > 0` benchmark-only override that extends the key with the first N post-task messages so repeated identical-prefix trials get separate latches. It seeds the same `CTX_SESSION_KEY` metadata slot, so every downstream consumer (affinity included) still sees one consistent key per request.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Agreed — done in 1c942d7. The default prompt now lives in package data at `switchyard/lib/processors/prompts/escalation_judge.md` (same pattern as the stage-router classifier prompt) and is loaded at import, so there's no longer a prompt hardcoded in the code file.

For runtime flexibility there are now two YAML overrides on the route: the existing inline `judge.prompt` and a new `judge.prompt_path` that reads the rubric from a file (the two are mutually exclusive, with a clear config error).

One guard rail to be aware of: this exact default text is the benchmark-validated line of record, so a unit test pins its hash. Editing the file is still easy — the test just makes it a deliberate act that prompts re-validation instead of a silent drift.

switchyard/cli/route_bundle.py
linj-glitch: Fair point — trimmed in e3fa30b. The docstring is down to a short summary, and the full YAML schema and knob reference live in `docs/routing_algorithms/escalation_router_routing.md`, which it now points to.

switchyard/lib/profiles/escalation_router_profile_config.py
linj-glitch: Done in 13c6ead — hoisted all of the `build()` imports to module top; there's no import cycle in practice, and the full test suite is green.

Small note for context: the deferred-import pattern was originally mirrored from `deterministic_routing_profile_config.py`'s `build()`, which still uses it. I left that file untouched to keep this PR scoped, but happy to align it in a follow-up if you'd like.

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+
+    first_user_seen = False
+    tail_taken = 0
+    for m in messages:
+        if not isinstance(m, dict):
+            continue
+        role = m.get("role")
+        if role in ("system", "developer"):
+            hasher.update(_message_text(m).encode("utf-8"))
+            hasher.update(b"\x00")
```
ayushag-nv: may be you should extend existing data stuctures and make them more robust rather than adding this helpers

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+            )
+    first_user_seen = False
+    for m in trimmed:
+        if not isinstance(m, dict):
+            continue
+        role = m.get("role")
+        text = _message_text(m)
+        if role in ("system", "developer"):
+            anchor_lines.append(f"[{role}] " + _truncate(text, config.system_chars))
+        elif role == "user" and not first_user_seen:
```
ayushag-nv, line 500: This kind of thing is also present in classifier code for summarizing request, trim request etc. May be try to re-use them or extend them in a generic way.

switchyard/lib/profiles/escalation_router_profile_config.py
```diff
@@ -0,0 +1,132 @@
...
+            strong_target,
+        )
+        weak_backend = maybe_wrap_anthropic_cache(
+            build_native_backend(weak_target),
+            weak_target,
+        )
+
+        backend = DeterministicRoutingLLMBackend(
+            tiers={
+                _TIER_STRONG: (strong_backend, strong_target.model),
```
ayushag-nv: Blocking: `EscalationRouterConfig` allows custom target IDs, but this backend is always keyed as `strong`/`weak`. I reproduced this with `cheap` and `frontier`: after weak overflowed, the retry selected `frontier`, the backend did not recognize it, and weak was called again until the pool was exhausted. Can we either key the backend with the configured IDs or restrict the config to these literal IDs, and add an overflow test through the full profile?

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+
+from switchyard.lib.backends.deterministic_routing_llm_backend import (
+    CTX_DETERMINISTIC_ROUTING_TIER,
+)
+from switchyard.lib.conversation_turn import conversation_turn_number
+from switchyard.lib.llm_client import OpenAILLMClient
+from switchyard.lib.processors.llm_classifier.request_processor import (
+    LLMClassifierClient,
+    OpenAIChatLLMClassifierClient,
+    _strip_markdown_fence,
```
ayushag-nv: I think this is the place to share more of the existing classifier path. The judge imports private classifier helpers, then reimplements structured-output parsing, fail-open handling, usage accounting, and logging. Could we extract a supported structured-decision client and keep only the escalation policy and confirmation logic here? That would give the classifier, stage router, and escalation router one common model-call path.

switchyard/lib/profiles/escalation_router_profile_config.py
```diff
@@ -0,0 +1,132 @@
...
+        judge_config = EscalationJudgeConfig(
+            model=judge_target.model,
+            api_key=judge_target.api_key,
+            base_url=judge_target.base_url,
+            timeout_s=judge_target.endpoint.timeout_secs or config.judge_timeout_s,
+            system_prompt=config.judge_system_prompt or ESCALATION_JUDGE_SYSTEM_PROMPT,
+            min_judge_turn=config.judge_min_turn,
+            escalate_confirmations=config.judge_escalate_confirmations,
+            confirmation_window=config.judge_confirmation_window,
+            disable_reasoning=config.judge_disable_reasoning,
```
ayushag-nv: Can we use `model_accepts_reasoning_hint()` here, as the classifier and stage router already do? With the current default, a Claude or Bedrock judge gets the vLLM-only `chat_template_kwargs` field, rejects the call, and every judged turn quietly falls back to weak. The existing Claude and DeepSeek compatibility tests should carry over as well.

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+    if isinstance(content, str):
+        return content
+    parts: list[str] = []
+    for block in content:
+        if isinstance(block, str):
+            parts.append(block)
+            continue
+        if not isinstance(block, dict):
+            continue
+        text = block.get("text")
```
ayushag-nv: Anthropic tool calls are lost here because `tool_use` blocks have neither `text` nor nested `content`. I tried a bash tool call and the judge transcript contained an empty assistant line, which removes exactly the repeated-command signal this router relies on. Can we use one format-aware transcript renderer for Chat, Responses, and Anthropic rather than flattening each shape here?

switchyard/lib/profiles/escalation_router_profile_config.py
```diff
@@ -0,0 +1,132 @@
...
+from switchyard.lib.processors.escalation_judge_request_processor import (
+    ESCALATION_JUDGE_SYSTEM_PROMPT,
+    EscalationJudgeConfig,
+    EscalationJudgeRequestProcessor,
+)
+from switchyard.lib.processors.reasoning_effort_normalizer import (
+    ReasoningEffortNormalizer,
+)
+from switchyard.lib.profiles.chain import ComponentChainProfile
+from switchyard.lib.profiles.deterministic_routing_profile_config import (
```
ayushag-nv: This duplicates the deterministic router's target resolution, timeout defaults, DeepSeek overrides, backend construction, and Anthropic cache wrapping, and it already needs private helpers from that profile. Could we pull that into a supported two-tier builder so the two profiles cannot drift as those rules change?

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+    pre-escalation turn; a slow judge taxes the whole session, so keep this
+    tight — any failure or timeout fails open to the weak tier."""
+
+    max_completion_tokens: int = Field(default=128, ge=16)
+    system_prompt: str = Field(default=ESCALATION_JUDGE_SYSTEM_PROMPT, min_length=1)
+    structured_output_mode: Literal["json_schema", "json_object"] = "json_schema"
+    disable_reasoning: bool = True
+    extra_headers: dict[str, str] | None = None
+
+    dump_verdicts_to_stderr: bool = True
```
ayushag-nv: This is still enabled in the shipping profile, including the launcher path, and the payload includes part of the first user message as `task_hint`. The classifier profile disables its stderr dump because it interferes with the agent TUI. Can we make this benchmark output explicitly opt-in and keep production routing telemetry in the normal stats path?

switchyard/lib/profiles/escalation_router_profile_config.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def build(self) -> ComponentChainProfile:
+        """Build the escalation-router profile runtime."""
+        config = self.config
+
+        # The affinity store IS the escalation latch, so it is always on.
+        # Warmup gating lives in the judge processor (min_judge_turn), not
+        # here — affinity warmup would delay a decided escalation from
+        # taking effect, not just from being decided.
+        affinity = SessionAffinity(
```
ayushag-nv, line 54: The profile describes escalation as one-way for the rest of the task, but this creates only an in-process affinity store, and the confirmation streak is another local LRU. A worker or pod change can lose the latch, and confirmations split across workers never accumulate. Is that intended? If this is a production profile, I think it should expose the existing shared affinity store and define how pending confirmations are shared.

switchyard/cli/route_bundle.py
```diff
@@ -1008,6 +1052,151 @@ def _deterministic_switchyard(
...
+    )
+    valid_ids = {strong.id, weak.id}
+    if fallback_target_on_evict not in valid_ids:
+        raise RouteBundleConfigError(
+            f"route {model_id!r}: fallback_target_on_evict="
+            f"{fallback_target_on_evict!r} must match one of {sorted(valid_ids)} "
+            f"(the configured strong/weak target ids)",
+        )
+
+    config_data: dict[str, object] = {
```
ayushag-nv, line 1137: We now have a second escalation config schema here, including copies of defaults already owned by `EscalationRouterConfig`, while the new router is only reachable through the deprecated route-bundle path. Could we register it with the primary profile loader and make this compatibility path delegate to the same parser? Otherwise these two config surfaces will drift.

switchyard/lib/processors/escalation_judge_request_processor.py
```diff
@@ -0,0 +1,635 @@
...
+
+    model: str = Field(min_length=1)
+    api_key: str | None = None
+    base_url: str | None = None
+    timeout_s: float = Field(default=5.0, gt=0.0)
+    """Judge wall-clock ceiling. The judge sits on the request path of every
+    pre-escalation turn; a slow judge taxes the whole session, so keep this
+    tight — any failure or timeout fails open to the weak tier."""
+
+    max_completion_tokens: int = Field(default=128, ge=16)
```
ayushag-nv, line 103: `judge_disable_reasoning=False` is exposed, but the completion budget remains fixed at 128 tokens and is not configurable from the profile. The classifier uses a much larger budget because reasoning tokens and the JSON verdict share the same limit. Should we reuse that budget policy or expose the limit here? As written, a reasoning judge can truncate before producing a verdict and silently fail open.

ayushag-nv (changes requested):

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Done in a6a5f89. The depth extension is now part of the canonical key derivation: Rust `session_key_from_body` gained a `with_depth` variant sharing one hashing core, the Python binding takes `depth=0`, and `resolve_session_key(ctx, request, depth=N)` passes it through. The judge now calls the existing structure; the private re-implementation is removed. Unifying also closed a gap — the Rust tail-hasher covers tool-call payloads, so trials whose early divergence is entirely in tool calls (empty `content`) now get separate latches.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Done in 155bc6b. The shared plumbing — message trimming, content flattening, truncation, output-fence stripping — moved to `switchyard/lib/processors/message_condensing.py` with public names, and the classifier, judge, and plan-execute planner all import from it (the planner had a third copy of the fence stripper). Only the output shape stays per-processor: condensed JSON body for the classifier, role-labeled transcript for the judge.

switchyard/lib/profiles/escalation_router_profile_config.py
linj-glitch: Done in 920190c, taking the first option: the backend tiers are keyed by the configured target ids. The judge processor takes `strong_tier`/`weak_tier` labels, the profile passes `config.strong.id`/`config.weak.id`, and the config now rejects duplicate ids. Your repro is covered by a full-profile test (`cheap`/`frontier`, weak overflows, retry lands on `frontier`).

switchyard/lib/profiles/escalation_router_profile_config.py
linj-glitch: Done in 817693b. The profile gates the hint with `model_accepts_reasoning_hint()`, same as the classifier presets and the stage router, so Claude/Bedrock judges never receive `chat_template_kwargs`. Added Claude and non-Anthropic judge tests alongside the existing DeepSeek gateway test.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Done in 9600504. The transcript renderer is already one shared, format-aware flattener (`message_condensing.py`) — the gap was the missing block type. `tool_use` blocks now render as `tool_call name(input)`, consistent with the OpenAI Chat and Responses shapes, and the Rust deep-key tail hashing got the same handling so Anthropic-format benchmark trials diverge correctly. Test covers a bash tool call surviving into the judge transcript.

switchyard/lib/profiles/escalation_router_profile_config.py
linj-glitch: Done in 91812c4. `switchyard/lib/profiles/tier_target_builders.py` now owns the tier-timeout default, DeepSeek overrides, `format='auto'` resolution, native backend construction, and Anthropic cache wrapping; both profiles call `build_tier_backend()` and the private cross-imports are gone.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Done in 8e986d1. `dump_verdicts_to_stderr` defaults off everywhere now; benchmark runs opt in with `judge.dump_verdicts: true` (existing benchmark configs need that one-line addition). Production routing telemetry stays on the stats path.

switchyard/lib/profiles/escalation_router_profile_config.py
linj-glitch: Done in 8e986d1 for the latch: `affinity_store: redis` + `affinity_store_url` share it through the existing `RedisPinStore` L2 (best-effort, circuit-breakered), same semantics as the latency route. The confirmation streak stays per-process by design — it's short-lived counting state, and a shared store on the verdict path would add a round trip per judged turn. The config docs now state that limitation and recommend `confirmations: 1` (the default) or session-sticky load balancing on multi-worker deployments. Open to revisiting if there's a deployment that needs distributed confirmations.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: Done in 8e986d1. `judge.max_completion_tokens` is exposed, and the default follows the reasoning mode: 128 with the thinking-off hint in effect, 4096 (the classifier's budget) when `judge.disable_reasoning: false` — so a reasoning judge no longer truncates and fails open by default. Explicit values win.

ayushag-nv (approved):

PR #67, opened 2026-07-14
fix: Do not advertise the model directly

### LINKED ISSUE #51: [feature] components-v2: allow the same upstream model across multiple targets/endpoints (duplicate registration hard-fails the gateway)
## Problem

In a components-v2 profile config, two `targets` that point at the **same upstream `model`** (e.g. the same model served through two endpoints with different API keys) cause `serve` to abort at load with `SwitchyardDuplicateRegistrationError`, even when the `targets` and `profiles` have **distinct ids**. This makes it impossible to serve one upstream model via two credentials in a single gateway — the only workaround today is running a separate gateway process per key.
?

## Reproduction

`profiles.json` — two endpoints (same upstream, different keys), two **distinctly-named** targets and profiles, both pointing at the same `model`:

```json
{
  "endpoints": {
    "team_a": { "base_url": "https://api.openai.com/v1", "api_key": "sk-team-a" },
    "team_b": { "base_url": "https://api.openai.com/v1", "api_key": "sk-team-b" }
  },
  "targets": {
    "gpt-5.4__a": { "endpoint": "team_a", "model": "gpt-5.4", "format": "responses" },
    "gpt-5.4__b": { "endpoint": "team_b", "model": "gpt-5.4", "format": "responses" }
  },
  "profiles": {
    "team_a__gpt-5.4": { "type": "passthrough", "target": "gpt-5.4__a" },
    "team_b__gpt-5.4": { "type": "passthrough", "target": "gpt-5.4__b" }
  }
}
```

```
$ switchyard serve --config profiles.json --port 4000
INFO - Switchyard components-v2 profile config loaded from profiles.json
Traceback (most recent call last):
  ...
  File ".../switchyard_rust/server.py", line 17, in run_profile_server
    _load_native().run_profile_server(config_path, host, port, backlog, dry_run)
_switchyard_rust.SwitchyardDuplicateRegistrationError: model "gpt-5.4" is already registered
```

## Proposed solution

When multiple targets declare the same upstream `model`, don't abort. The namespaced profile/target ids already disambiguate for routing — only the *bare-model alias* is ambiguous. Any of these would unblock it:

1. **Skip the bare-model alias when it's ambiguous** (register only the unique profile/target ids). Simplest; selection by profile id keeps working.
2. **Register the bare alias once to a designated default** (first target, or an explicit `"default": true` on a target).
3. **Gate the bare-model alias behind an opt-in flag** (per-target or global), defaulting to off.

Option 1 seems least surprising.

## Scope notes

- Is this a new role/component in the chain (RequestProcessor / LLMBackend / ResponseProcessor / ResponseTranslator), or an extension of an existing one? No
- Does it touch a public API listed in `switchyard/__init__.py.__all__`? No
- Backward-compatibility concerns? No

grahamking:
Only advertise the target names, for example 'classifier', 'weak' and 'strong'.

This allows the operator to adjust the models without breaking clients.
It also removes the bug that prevents using the same model for
classifier and weak (also for example).

The code change is in `registry.rs`.

The rest is two test removals:
- `target_id_and_target_model_aliases_are_advertised_and_routable`. The
  checked we advertise the model directly which we don't now. The other
  half of that test is covered by a different test.
- `duplicate_public_model_ids_are_rejected`. This prevented having a
  model named the same as a target. That's no longer a problem because
  we don't advertise the model directly.

Closes: https://github.com/NVIDIA-NeMo/Switchyard/issues/51

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated passthrough model resolution so inbound model names use the configured target ID.
  * Target model names are now presented as display names rather than additional aliases.
  * Removed support for routing through alternate target-model aliases.

### DIFF
--- a/crates/switchyard-server/src/registry.rs
+++ b/crates/switchyard-server/src/registry.rs
@@ -67,13 +67,6 @@ impl ProfileRegistry {
                 Arc::clone(&profile),
                 target.model.as_str(),
             )?;
-            if target.id.as_str() != target.model.as_str() {
-                registry.insert(
-                    target.model.clone(),
-                    profile,
-                    format!("target {}", target.id.as_str()),
-                )?;
-            }
         }
 
         Ok(registry)
--- a/crates/switchyard-server/tests/server.rs
+++ b/crates/switchyard-server/tests/server.rs
@@ -207,60 +207,6 @@ async fn translation_errors_do_not_emit_routing_metadata_headers() -> TestResult
 }
 
 #[tokio::test]
-async fn target_id_and_target_model_aliases_are_advertised_and_routable() -> TestResult {
-    let _stats_guard = stats_guard().await;
-    let Some(stub) = HttpStub::start(2)? else {
-        log_loopback_bind_skip();
-        return Ok(());
-    };
-    let app = build_switchyard_router(state_from_yaml(&format!(
-        r#"
-targets:
-  direct:
-    model: upstream-direct
-    format: openai
-    base_url: {base_url}
-profiles:
-  direct-profile:
-    type: passthrough
-    target: direct
-"#,
-        base_url = stub.base_url
-    ))?);
-
-    let models = app
-        .clone()
-        .oneshot(request("GET", "/v1/models", None)?)
-        .await?;
-    let model_ids = json_body(models).await?["model_pool"].clone();
-    assert_eq!(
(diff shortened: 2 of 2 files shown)


grahamking:
@coderabbitai full review

ryan-lempka (approved):

PR #68, opened 2026-07-14
feat(stage_router): corroborative tanh-scored routing with windowed self-clearing signals

sabhatinas:
## What this PR does

Reworks the stage_router scorer, picker, and signals into a cleaner, honestly-documented router — and moves the routing logic into one shared Rust core.

- **One shared Rust core.** The scorer + picker now live in `switchyard-components/stage_router.rs` (a short priority ladder: escalate → de-escalate → scorer → fall-open). Python calls it via a binding (`stage_pick_tier`); the Rust profile calls it directly. One source of truth. The dead prompt-classification path is deleted. (SWITCH-1056)
- **Two-axis scorer.** Error signals (a hard error, spinning, or exploring) push toward the strong model; making progress (writing code) pushes toward the weak model. Squashed with `tanh`, and `confidence = |score|`.
- **Corroborative.** One signal alone rarely flips the decision — it takes a signal plus backup. A critical error overrides everything.
- **Cleaner signals.** `spinning` and `exploring` are mutually exclusive. Dropped `repeated_cmd_ratio` from scoring — it fired on nearly every shell command and never caught a real loop.
- **Compaction holds strong.** Once the context is compacted, the router pins the strong model.
- **Handoff notes.** Optional short note at a tier switch telling the next model why it was switched (off by default).
- **Per-tier system prompts.** Optional prompt injected for the weak or strong tier.

Docs and tests updated; the old calibration scripts are replaced by the scorer skill.

> **Signals live in libsy.** Tool-signal extraction (`tool_signals`) now lives in `switchyard_libsy::ToolSignals`; `switchyard-components` re-exports it — no divergence. The scorer/picker core stays in `switchyard-components/stage_router.rs`.
> **count_tokens** was split out to its own PR (SWITCH-1048).

Linear: SWITCH-1056 (Rust consolidation), SWITCH-939 (scorer), SWITCH-983 (compaction), SWITCH-982 (handoff notes).


ryan-lempka:
@sabhatinas is the plan to refactor this to align with libsy design at a later date / does this take libsy into account?

switchyard/lib/endpoints/anthropic_messages_endpoint.py
```diff
@@ -120,4 +122,52 @@ async def anthropic_messages(
                     log_msg=f"POST /v1/messages chain raised model={model}",
                 )
 
+        @router.post("/v1/messages/count_tokens", response_model=None)
```
messiaen: Should adding cout_tokens be a separate PR?

crates/switchyard-py/src/component_bindings/dimension_collector.rs
```diff
@@ -541,6 +541,20 @@ impl PyToolResultSignal {
         self.inner.prompt_char_count
     }
 
+    /// Largest share taken by a single identical command-bearing tool call over the
+    /// recent window, in `[0, 1]` — a weak model looping the same command (churn).
+    #[getter]
```
messiaen: is `#[getter]` like `@property` in python?

benchmark/score_staged_run.py
```diff
@@ -0,0 +1,282 @@
+#!/usr/bin/env python3
+"""Score a benchmark run directory via the stage-router Rust scorer.
```
messiaen, line 4: Should this file be called `score_statged_run.py`

benchmark/score_run.py
```diff
@@ -0,0 +1,282 @@
...
+import argparse
+import asyncio
+import csv
+import json
+import sys
+from collections import defaultdict
+from pathlib import Path
+from statistics import mean
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
```
messiaen: Couldn't we just depend on switchyard instead of manually injecting the path?

switchyard/lib/processors/stage_router_request_processor.py
```diff
@@ -113,6 +131,37 @@ async def _record_decision_source(self, ctx: ProxyContext) -> None:
...
+        else:
+            # OpenAI: prepend a system message before existing messages
+            messages = body.get("messages")
+            if not isinstance(messages, list):
+                return
+            if messages and messages[0].get("role") == "system":
+                existing = messages[0].get("content", "")
+                messages[0] = {"role": "system", "content": prompt + "\n\n" + existing}
+            else:
+                messages.insert(0, {"role": "system", "content": prompt})
```
messiaen, line 159: Can this be format agnostic?

switchyard/server/switchyard_app.py
```diff
@@ -152,6 +152,32 @@ async def _invalid_request_handler(
...
+            if callable(iter_fn):
+                for child in iter_fn():
+                    found = _find_anthropic_target(child, depth + 1)
+                    if found is not None:
+                        return found
+            return None
+
+        app.state.anthropic_target = _find_anthropic_target(switchyard)
+    except Exception:
+        app.state.anthropic_target = None
```
messiaen: Same as the above. I think count_token stuff should be separate PR

sabhatinas:
> @sabhatinas is the plan to refactor this to align with libsy design at a later date / does this take libsy into account?

Yes, I plan to migrate to `libsy` once I have this checked in. I want to have a clean submission of the working algorithm to reproduce results from our submissions.

benchmark/score_staged_run.py
sabhatinas: Done — renamed to `score_staged_run.py`.

benchmark/score_run.py
sabhatinas: Done — removed the `sys.path.insert`; the script relies on the installed `switchyard` package (run via `uv run`).

crates/switchyard-py/src/component_bindings/dimension_collector.rs
sabhatinas: Yes — `#[getter]` is pyo3's equivalent of Python's `@property`. Note this getter (`repeated_cmd_ratio`) has since been removed: the signal was dropped from the scorer, so the field no longer exists.

switchyard/lib/processors/stage_router_request_processor.py
sabhatinas: Not fully — Anthropic uses a top-level `system` field while OpenAI uses a `system`-role message, so the two genuinely need different handling. The branches share one intent (prepend the tier prompt); kept them explicit.

switchyard/lib/endpoints/anthropic_messages_endpoint.py
sabhatinas: Agreed — count_tokens is being split into its own PR. Tracked as SWITCH-1048.

switchyard/server/switchyard_app.py
sabhatinas: Agreed — same as above; count_tokens is moving to its own PR (SWITCH-1048).

switchyard/lib/endpoints/anthropic_messages_endpoint.py
sabhatinas: Done — count_tokens is removed from this PR (commits dropped). It will land in its own PR with the proper stage_router target resolution — SWITCH-1048.

switchyard/server/switchyard_app.py
sabhatinas: Done — removed; the `anthropic_target` wiring went with it. count_tokens moves to its own PR (SWITCH-1048).

messiaen (approved):

PR #69, opened 2026-07-15
feat(libsy): add OTel metrics, tracing spans, and structured logging

eric-liu-nvidia:
## Summary

Implements SWITCH-928: the observability layer for the libsy algorithm layer, built on OpenTelemetry per the discussion on the ticket (opentelemetry-rust as the metrics sink; `tracing` for spans and structured logs, which hosts bridge into OTel with `tracing-opentelemetry`).

All instrumentation lives in the libsy core at the `Decision` hook and the offload boundary, so algorithms carry no telemetry code beyond the new `Algorithm::name()` label:

- **Metrics** (OTel global meter provider, scope `libsy`; the library owns no exporter): counters `libsy.runs`, `libsy.llm_calls`, `libsy.decisions`, `libsy.{input,output,total,reasoning}_tokens`; histograms `libsy.run_duration_ms`, `libsy.llm_call_duration_ms`. Attributes are `algorithm` / `selected_model` / `outcome` (`ok`|`error`), so failure rates fall out of the `outcome` split. Instrument naming and cardinality bounds follow the switchyard OTel conventions.
- **Spans** (`tracing`): one `libsy.run` per request carrying the `Metadata` correlation ids and outcome, with one child `libsy.llm_call` per offloaded call carrying the selected model, latency, outcome, and token counts.
- **Structured logs** (`tracing`, target `libsy`): an info event per published `Decision` with its reasoning; warn events for failed calls and runs.

API changes: new required `Algorithm::name(&self) -> &str` (implemented for the three reference algorithms), `Usage` re-exported from `libsy`, and the unused `Driver::default()` removed (drivers now carry the algorithm label).

With no OTel SDK or tracing subscriber installed, all instrumentation is a no-op.

## Testing

- New integration tests in `crates/libsy/tests/observability.rs` assert the success path (counter/histogram values, exact token counts, span parentage and fields, decision log) and the failure path (`outcome=error` at run and call level, error text on spans, warn logs) via an in-memory OTel exporter and a capturing `tracing` layer.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` all pass (50/50 test binaries).

Fixes SWITCH-928

## Summary by CodeRabbit

- **New Features**
  - Added observability for algorithm runs, model calls, and routing decisions through structured tracing, logs, and OpenTelemetry metrics.
  - Captures run outcomes, durations, model selections, errors, and token usage when available.
  - Added stable names for built-in algorithms to improve telemetry labeling.

- **Documentation**
  - Expanded guidance for exploring the Rust libraries.
  - Documented algorithm naming and observability setup, including available metrics and tracing behavior.

- **Tests**
  - Added coverage for successful and failed runs, telemetry data, spans, metrics, and structured logs.

### DIFF
--- a/crates/libsy/src/lib.rs
+++ b/crates/libsy/src/lib.rs
@@ -52,6 +52,20 @@
 //! responsible for its own thread-safety — stateless (like the reference routers) or
 //! interior mutability over just its own state.
 //!
+//! ## Observability
+//!
+//! The provided run methods instrument every algorithm from the outside — at the
+//! [`Decision`] hook and the offload boundary — so algorithms carry no telemetry
+//! code. Each run gets a `libsy.run` tracing span (correlation ids from
+//! [`Metadata`] attached) with a child `libsy.llm_call` span per model call;
+//! each [`Driver::info`] decision is logged with its reasoning; and OpenTelemetry
+//! metrics record run/call counts, latency, token usage, and published
+//! decisions, keyed by [`Algorithm::name`] plus `selected_model` and `outcome`.
+//! Metrics use the global meter provider and spans/logs the `tracing` facade: a
+//! host that installs an OTel SDK and a `tracing` subscriber (bridged with
+//! `tracing-opentelemetry` for OTLP spans) gets the full signal set; with
+//! neither installed, everything is a no-op.
+//!
 //! ## Algorithms
 //!
 //! Concrete algorithms live in [`algorithms`]:
@@ -65,3 +79,5 @@ mod core;
 pub use core::*;
 
 pub mod algorithms;
+
+mod observability;
--- a/crates/libsy/src/lib.rs
+++ b/crates/libsy/src/lib.rs
@@ -57,7 +57,9 @@
 //! The provided run methods instrument every algorithm from the outside — at the
 //! [`Decision`] hook and the offload boundary — so algorithms carry no telemetry
 //! code. Each run gets a `libsy.run` tracing span (correlation ids from
-//! [`Metadata`] attached) with a child `libsy.llm_call` span per model call;
+//! [`Metadata`] attached) with a child `libsy.llm_call` span per model call
+//! (fulfillment time as the algorithm observes it) plus a `libsy.client_call`
+//! span around the actual API call when [`run`](Algorithm::run) serves it;
 //! each [`Driver::info`] decision is logged with its reasoning; and OpenTelemetry
 //! metrics record run/call counts, latency, token usage, and published
 //! decisions, keyed by [`Algorithm::name`] plus `selected_model` and `outcome`.
--- a/crates/libsy/src/observability.rs
+++ b/crates/libsy/src/observability.rs
@@ -12,8 +12,9 @@
 //! no-op. Spans and logs use the `tracing` facade (the async-native surface the
 //! OpenTelemetry ecosystem bridges with `tracing-opentelemetry` /
 //! `opentelemetry-appender-tracing`), so the host's subscriber decides where
-//! they go. Spans are attached to futures with [`Instrument`], never a
-//! [`Span::enter`] guard held across an `.await`: a suspended task would leave
+//! they go. Method spans use `#[tracing::instrument]`; the `libsy.run` span is
+//! attached to the spawned run task with [`tracing::Instrument`]. Neither holds
+//! a [`Span::enter`] guard across an `.await` — a suspended task would leave
 //! the span entered on its executor thread, mis-parenting every span other
 //! tasks create there (see the `tracing` docs on spans in asynchronous code).
 //!
@@ -34,7 +35,7 @@ use std::time::{Duration, Instant};
 
 use opentelemetry::metrics::Meter;
 use opentelemetry::{global, KeyValue};
-use tracing::{Instrument, Span};
+use tracing::Span;
 
 use crate::{Context, Decision, Metadata, Response};
 
@@ -49,7 +50,7 @@ const SCOPE: &str = "libsy";
 pub(crate) const ALGORITHM_KEY: &str = "algorithm";
 
 /// The algorithm label carried by a request context; empty until stamped.
-fn algorithm_label<S>(ctx: &Context<S>) -> &str {
+pub(crate) fn algorithm_label<S>(ctx: &Context<S>) -> &str {
     ctx.values
         .get(ALGORITHM_KEY)
         .map(String::as_str)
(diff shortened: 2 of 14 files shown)


crates/libsy/src/lib.rs
```diff
@@ -214,23 +232,46 @@ impl CallLlmRequest {
...
 impl Driver {
-    /// Build an empty driver with its step channel ready. Created per call by
+    /// Build an empty driver with its step channel ready, labeled with the name
+    /// of the algorithm it serves. Created per call by
     /// [`run_stream`](Algorithm::run_stream).
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new(algorithm: &str) -> Self {
         Self {
             driver: TypeErasedDriver::new(),
+            algorithm: Arc::from(algorithm),
```
messiaen: We should use a Context object for things like this.
In this case we could also put it in the Decision object.

Probably `driver.{call_model,info, ...}` should take the context object `Context` that `run` `run_stream` and `create_run_task` take.

crates/libsy/src/lib.rs
```diff
@@ -214,23 +232,46 @@ impl CallLlmRequest {
...
     /// Offload a model call: publish `routed` as a [`Step::CallLlm`] and await the
     /// consumer's [`Response`]. Errors if the stream is closed or the call failed.
+    /// The await is wrapped in a `libsy.llm_call` span, and the call's latency,
+    /// outcome, and token usage are recorded when it resolves.
     pub async fn call_llm(&self, routed: RoutedRequest) -> Result<Response, BoxErr> {
-        self.driver
-            .fulfill_request::<RoutedRequest, Response>(routed)
-            .await
+        let selected_model = routed.decision.selected_model().to_string();
+        let span = observability::llm_call_span(&self.algorithm, &selected_model);
```
messiaen: We need to be careful with labeling here. This is the time, from the algorithms perspective on how long it took to fulfill the model request.

We should also have a span for the api call itself. ie in LlmClient.

crates/libsy/src/observability.rs
```diff
@@ -0,0 +1,218 @@
...
+    let span = tracing::info_span!(
+        target: SCOPE,
+        "libsy.run",
+        algorithm,
+        session_id = tracing::field::Empty,
+        agent_id = tracing::field::Empty,
+        task_id = tracing::field::Empty,
+        correlation_id = tracing::field::Empty,
+        outcome = tracing::field::Empty,
+        error = tracing::field::Empty,
```
messiaen, line 92: I think I would make these generic labels, Or at least have the ability to add more labels generically.

crates/libsy/src/lib.rs
```diff
@@ -214,23 +232,46 @@ impl CallLlmRequest {
...
+                &self.algorithm,
+                &selected_model,
+                started.elapsed(),
+                &result,
+                &tracing::Span::current(),
+            );
+            result
+        }
+        .instrument(span)
+        .await
```
messiaen: This seems ugly to me. Can we have a helper so there is not so much code just to add a span?  It's hard to tell what is really going on which is just a simple `driver.fulfill_request`

Is there not some rusty way of instrumenting the method.  Like using an raii object.

messiaen (dismissed):

crates/libsy/src/lib.rs
eric-liu-nvidia: the label now lives on Context (with an algorithm() accessor), run_stream stamps it, and the Driver carries the stamped Context

crates/libsy/src/lib.rs
eric-liu-nvidia: docs now state libsy.llm_call measures fulfillment as the algorithm observes it (host queueing/serving included), and a new libsy.client_call span wraps the actual provider call in the one place libsy performs it (run's default-client serve), with outcome/error recorded; hosts serving over their own transport are pointed at spanning their LlmClient equivalently.

crates/libsy/src/observability.rs
eric-liu-nvidia: addressed within tracing's constraint that span field names are static

crates/libsy/src/lib.rs
eric-liu-nvidia: new observe_llm_call / observe_run future-wrapping helpers own all span/timing/record plumbing; call_llm is now three lines that read as "fulfill_request, observed". (A literal RAII guard doesn't work here — tracing's entered-guards can't be held across .awaits — so the future adapter is the rusty idiom.)

messiaen: I still think this could be cleaner. Could we use something like https://crates.io/crates/tracing
and https://github.com/tokio-rs/tracing-opentelemetry

That way it could look something like 
```rust
use tracing::{info, info_span};

fn process_data() {
    let span = info_span!("process_data");
    let _guard = span.enter(); // RAII guard acquired

    info!("Doing work inside the OTel-backed span");
} // _guard is dropped here, automatically exiting the span
```

messiaen: nvm I missed the await spans + async thing.

messiaen: Looks like we can use `#[instrument] in the methods though.

https://github.com/tokio-rs/tracing#in-asynchronous-code

eric-liu-nvidia: span.enter() guards can't be held across .await — the tracing docs call this out directly: "Holding the drop guard returned by Span::enter across .await points will result in incorrect traces." (https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code) For async code, the docs prescribe Future::instrument, which is exactly what observe_llm_call uses — we're already on tracing + the tracing-opentelemetry bridge. I tried the guard version to be sure: our own span test then shows client_call mis-parented under llm_call (thread-state leakage from the suspended task). 1e7dbde adds an assertion that fails red under the guard pattern. 

Also, the call site is now three lines after the earlier hoist — this snippet is outdated.

grahamking:
I think this PR tries to do too much. Should we start by defining what metrics libsy should report? That's probably a conversation for Slack.

Then we could add one in a small PR (just the new dependency and the one metric line), and go from there.

crates/libsy/src/lib.rs
eric-liu-nvidia: Good call — #[instrument] works for the method spans, and it came out nicer: call_llm and serve now declare their spans as attributes, and the three span-constructor helpers are gone (net −23 lines). Two caveats: the libsy.run span keeps explicit Future::instrument because it wraps the spawned run task rather than a method body, and the timing/metrics recording stays in-body since it needs the result and elapsed time. The parentage tests pass unchanged, so the traces are identical. Done in af6f871.

linj-glitch (approved):

PR #70, opened 2026-07-15
feat(launch): startup-timing breakdown and faster AUTO backend-format detection

elyasmnvidian:
## What this adds

Tooling and a fix for slow `switchyard launch claude` startup, tied to SWITCH-872.

**TL;DR — `AUTO` backend-format detection is the startup cost.** With `format: auto` (the Claude Code / Codex launcher default), the resolver probes the upstream *live* before the agent starts to detect the wire format — Chat Completions first, then `/v1/messages`, then `/v1/responses`. Each probe is a round-trip, so a slow endpoint (e.g. a cold NVCF/Azure deployment on Inference Hub) adds real wall-clock latency to startup. This PR reduces the worst case and adds the tooling to see it.

**1. An opt-in, per-probe startup-timing breakdown.** Enable it two ways — the `--startup-timing` flag or the `SWITCHYARD_STARTUP_TIMING=1` env var; if neither is set it does nothing and a normal launch pays zero. When on, the launcher prints to stderr, right before it spawns the agent, how long each startup phase took — each backend-format probe on its own line:

```
switchyard startup timing:
       2.0 ms  config + credentials resolved
       0.1 ms  chain init
    1174.0 ms  probe: /v1/chat/completions
       0.4 ms  chain assembled
      47.2 ms  proxy thread started
      69.5 ms  proxy health-ready
       0.3 ms  child agent spawned
  --------
    1293.5 ms  total (launch invoked -> child spawn)
```

A slow line names the slow phase — three slow `probe:` lines mean a slow upstream made AUTO stack timeouts. Works for both single-model (`--model`) and default classifier-routing launches.

**2. A fix for the phase that breakdown exposed.** Before this change, if the Chat Completions probe **timed out** (reachable-but-slow endpoint), the resolver treated the timeout like a 404 and **serially ran the `/v1/messages` and `/v1/responses` probes too**, each stacking another timeout — one slow endpoint became several seconds (measured up to ~7 s). Now a probe **timeout** is detected via the raised `httpx.TimeoutException` (not a wall-clock heuristic) and falls back to Chat Completions immediately. A genuine fast 404 still falls through to the other probes, so detection is unchanged for Anthropic-native / Responses-only endpoints. Worst case drops from ~three stacked timeouts to one.

**3. Docs.** `docs/architecture.md` now states that `AUTO` costs startup latency (a live probe per format) and that setting `format:` explicitly removes it — with a pointer to `--startup-timing` to measure it. Also corrected the AUTO decision-tree diagram, which showed `/v1/messages` probed first; the resolver actually probes `/v1/chat/completions` first (with the timeout short-circuit).

## What stays the same

- Explicit `format:` values still skip probing. Setting `format:` is the real fix for slow startup.
- `anthropic/` and `claude…` model prefixes still short-circuit to Anthropic with no probe.
- A fast 404 on Chat Completions still falls through to the `/v1/messages` and `/v1/responses` probes. Only a probe *timeout* changes behavior.

## Tests

`tests/test_backend_format_resolver.py`:

- A Chat Completions probe that raises `httpx.ReadTimeout` resolves to OPENAI without running the other two probes.
- `AUTO` records each probe as its own startup-timing mark.

The existing probe-order tests are unchanged. The `--startup-timing` output is debug-only diagnostic tooling, so it ships without its own test.

## Not included

- The one-time Python import (~1 s) runs before launch dispatch, so it is not in the breakdown. Measure it with `switchyard launch claude --model X --dry-run`. Trimming it is separate work.
- No format cache. A stale cached format would misroute silently, and the timeout fix already removes the slow-startup case a cache would paper over.


elyasmnvidian:
Thanks @coderabbitai — addressing both findings:

**`backend_format_resolver.py` (timeout via elapsed time, Major):** already fixed. The stopwatch heuristic was reviewed on an earlier commit; a later commit (`40d8a6d6`) removed the `time.perf_counter()` elapsed-time check entirely. The probe now re-raises `httpx.TimeoutException`, and `_resolve_auto` branches on the caught exception — exactly the distinct-timeout-signal approach suggested. No `import time` remains in the resolver.

**`claude_code_launcher.py` (move "child agent spawned" mark into process creation):** intentionally left where it is. The mark and `dump()` sit just before the handoff to `ShellTUI.run()` / `_supervise_claude_plain()` on purpose: once the child grabs the terminal (interactive TUI), stderr output would interleave with the child's UI. The timing is meant to capture switchyard's own pre-spawn overhead, which ends exactly at this handoff — `exec` of the child is not switchyard-attributable. Adding a mark after process creation would print into the running child's terminal.

switchyard/cli/launchers/claude_code_launcher.py
```diff
@@ -386,6 +390,8 @@ def _run_claude_with_switchyard(
             env_overrides.get("ANTHROPIC_CUSTOM_MODEL_OPTION"),
             env_overrides.get("CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY"),
         )
+        startup_timing.mark("child agent spawned")
+        startup_timing.dump()
```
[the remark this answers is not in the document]
elyasmnvidian: Resolving on the merits (also discussed in the main thread, which you confirmed): the mark and `dump()` are deliberately placed at the pre-spawn handoff, immediately before `ShellTUI.run()` / `_supervise_claude_plain()` take over the terminal. Dumping after `exec` would interleave the timing output with the child's interactive TUI, and `exec` of the child isn't switchyard-attributable overhead — the measurement is scoped to switchyard's own pre-spawn work, which ends here. Keeping the label.

switchyard/lib/startup_timing.py
```diff
@@ -0,0 +1,85 @@
...
+is one launch per ``switchyard`` invocation.
+"""
+
+import os
+import sys
+import time
+from typing import TextIO
+
+_ENV_VAR = "SWITCHYARD_STARTUP_TIMING"
+_FALSEY = {"", "0", "false", "no"}
```
grahamking: Given that you wrote this and are the only user, does it really need to take such a broad range of values? `"1"` or `"0"` would work too right? And simplify the code.

switchyard/lib/startup_timing.py
```diff
@@ -0,0 +1,85 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Opt-in granular startup timing for ``switchyard launch``.
```
grahamking: Does this need a 25 line intro comment? Could you human-summarize it to just what a human reader would need to know, and not be able to quickly understand from the code?

switchyard/lib/startup_timing.py
```diff
@@ -0,0 +1,85 @@
...
+    """Turn timing on for this process (called by the ``--startup-timing`` flag)."""
+    global _forced
+    _forced = True
+
+
+def enabled() -> bool:
+    """Return whether timing is on — the CLI flag or a truthy env var enables it."""
+    if _forced:
+        return True
+    return os.environ.get(_ENV_VAR, "").strip().lower() not in _FALSEY
```
grahamking: This is called every time we `mark`. Env vars don't change during program run.

How about `_forced` becomes `enabled`, which defaults to the value of the env var?

```
enabled = os.environ.get("SWITCHYARD_STARTUP_TIMING") # also don't need const

def enable():
   global enabled
   enabled = True

def mark(..):
   global enabled
   if enabled:
     ...
```

tests/test_startup_timing.py
```diff
@@ -0,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
grahamking: I'm not sure this needs tests. It's debug tooling. 

AI is always far too enthusiastic when writing tests. Have a look at what's in here, it's crazy, testing falsey values and so on. I would delete the whole file.

grahamking (approved):

switchyard/lib/startup_timing.py
elyasmnvidian: Agreed, simplified. The gate is now just `os.environ.get("SWITCHYARD_STARTUP_TIMING", "0") != "0"` — on unless the var is unset or `0`. Dropped the whole truthy/falsey value set.

switchyard/lib/startup_timing.py
elyasmnvidian: Done — cut it to four lines: what it is, how to turn it on, what `mark()` and `dump()` do, and that Python's import time isn't counted. Everything else was re-derivable from the code.

switchyard/lib/startup_timing.py
elyasmnvidian: Done exactly as suggested. `_forced` is gone; `enabled` is now a module-level bool set once from the env var at import, `mark()` just checks `if enabled`, and the `--startup-timing` flag flips it via `enable()`. No env read per `mark()`.

tests/test_startup_timing.py
elyasmnvidian: Agreed — deleted the whole file. The startup-timing output is debug tooling, and the falsey-value and reset tests weren't earning their keep.

PR #77, opened 2026-07-16
feat(switchyard-translation): add neutral-IR encode/decode helpers

messiaen:
## Summary

Adds convenience helpers to `switchyard-translation` for translating between provider wire
formats and the neutral IR without threading an engine/policy through every call.

**Base:** `libsy-lib-proto-cleanup`

## Changes

- **`helpers.rs`** — free functions over a shared default engine: `decode_request`,
  `encode_request`, `decode_buffered_response`, `encode_buffered_response`, `encode_stream`,
  and a byte-based `decode_stream` (`Stream<Item = Result<Vec<u8>, _>>` → IR chunks — no HTTP
  client types), plus the `RawEventStream` type for unframed wire-event streams.
- **`sse.rs`** — a minimal SSE frame parser backing `decode_stream`.
- **`lib.rs`** — wires in the two modules and re-exports the helpers from the crate root.
- **`Cargo.toml`** — adds `futures` + `async-stream` for the streaming helpers.

## Tests
Unit tests for every helper (request/response round-trips, buffered & stream encode/decode)
and the SSE parser. Full workspace builds clean, clippy at zero.

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **New Features**
  * Added convenient request and response translation between supported wire formats.
  * Added streaming translation for real-time responses, including SSE event encoding and decoding.
  * Added support for buffering partial stream frames and recognizing stream completion markers.
  * Preserved model information when encoding buffered responses.
* **Bug Fixes**
  * Improved handling of incomplete UTF-8 data and non-standard streams that omit a final separator.

### DIFF
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -10,7 +10,8 @@ use std::pin::Pin;
 use std::sync::LazyLock;
 
 use async_stream::try_stream;
-use futures::{Stream, StreamExt};
+use futures::io::AsyncBufReadExt;
+use futures::{Stream, StreamExt, TryStreamExt};
 use serde_json::Value;
 
 use crate::sse;
@@ -39,7 +40,7 @@ pub fn encode_request(request: &LlmRequest, wire_format: WireFormat) -> Result<V
 }
 
 /// Decodes a buffered `wire_format` response body into the neutral aggregate.
-pub fn decode_buffered_response(body: &Value, wire_format: WireFormat) -> Result<AggLlmResponse> {
+pub fn decode_aggregated_response(body: &Value, wire_format: WireFormat) -> Result<AggLlmResponse> {
     Ok(DEFAULT_TRANSLATION_ENGINE
         .decode_response(wire_format, body, &DEFAULT_TRANSLATION_POLICY)?
         .response)
@@ -47,7 +48,7 @@ pub fn decode_buffered_response(body: &Value, wire_format: WireFormat) -> Result
 
 /// Encodes a buffered aggregate into `wire_format`'s JSON body, stamping
 /// `requested_model` over the upstream id so the caller sees the model it asked for.
-pub fn encode_buffered_response(
+pub fn encode_aggregated_response(
     agg: &AggLlmResponse,
     wire_format: WireFormat,
     requested_model: Option<&str>,
@@ -80,28 +81,19 @@ pub fn encode_stream(
     chunks: LlmResponseStream,
     target: WireFormat,
     requested_model: Option<String>,
-) -> RawEventStream {
+) -> std::result::Result<RawEventStream, Box<dyn std::error::Error + Send + Sync>> {
     // The target is always a built-in wire format, so this lookup cannot fail; a
-    // failure surfaces as a single error item rather than a panic.
-    let codec = match StreamCodecRegistry::with_builtins().codec(target) {
-        Ok(codec) => codec,
-        Err(error) => {
-            let message = error.to_string();
-            return Box::pin(futures::stream::once(async move {
-                Err(Box::new(std::io::Error::other(message))
-                    as Box<dyn std::error::Error + Send + Sync>)
-            }));
-        }
+    // failure returns as an `Err` rather than a panic.
+    let codec = StreamCodecRegistry::with_builtins().codec(target)?;
+
+    let mut state = StreamTranslationState {
+        target: Some(target.into()),
+        target_model: requested_model,
+        ..Default::default()
     };
+    let mut chunks = chunks;
 
     let events = try_stream! {
-        let mut state = StreamTranslationState {
-            target: Some(target.into()),
-            target_model: requested_model,
-            ..Default::default()
-        };
-
-        let mut chunks = chunks;
         while let Some(item) = chunks.next().await {
             let chunk = item?;
             for value in codec.encode_event(&mut state, chunk) {
@@ -113,7 +105,7 @@ pub fn encode_stream(
         }
     };
 
-    Box::pin(events)
+    Ok(Box::pin(events))
 }
 
 /// Decodes a byte stream of `source`-format SSE frames into neutral IR chunks.
@@ -122,7 +114,10 @@ pub fn encode_stream(
 /// transport's body stream into `Stream<Item = Result<Vec<u8>, _>>`. Frames are
 /// buffered across chunks (a partial frame waits for its boundary); the source
 /// stream codec is resolved once and reused for every frame.
-pub fn decode_stream<S>(bytes: S, source: WireFormat) -> LlmResponseStream
+pub fn decode_stream<S>(
+    bytes: S,
+    source: WireFormat,
+) -> std::result::Result<LlmResponseStream, Box<dyn std::error::Error + Send + Sync>>
 where
     S: Stream<Item = std::result::Result<Vec<u8>, Box<dyn std::error::Error + Send + Sync>>>
         + Send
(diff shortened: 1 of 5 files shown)


crates/switchyard-translation/src/helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+) -> RawEventStream {
+    // The target is always a built-in wire format, so this lookup cannot fail; a
+    // failure surfaces as a single error item rather than a panic.
+    let codec = match StreamCodecRegistry::with_builtins().codec(target) {
+        Ok(codec) => codec,
+        Err(error) => {
+            let message = error.to_string();
+            return Box::pin(futures::stream::once(async move {
+                Err(Box::new(std::io::Error::other(message))
+                    as Box<dyn std::error::Error + Send + Sync>)
```
grahamking: I've seen us use this construction elsewhere. It's a bit awkward.

Could we return `Result<RawEventStream>` instead?

grahamking: That would allow the one-liner `let codec = StreamCodecRegistry..codec(target)?;`

crates/switchyard-translation/src/helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+            }));
+        }
+    };
+
+    let events = try_stream! {
+        let mut state = StreamTranslationState {
+            target: Some(target.into()),
+            target_model: requested_model,
+            ..Default::default()
+        };
```
grahamking: Nit, but I like to do as much as possible outside the async stuff. I think we could create this before calling `try_stream!`.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+{
+    let marker = sse::done_marker(source);
+    // The source is always a built-in wire format, so this lookup cannot fail; a
+    // failure still surfaces as a single error item rather than a panic.
+    let codec = match StreamCodecRegistry::with_builtins().codec(source) {
+        Ok(codec) => codec,
+        Err(error) => {
+            let boxed: Box<dyn std::error::Error + Send + Sync> = error.to_string().into();
+            return Box::pin(futures::stream::once(async move { Err(boxed) }));
+        }
```
grahamking: Same here, we could return `Result<LlmResponseStream>`. That allows the caller to handle any errors before getting into the stream machinery.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+        let request = decode_request(WireFormat::OpenAiChat, &body).unwrap();
+        assert_eq!(request.model.as_deref(), Some("gpt"));
+
+        let encoded = encode_request(&request, WireFormat::OpenAiChat).unwrap();
+        assert_eq!(encoded["model"], "gpt");
+        assert_eq!(encoded["messages"][0]["content"], "hi");
+    }
+
+    #[test]
+    fn buffered_response_round_trips_and_restamps_model() {
```
grahamking: Is this a superset of the test above? Meaning we can remove that one.

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+pub(crate) type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// The outcome of parsing one SSE frame's data lines.
+pub(crate) enum ParsedSseFrame {
+    /// Frame contained a JSON payload.
+    Json(Value),
+    /// Frame contained the provider's terminal marker (e.g. `[DONE]`).
+    Done,
+    /// Frame had no data payload.
+    Empty,
```
grahamking: Wouldn't we skip those at the parsing stage, and hence not need to emit anything?

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Boxed, thread-safe error carried by a streamed item.
+pub(crate) type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// The outcome of parsing one SSE frame's data lines.
+pub(crate) enum ParsedSseFrame {
+    /// Frame contained a JSON payload.
+    Json(Value),
+    /// Frame contained the provider's terminal marker (e.g. `[DONE]`).
+    Done,
```
grahamking: The SSE stream would stop at the point presumably, so maybe we don't need to emit this. We end the stream.

Without Done and Empty, we can drop the whole enum.

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+    };
+    let frame = decode_sse_frame(&buffer[..index])?;
+    buffer.drain(..index + separator_len);
+    Ok(Some(frame))
+}
+
+/// Decodes one raw SSE frame as UTF-8.
+pub(crate) fn decode_sse_frame(frame: &[u8]) -> Result<String, BoxError> {
+    std::str::from_utf8(frame)
+        .map(str::to_string)
```
grahamking: You could replace the whole function with `String::from_utf8_lossy(frame)`. It replaces invalid UTF-8 with that weird invalid symbol. I've not seen invalid UTF-8 since Python 2, and even then it was alway a mistake.

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+        .map_err(|error| format!("stream emitted invalid UTF-8 frame: {error}").into())
+}
+
+/// Returns whether the buffer has any non-whitespace bytes.
+pub(crate) fn has_non_whitespace_bytes(buffer: &[u8]) -> bool {
+    buffer.iter().any(|byte| !byte.is_ascii_whitespace())
```
grahamking: Caller should `buffer.trim_ascii()`. Then `buffer.is_empty()` would tell us what presumably the caller wants to know.

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+    done_marker: Option<&str>,
+) -> Result<ParsedSseFrame, BoxError> {
+    let mut data_lines = Vec::new();
+    for line in frame.lines() {
+        // SSE comments and blank lines do not contribute data.
+        if line.is_empty() || line.starts_with(':') {
+            continue;
+        }
+        if let Some(data) = line.strip_prefix("data:") {
+            data_lines.push(data.trim_start().to_string());
```
grahamking: Do `strip_prefix("data: ")` (extra white space after colon) then you don't need to `trim_start()`.

grahamking: Here's what I do: https://github.com/grahamking/ort/blob/master/src/input/prompt.rs#L487-L527

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+        return Ok(ParsedSseFrame::Done);
+    }
+
+    let value = serde_json::from_str::<Value>(&data)
+        .map_err(|error| format!("stream emitted invalid JSON frame: {error}"))?;
+    Ok(ParsedSseFrame::Json(value))
+}
+
+/// Finds the next CRLF or LF SSE frame boundary.
+fn next_sse_boundary(buffer: &[u8]) -> Option<(usize, usize)> {
```
grahamking: You might not need this. Wrap the input in a `BufReader` and then using `BufRead` trait call `.lines()` which iterates over the lines.

crates/switchyard-translation/src/sse.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+
+    use super::*;
+
+    // Multi-byte UTF-8 split across network chunks should wait for a full frame.
+    #[test]
+    fn buffers_incomplete_utf8_until_a_complete_sse_frame_arrives() -> Result<(), BoxError> {
```
grahamking: `BufReader` should handle your buffering needs.

nachiketb-nvidia:
nothing extra to add apart from Graham, just an additional pass checking if there are more idiomatic or standard ways of doing anything in this PR would be good

nachiketb-nvidia (approved):

crates/switchyard-translation/src/helpers.rs
messiaen: yep. fixed

crates/switchyard-translation/src/helpers.rs
messiaen: fixed

crates/switchyard-translation/src/helpers.rs
messiaen: fixed

crates/switchyard-translation/src/helpers.rs
messiaen: Restructured tests.

crates/switchyard-translation/src/sse.rs
messiaen: yep removed

crates/switchyard-translation/src/sse.rs
messiaen: removed

crates/switchyard-translation/src/sse.rs
messiaen: yep, method removed

crates/switchyard-translation/src/sse.rs
messiaen: method removed

crates/switchyard-translation/src/sse.rs
messiaen: yes fixed this up.

crates/switchyard-translation/src/sse.rs
messiaen: used `futures::io::BufReader`

crates/switchyard-translation/src/sse.rs
messiaen: test removed. with `futures::io::BufReader` it's a bit easier

crates/switchyard-translation/src/helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+                    sse::ParsedSseFrame::Json(value) => {
+                        for event in codec.decode_event(&mut state, &value) {
+                            yield event;
+                        }
+                    }
+                    sse::ParsedSseFrame::Done => return,
+                    sse::ParsedSseFrame::Empty => {}
+                }
+            }
+        }
```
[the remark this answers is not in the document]
messiaen: OBE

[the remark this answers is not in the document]
messiaen: `futures::io::BufReader` is use now. I don't think this is needed.

grahamking (approved):

PR #81, opened 2026-07-16
chore(libsy): move random routing algorithm into core

ayushag-nv:
## Why

`RandomAlgo` is a generally useful single-call libsy routing algorithm, but its implementation currently lives in the non-published `libsy-examples` crate. Downstream libsy consumers should be able to use uniform random routing without depending on the examples package.

## What

- Move `RandomAlgo`, `RandomDecision`, and their focused tests into the core `libsy` algorithms module.
- Re-export both public types from the `libsy` crate root.
- Update the streaming example and libsy documentation to consume the core export.
- Move the `rand` dependency from `libsy-examples` to `libsy`.

## How

The implementation keeps the existing uniform selection and `Driver::call_llm_target` flow unchanged. The source moves to `crates/libsy/src/algorithms/rand.rs`, preserving its original basename, while the examples crate retains only algorithms that are still examples rather than core library functionality. The implementation relies on the `Algorithm` trait's default `process_signals` method added on `main`.

## Where to Start Review

1. `crates/libsy/src/algorithms/rand.rs` — promoted implementation and tests.
2. `crates/libsy/src/lib.rs` and `crates/libsy/src/algorithms.rs` — public export wiring.
3. `crates/libsy-examples/examples/streaming_agent.rs` and `crates/libsy/README.md` — updated consumers and documentation.

## Test Plan

- `cargo test -p libsy -p libsy-examples` — 42 passed.
- `cargo fmt --all --check` — passed.
- `cargo clippy --workspace --all-targets -- -D warnings` — passed.
- `cargo test --workspace` — passed.
- `cargo doc -p libsy -p libsy-examples --no-deps` — passed with existing warnings in untouched documentation.
- Live provider tests were not run because this is a Rust-only source move with no provider or wire-format changes.

### DIFF
--- a/crates/libsy/src/algorithms/rand.rs
+++ b/crates/libsy/src/algorithms/rand.rs
@@ -5,8 +5,8 @@
 //!
 //! Selects one target from the set uniformly at random and calls it. This is the
 //! simplest possible routing algorithm and the reference for the single-call
-//! shape: one `driver.call_llm_target` inside `create_run_task`. (Weighted selection could
-//! be layered on later; the set defines the candidates.)
+//! shape: one `driver.call_llm_target` inside `create_run_task`. Weighted selection
+//! can be layered on later; the set defines the candidates.
 
 use std::error::Error;
 use std::sync::Arc;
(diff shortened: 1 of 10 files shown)


crates/libsy/src/algorithms/rand.rs
```diff
@@ -42,10 +44,10 @@ pub struct RandomOrchAlgo {
 }
 
 impl RandomOrchAlgo {
```
messiaen:
```suggestion
impl RandomAlgo {
```

nachiketb-nvidia (approved):

messiaen (approved):
One rename. Otherwise looks good

PR #82, opened 2026-07-16
feat(libsy): expose Python run streams

nachiketb-nvidia:
## Stack

1. #88 Protocol bindings
2. #89 Target bindings
3. #87 Algorithm execution
4. **#82 Host-driven run streams (this PR)**

Review this PR last, after #87.

## Why

Some Python hosts own their model transport and need to drive every model call themselves instead of using target clients through Algorithm.run().

## What

- Add Algorithm.run_stream().
- Bind RunStream as a single-consumer async iterator.
- Bind Step as structural CallLlm, Decision, and ReturnToAgent variants.
- Expose one-shot LlmCall.respond() and LlmCall.fail().
- Release stream state on terminal steps, errors, exhaustion, and aclose().
- Add focused no-op and random-routing stream tests.
- Add a runnable example for run(), run_stream(), and random routing.

## How

Each Python RunStream owns one Rust StepStream. Model-call promises move into LlmCall and can be fulfilled once. Terminal results tear down the underlying stream promptly.

## What to review

1. Step variant shape and Python structural matching.
2. One-shot LlmCall ownership.
3. Cancellation, terminal teardown, and aclose() behavior.
4. The example as the complete public workflow.

## Validation

- cargo fmt --all -- --check
- cargo clippy -p switchyard-py --all-targets -- -D warnings
- uv run pytest all four libsy binding test files (16 passed)
- uv run python examples/libsy.py

PR #87, opened 2026-07-17
feat(libsy): expose Python algorithm execution

nachiketb-nvidia:
## Stack

1. #88 Protocol bindings
2. #89 Target bindings
3. **#87 Algorithm execution (this PR)**
4. #82 Host-driven run streams

Review after #89.

## Why

Python hosts need to run Rust-owned algorithms using configured target clients without exposing algorithm implementation or subclassing hooks.

## What

- Add an opaque Algorithm handle with run().
- Add noop() and random(targets) factories.
- Add typed Decision values and the RoutedLlmClient protocol.
- Adapt Python async clients to the Rust RoutedLlmClient trait.
- Convert target sets into Rust-owned snapshots at factory construction.
- Add no-op, random-routing, and client-error tests.

## How

The adapter releases Python while awaiting the host client and maps the typed response back into the Rust protocol. Algorithm implementations remain Rust-owned and factories return one shared handle type.

## What to review

1. Opaque algorithm and factory API.
2. Python client adapter and error propagation.
3. Target snapshot semantics.
4. Context, request, decision, and response conversions.

## Validation

- cargo fmt --all -- --check
- cargo clippy -p switchyard-py --all-targets -- -D warnings
- uv run pytest the protocol, target, and algorithm binding tests (13 passed)


nachiketb-nvidia:
Superseded by the minimal dictionary-based Python API in #95.

PR #88, opened 2026-07-17
feat(libsy): expose typed Python protocol bindings

nachiketb-nvidia:
## Stack

1. **#88 Protocol bindings (this PR)**
2. #89 Target bindings
3. #87 Algorithm execution
4. #82 Host-driven run streams

Review and merge this PR first.

## Why

Python hosts need typed access to the neutral request and response values already owned by switchyard-protocol. Untyped dictionaries would duplicate the Rust model and lose structural matching at the language boundary.

## What

- Bind protocol request, response, content, tool, media, metadata, usage, and streaming chunk values.
- Expose the types under switchyard.libsy.protocol.
- Add lazy native-module facades and precise Python stubs.
- Bridge Python async response sources into Rust LlmResponseStream values.
- Add focused protocol and response-stream lifecycle tests.

## How

PyO3 wrappers own or clone the Rust values and convert only raw JSON extension fields. Complex enums use nested Python variants for structural pattern matching.

## What to review

1. Fidelity of the protocol fields and enum variants.
2. Ownership of aggregate versus streaming responses.
3. Python async source cleanup and single-consumer behavior.
4. Stub precision and namespace placement.

## Validation

- cargo fmt --all -- --check
- cargo clippy -p switchyard-py --all-targets -- -D warnings
- uv run pytest tests/test_libsy_protocol_bindings.py -v -o addopts= (8 passed)
- uv run ruff check on the changed Python files

## Summary by CodeRabbit

* **New Features**
  * Added Python access to a typed, neutral LLM protocol.
  * Added support for structured requests, responses, content blocks, tools, metadata, and serialization.
  * Added aggregate and asynchronous streaming response handling.
  * Added a dedicated `LibsyError` exception for orchestration failures.
  * Added comprehensive type definitions and public exports for protocol values.

* **Tests**
  * Added coverage for protocol typing, round-tripping, response variants, streaming, and stream cleanup.

PR #89, opened 2026-07-17
feat(libsy): expose Python target bindings

nachiketb-nvidia:
## Stack

1. #88 Protocol bindings
2. **#89 Target bindings (this PR)**
3. #87 Algorithm execution
4. #82 Host-driven run streams

Review after #88.

## Why

Python needs a reusable way to describe the semantic targets consumed by Rust-owned routing algorithms without binding any algorithm yet.

## What

- Bind LlmTarget and LlmTargetSet.
- Keep llm_client assignable before an algorithm snapshots the target set.
- Validate the presence of a callable call method.
- Expose only the target API and its stubs.
- Add focused target collection and mutation tests.

## How

The Python objects retain target and client references. Conversion into Rust targets is deliberately added in the next PR, where algorithm execution first consumes them.

## What to review

1. Target naming and lookup behavior.
2. Client assignment and validation.
3. The snapshot boundary reserved for algorithm construction.
4. Separation from the existing components LlmTarget.

## Validation

- cargo fmt --all -- --check
- cargo clippy -p switchyard-py --all-targets -- -D warnings
- uv run pytest tests/test_libsy_protocol_bindings.py tests/test_libsy_target_bindings.py -q -o addopts= (9 passed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1653,6 +1653,7 @@ dependencies = [
  "switchyard-components",
  "switchyard-components-v2",
  "switchyard-core",
+ "switchyard-protocol",
  "switchyard-server",
  "switchyard-translation",
  "tokio",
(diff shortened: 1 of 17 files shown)


nachiketb-nvidia:
Superseded by the minimal dictionary-based Python API in #95.

PR #92, opened 2026-07-18
feat(switchyard-protocol): add common utility to hydrate Metadata from headers

messiaen:
## Summary

Extracts the correlation `Metadata` type from `envelope.rs` into a dedicated
`metadata` module and gives it a `Metadata::from_headers` constructor that
normalizes the harness-specific HTTP headers emitted by Claude Code, Codex,
NeMo Relay, Dynamo, OpenCode, and OpenAI-style clients into a single neutral
correlation/routing envelope.

## Changes

- **New `metadata` module** (`crates/libsy-protocol/src/metadata.rs`): moves
  `Metadata` out of `envelope.rs`, re-exported from `lib.rs`. `envelope.rs` now
  imports it rather than defining it.
- **Expanded `Metadata`**: adds agent-lineage fields (`parent_agent_id`,
  `is_subagent`, `agent_kind`, `agent_role`, `task_kind`, `turn_id`,
  `session_final`) alongside the existing session/agent/task/correlation fields.
- **`Metadata::from_headers`**: case-insensitive header normalization with a
  documented precedence order — explicit `x-switchyard-*` overrides win,
  followed by each harness's native headers. Handles subagent inference from
  parent/child lineage headers (with an explicit `x-switchyard-is-subagent`
  override), Codex structured turn metadata (JSON), and Claude Code / OpenCode
  parent-session inference.
- **Named header constants**: all wire header names are defined as `*_HEADER`
  constants grouped by originating harness.

## Tests

12 unit tests covering Codex child/root metadata, Claude Code agent lineage and
parent inference, Relay/Dynamo correlation, OpenCode parent-session scoping,
`session_final` parsing, and case-insensitive lookup.

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **New Features**
  * Added standardized metadata extraction from HTTP headers.
  * Added support for session, agent, task, turn, correlation, and subagent information.
  * Added recognition of multiple provider-specific header formats, including JSON turn metadata.
  * Added case-insensitive header handling and support for explicit metadata overrides.

* **Bug Fixes**
  * Improved consistency when determining parent sessions, subagent status, and final-session state across supported integrations.

### DIFF
--- a/crates/protocol/src/envelope.rs
+++ b/crates/protocol/src/envelope.rs
@@ -4,7 +4,7 @@
 //! The request/response envelope: the normalized [`LlmRequest`]/[`LlmResponse`] paired
 //! with the original provider payload and correlation [`Metadata`].
 
-use crate::{LlmRequest, LlmResponse, WireFormat};
+use crate::{LlmRequest, LlmResponse, Metadata};
 use std::collections::HashMap;
 
 /// Per-request state threaded to an algorithm alongside a request. A placeholder
@@ -22,29 +22,6 @@ pub struct Context {
 /// grows without changing the orchestrator contract.
 #[derive(Clone)]
 pub struct Signals {}
-
-/// Correlation and routing metadata attached to a request or response.
-///
-/// All fields are optional; algorithms and observers use whichever are present
-/// (e.g. to key per-session state or emit correlated telemetry). `extra_metadata`
-/// is a free-form escape hatch for host-specific keys.
-#[derive(Clone)]
-pub struct Metadata {
-    /// Stable id for a multi-request session/conversation.
-    pub session_id: Option<String>,
-    /// Id of the agent making the request.
-    pub agent_id: Option<String>,
-    /// Id of the task the request belongs to.
-    pub task_id: Option<String>,
-    /// External trace/request id for joining with the host's telemetry.
-    pub correlation_id: Option<String>,
-    /// Arbitrary host-defined key/value metadata.
-    pub extra_metadata: Option<std::collections::BTreeMap<String, String>>,
-    /// HTTP headers to attach when forwarding the request/response, if any.
-    pub http_headers: Option<std::collections::BTreeMap<String, String>>,
-    /// The wire format the request/response was originally encoded in, if known.
-    pub wire_format: Option<WireFormat>,
-}
 
 /// A request an algorithm routes: the normalized [`LlmRequest`] plus the original
 /// provider payload and correlation [`Metadata`].
--- a/crates/protocol/src/metadata.rs
+++ b/crates/protocol/src/metadata.rs
@@ -1,0 +1,606 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Correlation metadata and harness header normalization.
+//!
+//! [`Metadata`] is the correlation/routing envelope carried alongside a request or
+//! response. [`Metadata::from_headers`] normalizes the harness-specific HTTP headers
+//! that Claude Code, Codex, NeMo Relay, and Dynamo attach into that neutral shape.
+
+use std::collections::BTreeMap;
+
+use crate::WireFormat;
+
+// Dotted paths addressing fields inside Codex's turn-metadata header JSON value.
+const CODEX_SESSION_ID_PATH: &str = "x-codex-turn-metadata.session_id";
+const CODEX_THREAD_ID_PATH: &str = "x-codex-turn-metadata.thread_id";
+const CODEX_PARENT_THREAD_ID_PATH: &str = "x-codex-turn-metadata.parent_thread_id";
+const CODEX_TURN_ID_PATH: &str = "x-codex-turn-metadata.turn_id";
+const CODEX_SUBAGENT_KIND_PATH: &str = "x-codex-turn-metadata.subagent_kind";
+const CODEX_AGENT_ROLE_PATH: &str = "x-codex-turn-metadata.agent_role";
+const CODEX_TASK_ID_PATH: &str = "x-codex-turn-metadata.task_id";
+const CODEX_TASK_KIND_PATH: &str = "x-codex-turn-metadata.task_kind";
+
+// Explicit Switchyard override headers; these take precedence over harness-native headers.
+const SWITCHYARD_SESSION_ID_HEADER: &str = "x-switchyard-session-id";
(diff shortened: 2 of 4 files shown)


crates/protocol/src/metadata.rs
```diff
@@ -0,0 +1,475 @@
...
+const OPENAI_SUBAGENT_HEADER: &str = "x-openai-subagent";
+
+// Claude Code agent-lineage headers.
+const CLAUDE_SESSION_ID_HEADER: &str = "x-claude-code-session-id";
+const CLAUDE_AGENT_ID_HEADER: &str = "x-claude-code-agent-id";
+const CLAUDE_PARENT_AGENT_ID_HEADER: &str = "x-claude-code-parent-agent-id";
+
+// OpenCode session headers.
+const OPENCODE_SESSION_ID_HEADER: &str = "x-session-id";
+const OPENCODE_PARENT_SESSION_ID_HEADER: &str = "x-parent-session-id";
```
ayushag-nv, line 59: I think this will become cumbersome to maintain if we want to add more harness support. Also, do we need these many switchyard headers ?? are they derivable from harness ones ??

ayushag-nv:
@messiaen The overall idea of PR looks good to me. Just have concerns on maintainability of these headers which I highlighted in the comment

crates/protocol/src/envelope.rs
```diff
@@ -4,7 +4,7 @@
 //! The request/response envelope: the normalized [`LlmRequest`]/[`LlmResponse`] paired
 //! with the original provider payload and correlation [`Metadata`].
 
-use crate::{LlmRequest, LlmResponse, WireFormat};
+use crate::{LlmRequest, LlmResponse, Metadata};
```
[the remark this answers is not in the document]
messiaen: Metadata is exported from lib.rs so we are good.

crates/protocol/src/metadata.rs
messiaen: the switchyard headers are used as lookup keys for metadata values, and they can be set directly.

crates/protocol/src/envelope.rs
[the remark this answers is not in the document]
messiaen: Users should `use switchyard_protocol::Metadata` so yes intentional

messiaen:
> @messiaen The overall idea of PR looks good to me. Just have concerns on maintainability of these headers which I highlighted in the comment

@ayushag-nv I refactored so that that the harness headers are now in a generic config.  So almost all the logic in from_headers is generic. The one exception is parsing `is_subagent` and `paraent_agent_id` which is too harness specific.

wdyt?

crates/protocol/src/metadata.rs
```diff
@@ -0,0 +1,606 @@
...
+
+// Dotted paths addressing fields inside Codex's turn-metadata header JSON value.
+const CODEX_SESSION_ID_PATH: &str = "x-codex-turn-metadata.session_id";
+const CODEX_THREAD_ID_PATH: &str = "x-codex-turn-metadata.thread_id";
+const CODEX_PARENT_THREAD_ID_PATH: &str = "x-codex-turn-metadata.parent_thread_id";
+const CODEX_TURN_ID_PATH: &str = "x-codex-turn-metadata.turn_id";
+const CODEX_SUBAGENT_KIND_PATH: &str = "x-codex-turn-metadata.subagent_kind";
+const CODEX_AGENT_ROLE_PATH: &str = "x-codex-turn-metadata.agent_role";
+const CODEX_TASK_ID_PATH: &str = "x-codex-turn-metadata.task_id";
+const CODEX_TASK_KIND_PATH: &str = "x-codex-turn-metadata.task_kind";
```
ayushag-nv, line 22: just a suggestion: if any of these are not to be used or provided by other harnesses, we can remove them for now. Can always add back whenever necessary. If you feel they will be used, then we can keep them.

crates/protocol/src/metadata.rs
```diff
@@ -0,0 +1,606 @@
...
+const SWITCHYARD_AGENT_ID_HEADER: &str = "x-switchyard-agent-id";
+const SWITCHYARD_PARENT_AGENT_ID_HEADER: &str = "x-switchyard-parent-agent-id";
+const SWITCHYARD_IS_SUBAGENT_HEADER: &str = "x-switchyard-is-subagent";
+const SWITCHYARD_AGENT_KIND_HEADER: &str = "x-switchyard-agent-kind";
+const SWITCHYARD_AGENT_ROLE_HEADER: &str = "x-switchyard-agent-role";
+const SWITCHYARD_TASK_ID_HEADER: &str = "x-switchyard-task-id";
+const SWITCHYARD_TASK_KIND_HEADER: &str = "x-switchyard-task-kind";
+const SWITCHYARD_TURN_ID_HEADER: &str = "x-switchyard-turn-id";
+const SWITCHYARD_REQUEST_ID_HEADER: &str = "x-switchyard-request-id";
+const SWITCHYARD_SESSION_FINAL_HEADER: &str = "x-switchyard-session-final";
```
ayushag-nv, line 35: same here if we need all, then keep them otherwise keep selected.

crates/protocol/src/metadata.rs
```diff
@@ -0,0 +1,606 @@
...
+}
+
+/// Returns `(parent_agent_id, is_subagent)` from the headers
+fn parse_sub_agent(headers: &BTreeMap<String, String>) -> (Option<String>, bool) {
+    let mut is_subagent = false;
+    let sy_is_sub_agent = header(headers, SWITCHYARD_IS_SUBAGENT_HEADER).and_then(parse_bool);
+    let mut parent: Option<String> = sy_header(headers, SWITCHYARD_PARENT_AGENT_ID_HEADER);
+    if let Some(is_sub) = sy_is_sub_agent {
+        if parent.is_some() {
+            return (parent, is_sub);
```
ayushag-nv, line 240: This might be a bug. If switchyard sub agent header is false or not set, but harness level headers are set, then is_subagent will be false. and in the from_headers in metadata, we are just directly consuming this and not going via HEADER_CONFIG

ayushag-nv (approved):
Left some minor comments. Rest looks good.

crates/protocol/src/metadata.rs
messiaen: We can change this, but this behavior is by design.

when both switchyard sub agent and switchyard parent headers are present we use those values.

crates/protocol/src/metadata.rs
messiaen: These are all used.

crates/protocol/src/metadata.rs
messiaen: all used

PR #96, opened 2026-07-20
fix(launch): tier backend formats come from the preset, not launcher overrides

ryan-lempka:
## Summary

The launcher route builder rebuilt every deterministic tier with caller-supplied `backend_format` / `strong_backend_format` params, discarding the preset's per-tier formats. The preset's strong tier is a Claude model, so `launch codex` and `launch openclaw` forced it onto OpenAI format, which strips `cache_control` and silently disables prompt caching.

The preset now owns tier formats: strong is `AUTO` (resolves to Anthropic by model prefix; probes otherwise), weak and classifier stay `OPENAI`. Both override params are deleted. `launch claude` is unchanged (it already overrode to `AUTO`); `switchyard serve` is untouched (YAML formats come from the config schema, not the preset). Launchers never exposed a user-facing format flag, so no capability is removed; explicit per-tier formats remain available via `--routing-profiles`.

## How tested

- [x] `uv run pytest tests/` green (1998 passed; builder tests now assert formats are preset-owned)
- [x] New e2e regression guard `tests/test_launch_preset_cache_e2e.py`: real builder output, real profile build, real HTTP upstream — asserts the strong tier's wire body carries `cache_control` ephemeral markers on the zero-flag launcher path, and the weak tier's does not.

## Checklist

- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (none added)
- [x] README / `--help` updated if customer-facing surface changed. (no surface change)

### DIFF
--- a/switchyard/cli/routing/route_builder.py
+++ b/switchyard/cli/routing/route_builder.py
@@ -90,9 +90,7 @@ def build_deterministic_routing_config(
     classifier_model: str | None,
     profile_name: str | None,
     classifier_min_confidence: float | None,
-    backend_format: BackendFormat,
     timeout: float | None,
-    strong_backend_format: BackendFormat | None = None,
 ) -> DeterministicRoutingConfig:
     """Layer LLM-as-classifier CLI overrides on top of the shipping preset.
 
@@ -116,10 +114,6 @@ def build_deterministic_routing_config(
             / ``coding_agent`` / ``openclaw``).
         classifier_min_confidence: User override for the tier selector's
             confidence floor.
-        backend_format: Wire format for the weak + classifier tiers (and the
-            strong tier when ``strong_backend_format`` is ``None``).
-        strong_backend_format: Strong-tier wire format override; ``AUTO``
-            probes the backend for Anthropic Messages support at runtime.
         timeout: Per-request HTTP timeout for the tier backends
             (seconds).
     """
@@ -158,7 +152,7 @@ def build_deterministic_routing_config(
     strong_target = LlmTarget(
         id="strong",
         model=strong_model,
-        format=strong_backend_format or backend_format,
+        format=preset.strong.format,
         api_key=primary.api_key,
         base_url=primary.base_url,
         timeout_secs=timeout,
@@ -166,7 +160,7 @@ def build_deterministic_routing_config(
     weak_target = LlmTarget(
         id="weak",
         model=weak_model,
-        format=backend_format,
+        format=preset.weak.format,
         api_key=weak.api_key,
         base_url=weak.base_url,
         timeout_secs=timeout,
@@ -174,7 +168,7 @@ def build_deterministic_routing_config(
     classifier_target = LlmTarget(
         id="classifier",
         model=classifier_model_resolved,
-        format=backend_format,
+        format=preset.classifier.format,
         api_key=primary.api_key,
         base_url=primary.base_url,
         timeout_secs=preset.classifier_timeout_s,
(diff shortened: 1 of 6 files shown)


switchyard/cli/routing/route_builder.py
```diff
@@ -158,23 +152,23 @@ def build_deterministic_routing_config(
     strong_target = LlmTarget(
         id="strong",
         model=strong_model,
-        format=strong_backend_format or backend_format,
+        format=preset.strong.format,
```
elyasmnvidian, line 155: Small heads-up on the override case. When someone runs `launch codex --model <a non-Claude model>`, the strong tier now takes `format=preset.strong.format`, which is `AUTO`. Before this change it was a fixed `OPENAI`. `AUTO` makes one `/v1/chat/completions` probe to the gateway at startup; on the NVIDIA gateway that probe answers yes quickly, so the model still ends up on OpenAI format — same result, one extra round-trip at launch. Not a correctness problem, just flagging that override launches now pay for a probe they used to skip. Fine to leave as-is.

tests/test_launch_preset_cache_e2e.py
```diff
@@ -0,0 +1,121 @@
...
+    CAPTURED.clear()
+    request = ChatRequest.openai_chat({
+        "model": "inbound",
+        "messages": [
+            {"role": "system", "content": "You are a coding agent."},
+            {"role": "user", "content": "hello"},
+        ],
+    })
+    ctx = ProxyContext()
+    response = await backend._backends["strong"].call(ctx, request)
```
elyasmnvidian, line 106: This calls the strong backend directly, so the test skips the classifier and router that decide which tier a request goes to. It proves that once the strong tier is picked, it sends Anthropic Messages with `cache_control` — exactly the bug you're guarding against, so this is good. It does not prove that a normal incoming request actually gets routed to the strong tier. If you want the guard to cover the full path, send the request through the deterministic routing backend instead of reaching into `_backends["strong"]`. Optional — the current test already locks in the format fix.

elyasmnvidian (approved):
This fixes a real problem: `launch codex` and `launch openclaw` were sending the Claude strong model over OpenAI Chat Completions, which drops `cache_control` and quietly turns off prompt caching. Moving the format onto the preset (strong = `AUTO`) fixes it, and `launch claude` still behaves exactly as before. I checked the resolver: `anthropic/claude-opus-4.7` matches the `anthropic/` prefix, so `AUTO` picks Anthropic Messages with no startup probe and keeps the cache markers. Small, clean change. GitHub now shows it merges cleanly into `main` after the rebase.

One thing to confirm before merge. It lives in a file this PR does not change, so I can't leave it on a line: `switchyard/lib/profiles/plan_execute_presets.py` line 76 has the same bug you're fixing here. The plan-execute planner is `anthropic/claude-opus-4.6` (a Claude model) but is pinned to `format=BackendFormat.OPENAI`, so the plan-execute path drops `cache_control` on its planner exactly like codex/openclaw did. If `switchyard launch ... --plan-execute` ships with these defaults, its planner loses prompt caching too. Can you confirm whether that path ships? If it does, the planner tier needs the same `AUTO` treatment.

While tracing that: `build_plan_execute_config` in `route_builder.py` still takes a `backend_format` argument, but nothing in the app calls it anymore (only tests do), so that argument and its docstring now describe code that never runs. Worth deleting in a follow-up.

Two smaller notes are on the lines themselves.

switchyard/cli/routing/route_builder.py
ryan-lempka: This PR does not touch the `--model` path. Behavior there is unaffected.

ayushag-nv:
@ryan-lempka seems like touching deprectaed code. Closing this.

PR #13, opened 2026-07-01
docs(local-models): add vllm hidden-state serving notes

sabhatinas:
## Summary

- Add internal local vLLM serving notes for hidden-state extraction.
- Document both the Docker launch path and the equivalent direct `vllm serve` command.
- Include a one-request safetensors smoke test and local serving troubleshooting notes.

## Validation

- `uv run --only-group docs mkdocs build --strict`

## Summary by CodeRabbit

* **Documentation**
  * Added a new guide for serving a local vLLM model with hidden-state extraction enabled.
  * Included setup options for Docker and direct CLI usage, plus example launch commands.
  * Added steps to verify output files and inspect saved hidden states, along with troubleshooting tips.
  * Updated the documentation navigation to point to the new guide.

PR #97, opened 2026-07-20
feat(prefill-probe): add learned prefill complexity router

mich-xu:
## What

Adds a learned `prefill-probe` profile that routes requests between configured weak and strong completion targets using prompt hidden states from a separate vLLM probe model.

The profile:

- Requests all-layer prompt hidden states from a dedicated Qwen3.6 vLLM endpoint.
- Mean-pools each layer across prompt tokens and concatenates the layer vectors.
- Applies an artifact-provided scaler and PCA-200 projection.
- Runs a five-member shared-trunk ensemble that produces four correctness probabilities.
- Uses only the two checkpoint heads mapped to the configured weak and strong targets.
- Selects between those targets using a cost-aware utility policy controlled by `lambda`.
- Caches successful decisions by resolved probe input and falls back to strong on probe or scoring failures.
- Loads the learned checkpoint from an external inference-artifact directory; no training or online learning occurs in Switchyard.

This also adds a Terminal-Bench Terminus 2 adapter that supplies the exact raw task instruction through the private `_switchyard_prefill_probe_input` field. The profile uses that value for scoring and caching, then removes it before calling the selected completion backend.

A sample local profile and vLLM deployment documentation are included.

## Why

The trained complexity router expects prompt-only hidden-state features, while Terminal-Bench's standard Terminus 2 agent wraps the raw task instruction with its command protocol and terminal state.

Passing the raw task instruction explicitly reproduces the checkpoint's expected representation without changing the conversation received by the selected completion model.

The probe model is intentionally independent of the completion targets. It only needs to match the encoder metadata stored in the inference artifact.

Closes #

## How tested

- [x] `uv run mypy switchyard` clean: 162 source files checked
- [x] `uv run pytest tests/ -v -m "not integration"` green: 1,978 passed, 12 skipped, 43 deselected
- [x] Python SPDX header gate
- [x] `mkdocs build --strict`
- [x] Manual Qwen3.6 hidden-state probe startup and artifact inference
- [x] Manual three-task verifier smoke: all three tasks completed successfully
- [x] Two 89-task routing runs with explicit raw input:
  - `lambda = 0.7`: 46 weak, 43 strong, 0 probe failures
  - `lambda = 0.8`: 6 weak, 83 strong, 0 probe failures

The full live runs encountered upstream provider rate limits, but all 89 routing decisions completed in each run and no probe failures were observed.

## Checklist

## Notes for reviewers

This is a stacked PR based on and targeting #13. It can be retargeted to `main` after #13 merges.

The router artifact is deployment data and is not packaged or downloaded by Switchyard. At startup, the profile validates the artifact metadata and tensor shapes, including encoder identity, layer count, hidden width, PCA dimension, output heads, and ensemble architecture.

The learned profile uses a fixed public threshold of `0.5`: score `1.0` selects weak and score `0.0` selects strong. Routing should be tuned through `lambda`, not through an additional confidence threshold.

The private Terminal-Bench probe-input field is rejected when malformed and is always removed before target execution.


benchmark/routing-profiles/prefill-probe-local.yaml
```diff
@@ -0,0 +1,50 @@
...
+    strong: strong
+    strong_checkpoint_head: opus-4.7
+    weak: weak
+    weak_checkpoint_head: nemotron-3-super
+    hidden_states_dir: /tmp/switchyard-hidden-states
+    inference_artifact_dir: ${PREFILL_ROUTER_ARTIFACT_DIR}
+    routing_policy:
+      type: cost-aware
+      lambda: 0.5
+      weak_cost: 0
```
sabhatinas, line 49: Curious what this field translates to in the algorithm.

benchmark/routing-profiles/prefill-probe-local.yaml
```diff
@@ -0,0 +1,50 @@
...
+profiles:
+  router:
+    type: prefill-probe
+    probe: probe
+    strong: strong
+    strong_checkpoint_head: opus-4.7
+    weak: weak
+    weak_checkpoint_head: nemotron-3-super
+    hidden_states_dir: /tmp/switchyard-hidden-states
+    inference_artifact_dir: ${PREFILL_ROUTER_ARTIFACT_DIR}
```
sabhatinas: are hidden_states_dir and inference_artifact_dir the same thing?

docs/vllm-serve-hidden-state.md
```diff
@@ -1,61 +1,203 @@
...
+The `prefill-probe` profile uses a non-empty string value for both scoring and
+the decision-cache key. It removes the field before calling the selected
+completion target, so that target still receives the unmodified Terminus
+conversation. A present empty or non-string value is rejected as an invalid
+request instead of silently falling back to the wrapped message.
+
+Run Harbor from the repository root and select the adapter by import path:
+
+```bash
+uv run --no-sync harbor run \
```
sabhatinas: we have baseline benchmarking scripts in the repo to run closed book setup. can you update docs to use that instead ?

sabhatinas (changes requested):
LGTM overall, we just need e2e testing done using the benchmarks.

benchmark/routing-profiles/prefill-probe-local.yaml
mich-xu: hidden_states_dir is where the VLLM endpoint keeps the extracted prefills, inference_artifact_dir is the actual router checkpoint location

docs/vllm-serve-hidden-state.md
mich-xu: addressed this in latest 2 commits, also moved the task extraction to be internal.

sabhatinas (approved):

benchmark/routing-profiles/prefill-probe-local.yaml
sabhatinas: NIT: rename to checkpoint dir ?

mich-xu:
## Live smoke test — single terminal bench 2.1 task

Tested commit `83d55a7` end-to-end against real completion backends.

### Configuration

| Setting | Value |
|---|---|
| Dataset | `terminal-bench/terminal-bench-2-1` |
| Task | `terminal-bench/kv-store-grpc` |
| Agent | stock `terminus-2` |
| Custom agent/import path | none |
| Lambda | `0.75` |
| Weak/strong costs | `0 / 1` |
| Probe | `Qwen/Qwen3.6-35B-A3B` |

### Result

| Metric | Result |
|---|---|
| Completed trials | 1 |
| Errored trials | 0 |
| Retries | 0 |
| Verifier reward | `1.0` |
| Probe failures | 0 |
| Selected tier | strong |
| Weak probability | `0.2228` |
| Strong probability | `0.6167` |
| Weak utility | `0.1671` |
| Strong utility | `0.2126` |
| Public routing score | `0.0` |

The stock Terminus request was accepted without a custom agent adapter. Switchyard
internally extracted the raw task instruction, probed it once, selected the strong
target, and reused the cached decision for subsequent turns.

benchmark/routing-profiles/prefill-probe-local.yaml
mich-xu: renamed

PR #103, opened 2026-07-21
fix: resolve open OSS QA bugs across libsy, v2 server, CLI, and docs

elyasmnvidian:
Clears the open QA-filed bugs against the OSS builds, grouped by area. Verified against current `main`: `cargo build`, `cargo clippy -D warnings`, strict rustdoc (`RUSTDOCFLAGS="-D warnings" cargo doc`), `cargo test`, `ruff`, `mypy --strict`, and the full non-integration `pytest` suite (one unrelated pre-existing failure in `test_verify.py::test_secrets_file_wins_when_no_cli_or_env`, which also fails on clean `main` on this dev box).

## libsy (Rust)
- The LLM classifier drains a streamed classifier response before reading its score, so a streaming classifier target routes on its real verdict instead of always failing open to the strong tier.
- Invalid-but-parseable scores (NaN, ±inf, out of `[0,1]`) now fail open to strong instead of being treated as a below-threshold verdict and silently downgraded to weak.
- `cargo package -p switchyard-libsy` succeeds: the `switchyard-protocol` path dependency now carries a version.
- Strict rustdoc passes: fixed intra-doc links to private items and redundant explicit link targets.

## v2 server / routing (Rust)
- `GET /v1/models` reports real `tool_calling` and `context_window` capability metadata instead of `null`.
- A context-window overflow returns HTTP 400 (`context_length_exceeded`) instead of a 500 `server_error`, on all inbound APIs.
- The v2 random-routing profile accepts `fallback_target_on_evict` and retries once against the fallback target on a context-window overflow, matching `stage_router` / `llm_routing`.

## CLI
- `launch <agent> -- <args>` forwards documented harness args after a single `--` (the separator is only stripped before the subcommand, so the doubled `--` workaround is no longer needed).
- Zero-config `launch` against a non-OpenRouter provider fails fast with guidance instead of launching the OpenRouter-only default trio against an incompatible endpoint.
- `launch --dry-run` reports a routing-profiles bundle as `route: bundle` (with the bundle path and route ids) instead of `route: single`.
- `configure --list-models` and `configure` honor the saved `default_provider` (the `--provider` flag no longer shadows it), and non-interactive `configure` resolves an API key from the environment/secrets instead of requiring `--api-key`.
- `serve --config` with a duplicate model id prints a one-line diagnostic instead of a raw traceback.
- `serve` with no config points at `--config` first (the v2 path), with `--routing-profiles` as the legacy option.
- Local readiness probes (verify round-trip, launcher health) bypass env proxies, so a configured `HTTP_PROXY` no longer breaks the `127.0.0.1` checks.

## Stats
- noop routes appear in `/v1/routing/stats` (request count and tokens under the `noop` model) instead of being invisible.

## Docs / examples
- Corrected the AUTO backend-format probe order (Chat Completions probed first), the `[all]` extra description, and the in-repo example `route.yaml`.
- Documented that saved routing-profile bundles keep `${VAR}` references literal, so the referenced env vars must be set at `serve` / `launch` time.
- Removed stale references to the removed `SwitchyardRecipes` API.

## Already correct on `main` (no change here)
The HTTP error-envelope consolidation, fail_open classifier observability, and v2 Codex Responses token stats were already fixed on `main`; they only need QA verification.

## Summary by CodeRabbit

- **New Features**
  - Random routing can retry with a fallback target after context-window eviction.
  - Model listings now report tool-calling support and inferred context-window sizes.
  - No-op requests can contribute to routing statistics.
  - Dry runs now display routing-profile bundle details.

- **Bug Fixes**
  - Context-length failures now return clear HTTP 400 client errors.
  - Streaming classifier scores are validated reliably.
  - Local health checks bypass configured proxies.
  - CLI configuration preserves saved providers and can use environment API keys.

- **Documentation**
  - Updated installation, configuration, profile, launcher, and routing guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,9 +34,10 @@ traffic that sits between client applications and LLM backends.
 - **Observability** — Prometheus `/metrics`, a JSON `/v1/stats`
   (`/v1/routing/stats` alias), and per-request cost/token/latency stats. See
   [Metrics Reference](docs/METRICS_REFERENCE.md).
-- **Python library** — `SwitchyardRecipes` (`passthrough_recipe`,
-  `random_routing_recipe`, `stage_router_recipe`, `deterministic_routing_recipe`,
-  …) and typed `ChatRequest` / `ChatResponse` containers for in-process use.
+- **Python library** — `ProfileSwitchyard` driven by typed profile configs
+  (`PassthroughProfileConfig`, `RandomRoutingProfileConfig`,
+  `StageRouterProfileConfig`, …) and typed `ChatRequest` / `ChatResponse`
+  containers for in-process use.
 - **Rust core** (PyO3) — chain execution, the latency-aware router, and the
   tool-result signal collector are implemented in Rust and re-exported to
   Python.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -46,7 +46,7 @@ switchyard/
 │   ├── lib/                       # Core library
 │   │   ├── roles.py               # RequestProcessor, LLMBackend, ResponseProcessor ABCs
 │   │   ├── switchyard.py          # Switchyard chain executor
-│   │   ├── recipes.py             # SwitchyardRecipes (passthrough, random_routing, …)
+│   │   ├── profiles/              # Profile configs/runtimes (passthrough, random_routing, …)
 │   │   ├── proxy_context.py       # ProxyContext — per-request state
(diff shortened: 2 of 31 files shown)


elyasmnvidian:
Superseded — split into per-area PRs for independent review/merge: #104 (libsy), #105 (v2 server), #106 (v2 random-routing), #107 (cli), #108 (launch), #109 (stats), #110 (docs).

PR #105, opened 2026-07-21
chore(server): guard that /v1/models reports null model capabilities

elyasmnvidian:
## What

`GET /v1/models` leaves each model's `context_window` and `tool_calling` as `null`, and this PR keeps it that way on purpose, with a test that locks it in.

## Why

The `id` in `/v1/models` is a symbolic route name (like `random` or `smart`), not the upstream model behind it — so there's nothing to read a real context window or tool-calling flag from. Inferring one from the route name (or from a target) would be a guess, and a wrong one. The operator picks the models they route to and already knows their limits, so Switchyard doesn't need to advertise them.

This replaces the earlier version of this PR, which inferred the context window from the model id and broke once the id became a route name (flagged in review).

## Change

- A comment on `model_entry_json` explaining why the two fields are `null`.
- A test that serves a route named `random` over a big-context target and asserts both fields are `null` — written so any future attempt to infer a value (from the route name or the target) fails the test.
- `streaming` and the supported inbound formats stay populated; those are facts about this server, not the upstream model.

If someone later needs these reported, the operator can declare them in the server config and we report exactly what they set — no guessing.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -481,6 +481,10 @@ fn model_list_payload<'a>(models: impl IntoIterator<Item = &'a str>) -> Value {
     })
 }
 
+// One OpenAI-compatible `/v1/models` entry. `model` is a symbolic route/algorithm id
+// (e.g. "random"), not an upstream model, so the served model's context window and
+// tool-calling support are unknown here — report null rather than inferring from the id,
+// which would be wrong. `streaming` and the inbound formats are facts about this server.
 fn model_entry_json(model: &str) -> Value {
     json!({
         "id": model,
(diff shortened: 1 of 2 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -501,6 +501,27 @@ fn model_entry_json(model: &str) -> Value {
     })
 }
 
+/// Infers an advertised context window from a model-name string.
+///
+/// Uses the same known-model fragments as the Python `model_listing` table:
+/// the first matching substring wins, else a conservative 128k default.
+fn inferred_context_window(model: &str) -> u32 {
```
grahamking: Won't this be a symbolic / router name like `smart` or `random`? We don't advertise the upstream models any more.

PR #110, opened 2026-07-21
docs: correct AUTO probe order, extras, saved-bundle note, and stale API references

elyasmnvidian:
Docs and examples had several claims that no longer match the shipped package, so a new user copying from the README, INSTALLATION, or `examples/` would hit an ImportError or a config that won't load. This PR corrects them.

What changed:

- **`examples/route.yaml` now loads.** The `type: deterministic` (LLM-classifier) route was missing `fallback_target_on_evict`, which the route-bundle builder requires. Before this, building the table raised `RouteBundleConfigError: llm-classifier.fallback_target_on_evict must be a non-empty string`. Added `fallback_target_on_evict: strong`.
- **`examples/minimal.py` and the INSTALLATION library snippet use the current API.** `SwitchyardRecipes.passthrough_recipe(...)` was removed from the package, so the old example fails at import. Replaced with `ProfileSwitchyard(PassthroughProfileConfig(...).build())`. `switchyard.call()` now returns the response body directly, so the example reads `response['choices'][0]['message']['content']` instead of `response.body[...]`.
- **CHANGELOG "Python library" bullet** now names `ProfileSwitchyard` and the typed profile configs (`PassthroughProfileConfig`, `RandomRoutingProfileConfig`, `StageRouterProfileConfig`) instead of the removed `SwitchyardRecipes`.
- **DEVELOPMENT.md** points at `switchyard/lib/profiles/` instead of the deleted `recipes.py`.
- **`[all]` extra described correctly** in README and INSTALLATION: it pulls in `server, cli, tracing, intake, affinity-redis` (was "server, cli" / "tracing extras").
- **AUTO backend-format probe order** in `docs/cli_reference.md` fixed to match the resolver: probe `/v1/chat/completions` first (use `openai`), then `/v1/messages` (`anthropic`), then `/v1/responses` (`responses`), then fall back to `openai`.
- **`docs/cli_reference.md`** now notes that a saved routing-profile bundle keeps `${VAR}` references literal — the env vars must be present at `serve`/`launch` time, or Switchyard aborts with `missing environment variable(s): NAME`.

How to try it:
- `python examples/minimal.py` with `OPENAI_API_KEY` set now imports and runs; against the current package the old file raised `ImportError: cannot import name 'SwitchyardRecipes'`.
- Loading `examples/route.yaml` through the route-bundle builder no longer raises on the `llm-classifier` route.
Audit follow-up in this push: the CHANGELOG packaging line listed a `[gpu]` extra that pyproject.toml does not define and omitted `[tracing]`/`[intake]`/`[affinity-redis]`; the README `[all]` comment named `redis-affinity` (the real extra is `affinity-redis`); and CONTRIBUTING.md still pointed at the removed `switchyard/lib/recipes.py` (now `switchyard/lib/profiles/`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,14 +34,16 @@ traffic that sits between client applications and LLM backends.
 - **Observability** — Prometheus `/metrics`, a JSON `/v1/stats`
   (`/v1/routing/stats` alias), and per-request cost/token/latency stats. See
   [Metrics Reference](docs/METRICS_REFERENCE.md).
-- **Python library** — `SwitchyardRecipes` (`passthrough_recipe`,
-  `random_routing_recipe`, `stage_router_recipe`, `deterministic_routing_recipe`,
-  …) and typed `ChatRequest` / `ChatResponse` containers for in-process use.
+- **Python library** — `ProfileSwitchyard` driven by typed profile configs
+  (`PassthroughProfileConfig`, `RandomRoutingProfileConfig`,
+  `StageRouterProfileConfig`, …) and typed `ChatRequest` / `ChatResponse`
+  containers for in-process use.
 - **Rust core** (PyO3) — chain execution, the latency-aware router, and the
   tool-result signal collector are implemented in Rust and re-exported to
   Python.
 - **Packaging** — `pip install nemo-switchyard` with optional extras `[server]`,
-  `[cli]`, `[gpu]`, `[all]`. See [Installation](INSTALLATION.md).
+  `[cli]`, `[tracing]`, `[intake]`, `[affinity-redis]`, `[all]`. See
+  [Installation](INSTALLATION.md).
 
 ### Deprecated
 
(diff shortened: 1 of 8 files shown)


elyasmnvidian:
### Live before/after — the shipped example loads, and the extras match pyproject

**`examples/route.yaml`** — building the route bundle from the shipped example (`OPENROUTER_API_KEY` set to a dummy so env interpolation passes):

```
# main — the example fails to build
RouteBundleConfigError: llm-classifier.fallback_target_on_evict must be a non-empty string

# this branch — passes that validation and proceeds to build backends
passed the fallback validation — reached backend construction
```

The `type: deterministic` (llm-classifier) route requires `fallback_target_on_evict`; the example was missing it. Added `fallback_target_on_evict: strong`.

**Extras** — `[all]` in `pyproject.toml` resolves to `[server, cli, tracing, intake, affinity-redis]`. Confirmed and matched the docs:

- CHANGELOG listed a non-existent `[gpu]` extra and omitted `[tracing]`/`[intake]`/`[affinity-redis]` — now lists the real ones.
- README named `redis-affinity`; the actual extra is `affinity-redis` — fixed.
- CONTRIBUTING pointed at the removed `switchyard/lib/recipes.py` — now `switchyard/lib/profiles/`.

The AUTO probe-order correction and the `SwitchyardRecipes` removals are text/example-only (no runtime behavior to A/B).

grahamking (approved):

PR #114, opened 2026-07-22
fix(libsy): propagate errors and cancel the algorithm task in `run_stream`

messiaen:
## Summary

`Algorithm::run_stream` spawn `create_run_task`, but discards the `JoinHandle`.
This means that if `create_run_task` panics `runs_stream` will just silently end without passing those panics back as errors. This PR fixes that by chaining the primary step stream from the driver with an error stream created from `create_run_task` panics or other errors.

Second this PR adds a AbortGaurd so that if a caller of `run_stream` drops the stream, the task spawned in run_stream is aborted to shutdown any ongoing processing.

## Bugs fixed

- NVBUG: 6492419
- NVBUG: 6492420

## Tests

added unit tests for run_stream and run.

## Checklist

## Notes for reviewers

### DIFF
--- a/crates/libsy/src/core/algorithm.rs
+++ b/crates/libsy/src/core/algorithm.rs
@@ -296,14 +296,42 @@ pub trait Algorithm: Send + Sync + 'static {
         // the algorithm task, and keep one to emit the terminal step. The task blocks
         // publishing a step until the consumer pulls the previous one.
         let driver = Driver::new();
-        let stream = driver.stream();
-        tokio::spawn(async move {
-            let outcome = self
-                .create_run_task(ctx.clone(), driver.clone(), request)
-                .await;
-            let _ = driver.finish(ctx, outcome).await;
+        let task_driver = driver.clone();
+        let task_ctx = ctx.clone();
+        let stream = task_driver.stream();
+        let handle = tokio::spawn(async move {
+            self.create_run_task(task_ctx, task_driver, request).await
         });
-        Box::pin(stream)
+
+        // On its own task, await the algorithm's completion and emit the terminal step:
+        // a panic (the `JoinError`) becomes a failed `finish`, a normal return is
+        // forwarded as-is. Returns the `finish` result so a failed terminal emit surfaces
+        // on the stream rather than panicking an unobserved task.
+        let finish_driver = driver.clone();
+        let finish_ctx = ctx.clone();
+        let err_handle = tokio::spawn(async move {
+            let result = match handle.await {
+                Ok(response) => response,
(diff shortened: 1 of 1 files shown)


crates/libsy/src/core/algorithm.rs
```diff
@@ -285,37 +294,42 @@ pub trait Algorithm: Send + Sync + 'static {
...
+            .filter_map(|finish_result| async move { finish_result.err().map(Err) }),
+        );
+
+        let stream: StepStream = Box::pin(stream);
+        Box::pin(futures::stream::select(stream, tail).map(move |step| {
+            // link abort guard to stream
+            let _keep_alive = &abort_guard;
+            step
+        }))
+    }
```
grahamking, line 329: That's a lot of Box::pin.

I wonder if later we could rebuild this with much less async.

crates/libsy/src/core/algorithm.rs
```diff
@@ -285,37 +294,42 @@ pub trait Algorithm: Send + Sync + 'static {
...
+                finish_driver.finish(finish_ctx, result).await
+            })
+            .filter_map(|finish_result| async move { finish_result.err().map(Err) }),
+        );
+
+        let stream: StepStream = Box::pin(stream);
+        Box::pin(futures::stream::select(stream, tail).map(move |step| {
+            // link abort guard to stream
+            let _keep_alive = &abort_guard;
+            step
```
grahamking, line 327: Can you explain this part? `_keep_alive` is dropped at the end of `map`.

grahamking (approved):

crates/libsy/src/core/algorithm.rs
grahamking: I worry that libsy was born last week and it's already too complicated, with all the spawn, async move, pin and stuff. I don't have a good idea to simplify though.

crates/libsy/src/core/algorithm.rs
messiaen: This so that the `async move` captures `abort_guard`.

As I understand it  `&abort_guard` get dropped in each map step, but that is a noop. So `abort_drop` itself will be dropped when the stream is dropped since it's captured in the `async move`.

I think the alternative would be to have some `Guard` struct to wrap the stream and hold the `abort_guard`.

crates/libsy/src/core/algorithm.rs
messiaen: Yeah TBH I think this is one case were rust + async is making things harder than it needs to be.  It needs to be async.

Hopefully most of the complexity will be contained to `run_stream` and `run`, which shouldn't change much once it stabilizes.

PR #115, opened 2026-07-22
feat(server): load libsy algorithms from explicit TOML

nachiketb-nvidia:
## What

- add a strict, versioned TOML schema for server deployments
- construct a shared default LLM client or explicitly named clients, target-local backends, and libsy algorithms
- support noop, random, and LLM-classifier routes with OpenAI Chat, OpenAI Responses, and Anthropic targets
- replace the temporary random-route CLI flags with a required `--config` TOML file

## Why

The Rust server needs a clear deployment format now that it serves libsy directly. Keeping TOML parsing in `switchyard-server` avoids coupling libsy to a configuration format, while explicit tagged variants make supported clients, backends, and algorithms discoverable and reviewable.

## How

Configuration is deserialized with unknown-field rejection, validated up front, and built in a fixed order: clients, targets, then routes. Targets share the translating HTTP client when no client is specified; named clients remain available when isolation or a future client type requires one. API keys are referenced only through `api_key_env`, and target names are the exact upstream model IDs. Adding an algorithm requires one typed config variant, one construction match arm, and focused tests.

## What to review

- whether the initial TOML shape is explicit enough without adding abstraction
- default-client behavior, environment-only API-key handling, and target/client ownership
- validation behavior for references, duplicate targets, and classifier thresholds
- the end-to-end test that serves random and classifier routes from TOML

## Validation

- `cargo fmt --all --check`

Tracks [SWITCH-930](https://linear.app/nvidia/issue/SWITCH-930).

### DIFF
--- a/crates/switchyard-server/Cargo.toml
+++ b/crates/switchyard-server/Cargo.toml
@@ -19,6 +19,8 @@ rustls = { version = "0.23" }
 clap = { version = "4", features = ["derive", "env"] }
 futures-util = "0.3"
 libsy = { package = "switchyard-libsy", path = "../libsy" }
+serde = { version = "1", features = ["derive"] }
+yaml_serde = "0.10"
 switchyard-llm-client = { path = "../libsy-llm-client" }
 serde_json = "1"
 switchyard-translation = { path = "../switchyard-translation" }
--- a/crates/switchyard-server/Cargo.toml
+++ b/crates/switchyard-server/Cargo.toml
@@ -20,7 +20,7 @@ clap = { version = "4", features = ["derive", "env"] }
 futures-util = "0.3"
 libsy = { package = "switchyard-libsy", path = "../libsy" }
 serde = { version = "1", features = ["derive"] }
-toml = "0.8"
+toml = "1.1"
 switchyard-llm-client = { path = "../libsy-llm-client" }
 serde_json = "1"
 switchyard-translation = { path = "../switchyard-translation" }
--- a/crates/switchyard-server/Cargo.toml
+++ b/crates/switchyard-server/Cargo.toml
@@ -30,5 +30,6 @@ tracing = { version = "0.1", default-features = false, features = ["std"] }
 [dev-dependencies]
 async-trait = "0.1"
 http-body-util = "0.1"
+tempfile = "3"
 tokio = { version = "1", features = ["macros", "rt", "net"] }
 tower = { version = "0.5", features = ["util"] }
--- a/crates/switchyard-server/Cargo.toml
+++ b/crates/switchyard-server/Cargo.toml
@@ -20,7 +20,7 @@ clap = { version = "4", features = ["derive", "env"] }
 futures-util = "0.3"
 libsy = { package = "switchyard-libsy", path = "../libsy" }
 serde = { version = "1", features = ["derive"] }
-yaml_serde = "0.10"
+toml = "0.8"
 switchyard-llm-client = { path = "../libsy-llm-client" }
 serde_json = "1"
 switchyard-translation = { path = "../switchyard-translation" }
--- a/crates/switchyard-server/src/config.rs
+++ b/crates/switchyard-server/src/config.rs
@@ -1,0 +1,415 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Typed YAML configuration and explicit construction for the Rust server.
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs;
+use std::path::Path;
+use std::sync::Arc;
+
+use libsy::algorithms::{LlmClassifier, Noop, Random};
+use libsy::{Algorithm, LlmTarget, LlmTargetSet, RoutedLlmClient};
+use serde::Deserialize;
+use switchyard_llm_client::{Backend, HttpBackendConfig, ModelConfig, TranslatingLlmClient};
+
+use crate::{ServerError, ServerResult, ServerState};
+
+const SUPPORTED_SCHEMA_VERSION: u32 = 1;
+
+/// Loads a YAML deployment file and constructs the complete server state.
+pub fn load_server_state(path: impl AsRef<Path>) -> ServerResult<ServerState> {
+    let path = path.as_ref();
+    let yaml = fs::read_to_string(path).map_err(|error| {
+        ServerError::new(format!(
+            "failed to read server config {}: {error}",
+            path.display()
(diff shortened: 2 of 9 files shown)


crates/switchyard-server/tests/server.rs
```diff
@@ -144,6 +151,17 @@ async fn test_app(routes: &[(&str, &[&str])]) -> TestResult<(MockUpstream, Route
...
+    let path = std::env::temp_dir().join(format!(
+        "switchyard-server-config-{}.yaml",
+        std::process::id()
+    ));
+    fs::write(&path, yaml)?;
+    let state = load_server_state(&path);
+    fs::remove_file(path)?;
+    Ok(state?)
+}
+
```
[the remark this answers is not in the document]
grahamking: Ooh good catch! This is the kind of thing that causes intermittent CI failures that are hell to debug.

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,415 @@
...
+            ];
+            let target_set = resolve_targets(route_name, names, targets)?;
+            Ok(Arc::new(LlmClassifier::new(
+                classifier_target,
+                strong_target,
+                weak_target,
+                *threshold,
+                target_set,
+            )))
+        }
```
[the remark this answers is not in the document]
nachiketb-nvidia: we dont care about this, its upto the user. Overdoing validation often limits good software engineering in place

crates/switchyard-server/tests/server.rs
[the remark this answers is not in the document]
nachiketb-nvidia: Fixed in a915291b by using a unique `tempfile::NamedTempFile` with automatic cleanup, eliminating collisions between parallel test invocations.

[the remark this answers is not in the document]
nachiketb-nvidia: @coderabbitai close this thread if satisfied

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,469 @@
...
+}
+
+fn validate_id(kind: &str, name: &str, id: &str) -> ServerResult<()> {
+    if id.trim().is_empty() || id.trim() != id {
+        return Err(ServerError::new(format!(
+            "{kind} {name} id must be non-empty and have no surrounding whitespace"
+        )));
+    }
+    Ok(())
+}
```
grahamking, line 314: `validate_name` and `validate_id` are the same function. Maybe pass a label "id" or "name" for the error string?

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,469 @@
...
+    names: impl IntoIterator<Item = &'a str>,
+    targets: &BTreeMap<String, LlmTarget>,
+) -> ServerResult<LlmTargetSet> {
+    let resolved = names
+        .into_iter()
+        .map(|name| {
+            targets.get(name).cloned().ok_or_else(|| {
+                ServerError::new(format!(
+                    "route {route_name} references unknown target {name}"
+                ))
```
grahamking: Should this inner body call `resolve_target` function below? It's the same code.

Or you could leave this here, and delete `resolve_target`. Put the single item in an array before calling it.

crates/switchyard-server/Cargo.toml
```diff
@@ -19,6 +19,8 @@ rustls = { version = "0.23" }
 clap = { version = "4", features = ["derive", "env"] }
 futures-util = "0.3"
 libsy = { package = "switchyard-libsy", path = "../libsy" }
+serde = { version = "1", features = ["derive"] }
+toml = "1.1"
```
grahamking, line 23: Can you also upgrade this one to 1.1? Otherwise we have to include two versions of the same library.
```
./crates/switchyard-components-v2/Cargo.toml:26:toml = "0.8"
```

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,469 @@
...
+            .collect()
+    }
+}
+
+#[derive(Clone, Copy, Debug, Default, Deserialize)]
+#[serde(rename_all = "snake_case")]
+enum LlmClientType {
+    #[default]
+    Translating,
+}
```
grahamking: Do we really need this? Maybe introduce it when we have a second kind?

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,469 @@
...
+
+fn validate_name(kind: &str, name: &str) -> ServerResult<()> {
+    if name.trim().is_empty() || name.trim() != name {
+        return Err(ServerError::new(format!(
+            "{kind} name must be non-empty and have no surrounding whitespace"
+        )));
+    }
+    Ok(())
+}
```
grahamking: I'm pretty sure TOML won't give you surrounding whitespace (unless you quote it maybe). So likely delete this and not validate names, it's already done.

crates/switchyard-server/src/config.rs
```diff
@@ -0,0 +1,469 @@
...
+                    "llm client {client_name} api_key_env must not be empty"
+                )));
+            }
+            std::env::var(variable).map_err(|error| {
+                ServerError::new(format!(
+                    "llm client {client_name} could not read api_key_env {variable}: {error}"
+                ))
+            })
+        })
+        .transpose()?;
```
grahamking, line 217: Nit, but we should probably also `if api_key.empty() { bail }`. In the weird case it is set, but empty (bash script failure somehow).

crates/switchyard-server/src/config.rs
nachiketb-nvidia: addressed! Consolidated into one labeled validator.

crates/switchyard-server/src/config.rs
nachiketb-nvidia: addressed! resolve_targets now reuses resolve_target.

crates/switchyard-server/Cargo.toml
nachiketb-nvidia: addressed! Components-v2 now uses TOML 1.1.

crates/switchyard-server/src/config.rs
nachiketb-nvidia: addressed! Removed the premature client-type enum.

crates/switchyard-server/src/config.rs
nachiketb-nvidia: addressed! Quoted keys still need validation.

crates/switchyard-server/src/config.rs
nachiketb-nvidia: addressed! Empty API keys now fail.

grahamking (approved):

PR #117, opened 2026-07-22
fix(translation): forward prompt-cache token usage across format translation

BrianNewsom:
## What

Forward prompt-cache token usage (cached / cache-creation input tokens) faithfully through response translation, for every inbound↔outbound format pair, buffered and streaming.

## Why — two bugs today

Before this change the normalized `Usage` struct carried no cache fields, so cache-token detail was handled inconsistently and, in most paths, wrong:

1. **Buffered translation dropped cache tokens entirely.** None of `decode_openai_usage`, `decode_anthropic_usage`, or `decode_responses_usage` read `cached_tokens` / `cache_read_input_tokens` / `cache_creation_input_tokens`. Any non-streaming response translated between formats lost the cache breakdown. Additionally, `decode_anthropic_usage` computed `total_tokens = input + output`, which **undercounts** by the entire cache amount (Anthropic reports cache tokens *separately* from `input_tokens`).

2. **Streaming to Anthropic double-counted.** The one path that carried cache (via a stringly-typed `usage_extras` map) never adjusted `input_tokens`. For a 1000-token prompt with 800 cached, Anthropic output emitted `input_tokens: 1000` **and** `cache_read_input_tokens: 800` — so anything reconstructing the prompt total (`input + cache_read`, Anthropic's own model) got 1800.

## Design — one canonical representation

The providers disagree on what "input tokens" means:

- **Anthropic** `input_tokens` **excludes** cache; cache read/creation are separate siblings.
- **OpenAI Chat** `prompt_tokens` / **Responses** `input_tokens` **include** cache (`cached_tokens` is a subset).

`Usage` now stores the **cache-exclusive** input count plus an explicit, boxed `InputCacheUsage` detail (kept behind `Option<Box<…>>` so the common cache-free path stays a small allocation). This is the only lossless canonical form: decoders subtract cache from inclusive providers; encoders re-aggregate for inclusive providers and pass through for Anthropic. The old `usage_extras: BTreeMap<String,u64>` on the stream state is deleted in favor of typed fields.

## Per-format wire fidelity

Each encoded output matches that provider's own convention:

- **Anthropic** → `input_tokens` (cache-excluded) + `cache_read_input_tokens` + `cache_creation_input_tokens`, and `total_tokens` now includes cache.
- **OpenAI Responses** → `input_tokens` (inclusive) + `input_tokens_details.cached_tokens`.
- **OpenAI Chat** → `prompt_tokens` (inclusive) + `prompt_tokens_details.cached_tokens`.

**One deliberate deviation:** on Anthropic-backend → OpenAI-Chat-client, the codec also emits `prompt_tokens_details.cache_creation_tokens`, which OpenAI itself never returns (OpenAI chat has no cache-creation concept). This is an additive, non-standard key that preserves Anthropic's cache-creation accounting through translation; it only appears when there is Anthropic cache-creation data to carry (never on pure OpenAI passthrough). If strict OpenAI fidelity is preferred over round-trip preservation, drop that one field on encode.

## Tests — each fails on `main`, passes here

- `openai_chat_cache_usage_translates_to_responses_usage_details` — buffered, cache dropped on `main` (`Null` vs `{cached_tokens: 80}`).
- `openai_chat_cache_usage_translates_to_anthropic_usage_fields` — buffered, double-count on `main` (`input_tokens` 100 vs corrected 20).
- `openai_chat_stream_cache_usage_translates_to_responses_usage_details` — streaming, cache dropped on `main`.
- `test_response_translation_engine.py::test_cached_tokens_propagated` — updated to assert the corrected split (`input_tokens == 200`, was `1000`).
- `server.rs::all_inbound_formats_run_libsy…` — e2e: a single upstream OpenAI response with `cached_tokens: 7` now surfaces correctly in all three inbound formats.

Verified locally with rustc 1.96.1: `switchyard-translation` suite green, `switchyard-components` builds (protocol change is additive), and the e2e server test passes. Confirmed each new test fails against `main` with the exact diff shown above.

## Summary by CodeRabbit

* **New Features**
  * Added support for cached and cache-creation input token usage details.
  * Preserved cache token breakdowns when translating between OpenAI Chat, OpenAI Responses, and Anthropic formats.
  * Updated token totals to distinguish cached input from regular input.

* **Bug Fixes**
  * Corrected streamed and buffered usage reporting for cached tokens.
  * Prevented cached tokens from being counted twice in input totals.

* **Tests**
  * Added coverage for cached-token propagation across supported response and streaming formats.

### DIFF
--- a/crates/protocol/src/llm.rs
+++ b/crates/protocol/src/llm.rs
@@ -232,10 +232,60 @@ pub struct LlmRequest {
 /// Normalized token usage counts.
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct Usage {
+    /// Non-cached input tokens. Provider codecs normalize aggregate OpenAI
+    /// input counts by subtracting the cache detail fields.
     pub input_tokens: Option<u64>,
+    #[serde(flatten)]
+    pub cache: Option<Box<InputCacheUsage>>,
     pub output_tokens: Option<u64>,
     pub total_tokens: Option<u64>,
     pub reasoning_tokens: Option<u64>,
+}
+
+/// Optional cache-token detail kept out of the common, cache-free usage allocation.
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct InputCacheUsage {
+    pub cached_input_tokens: Option<u64>,
+    pub cache_creation_input_tokens: Option<u64>,
+}
+
+impl Usage {
+    pub fn cache_details(
+        cached_input_tokens: Option<u64>,
+        cache_creation_input_tokens: Option<u64>,
+    ) -> Option<Box<InputCacheUsage>> {
+        if cached_input_tokens.is_none() && cache_creation_input_tokens.is_none() {
+            return None;
+        }
+        Some(Box::new(InputCacheUsage {
+            cached_input_tokens,
+            cache_creation_input_tokens,
+        }))
+    }
+
+    pub fn cached_input_tokens(&self) -> Option<u64> {
+        self.cache
+            .as_ref()
+            .and_then(|cache| cache.cached_input_tokens)
+    }
+
+    pub fn cache_creation_input_tokens(&self) -> Option<u64> {
+        self.cache
+            .as_ref()
+            .and_then(|cache| cache.cache_creation_input_tokens)
+    }
+
+    pub fn set_cached_input_tokens(&mut self, value: u64) {
+        self.cache
+            .get_or_insert_with(|| Box::new(InputCacheUsage::default()))
+            .cached_input_tokens = Some(value);
+    }
+
+    pub fn set_cache_creation_input_tokens(&mut self, value: u64) {
+        self.cache
+            .get_or_insert_with(|| Box::new(InputCacheUsage::default()))
+            .cache_creation_input_tokens = Some(value);
+    }
 }
 
 /// Normalized reason a model stopped producing output.
--- a/crates/switchyard-translation/src/codecs/responses/stream.rs
+++ b/crates/switchyard-translation/src/codecs/responses/stream.rs
@@ -522,15 +522,22 @@ fn encode_responses_tool_delta(
 
 // Normalizes OpenAI Responses token usage fields.
 fn responses_usage(usage: &serde_json::Map<String, Value>) -> Usage {
-    let input_tokens = usage.get("input_tokens").and_then(Value::as_u64);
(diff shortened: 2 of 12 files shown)


crates/switchyard-translation/src/codecs/responses/stream.rs
```diff
@@ -522,10 +522,17 @@ fn encode_responses_tool_delta(
...
+    let cached_input_tokens = usage
+        .get("input_tokens_details")
+        .and_then(|details| details.get("cached_tokens"))
+        .and_then(Value::as_u64);
+    let input_tokens = aggregate_input_tokens
+        .map(|tokens| tokens.saturating_sub(cached_input_tokens.unwrap_or(0)));
     let output_tokens = usage.get("output_tokens").and_then(Value::as_u64);
     Usage {
         input_tokens,
+        cache: Usage::cache_details(cached_input_tokens, None),
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in 7475f4f0 — the streaming fallback now uses `aggregate_input_tokens` (matching the buffered `decode_responses_usage`), so cached tokens stay in `total_tokens` when upstream omits it. Added a regression test (`responses_stream_usage_without_total_keeps_cached_tokens_in_total`) that fails on the old code (total 25) and passes now (105).

crates/protocol/src/llm.rs
```diff
@@ -232,12 +232,62 @@ pub struct LlmRequest {
...
     pub output_tokens: Option<u64>,
     pub total_tokens: Option<u64>,
     pub reasoning_tokens: Option<u64>,
 }
 
+/// Optional cache-token detail kept out of the common, cache-free usage allocation.
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct InputCacheUsage {
+    pub cached_input_tokens: Option<u64>,
+    pub cache_creation_input_tokens: Option<u64>,
```
ayushag-nv, line 249: what is the diff b/w two ? asking for my knowledge only

ayushag-nv (approved):

crates/protocol/src/llm.rs
BrianNewsom: This is for Anthropic prompt_caching (https://platform.claude.com/docs/en/build-with-claude/prompt-caching) 

Basically you pay extra for cache_creation tokens to add them to the cache so they're cheaper later on (the cached_input_tokens)

ayushag-nv:
Nice PR. Thanks for working on this. Really appreciate fixing this

PR #120, opened 2026-07-23
feat(benchmark): net out retried attempts by trial id in task-level routing stats

ryan-lempka:
Extends the task-level routing stats (#116) so retried attempts can be netted out of cost, plus a token schema that matches the global stats file.

**What changed**

1. **Retry netting by trial.** The proxy sidecar stamps `x-switchyard-trial-id` (Harbor's trial name, read from the container env) alongside the existing task and session headers. At finalize, requests are grouped by trial, and within a trial the latest attempt is the one Harbor kept (it retries sequentially and discards earlier attempts). So tokens from retried-away attempts land in a `retries` section and can be subtracted. No timestamps, no result.json parsing.
2. **Six-field token schema.** Records and the rollup now carry `cached_tokens`, `cache_creation_tokens`, and `reasoning_tokens` alongside prompt/completion/total, matching `routing_stats_final.json` so the existing cost estimator prices either file. `model` is the served model id, so buckets reconcile with the global file.

**Schema, before → after** (per task in `routing_stats_by_task.json`):

```jsonc
// before
"build-system-task-ordering": {
  "requests": 24,
  "sessions": ["8c54..."],
  "models": [
    {"model": "gpt-5.4-mini", "tier": "strong", "calls": 12,
     "prompt_tokens": 349072, "completion_tokens": 22239, "total_tokens": 371311}
  ]
}

// after
"build-system-task-ordering": {
  "requests": 29,
  "n_retries": 1,
  "final": {
    "calls": 24,
    "totals": {"prompt_tokens": 349072, "cached_tokens": 300112, "cache_creation_tokens": 0,
               "completion_tokens": 22239, "reasoning_tokens": 18110, "total_tokens": 371311},
    "models": [
      {"model": "gpt-5.4-mini", "tier": "strong", "calls": 12,
       "prompt_tokens": 349072, "cached_tokens": 300112, "cache_creation_tokens": 0,
       "completion_tokens": 22239, "reasoning_tokens": 18110, "total_tokens": 371311}
    ]
  },
  "retries": { "calls": 5, "totals": { /* same six fields */ }, "models": [ /* ... */ ] }
}
```

`final` and `retries` share the same shape, so the cost estimator runs on either: net cost = price(final), retry overhead = price(retries).

**Tested end to end:** Ran a real Harbor benchmark, forced a task to time out and retry, and confirmed the stamped trial id matches Harbor's trial name and that the retried attempt's tokens land in `retries` while the kept attempt's stay in `final`.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -286,10 +286,16 @@ deterministic LLM-classifier routes, `server.classifier_prompts` records each ro
 prompt, prompt SHA-256, `max_request_chars`, and `recent_turn_window` for reproducibility. Direct
 runs mark routing stats as `not-requested`.
 
-Switchyard runs also write `routing_requests.jsonl` (one record per router request: task, session,
-selected model, tier, token usage) and roll it up into `routing_stats_by_task.json` at finalize.
-The task identity comes from the per-task proxy sidecar, which stamps `x-switchyard-intake-task`
-and a per-trial `proxy_x_session_id` on model-bound requests.
+Switchyard runs also write `routing_requests.jsonl` (one record per router request: task, trial,
+session, served model, tier, and the six token fields from the global schema) and roll it up into
+`routing_stats_by_task.json` at finalize. The per-task proxy sidecar stamps the identity headers on
+model-bound requests: `x-switchyard-intake-task`, `x-switchyard-trial-id` (Harbor's trial name), and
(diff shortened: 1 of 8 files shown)


ayushag-nv (approved):
Looks good. Aligned with what I was trying to do with logging per request routing decisions

Issue #121, opened 2026-07-23
vLLM chat_template_kwargs hint is injected for non-vLLM providers — HTTP 400 on Fireworks-served deepseek-v4 targets


himorishige:
## Summary

Routing deepseek-v4 models served by Fireworks AI through the Python deterministic (LLM-classifier) profile fails on every request with:

```
HTTP 400 {"error":{"message":"Extra inputs are not permitted, field: 'chat_template_kwargs'", ...}}
```

`chat_template_kwargs.enable_thinking=false` is a vLLM-side hint. Fireworks' OpenAI-compatible API validates request bodies strictly and rejects the unknown field. Two Python-side code paths inject the hint for provider-agnostic targets:

1. **`apply_deepseek_overrides()`** (`switchyard/lib/profiles/tier_target_builders.py`) injects it into **tier calls** for any target whose model id contains `deepseek-v4`, regardless of provider. It does not consult `model_accepts_reasoning_hint()`, the model-id compatibility check the LLM classifier already uses for the same hint.
2. **`model_accepts_reasoning_hint()`** (`switchyard/lib/processors/reasoning_hint.py`) deny-lists only `anthropic` / `bedrock` / `claude`, so the classifier's `disable_reasoning` auto-detect injects the hint for Fireworks-hosted classifier models (`accounts/fireworks/models/...`).

## Reproduction

`profiles.yaml`:

```yaml
endpoints:
  fireworks:
    base_url: https://api.fireworks.ai/inference/v1
    api_key: ${FIREWORKS_API_KEY}
targets:
  weak:
    endpoint: fireworks
    model: accounts/fireworks/models/deepseek-v4-flash
    format: openai
```

Any profile that routes a tier call through `build_tier_backend()` (deterministic / escalation router) to the `weak` target returns the 400 above for every request.

## Expected

Provider-specific defaults should not be sent to providers that reject them; at minimum the existing compatibility check should be honored consistently.

## Workaround

Declaring an explicit empty `extra_body: {}` on the target suppresses the tier-call injection (caller-supplied extras win). The classifier-side injection has no config-level escape in the v2 profile config path.

## Proposed fixes (PRs to follow)

- Gate the `apply_deepseek_overrides()` body default on `model_accepts_reasoning_hint()` — the same check the classifier applies.
- Add `fireworks` to the reasoning-hint deny list (Fireworks ids always carry the `accounts/fireworks/models/` namespace).

## Related observation (out of scope here)

[...]

PR #122, opened 2026-07-23
fix(profiles): gate DeepSeek thinking-off default on reasoning-hint compatibility

himorishige:
Fixes the tier-call half of #121.

## What

`apply_deepseek_overrides()` unconditionally injects the vLLM-only `chat_template_kwargs.enable_thinking=false` hint for any target whose model id contains `deepseek-v4`. Providers with strict request validation (observed: Fireworks AI serverless) reject the unknown field with HTTP 400 `Extra inputs are not permitted`, breaking every routed call for such tiers.

This PR gates the body default on `model_accepts_reasoning_hint()` — the same model-id compatibility check `LLMClassifierConfig.disable_reasoning` auto-detect already applies before sending the identical hint on classifier calls. The provider-neutral `X-Inference-Priority` header default is unchanged, and explicitly provided `extra_body` still wins as before.

Pairs with #123 (adds `fireworks` to the deny list); together they stop the 400 for Fireworks-served deepseek-v4 targets.

## Validation

- New unit test: deny-listed model id keeps `extra_body` untouched while still receiving the batch-priority header; existing `TestDeepSeekOverrides` cases unchanged and passing.
- `tests/test_deterministic_routing_profile.py` + `tests/test_reasoning_hint.py`: 47 passed. Two failures (`TestProfileStructure::test_shared_stats_accumulator`, `test_overflow_reroutes_to_custom_strong_id_through_full_profile`) are pre-existing on unmodified `main` in my environment (same class of stale-build failure noted in #118's validation notes).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for DeepSeek models that do not support reasoning hints.
  * Prevented unsupported “thinking off” request parameters that could cause HTTP 400 errors.
  * Preserved batch inference prioritization for affected requests.

* **Tests**
  * Added coverage to verify incompatible DeepSeek models receive valid request settings.

### DIFF
--- a/switchyard/lib/profiles/tier_target_builders.py
+++ b/switchyard/lib/profiles/tier_target_builders.py
@@ -20,6 +20,7 @@ from switchyard.lib.backends.multi_llm_backend import (
     build_native_backend,
     resolve_llm_target,
 )
+from switchyard.lib.processors.reasoning_hint import model_accepts_reasoning_hint
 from switchyard.lib.profiles.deterministic_routing_config import (
     DEFAULT_DETERMINISTIC_TIER_TIMEOUT_S,
 )
@@ -27,10 +28,18 @@ from switchyard.lib.roles import LLMBackend
 
 
 def apply_deepseek_overrides(target: LlmTarget) -> LlmTarget:
-    """Apply benchmark-specific DeepSeek extras without clobbering callers."""
+    """Apply benchmark-specific DeepSeek extras without clobbering callers.
+
+    The thinking-off default is a vLLM-side hint (``chat_template_kwargs``);
+    serving stacks on the :func:`model_accepts_reasoning_hint` deny list
+    reject the field outright (HTTP 400 ``Extra inputs are not permitted``),
+    so the default is gated on the same model-id check the LLM classifier
+    already uses for its own calls. The batch-priority header default is
+    provider-neutral and stays unconditional.
+    """
     default_body = (
         {"chat_template_kwargs": {"enable_thinking": False}}
-        if "deepseek-v4" in target.model
+        if "deepseek-v4" in target.model and model_accepts_reasoning_hint(target.model)
         else None
     )
     default_headers = (
--- a/tests/test_deterministic_routing_profile.py
+++ b/tests/test_deterministic_routing_profile.py
@@ -391,6 +391,21 @@ class TestDeepSeekOverrides:
         out = apply_deepseek_overrides(target)
         assert out.extra_body == {}
 
+    def test_hint_incompatible_model_id_skips_thinking_off(self) -> None:
+        """Deny-listed serving stacks reject the vLLM hint with HTTP 400,
+        so the thinking-off default must not be injected for their ids.
+        The provider-neutral batch-priority header still applies."""
+        target = LlmTarget(
+            id="weak",
+            model="bedrock/deepseek-ai/deepseek-v4-flash",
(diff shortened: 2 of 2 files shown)


switchyard/lib/profiles/tier_target_builders.py
```diff
@@ -20,17 +20,26 @@
...
+    serving stacks on the :func:`model_accepts_reasoning_hint` deny list
+    reject the field outright (HTTP 400 ``Extra inputs are not permitted``),
+    so the default is gated on the same model-id check the LLM classifier
+    already uses for its own calls. The batch-priority header default is
+    provider-neutral and stays unconditional.
+    """
     default_body = (
         {"chat_template_kwargs": {"enable_thinking": False}}
-        if "deepseek-v4" in target.model
+        if "deepseek-v4" in target.model and model_accepts_reasoning_hint(target.model)
```
elyasmnvidian, line 42: Nice — gating the tier default on `model_accepts_reasoning_hint` matches what the classifier already does, so both paths now agree on when to send `enable_thinking=false`. (For context: `apply_deepseek_overrides` sets default request options for DeepSeek tier calls, and this hint is one of them.)

One thing to flag for later: both the tier default (this function) and the classifier default now depend on that one check being complete. The check works by looking for provider names inside the model id, so a new backend that rejects the hint will still receive it until someone remembers to add its name to `_NO_REASONING_HINT_TAGS`. That is easy to forget, and the failure shows up as a hard 400 in production.

Concrete suggestion: consider deciding this on the target/endpoint config rather than the model name. For example, add an optional `accepts_reasoning_hint` flag to the target (`LlmTarget` in `llm_target.py`), default it from `model_accepts_reasoning_hint(model)` when the config doesn't set it, and have both `apply_deepseek_overrides` and the classifier presets read that flag. Then a strict backend can be marked incompatible in its own config, with no code change and no name list to keep in sync.

elyasmnvidian (changes requested):

switchyard/lib/profiles/tier_target_builders.py
himorishige: Thanks for the review — agreed on both points.

On the fragility: yes, with this PR both the tier default and the classifier default hang off `_NO_REASONING_HINT_TAGS`, and a new strict backend silently inherits the hint until someone adds its name. Your Azure example on #123 already demonstrates a miss.

On moving the decision to the target config — one implementation note from reading current `main`: `LlmTarget` is owned by the Rust extension (`switchyard_rust.components`); `llm_target.py` only hosts `coerce_llm_target()` and helpers. So an `accepts_reasoning_hint` field on the target itself means a Rust-side schema change, not just a Python default. Two Python-side shapes that stay closer to the current layout:

- extend `llm_target_with_runtime_defaults()`, which already maps model-name fragments (`nemotron-3-super`) to a thinking-off `extra_body` with an explicit-`extra_body` guard — so per-model request defaults keep living in one place; or
- the per-provider settings map you sketched on #123. I posted measurements there suggesting the map earns its keep: for Fireworks the right move is to *send a different knob* (`reasoning_effort: "none"`), not merely to suppress this one, and a boolean flag can never express that.

Happy to keep this PR as the minimal consistency fix (both paths agreeing on one check) and follow up with the config/map work as a separate PR — your call on which shape you'd want.

ayushag-nv:
@himorishige deprecated code. Please create a new one if you find the same bug in the latest code.

PR #124, opened 2026-07-23
fix(libsy): run should return when stream ends

messiaen:
## What

`Algorithm::run` should return when the stream has a final result canceling any in_flight `CallLlm` futures.

Before this MR `run` would continue running until all the futures were done even after the final result. 

Also simplified the logic a bit and remove the arbitrary in flight cap. (no thanks to claude)

## Why

- nvbug 6501297
- nvbug 6501299
- nvbug 6502618

## How tested

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent processing so results are returned promptly when multiple operations are in flight.
  * Ensured the first successful response is preserved without being overwritten by slower operations.
  * Prevented hangs when a concurrent operation never completes.
  * Terminal errors are now surfaced promptly, even during heavy concurrent activity.

### DIFF
--- a/crates/libsy/src/core/algorithm.rs
+++ b/crates/libsy/src/core/algorithm.rs
@@ -344,10 +344,6 @@ where
         ctx: Context<S>,
         request: Request,
     ) -> Result<(Vec<Arc<dyn Decision>>, Response), Box<dyn Error + Send + Sync>> {
-        // Serve up to this many offloaded calls concurrently, so an algorithm that
-        // fans out (e.g. an ensemble) isn't serialized on the client.
-        const MAX_CONCURRENT_CALLS: usize = 10;
-
         // Serve one offloaded call with its target's default client. A failed *model*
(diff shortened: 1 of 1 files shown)


grahamking (approved):

crates/libsy/src/core/algorithm.rs
```diff
@@ -372,31 +368,25 @@ where
...
-        let mut stream_open = true;
 
-        while stream_open || !in_flight.is_empty() {
+        loop {
             tokio::select! {
-                // Surface a failed serve as soon as it completes.
-                Some(result) = in_flight.next(), if !in_flight.is_empty() => result?,
-                // Pull the next step, unless the stream ended or we're at the cap.
-                step = stream.next(), if stream_open && in_flight.len() < MAX_CONCURRENT_CALLS => {
+                Some(result) = in_flight.next() => result?,
```
grahamking: `Some(Err(err))` maybe

PR #132, opened 2026-07-23
feat(server): expose libsy OpenTelemetry metrics

nachiketb-nvidia:
## What

- install one process-wide OpenTelemetry Prometheus exporter in `switchyard-server`
- expose libsy's existing counters and histograms through `GET /metrics`
- verify the endpoint before traffic and after a routed LLM request
- document the scrape path and the no-parallel-accumulator rule

## Why

Libsy already records algorithm runs, LLM calls, latency, token usage, outcomes, and routing
decisions through the global OpenTelemetry meter provider. The Rust server did not install a
provider or exporter, so those instruments were no-ops when served directly.

## Scope

This intentionally adds conventional Prometheus exposition only. It does not recreate the
legacy `/v1/stats` JSON schema, reset operations, cost calculations, or a second stats
accumulator. Existing limitations in libsy's instruments, including unavailable token usage for
streaming responses, remain unchanged.

## Review

- process-wide provider and registry initialization in `metrics.rs`
- `/metrics` content type and exporter encoding
- the integration test's OTel instrument and bounded-label assertions

## Validation

- `cargo fmt --all --check`
- compiled-binary smoke: TOML load, routed OpenAI request, and `/metrics` scrape
- skill frontmatter validation
- `git diff --check`


grahamking:
Will start fresh here https://github.com/NVIDIA-NeMo/Switchyard/pull/165

PR #133, opened 2026-07-23
test: add reusable end-to-end LLM mock server

nachiketb-nvidia:
## What

- add a private, non-publishable `switchyard-test-server` workspace crate
- serve OpenAI Chat Completions, OpenAI Responses, and Anthropic Messages
- support buffered and SSE streaming responses through the existing translation codecs
- provide a curated random response bank plus deterministic per-test overrides
- support model/header-based HTTP error injection and request capture
- expose both an in-process fixture and a standalone binary
- replace the inline mock implementation in `switchyard-server` integration tests

## Why

Switchyard integration tests need one reusable provider mock rather than independent,
partially compatible servers embedded in each test suite. Generating responses through
`switchyard-translation` keeps the mock aligned with the same wire formats the product
supports.

## Review

- confirm the fixture API is intentionally small and sufficient for server/client tests
- confirm the built-in response bank covers useful text shapes without making exact tests flaky
- confirm `publish = false` and dev-only consumption keep this out of production artifacts

## Validation

- `cargo test -p switchyard-test-server`
- `cargo test --workspace`
- `cargo fmt --all --check`
- standalone binary HTTP smoke against `/v1/chat/completions`

## Summary by CodeRabbit

* **New Features**
  * Added a reusable mock LLM server for integration testing across OpenAI and Anthropic-compatible endpoints.
  * Supports configurable responses, streaming output, request capture, and simulated errors.
  * Added a standalone mode for running the mock server locally.

* **Tests**
  * Updated server integration tests to use the shared mock server.
  * Added coverage for buffered and streaming responses, request recording, default responses, and error handling.

* **Documentation**
  * Added guidance recommending the shared mock server for Rust HTTP end-to-end tests.

### DIFF
--- a/.agents/skills/switchyard-testing-ci/SKILL.md
+++ b/.agents/skills/switchyard-testing-ci/SKILL.md
@@ -29,6 +29,7 @@ use `cargo test --workspace` before calling a broad Rust MR ready.
 | Mirror CI pytest with no live creds | `env -u OPENROUTER_API_KEY -u NVIDIA_API_KEY -u OPENAI_API_KEY -u ANTHROPIC_API_KEY uv run pytest tests/ -v -m "not integration"` |
 | Rust component crate change | `cargo fmt --all --check && cargo clippy --workspace --all-targets -- -D warnings && cargo test -p switchyard-components` |
 | Rust server crate change | `cargo fmt --all --check && cargo clippy --workspace --all-targets -- -D warnings && cargo test -p switchyard-server` |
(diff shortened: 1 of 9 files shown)


grahamking:
Could we try https://github.com/vidaiUK/VidaiMock or https://github.com/StacklokLabs/mockllm instead, to avoid maintaining it ourselves, and get much higher API coverage?

In my experience these tend to turn into a full time job.

PR #134, opened 2026-07-24
feat(protocol): Add an error type

grahamking:
The RoutedLlmClient implementations now have a concrete error `enum` to
return. This allows the caller to identify what happened. No more
`downcast`. Particularly, it allows the caller to identify timeouts
separately from other errors.

- Before RoutedLlmClient::call returned `Result<Response, Box<dyn Error + Send + Sync>>`
- Now it returns `Result<Response, LlmClientError>`

Which has these variants (see code for messages, this is edited) (note some more added in later commit):

```
pub enum LlmClientError {
    InvalidRequest { message: String },
    Configuration { message: String },
    Transport { #[source] source: BoxError},
    Timeout { #[source] source: BoxError },
    ContextWindowExceeded { .. },
    UpstreamHttp { status: u16, body: String, },
    InvalidResponse { #[source] source: BoxError },
    Other(#[from] BoxError),
}
```

`libsy-llm-client` uses this error type directly.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -215,13 +215,10 @@ impl TranslatingLlmClient {
                 .map_err(|error| LlmClientError::Stream(error.to_string()))?;
             LlmResponse::Stream(chunks)
         } else {
-            let body = http_response.json::<Value>().await.map_err(|error| {
-                if error.is_timeout() {
-                    LlmClientError::Timeout(error)
-                } else {
-                    LlmClientError::ResponseTranslation(format!("invalid upstream JSON: {error}"))
-                }
-            })?;
+            let body = http_response
+                .json::<Value>()
+                .await
+                .map_err(classify_json_response_error)?;
             let agg = decode_aggregated_response(&body, wire_format)
                 .map_err(|error| LlmClientError::ResponseTranslation(error.to_string()))?;
             LlmResponse::Agg(agg)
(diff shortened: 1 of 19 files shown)


grahamking:
@coderabbitai full review

crates/libsy-llm-client/src/client.rs
```diff
@@ -84,9 +85,9 @@ impl TranslatingLlmClient {
     /// Builds a client over the given [`ModelConfig`]s, with a fresh shared HTTP
     /// client and the built-in translation codecs.
     pub fn new(model_configs: &[ModelConfig]) -> Result<Self> {
-        let client = reqwest::Client::builder().build().map_err(|error| {
-            LlmClientError::Transport(format!("failed to build HTTP client: {error}"))
-        })?;
+        let client = reqwest::Client::builder()
+            .build()
+            .map_err(LlmClientError::Transport)?;
```
messiaen: Could we implement `Into` for out client error types / result type from reqwest error types, it could look like

```
let client = reqwest::Client::builder()
            .build().into()?
```

messiaen (approved):

crates/libsy-llm-client/src/client.rs
grahamking: Now that I merged the two errors, we can't. We don't want `protocol` to depend on `reqwest` so we can't put the `Into` impl in there. And `libsy-llm-client` can't have it, because it doesn't own either of the types.

PR #135, opened 2026-07-24
feat(libsy): stage router (signals only)

messiaen:
## What

- Implements stage_router tier_picker piece as a Classifier in libsy.
- No llm fallback yet, once we have llm_classifier processor / classifier, we can join them with fall_through
- bridge for switchyard-components

## How tested

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **New Features**
  * Added automatic stage routing that selects efficient or capable processing based on request signals.
  * Added escalation for critical conditions and de-escalation when work is settled.
  * Added confidence-based fallback classification when routing signals are inconclusive.
  * Added processing to capture recent tool activity for routing decisions.
* **API Updates**
  * Exposed stage-routing scores, dimensions, outcomes, and tier-selection utilities through the library interface.


crates/libsy/src/algorithms/util/stage_router.rs
```diff
@@ -0,0 +1,527 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
nachiketb-nvidia, line 3: can we make the structure src/algorithsm/<algorithm-name>/*.rs
with src/algorithsm/<algorithm-name>.rs

crates/switchyard-components/src/stage_router.rs
```diff
@@ -1,340 +1,15 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Stage-router scoring and tier selection — the shared routing core.
+//! Stage-router tier selection — thin re-export of libsy's decision core.
 //!
-//! Decides whether a coding-agent turn goes to the **capable** (strong) or
-//! **efficient** (weak) tier from its [`ToolResultSignal`]. Single source of
```
nachiketb-nvidia, line 7: we should also replace/delete these stage router files in `switchyard-components`

sabhatinas:
Will you also be updating docs?

messiaen:
> Will you also be updating docs?

My plan was to update the user docs when we have the full router working.

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -249,6 +256,30 @@ enum ToolCategory {
     Other,
 }
 
+#[derive(Debug, Clone)]
+pub struct ToolSignalProcessor {
+    pub recent_window: usize,
+}
+
```
nachiketb-nvidia, line 267: we should also replace any parsing happening here, to switchyard-protocol/IR
too much API specific parsing happening here

nachiketb-nvidia (changes requested):

crates/switchyard-components/src/stage_router.rs
messiaen: The only dependency is dimension_collector I think. Which I don't think we need to move.

tool_signal was moved previous PR

crates/libsy/src/algorithms/util/tool_signals.rs
messiaen: Yes, I agree. I think some of it is necessary.  @sabhatinas can clarify.

Let's do this is a separate PR.

crates/libsy/src/algorithms/util/stage_router.rs
messiaen: My goal was to separate the public stuff (algorithms) and the non-public (util).

The public part of this would be some instantiation `FallThrough` ... probably.

crates/libsy/src/algorithms/util/stage_router.rs
```diff
@@ -0,0 +1,527 @@
...
+                state.extra.insert(
+                    "default_target".to_string(),
+                    StateValue::String(target.to_string()),
+                );
+                let conf = (score + 1.0) / 2.0;
+                // TODO add the non-target to this score set?
+                Ok(Classification::Ambiguous(vec![Score {
+                    target: target.to_string(),
+                    confidence: conf,
+                }]))
```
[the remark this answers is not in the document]
messiaen: confidence is just the magnitude of the score.  The logic is correct here.

crates/libsy/src/algorithms/util/stage_router.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        // TOOD make mode and thresh configurable
+        let outcome = pick_tier(signal, PickerMode::EfficientFirst, 0.5);
+        match outcome {
+            PickOutcome::Resolved { tier, score, .. } => {
+                let target = match tier {
+                    Tier::Capable => "strong",
+                    Tier::Efficient => "weak",
+                };
+                let conf = (score + 1.0) / 2.0;
```
messiaen: just put source score here.

crates/libsy/src/algorithms/util/stage_router.rs
```diff
@@ -0,0 +1,527 @@
...
+/// the decision; corroboration across the two axes is what raises confidence.
+const SIGNAL_UNIT: f64 = 0.10;
+/// Critical severity forces the capable tier regardless of the scorer.
+const SEVERITY_CRITICAL: f32 = 1.0;
+
+/// Signed, fixed weights over the two axes. Error signals (`severity`,
+/// `spinning`, `exploring`) push toward capable (+); `production_intensity`
+/// pushes toward efficient (−). `severity` is normalised by its hard cap so it
+/// too contributes one `SIGNAL_UNIT` when maxed.
+const DEFAULT_WEIGHTS: &[(&str, f64)] = &[
```
sabhatinas: Non-blocking / craft: this reintroduces the string-keyed `DEFAULT_WEIGHTS` + `value()` lookup that PR #68 deliberately removed. A misspelled or dropped key silently returns `0.0` (the `_ => 0.0` arm in `value()`) with no compile error, so a future edit could zero out a term unnoticed. The four keys map 1:1 to the dimensions today, so the score is currently identical to the direct form — but it's a fragility regression vs the code it replaces.

Suggest collapsing back to the compile-checked direct expression:

```rust
let raw = SIGNAL_UNIT
```

Everything else is a clean, verified 1:1 port of the v2 scorer/picker — constants, `dimensions_from_signal`, `should_escalate`/`should_deescalate` (incl. the `compacted` override), and the `pick_tier` ladder all match, and the libsy stage_router tests pass. Nice port.

sabhatinas:
Pushed a commit to this branch (`3ba1804`):

- **Addressed the `DEFAULT_WEIGHTS` comment** — `score_signal` now uses the direct, compile-checked expression; dropped the string-keyed weights table and `CodingAgentDimensions::value()`.
- **Made the picker configurable** — `StageClassifier` was hardcoding `PickerMode::EfficientFirst, 0.5`; it now takes `mode` + `confidence_threshold` via `StageClassifier::new(...)` and is exported from `switchyard_libsy::stage_router`. Added a test proving the mode is honored (capable_first falls open to strong, efficient_first to weak).

All 9 libsy stage_router tests pass; cargo fmt + clippy clean. Parity with the v2 scorer/picker unchanged.

sabhatinas:
`9f671cc`: resolved the **no-signal TODO** using the new config — a turn with no tool activity yet now falls open to the picker's **configured default tier** (`efficient_first → weak`, `capable_first → strong`) and records `default_target`, instead of abstaining with an empty set. Matches the reference stage_router's `no_signal → default_tier` behavior. Updated the test accordingly (`classifier_defaults_without_tool_signals`); 9 libsy tests pass, clippy clean.

Left the other TODO (`add the non-target to this score set`) for you — that's a scoring-shape design call.

crates/libsy/src/algorithms/util/tool_signals.rs
sabhatinas: agreed as follow up

sabhatinas:
Design note for the follow-up parity work (tracked in SWITCH-1110), flagging it here since it may shape the `Classification` shape you land:

**`StageClassifier` currently drops the `DecisionSource`.** `score()` maps the outcome to `strong`/`weak`, but the *reason* (`Override` / `Dimensions` / `TestsPassed` / `FallOpen`) doesn't travel downstream. The **handoff-notes** feature needs it: the escalation note is direction-specific (weak→strong) **and** gated by `only_on_wrong_signal_escalation` — it should fire only when the escalation was signal-driven (`Dimensions`/`Override`), never on a `fall_open` default. Without the source surfaced (e.g. stashed in `state.extra` alongside `default_target`), a downstream handoff-note processor can't tell a real escalation from a defaulted one.

Not asking you to build handoff notes here — just consider surfacing the `DecisionSource` so the signals-only picker doesn't have to change shape later.

sabhatinas:
Addressed in `8c627b3` on this branch — `StageClassifier` now surfaces the `DecisionSource` in `state.extra["decision_source"]` (`override` / `dimensions` / `tests_passed` for resolved turns, `fall_open` for fall-open/no-signal). So the source **does** travel downstream now, and a handoff-note processor can gate `only_on_wrong_signal_escalation` on `decision_source ∈ {dimensions, override}`. Tests assert it. Retracting my earlier "drops the DecisionSource" note — it's handled.

sabhatinas (approved):

sabhatinas:
Addressed the CodeRabbit findings in `184c178`:

- **Expose `StageClassifier` through the bridge** — `switchyard-components::stage_router` now re-exports `StageClassifier` (it was already `pub` + exported from `switchyard_libsy::stage_router`).
- **Confidence mapping** — `score()` now emits **`score.abs()`** (the magnitude) for resolved outcomes instead of the signed/remapped value, so a resolved-weak `-0.6` reports `0.6`. (Direction still travels via the `target` string + `decision_source`.)
- **Public-item docs** — added `///` docs for `ToolSignalProcessor` + `recent_window`, `State.tool_signals` (lifecycle / absence semantics), and the `stage_router` module.

libsy + components build clean; 9 stage_router tests pass; clippy clean; no new rustdoc warnings.

Not touching Nachiket's structural asks here (module layout, deleting the old `switchyard-components` files, moving `tool_signals` parsing to `switchyard-protocol`/IR) — flagging those for @messiaen to scope on this PR vs a follow-up.

nachiketb-nvidia (approved):

PR #138, opened 2026-07-24
feat(server): /v1/messages/count_tokens forwarding on the Rust server

sabhatinas:
## What

Adds `POST /v1/messages/count_tokens` to the Rust `switchyard-server`. Claude Code calls it to size a request before sending; without it the endpoint 503'd.

## How

`count_tokens` is a **direct passthrough**, not a routed call. Completions run the router; count_tokens does not:

| endpoint | behavior |
| -- | -- |
| `/v1/messages` (completion) | Routes normally — classifier cascade, StageClassifier, escalate/abstain logic |
| `/v1/messages/count_tokens` | **Direct passthrough** — forwards straight to the route's Anthropic target's `/v1/messages/count_tokens`, no routing |

Token counting is a pre-flight estimate with no routing decision, so running the classifier cascade for it is wrong (on a stage router a signal-less request would abstain, or fire a billable judge call). Instead:

- `Algorithm::count_tokens(request)` forwards via `count_tokens_client()` — the route's first Anthropic-capable target. 
- `RoutedLlmClient::count_tokens(request)` forwards to that client's Anthropic backend, restamping the model to the upstream id. It shares the encode/POST helper (`send_encoded`) with the completion path.
- No Anthropic target in the route → 400.

**Caveat (deferred):** for a route with *multiple* Anthropic tiers, "first" is arbitrary — choosing which tier count_tokens should reflect is left for later.

## Live snapshots

Run against real NVIDIA Anthropic (`claude-opus-4-8` @ inference-api) through `switchyard-server` on a single-target route:

| request | response |
| -- | -- |
| user: `"hi"` | `{"input_tokens": 8}` · 200 |
| user: `"count these tokens please"` (+ system) | `{"input_tokens": 19}` · 200 |
| longer user prompt (~1 paragraph) | `{"input_tokens": 44}` · 200 |
| system + 1 user msg + 1 tool (Claude Code shape) | `{"input_tokens": 370}` · 200 |
| `/v1/messages` completion on the same route | 200 |

Counts scale with content and include system + tools, matching Anthropic's own count_tokens.

## Stage router

A unit test confirms the fix: on an efficient-first stage router, a **completion abstains** (no signals to route on) but **count_tokens succeeds** — it passes through to the strong (Anthropic) tier and returns a count.

## Deferred to a follow-up

The **Python server** keeps its current forwarder for now. Making it symmetric means adding count_tokens to the `LlmBackend` stack (switchyard-core / switchyard-components) and reaching the backend through the chain. Tracked off SWITCH-1048.

## Testing

231 unit/integration tests green (incl. the stage-router passthrough test and the shared `send_encoded` path), clippy clean, plus the live checks above.

Linear: SWITCH-1048

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -198,6 +198,7 @@ pub fn build_switchyard_router(state: ServerState) -> Router {
     Router::new()
         .route("/v1/chat/completions", post(openai_chat_completions))
         .route("/v1/messages", post(anthropic_messages))
+        .route("/v1/messages/count_tokens", post(anthropic_count_tokens))
         .route("/v1/responses", post(openai_responses))
         .route("/v1/models", get(models))
         .route("/metrics", get(prometheus_metrics))
@@ -257,6 +258,52 @@ async fn openai_responses(
     handle_endpoint(state, headers, body, WireFormat::OpenAiResponses).await
 }
 
+/// Anthropic token counting. Resolves the route named by `model`, then does a
+/// **direct passthrough** via [`Algorithm::count_tokens`] to that route's
+/// Anthropic target — it does *not* run the routing cascade (count_tokens is a
+/// pre-flight estimate with no routing decision). Unknown route → 404; a route
+/// with no Anthropic target → 400.
+async fn anthropic_count_tokens(
+    State(state): State<ServerState>,
+    headers: HeaderMap,
+    body: std::result::Result<Json<Value>, JsonRejection>,
+) -> Response {
+    let body = match llm_json_body(body) {
+        Ok(body) => body,
+        Err(message) => return invalid_body_error(message),
+    };
+    let (algorithm, request, _) = match resolve_route(
+        &state,
+        metadata_from_headers(&headers),
+        body,
+        WireFormat::AnthropicMessages,
+    ) {
+        Ok(resolved) => resolved,
+        Err(response) => return response,
+    };
+    match algorithm.count_tokens(request).await {
+        Ok(payload) => (StatusCode::OK, Json(payload)).into_response(),
+        Err(error) => count_tokens_error(error),
+    }
+}
+
+/// Map a [`count_tokens`](Algorithm::count_tokens) failure to an HTTP response:
+/// the route has no Anthropic target → 400, an upstream HTTP error → its own
+/// status, anything else → 502.
+fn count_tokens_error(error: LibsyError) -> Response {
+    // The one count_tokens-specific case is "no Anthropic target in the route";
+    // every upstream/client failure gets the same mapping completions use.
+    match &error {
+        LibsyError::AlgorithmError { message } => error_response(
+            StatusCode::BAD_REQUEST,
+            message.clone(),
+            "invalid_request_error",
+            "count_tokens_unsupported",
+        ),
+        _ => algorithm_error(error),
+    }
+}
+
 async fn handle_endpoint(
     state: ServerState,
     headers: HeaderMap,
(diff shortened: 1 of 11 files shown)


grahamking:
> A routing chain can't count tokens, so the handler forwards the request to the route's Anthropic tier and returns its answer. The full target (base_url + key) is gone by request time, so the config capture grabs the first Anthropic-format target at build time and the endpoint forwards to it — swapping the route name for the real model and sending x-api-key + anthropic-version. No hardcoded keys. No Anthropic tier → clean 400.

@sabhatinas Can you re-write that? I think that's AI stringing words together. I've been squinting at it but it doesn't really make any sense.

crates/switchyard-server/src/config.rs
```diff
@@ -67,7 +67,36 @@ impl ServerConfig {
...
+    fn anthropic_count_tokens_target(&self) -> ServerResult<Option<CountTokensTarget>> {
+        for target in self.targets.values() {
+            let client = self.llm_clients.get(&target.llm_client).ok_or_else(|| {
+                ServerError::new(format!(
+                    "target references unknown llm client {}",
+                    target.llm_client
+                ))
+            })?;
+            if matches!(client.format, ClientFormat::AnthropicMessages) {
+                return Ok(Some(CountTokensTarget {
```
grahamking: If both "strong" and "weak" are `messages`, does it matter which one we choose? This uses the first match.

crates/switchyard-server/src/config.rs
```diff
@@ -67,7 +67,36 @@ impl ServerConfig {
...
+                    base_url: client.base_url.trim().to_string(),
+                    api_key: resolve_api_key(&target.llm_client, client)?,
```
grahamking: Having to trim the base_url twice and call `resolve_api_key` twice (both are done in `build_backend`) feels like a design smell. Not a blocker though.

crates/switchyard-server/src/lib.rs
```diff
@@ -246,6 +273,87 @@ async fn openai_responses(
...
+                StatusCode::BAD_GATEWAY,
+                format!("count_tokens upstream request failed: {error}"),
+                "upstream_error",
+                "upstream_error",
+            );
+        }
+    };
+    let status =
+        StatusCode::from_u16(upstream.status().as_u16()).unwrap_or(StatusCode::BAD_GATEWAY);
+    let payload = match upstream.bytes().await {
```
grahamking: ~You could do~

Ignore previous message. I forgot we were proxying.

grahamking (approved):

sabhatinas:
> > A routing chain can't count tokens, so the handler forwards the request to the route's Anthropic tier and returns its answer. The full target (base_url + key) is gone by request time, so the config capture grabs the first Anthropic-format target at build time and the endpoint forwards to it — swapping the route name for the real model and sending x-api-key + anthropic-version. No hardcoded keys. No Anthropic tier → clean 400.
> 
> @sabhatinas Can you re-write that? I think that's AI stringing words together. I've been squinting at it but it doesn't really make any sense.

I updated the PR body now, it should be easier to understand now.

crates/switchyard-server/src/config.rs
sabhatinas: It would matter, I am looking into using the same target as original request.

sabhatinas: I deferred it as follow up. In this PR we will skip routing and directly passthrough the client's request by picking the first anthropic tier. If we are missing an anthropic tier, 400 status error is thrown

crates/switchyard-server/src/config.rs
sabhatinas: Good call. The rework removed the build-time target capture that caused this duplication, so `resolve_endpoint` had a single caller and a stale comment. I've reverted `config.rs` entirely — `build_backend` / `resolve_api_key` are back to main's version and the PR no longer touches this file. count_tokens now lives in `lib.rs` + `Algorithm::count_tokens` (direct passthrough), not the config layer.

PR #140, opened 2026-07-24
feat(prefill-router): add learned prefill complexity routing

mich-xu:
## What

Adds a learned `prefill_probe` routing profile using Switchyard’s supported v1 component chain.

- Adds a Rust request processor that:
  - Extracts the first user task instruction internally without requiring a custom benchmark agent or request field.
  - Sends only the extracted task text to a dedicated Qwen3.6 hidden-state probe.
  - Mean-pools prompt tokens independently across all hidden-state layers, concatenates the layer vectors, and applies the external checkpoint’s scaler and PCA-200 transform.
  - Runs the learned ensemble and compares only the configured weak and strong checkpoint heads.
  - Applies a lambda-controlled, cost-aware utility decision.
  - Caches successful task-level decisions and routes probe failures to strong without caching the failure.
- Adds the direct PyO3 binding and Python profile configuration.
- Registers `prefill_probe` in the current route-bundle configuration and model discovery paths while keeping the probe model internal.
- Builds completion tiers through the standard tier backend path, preserving Anthropic cache breakpoints and canonical routing statistics.
- Adds unit tests, a sample deployment profile, and updated hidden-state deployment documentation.

The checkpoint remains external deployment data. Switchyard does not download checkpoints, fit PCA, train models, or perform online learning.

## Why

The earlier prefill-router implementation targeted the removed components-v2 architecture and is not present on canonical `main`. This ports the learned router onto the supported v1 request-component and deterministic-backend chain.

The router enables task-level strong/weak selection using learned prompt complexity while preserving the original completion request and Switchyard’s existing backend, statistics, translation, and provider-cache behavior.

Related to #97 , #13 , and #119 .

## How tested

Additional validation:

- `cargo fmt --all --check`
- `cargo test --workspace`
- Verified direct Qwen3.6 probe startup, `[prompt_tokens, layers, hidden_size]` artifact handling, PCA-200 checkpoint inference, and hidden-state file cleanup.
- Verified both strong and weak decisions.
- Verified repeated turns reuse one cached task-level decision.
- Verified induced probe failures route to strong and are retried rather than cached.
- Verified Anthropic cache breakpoints and provider cache accounting: the repeated strong-tier smoke reported `15,420` cache-creation tokens on the first request and `15,420` cached tokens on the second.
- Ran one complete Terminal-Bench 2.1 `mteb-leaderboard` task at `lambda: 0.75` with stock Terminus 2. The task made 14 successful weak-tier Nemotron calls from one probe decision, recorded zero endpoint errors, and cleaned up the hidden-state artifact. The verifier ran successfully; the task reward was `0.0` because the agent produced an incorrect answer.

## Checklist

## Notes for reviewers

- The cost-aware policy semantics: `lambda` is the only continuous routing knob, an exact utility tie selects weak, and checkpoint probabilities remain private.
- The v1 profile composition through `build_tier_backend`, which preserves standard Anthropic cache wrapping and statistics.
- Route-bundle discovery: strong and weak completion targets are exposed normally, while the Qwen probe remains internal.

Intentional minimal-release trade-offs:
- The successful-decision cache remains in-process and unbounded.
- Probe and scoring failures fall back to strong and are not cached.
- The checkpoint and Qwen3.6 probe are externally managed deployment dependencies.
- No components-v2 code is restored.

## Summary by CodeRabbit

* **New Features**
  * Added learned prefill-probe routing with cost-aware selection between strong and weak completion targets.
  * Added support for configuring prefill-probe routes through route bundles and Python APIs.
  * Added local benchmark configuration for prefill-complexity routing.
  * Probe models remain internal while completion tiers are exposed for discovery.
* **Documentation**
  * Added setup and configuration guidance for hidden-state extraction and learned routing.
* **Tests**
  * Added coverage for routing behavior, validation, fallback handling, caching, and model discovery.


nachiketb-nvidia (changes requested):
Pls implement this algorithm directly into libsy, see `crates/libsy/src/algorithms.rs` and `crates/libsy/src/algorithms` for precedent.

nachiketb-nvidia (changes requested):
To add, can we break this up?

It is much harder to review with a 4k diff, with and w/o agents.

Apart from that, if I understand correct, a vllm based service as a sidecar is required. Which goes through a complete model forward pass and we pick only the attention activations which is written to disk. Correct me if im wrong. 
This is not the most efficient but the question is, if this meant to be this way as a first pass?

anniesurla:
You are right on it doing a forward pass on every request, and as of today, given the vllm implementation, we have the activations written to disk. Will let @sabhatinas add to the above on if there is a more efficient way to do this with Switchyard.

nachiketb-nvidia:
closing as it will be reimplemented during prefill-router rewrite, will be kept as reference!

PR #145, opened 2026-07-26
feat(routing): sub-agent override as a libsy classifier, restored on the serve path

linj-glitch:
## What

Two halves of the same story: sub-agent routing becomes a composable classifier in libsy, **and** returns to the serve path where traffic actually flows.

**libsy** — `SubagentOverride` moves from an `Algorithm` combinator to a `Classifier`, so it composes with `AffinityRouter` in a `FallThrough` cascade:

```rust
FallThrough::new(targets)
    .with_component(Arc::new(AffinityRouter::for_subagents()))   // replay an existing pin
    .with_classifier(Arc::new(SubagentOverride::new("worker")))  // else seed one for delegated work
    .with_classifier(base)                                       // else route normally
```

**serve** — the `subagent_target` route envelope key returns, valid on any route type:

```yaml
routes:
  qa-subagent-split:
    type: model
    target:          { model: aws/anthropic/bedrock-claude-opus-5 }
    subagent_target: { model: nvidia/nemotron-3-nano-30b-a3b }
```

## Why

Closes SWITCH-1068. Per @grclark: *"I think SubagentOverride could be a classifier in libsy terms. That way we can easily combine affinity and override."*

The two sat at different layers and could not be combined: `SubagentOverride` was `Algorithm<()>` while `FallThrough` is `Algorithm<SharedState>`, so wrapping did not typecheck. **And the arrangement described earlier in that thread — override wrapping an inner algorithm with affinity inside — could not have worked even with matching types**: the override short-circuited to `driver.call_llm_target`, so a sub-agent request never entered the inner algorithm, the processor chain never saw `Event::Request`, and affinity never latched. Sub-agent traffic was invisible to every stateful component. As sibling classifiers this resolves itself — `FallThrough` replays the winning decision to the processors, so the override's turn-1 choice is latched for free.

**The serve half is a restoration, not a new feature.** `subagent_target` shipped in #112 and was deleted by #119 ("nuke components-v2"), where it appears to be collateral rather than a decision — the commit targets the crate, not this feature. #112 also created libsy's `subagent_override.rs`, which survived; that asymmetry is why the classifier had no serve-path counterpart. Restoring it needs review from whoever drove #119 in case dropping it was intentional.

## Measured on real traffic

Both halves were exercised end to end against four SWE-Atlas QnA tasks driven by Claude Code through Switchyard, closed-book. This also confirmed the thing every sub-agent key depends on: **`x-claude-code-agent-id` does reach Switchyard**, so `is_subagent_request` fires on live traffic rather than only in tests.

The serve-path arm below routes delegated work to an open-weights worker on a self-hosted vLLM deployment (`subagent_target`), against an otherwise identical single-model control:

| Arm | Requests | Cost | $/task | mean agg_score | reward |
|---|---:|---:|---:|---:|:---:|
| Opus 5 solo | 527 | $36.62 | $9.15 | 0.962 | 3/4 |
| Opus 5 + GLM-5.2 worker | 506 | **$18.71** | $4.68 | 0.962 | 3/4 |

**49% cheaper at identical quality** — same mean `agg_score`, same reward pass rate, and the only task below ceiling (grafana, 0.846) scored the same in both arms. 322 of 506 requests (64%) were served by the worker for $4.19, while the root model handled 184 for $14.52.

Traffic shape, which is what motivates pairing the override with affinity: the split arm spawned **17 sub-agents, none single-shot** — median 39 turns, max **137**. A 137-turn child under a re-scoring classifier is 137 routing decisions where one would do; affinity-first collapses that to 1 decision plus 136 replays.

Caveats, stated plainly: 4 tasks, one trial per arm, and three of the four sit at the 1.000 `agg_score` ceiling, so the quality claim is mostly "ceiling tasks stayed at ceiling". Spawn counts varied 5-17 across runs on identical inputs, so run variance is large relative to the differences. The cost direction is solid — a big effect with a visible mechanism — but the magnitude is indicative, not decision-grade.

Also observed: the routing log records **one `session_id` per task** covering parent and all descendants, with no `agent_id` field. Sub-agent routing is currently unobservable in our own telemetry, and it confirms why affinity keys on `session + agent` — keying on session alone would pin a whole task, parent and every descendant, to one model. Telemetry gap left for a follow-up.

## How tested

- [x] `cargo test` green, `cargo clippy --all-targets` clean, `cargo fmt --all`
- [x] `uv run mypy switchyard` clean (161 files)
- [x] `uv run pytest tests/ --ignore=tests/e2e` green (1954 passed, 9 skipped)
- [ ] `tests/e2e/test_classifier_planner_chain_e2e.py` fails on a live-upstream 401 (`LiteLLM Virtual Key expected. Received=nvap****`). Pre-existing env key-format issue, unrelated.

New coverage:

**`crates/libsy/tests/subagent_affinity.rs` (8 tests)** — drives the cascade through the public API. The key one is **a pin outliving the policy that seeded it**: two cascades share one `AffinityRouter` but disagree on the worker, and the second override is never consulted. Paired with a negative control (independent routers ⇒ the second override *does* win), so it cannot pass for the wrong reason. Also: a Codex `compact` turn is sub-agent *lineage* but not delegated *work*, so it is keyed by affinity yet not forced to the worker; and dropping either classifier still leaves a valid cascade.

**`tests/test_subagent_routing.py` (8 tests)** — detection policy, runtime branching, and route-bundle wiring, including that an unknown envelope key is still rejected.

## Notes for reviewers

**One policy, two plumbings.** The serve wrapper never sniffs headers: `is_subagent_request` is a 29-line binding over `switchyard_protocol::Metadata::from_headers(..).is_subagent_work()` — the same function libsy's classifier calls. So `compact` cannot count as delegated work in one engine and not the other. This is deliberately unlike the LLM classifier, where the libsy (390-line) and serve (2055-line) implementations have genuinely diverged.

**Public API removed:** the Python `subagent_override(inner, worker)` libsy binding is gone. A classifier scores a target *name* resolved from the cascade's `LlmTargetSet`, so the old `(inner: Algorithm, worker: LlmTarget)` shape is not representable, and libsy ships no terminal `Classifier` — a Python-built cascade would fail with "every classifier abstained" on root traffic. Zero consumers across Gym, log2-infra, nv-OpenHands, open_source_model_improvement, and switchyard-bench. Exposing composition to Python needs a terminal classifier; left to a follow-up.

**`FallThrough::with_component`** is new: `AffinityRouter` holds assignments on the instance, so both roles must be the same `Arc`. Registering them separately compiles, runs, and silently never latches. This makes that unrepresentable and matches what the affinity module docs already claimed.

**Serve-path scope — one wrap, no per-type branching.** The whole rule is three lines in `_build_switchyard_for_route`:

```python
chain  = _build_route_chain(...)
worker = _subagent_worker_runtime(...)   # None when subagent_target is unset
return chain if worker is None else SubagentOverrideRuntime(chain, worker)
```

`model`, `passthrough`, `stage_router`, `deterministic`, and `escalation_router` all funnel through that site, so they get it for free. `random_routing` expands into its table entries on a separate path and so is unaffected — the key parses but does nothing there. That is deliberate: sub-agent logic branching on `route_type` was not worth the code, so there is none.

**Skill drift, separate commit:** `switchyard-lib-core/SKILL.md` documented `SubagentOverrideProfile`, `switchyard/lib/profiles/loader.py`, `is_subagent_request`, and `crates/switchyard-components-v2/` — all removed by #119, so the skill was telling agents to set a YAML key that did nothing. Rewritten to match reality.

**Benchmark artifacts included:** five routing profiles under `benchmark/routing-profiles/` (the `qa-subagent-*` ones are the usage documentation for `subagent_target`, including the cross-provider case where the worker carries its own `base_url`/`api_key`), plus `benchmark/run_cost.py` for per-model cost reporting. Cost-table additions: Claude Opus 5 at the published $5/$25 Opus-tier rate, and GLM-5.2 at **third-party reference rates** — that model runs on a self-hosted deployment with no per-token billing, so the entry answers "what would this traffic have cost at market rates", not what we are billed; provider rates vary about 2x. Run outputs and generated datasets are gitignored and not included.


benchmark/routing-profiles/qa-single-opus-5.yaml
ayushag-nv, line 1: @linj-glitch we don't need all the routing profiles with internal endpoint names. This should be given in the docs only. No need to add 4 files for this

linj-glitch: Agreed on both counts — removed all four profiles. They live in gitignored `scratch/` now.

Replaced with a `## Sub-agent override (`subagent_target`)` section in `docs/routing_algorithms/overview.md`, using placeholder endpoints. That restores documentation #112 originally shipped and #119 removed along with the feature — I had restored the code without the docs, so this is the right home for it.

One thing worth flagging given the repo is public: the internal hostname was live here for about an hour. I rewrote the branch history so it no longer appears in any commit, but a force-push does not retract what GitHub may retain. Happy to escalate if you think it warrants it.

eric-liu-nvidia (approved):

PR #155, opened 2026-07-27
feat(libsy): add weighted reproducible random routing

nachiketb-nvidia:
## Summary

- add relative weighted selection and reproducible seeds to libsy `Random`
- expose optional `weights` and `seed` fields for random routes in server TOML
- expose `algorithms.random(targets, *, weights=None, seed=None)` in Python
- keep algorithm validation inside libsy constructors instead of the server config layer
- move the existing LLM classifier threshold validation into `LlmClassifier`

## Behavior

- `Random::new(target_set, weights, seed) -> Result<Random>` owns target, weight, and seed setup.
- Weights follow target order, default to `1` per target, and do not need to sum to `1`.
- A zero weight disables that target; weights must otherwise be finite and nonnegative, with at
  least one positive value.
- A seed reproduces selections for the same ordered calls; omitting it uses entropy.
- `LlmClassifier::new(...) -> Result<LlmClassifier>` now validates its threshold and required
  targets instead of relying on server configuration.

## Validation

- `cargo fmt --all --check`
- `cargo test -p switchyard-libsy -p switchyard-server`
- `uv run pytest tests/test_libsy_minimal_bindings.py -v -o addopts=`

Linear: https://linear.app/nvidia/issue/SWITCH-1102/add-weighted-and-reproducible-random-routing

### DIFF
--- a/crates/libsy/src/algorithms/rand.rs
+++ b/crates/libsy/src/algorithms/rand.rs
@@ -3,18 +3,20 @@
 
 //! Random router built on the [`Algorithm`] interfaces.
 //!
-//! Selects one target from the set uniformly at random and calls it. This is the
-//! simplest possible routing algorithm and the reference for the single-call
-//! shape: one `driver.call_llm_target` inside `create_run_task`. Weighted selection
-//! can be layered on later; the set defines the candidates.
+//! Selects one target from the set at random and calls it. Selection is uniform
+//! by default and can use relative weights and a reproducible seed.
 
-use std::sync::Arc;
+use std::collections::BTreeSet;
+use std::sync::{Arc, Mutex};
 
 use async_trait::async_trait;
-use rand::seq::SliceRandom;
+use rand::distributions::{Distribution, WeightedIndex};
+use rand::rngs::StdRng;
+use rand::SeedableRng;
 
 use crate::{
-    Algorithm, Context, Decision, Driver, LibsyError, LlmTargetSet, Request, Response, Result,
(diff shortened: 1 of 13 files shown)


crates/libsy/src/algorithms/rand.rs
```diff
@@ -39,18 +41,85 @@ impl Decision for RandomDecision {
...
+        let rng = match seed {
+            Some(seed) => StdRng::seed_from_u64(seed),
+            None => StdRng::from_entropy(),
+        };
+        Ok(Self {
+            target_set,
+            distribution,
+            rng: Mutex::new(rng),
+        })
+    }
```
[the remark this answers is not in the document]
nachiketb-nvidia: Addressed with explicit error documentation.

crates/libsy/src/algorithms/rand.rs
```diff
@@ -39,18 +41,91 @@ impl Decision for RandomDecision {
...
+        let weights = weights.unwrap_or_else(|| vec![1.0; target_count]);
+        if weights.len() != target_count {
+            return Err(invalid_weights(format!(
+                "expected {target_count} weights, got {}",
+                weights.len()
+            )));
+        }
+        if weights
+            .iter()
+            .any(|weight| !weight.is_finite() || *weight < 0.0)
```
grahamking, line 92: `is_finite()` :-)  Is that Opus 5? It goes above and beyond.

grahamking (approved):

crates/libsy/src/algorithms/rand.rs
nachiketb-nvidia: this is GPT 5.6 Sol :rocket:

PR #158, opened 2026-07-27
refactor(libsy): generalize fall-through state

nachiketb-nvidia:
## Summary

- make `Processor<S>` and `Classifier<S>` generic over composition-owned state
- retain one private `S` per session inside `FallThrough<S>`
- use fresh, unretained state for requests without a session ID
- implement weighted random routing as `FallThrough<()>` plus `RandomClassifier`
- make decision replay request-aware so affinity remains correct across concurrent turns
- wire the merged LLM classifier and switchyard-server through the generic state design

## Design

`FallThrough<S>` is the reusable composition primitive. Stateful compositions keep a private `HashMap` from session ID to `Arc<tokio::sync::Mutex<S>>`. The registry lock is held only while looking up or creating an entry; each session then serializes its processor, classifier, and decision-replay sequence without blocking other sessions. The state guard is dropped before the selected target is called.

Zero-sized state types such as `()` bypass the registry. Requests without a session ID receive a fresh `S::default()` for that run rather than sharing anonymous global state.

The built-in `State` remains in `core/state.rs` for algorithms using standard routing facts. `LlmTaskClassifier` implements `Classifier<State>`, and switchyard-server constructs that route as `FallThrough<State>` while storing every route uniformly as `Arc<dyn Algorithm>`. State is no longer threaded through `Context` or the server route table.

## Validation

- `cargo fmt --all --check`
- `cargo test -p switchyard-libsy`
- `cargo test --workspace`
- `git diff --check`

No Python behavior changed and no live provider calls were made.

### DIFF
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -1,23 +1,23 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Fall-through classifier routing: a stateful [`Algorithm`] that routes each turn
+//! Fall-through classifier routing: a composable [`Algorithm`] that routes each turn
 //! through a processor chain and a classifier cascade.
 //!
-//! Each turn: request-side [`Processor`]s fold facts into the session [`State`](crate::State); the
+//! Each turn: request-side [`Processor`]s fold facts into the composition's state; the
 //! [`Classifier`] cascade is consulted in order and the first to score decides the target
 //! (its `argmax`); the [`Decision`] is published and then replayed to the processors so
 //! stateful ones (latch, affinity) can bind it.
 //!
-//! [`FallThrough`] is generic over the state handle carried in [`Context`]. The default
-//! [`Shared<State>`] persists state across turns; `FallThrough<()>` is stateless and acquires
-//! no lock.
+//! [`FallThrough`] owns its state handle. The default `FallThrough<()>` is stateless and
+//! acquires no lock; [`Shared<S>`] persists custom state across turns. Request-scoped values
+//! continue to travel through an ordinary [`Context`].
 
 use std::sync::Arc;
 
 use async_trait::async_trait;
 
-use crate::core::{Classifier, Event, Processor, Score, Shared, State, StateGuard, StateHandle};
+use crate::core::{Classifier, Event, Processor, Score, StateGuard, StateHandle};
 use crate::{
     Algorithm, Context, Decision, Driver, LibsyError, LlmTargetSet, Request, Response, Result,
     RoutedLlmClient,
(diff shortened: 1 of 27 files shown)


crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -46,36 +47,72 @@ impl Decision for FallThroughDecision {
...
-/// Stateless: the per-session [`State`](crate::State) lives in the threaded [`Context<SharedState>`], so one
-/// shared instance serves every session (see the module docs).
-pub struct FallThrough {
-    processors: Vec<Arc<dyn Processor>>,
-    classifiers: Vec<Arc<dyn Classifier>>,
+/// The generic state type is shared by every processor and classifier in the composition.
+pub struct FallThrough<S = ()> {
+    name: String,
+    decision_reason: fn(&str, &Score) -> String,
+    state: FallThroughState<S>,
```
messiaen: @nachiketb-nvidia If the algorithm owns the state, we have to recreate it to clear the state, right?

It also means we'd have to have an Algorithm per session.

Ideally I think we could have a single algorithm serve many sessions via different passed in state.  This used to be done via context.  Maybe we could create the state from the context...

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -84,46 +107,78 @@ impl FallThrough {
...
+            .call_llm_target(ctx, &target, request, decision)
+            .await
+    }
 
-        // 1. Processor chain accumulates request-side facts into the session State.
+    /// Returns this request's retained state without holding the registry lock.
+    fn session_state(&self, request: &Request) -> Option<Arc<AsyncMutex<S>>> {
+        if std::mem::size_of::<S>() == 0 {
+            return None;
+        }
```
messiaen:
```suggestion
        if std::any::TypeId::of::<S>() == std::any::TypeId::of::<()> {
            return None;
        }
```

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -84,46 +107,78 @@ impl FallThrough {
...
+            .as_ref()?
+            .session_id
+            .as_deref()
+            .filter(|session_id| !session_id.is_empty())?;
+        let mut states = self.session_states.lock();
+        Some(
+            states
+                .entry(session_id.to_string())
+                .or_insert_with(|| Arc::new(AsyncMutex::new(S::default())))
+                .clone(),
```
messiaen: Do we have to `clone`? can we return a reference.

messiaen (approved):

crates/libsy/src/algorithms/fall_through.rs
nachiketb-nvidia: removed this code smell entirely

crates/libsy/src/algorithms/fall_through.rs
nachiketb-nvidia: apparently yes, some type of clone. Did an Arc::clone instead which makes it cheaper, we still have the same updating happening though across requests for the same session.

PR #159, opened 2026-07-27
feat(server): add structured request logging

nachiketb-nvidia:
## What

- Initialize structured `tracing` output in `switchyard-server`.
- Emit exactly one terminal request event for each non-stream failure/success path.
- Log handled requests and client errors at `INFO`.
- Log algorithm, translation, and upstream failures at `WARN` with the propagated error.
- Keep libsy routing decisions and nested call details available through opt-in debug logging.

Streaming failures remain a separate `WARN` because they can occur after the HTTP response has started.

## Why

The Rust server records spans and OpenTelemetry metrics but did not install a tracing subscriber, so operators received no request logs. The server needs a concise default signal: one event that says what route was requested, what model was selected, how the request ended, and how long handling took. Internal routing detail should remain available without making routine logs noisy.

## How

- Install `tracing-subscriber` in the server binary with the default filter `switchyard_server=info`.
- Capture request metadata before dispatch and retain propagated terminal error context alongside the HTTP response.
- Emit the terminal event centrally after response construction:
  - `INFO ... LLM request handled` for successful and client-error responses.
  - `WARN ... LLM request failed` for algorithm/upstream failures.
- Include wire format, HTTP status, requested model, selected model, streaming flag, session ID, correlation ID, handling duration, and the propagated error on failures.
- Keep libsy warning semantics unchanged while excluding nested libsy events from the default server filter, preventing duplicate failure lines.
- Write logs to stderr without ANSI formatting for container and service log collectors.
- Leave OpenTelemetry counters, histograms, and span recording unchanged; log filtering does not gate metric recording.

This MR intentionally does not add disk persistence, RL records, session aggregation, or new endpoints.

## Operator know-how

Default operation needs no environment variable and emits one terminal server event per request:

```text
INFO switchyard_server::request: LLM request handled ... status=200
WARN switchyard_server::request: LLM request failed ... status=502 error="..."
```

Enable Switchyard-specific routing and nested call diagnostics without enabling debug logs from every dependency:

```bash
RUST_LOG=switchyard_server=debug,libsy=debug switchyard-server --config routes.toml
```

## What to review

- Confirm the terminal event carries useful routing metadata without prompt, response, API-key, or raw-header content.
- Confirm propagated failures produce one default `WARN`, not duplicate model-call and algorithm-run warnings.
- Confirm handled requests and client errors produce one `INFO` event.
- Confirm the streaming-error exception is appropriate because stream consumption outlives response construction.
- Confirm libsy OpenTelemetry metrics and standalone warning semantics remain unchanged.

## Validation

- `cargo test -p switchyard-libsy --test observability`
- `cargo fmt --all --check`
- Live real-binary smoke test:
  - no-op route returned HTTP 200 and emitted one `INFO`
  - unreachable passthrough returned HTTP 502 and emitted one `WARN` with the propagated transport error
  - zero nested `libsy` lines appeared under the default filter

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -252,36 +252,31 @@ async fn handle_endpoint(
     body: std::result::Result<Json<Value>, JsonRejection>,
     wire_format: WireFormat,
 ) -> Response {
-    let started = Instant::now();
     let metadata = metadata_from_headers(&headers);
-    let session_id = metadata.session_id.clone();
-    let correlation_id = metadata.correlation_id.clone();
-    let requested_model = body
-        .as_ref()
-        .ok()
-        .and_then(|body| body.0.get("model"))
-        .and_then(Value::as_str)
-        .map(str::to_string);
-    let streaming = body
-        .as_ref()
-        .ok()
-        .and_then(|body| body.0.get("stream"))
-        .and_then(Value::as_bool)
-        .unwrap_or(false);
-
-    let response = match llm_json_body(body) {
-        Ok(body) => handle_llm_request(state, metadata, body, wire_format).await,
-        Err(message) => invalid_body_error(message),
-    };
-    log_llm_request(
-        &response,
+    let request_log = RequestLogContext {
+        started: Instant::now(),
         wire_format,
-        requested_model.as_deref(),
-        session_id.as_deref(),
-        correlation_id.as_deref(),
-        streaming,
-        started.elapsed(),
-    );
+        requested_model: body
+            .as_ref()
+            .ok()
+            .and_then(|body| body.0.get("model"))
+            .and_then(Value::as_str)
+            .map(str::to_string),
+        streaming: body
+            .as_ref()
+            .ok()
+            .and_then(|body| body.0.get("stream"))
+            .and_then(Value::as_bool)
+            .unwrap_or(false),
+        session_id: metadata.session_id.clone(),
+        correlation_id: metadata.correlation_id.clone(),
+    };
+
+    let (response, error) = match llm_json_body(body) {
+        Ok(body) => handle_llm_request(state, metadata, body, wire_format).await,
+        Err(message) => (invalid_body_error(message), None),
+    };
+    request_log.emit(&response, error.as_deref());
     response
 }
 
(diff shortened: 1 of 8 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -252,11 +252,32 @@ async fn handle_endpoint(
...
+        session_id: metadata.session_id.clone(),
+        correlation_id: metadata.correlation_id.clone(),
     };
-    handle_llm_request(state, headers, body, wire_format).await
+
+    let (response, error) = match llm_json_body(body) {
+        Ok(body) => handle_llm_request(state, metadata, body, wire_format).await,
+        Err(message) => (invalid_body_error(message), None),
+    };
+    request_log.emit(&response, error.as_deref());
```
grahamking: Instead of adding the `Option<String>` to `handle_llm_request` could we have `request_log.emit` handle it? By checking `response.status().is_success()`. If it is not success, log the request body as the error. Something like that.

crates/switchyard-server/src/lib.rs
```diff
@@ -271,53 +292,113 @@ fn llm_json_body(
...
+
+impl RequestLogContext {
+    fn emit(self, response: &Response, error: Option<&str>) {
+        let selected_model = response
+            .headers()
+            .get(HEADER_SELECTED_MODEL)
+            .and_then(|value| value.to_str().ok())
+            .unwrap_or("");
+        let duration_ms = self.started.elapsed().as_secs_f64() * 1_000.0;
+        match error {
```
grahamking: I think we can use `response.status()` instead to detect errors.

We could even do `status().is_server_error()` -> Log at `tracing::error!` level.
And `is_client_error()` (and maybe others) at `warn!` level.

crates/switchyard-server/src/lib.rs
```diff
@@ -271,53 +292,113 @@ fn llm_json_body(
...
         raw_request: Some(body),
-        metadata: Some(metadata_from_headers(&headers)),
+        metadata: Some(metadata),
     };
     let (trace, response) = match algorithm.run(Context::default(), request).await {
         Ok(result) => result,
-        Err(error) => return algorithm_error(error),
+        Err(error) => {
+            let message = error.to_string();
+            return (algorithm_error(error), Some(message));
```
grahamking: `error` is already inside the `Response` that `algorithm_error` returns. It would be neat if we didn't have to return the tuple.

crates/switchyard-server/src/lib.rs
```diff
@@ -271,53 +292,113 @@ fn llm_json_body(
...
+                wire_format = %self.wire_format,
+                status = response.status().as_u16(),
+                requested_model = self.requested_model.as_deref().unwrap_or(""),
+                selected_model,
+                streaming = self.streaming,
+                session_id = self.session_id.as_deref().unwrap_or(""),
+                correlation_id = self.correlation_id.as_deref().unwrap_or(""),
+                handling_duration_ms = duration_ms,
+                "LLM request handled"
+            ),
```
grahamking: Use `tracing::event!`.

```
let level = match error { Some(err) => Level::WARN, None => Level::INFO };
tracing::event!(level, ..all those other fields only once ..)
```

https://docs.rs/tracing/latest/tracing/macro.event.html

grahamking (approved):

grahamking (approved):

PR #168, opened 2026-07-28
feat: task affinity based llm classifier with message hash fallback

ayushag-nv:
## What

- Extend the base llm classifier implementation to work with AffinityRouter for doing task based routing
- Added message_hash_fallback to treat first message hash as the AffinityKey in case of absence of headers
- Make FallThrough as the internal implementation detail , so that users just export the algorithm directly
- Earlier in the affinity Router, if concurrent sessions hit the same server their session keys were getting stashed because they were not stored in state, so fixed that

## Why

The motivation — Completes Sticky Task Based LLM Classifier implementation 

Closes #

## Where to Start Review

1. `crates/libsy/src/algorithms/llm_class.rs` — public algorithm shape and internal routing composition.
2. `crates/libsy/src/algorithms/util/affinity.rs` — affinity-key selection and message-hash fallback behavior.
3. `crates/switchyard-server/src/config.rs` — TOML wiring and validation.
4. `crates/libsy/src/algorithms/fall_through.rs` — internal execution delegation.

## How Tested

- `cargo test -p switchyard-libsy`

## Summary by CodeRabbit

* **New Features**
  * Added optional session affinity for classifier-routed requests.
  * Added message-based affinity fallback when no session identifier is available.
  * Improved request consistency by reusing prior routing decisions across related requests.

* **Bug Fixes**
  * Explicit session metadata now takes precedence over message-based affinity.
  * Subagent requests correctly avoid message-hash fallback routing.

* **Documentation**
  * Updated examples and guidance to reflect the simplified classifier configuration and routing behavior.

### DIFF
--- a/crates/libsy/src/algorithms/util/affinity.rs
+++ b/crates/libsy/src/algorithms/util/affinity.rs
@@ -83,7 +83,7 @@ impl AffinityRouter {
         }
     }
 
-    /// Uses a stable system/developer and first-user-message hash when metadata has no session.
+    /// Uses the latest user-message text as a fallback key when metadata has no session.
     pub fn with_message_hash_fallback(mut self) -> Self {
         self.message_hash_fallback = true;
         self
@@ -127,7 +127,7 @@ impl AffinityRouter {
             .is_some_and(|metadata| metadata.is_subagent);
         (!self.subagents_only && !is_subagent && self.message_hash_fallback)
             .then(|| {
-                message_hash(request).map(|hash| {
+                last_user_message_hash(request).map(|hash| {
                     tracing::debug!(affinity_key = %hash, "affinity using message hash fallback");
                     AffinityKey::Session(hash)
                 })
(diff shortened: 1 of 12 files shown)


linj-glitch:
@ayushag-nv one thing to fix before this lands:

**`message_hash` hashes the full `Message` struct** (`crates/libsy/src/algorithms/util/affinity.rs`):

```rust
serde_json::to_string(instruction).ok()?.hash(&mut hasher);
```

`serde_json` serializes every field including `cache_control`, `type`, `id`, etc. Two requests with identical text but different cache metadata will hash to different affinity keys, so the sticky-routing guarantee silently breaks for any client that sets cache hints. The existing `switchyard-core::session_key_from_body` avoids this by hashing only `(role, text content)` — suggest the same approach here.

One-function fix, doesn't touch the architecture. Otherwise LGTM.

crates/libsy/src/algorithms/util/affinity.rs
```diff
@@ -152,6 +177,32 @@ impl Processor for AffinityRouter {
...
+    {
+        serde_json::to_string(instruction).ok()?.hash(&mut hasher);
+    }
+    for message in &request.llm_request.messages {
+        match message.role {
+            Role::System | Role::Developer => {
+                serde_json::to_string(message).ok()?.hash(&mut hasher);
+            }
+            Role::User => {
+                serde_json::to_string(message).ok()?.hash(&mut hasher);
```
messiaen: Do we really want all the fields in message in the hash?

ayushag-nv: actually no. Needs the first system or user message only

linj-glitch (approved):
LGTM — the concurrent session bug is fixed by moving the pending key into per-request State.

linj-glitch:
Fixed in commit 67b9ac4 — pending key moved from the shared AffinityAssignments mutex into per-request State, so concurrent sessions can no longer clobber each other.

crates/libsy/src/algorithms/util/affinity.rs
ayushag-nv: for now removed this method. For task based classifier, I changed it to last user message only.

nachiketb-nvidia (approved):
looks right to me

PR #170, opened 2026-07-28
feat(libsy): stage router with handoff notes, per-tier prompts, and LLM fallback

sabhatinas:
A stage router cascade algorithm: tool signals decide most turns for free, and only the ones they can't fall through to the LLM classifier. If LLM classifier can not decide, default target is used as fall open. 

## The cascade

```
tool signals → capability classifier (optional) → picker's default tier
```

Each step decides or abstains, and the next one gets the turn. The last never abstains, so a turn is never left unrouted.

| Turn | Decided by | Recorded as |
| --- | --- | --- |
| signals confident | signals | `override` / `tests_passed` / `dimensions` |
| signals unsure, classifier answers | classifier | `llm-classifier` |
| signals unsure, classifier can't tell or absent | picker default | `fall_open` |

A turn the signals settle never reaches the classifier, so it costs nothing. `StageRouter` wraps the cascade and implements `Algorithm`, so `FallThrough` stays an internal detail.

The two tiers are **capable** and **efficient**; their targets are whatever the deployment names them.

## Changes to `llm_class`

- Abstains when it can't tell — unusable verdict, timeout, or the judge's own `abstain` — instead of quietly forcing the capable tier. The composition picks the fallback, so a judge failure now errs cheap under `efficient_first` rather than expensive.
- Its own cascade closes with the capable tier, so the standalone route behaves exactly as before.
- `recent_turn_window` is now a `TaskClassifierConfig` knob. `None` keeps today's newest-user-message-only judging; `Some(n)` widens to the client instructions, the opening task, and the last `n` turns after it.
- `trim_messages` selects by reference and clones only what survives, rather than copying the whole transcript to keep a window of it.

## Changes to tool signals

- Signals are read off the decoded conversation, not the raw body — three per-format parsers become one.
- Found live: the endpoint records no inbound wire format, so the old parser saw no signals and every turn routed as if the conversation had none.
- Recording the format is not the fix — the LLM client reads that field to pick the *upstream* format.
- The 40 signal tests are unchanged, still written against real wire bodies, now decoded through the codec.

## Changes to `llm_class`

- Abstains when it can't tell — unusable verdict, timeout, or the judge's own `abstain` — instead of forcing the capable tier. The composition picks the fallback, so a judge failure errs cheap under `efficient_first`.
- Its own cascade closes with the capable tier, so the standalone route is unchanged.
- `recent_turn_window` is now a `TaskClassifierConfig` knob: `None` judges the newest user message alone, `Some(n)` adds the instructions, opening task, and last `n` turns.
- `trim_messages` clones only what survives instead of copying the transcript.

## Changes to `fall_through`

- **A post-decision processor block.** Two passes: every processor sees the decision, then the outbound request is offered for rewriting. Order matters — the second pass sees the state the first settled.
- `Event::ModelRequest` now fires and carries the decision. Declared but never emitted before; it is the only hook that sees both, so a processor can act on the chosen target without keeping state between turns.
- The tier label now comes from any classifier in the cascade, not just the one that scored first, so a terminal fallback no longer leaves the call untiered.
- `DefaultTarget` — a terminal classifier that always picks one target. Used by this router and the capability route to close their cascades now that classifiers can abstain.

## Configuration

Thresholds are validated at construction, and the fall-back target is resolved once from the picker mode rather than per turn.

```toml
[routes.stage]
id = "switchyard/stage"
type = "stage_router"
capable_target = "opus"
efficient_target = "nemotron"
picker = "efficient_first"          # tier an undecided turn falls back to
confidence_threshold = 0.5          # agreement the signals need to decide
recent_turn_window = 3              # recent tool results the signals read
capable_system_prompt = "Diagnose the root cause quickly and act decisively."
efficient_system_prompt = "Work efficiently. Do not re-read files you have examined."

[routes.stage.handoff_notes]
escalation_note = "The previous model was stalling; pick up the diagnosis."
# deescalation_note = "..."                 # optional capable→efficient note
# only_on_wrong_signal_escalation = true    # default

[routes.stage.classifier]                   # omit for a signals-only router
target = "judge"
base_threshold = 0.5
recent_turn_window = 3                      # conversation turns the judge reads
```

## Handoff notes

| | |
| --- | --- |
| Fires on | signal-driven escalation; hand-back to efficient if configured |
| Placement | end of the last user turn, after the tool result |
| State | none — describes this turn's signals, like the Python router |
| Lifetime | forwarded request only, never in history |

## Tier system prompts

| | |
| --- | --- |
| Applies on | every turn that target serves, whichever step of the cascade picked it |
| Placement | ahead of the client's own instructions, so the framing is read first |
| Keyed by | target name, not tier — any router can use it, not just this one |
| Lifetime | forwarded request only, never in history |

## Follow-ups left as TODOs

- **Session affinity should bind on the post-decision hook** (`llm_class.rs`), the way the per-target system prompt does. Affinity is a fact about the decision, not a classification, so recording it there would let a pinned session skip the cascade outright on the next turn. It would also make the pin survive composition: a cascade that uses this classifier as one member — the stage router does — never runs the affinity wired inside it, because only the inner classifier is consulted. That is why `session_affinity` in a `stage_router`'s `[classifier]` block parses today but has no effect.

## Testing

- `tests/stage_router.rs` — multi-turn sessions through the assembled router: one note per signal-driven turn and none otherwise, no classifier call when the signals settled, a fresh classifier answer each turn, an unusable verdict landing on the picker default, the configured window reaching the judge, and the correct record of who decided.
- `tests/prompts.rs` — the request-text helpers on a plain cascade with no stage routing, which is what proves they are reusable: per-target prompts, one target's prompt not leaking onto another, the prompt following the target whichever classifier picked it, and a note reaching the model in the conversation.
- The server tests build a `stage_router` route with notes, tier prompts, and a classifier, and pin that a critical tool error escalates on the signals alone — the case that was silently broken.
- Validated against a live server (NVIDIA Inference Hub, Opus 4.7 capable / Nemotron-3-Super efficient) over 20 turns, run twice: signals-only, and with the classifier. All 10 error turns escalate identically in both configs, so the classifier is genuinely never consulted when the signals resolve a turn; the 10 clean turns fall open to the picker default without a classifier, and split on the judge's own reading with one.
- All green, plus lint and docs.

Refs SWITCH-1120, SWITCH-982, SWITCH-1110.


crates/libsy/src/algorithms/stage.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        Ok(Self {
+            inner: Arc::new(router),
+        })
+    }
+}
+
+#[async_trait]
+impl Algorithm<SharedState> for StageRouter {
```
nachiketb-nvidia: we should use this directly in config.rs

crates/libsy/src/algorithms/stage.rs
```diff
@@ -0,0 +1,305 @@
...
+/// judge takes the turns they cannot, and the picker's default tier closes the
+/// cascade so a turn is never left unrouted.
+///
+/// `targets` must hold the `strong` and `weak` tiers the classifier scores onto.
+/// The judge, when configured, is called through its own target and is not a
+/// routing destination.
+///
+/// Errors if either threshold in `config` is outside `[0.0, 1.0]` or a tier is
+/// missing from `targets`.
+pub fn stage_router(
```
messiaen: Do we want a factory function like this or a `algorithm::Stage` wrapper around `FallThrough`?
`LlmTaskClassifier` uses a wrapper.  I'm guessing that is easier to configure on the server size?

@nachiketb-nvidia what are your thoughts?

nachiketb-nvidia: I prefer not have function,

something like this is better
```
struct StageRouter {
 inner: FallThrough<...>
}

impl Algorithm for StageRouter {
... # use inner directly here
}
```
Using a function means we move Fallthrough from an internal implementation detail to public API (since it is crossing crates right now)

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -230,6 +265,21 @@ where
                 .await?;
         }
 
+        // 5. Offer the outbound request to the processors, paired with the decision that
```
sabhatinas: @messiaen for clarity, I added this to append system prompts for instance. operations that do not need to go through routing algorithm and depend on routing decision can be here?

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -81,17 +124,20 @@ impl Judge for CapabilityJudge {
...
-            .messages
-            .iter()
-            .rev()
-            .find(|message| message.role == Role::User)
-            .cloned()
-            .into_iter()
-            .collect::<Vec<_>>();
+        // Task-based routing judges the newest user message alone. A configured
+        // window widens that to the surrounding conversation.
+        let mut messages = match self.recent_turn_window {
```
sabhatinas, line 98: @ayushag-nv please review this new method for recent window trimming.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -147,18 +193,25 @@ impl JudgePolicy for TaskClassifierPolicy {
...
-            Some(verdict)
-                if verdict.is_valid()
-                    && !verdict.abstain
-                    && verdict.confidence >= self.config.min_confidence
-                    && verdict.p_solve >= self.threshold(verdict) =>
-            {
-                &self.efficient_target
-            }
-            _ => &self.capable_target,
+        // Judge output is untrusted, so only a complete, valid, non-abstained verdict that
```
sabhatinas, line 165: @messiaen  @ayushag-nv please should review this change as well. When an algorithm does not have a clear decision, there should be an ambigious decision so that cascades and fallbacks can move the needle.

crates/libsy/src/algorithms/stage.rs
sabhatinas: Got it, I am pushing this change

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -54,7 +54,42 @@ impl Decision for FallThroughDecision {
...
+    /// Close a cascade with `target`.
+    pub fn new(target: impl Into<String>) -> Self {
+        Self {
+            target: target.into(),
+        }
+    }
+}
+
+#[async_trait]
+impl<S: Send> Classifier<S> for DefaultTarget {
```
sabhatinas, line 77: No preference on how we are setting default target when all algorithms fail to provide a score. Please help review.

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -230,6 +265,21 @@ where
                 .await?;
         }
 
+        // 5. Offer the outbound request to the processors, paired with the decision that
+        //    routed it. This is the only hook that sees both, so it is where a
+        //    decision-dependent rewrite belongs.
+        for processor in &self.processors {
```
sabhatinas, line 319: @linj-glitch please grab this for your latching mechanism with escalation router. For confirmation count and session affinity, this will be the path for next request to skip all classifiers.

linj-glitch: Sounds good!

crates/libsy/src/algorithms/stage.rs
sabhatinas: done

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: cc: @messiaen for confirmation

messiaen: A decision event contains a request.
And we want to process the decision also, no?

crates/libsy/src/algorithms/fall_through.rs
messiaen: This seems reasonable.

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -208,26 +241,39 @@ where
             });
         };
 
-        // 3. Resolve the target and publish the decision.
+        // 3. Resolve the target and publish the decision. A tier is a fact about the
+        //    target, not about which classifier happened to name it first, so any member
+        //    of the cascade may answer — a terminal fallback that decides a turn would
+        //    otherwise leave the call untiered.
```
messiaen: don't need the paragraphs here.

crates/libsy/src/algorithms/fall_through.rs
messiaen: Yeah we want to pass Decision event here, request event implies it's preprocessing I think

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -512,6 +559,7 @@ mod tests {
                 let kind = match event {
                     Event::Request(_) => "request",
                     Event::Decision { .. } => "decision",
+                    Event::ModelRequest { .. } => "model_request",
```
messiaen: Decision has the request already.

crates/libsy/src/algorithms/llm_class.rs
messiaen: Looks correct to me

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -147,18 +173,25 @@ impl JudgePolicy for TaskClassifierPolicy {
...
+        // reported as ambiguous so the composition around this classifier chooses the
+        // fallback, rather than this policy silently imposing one. The capable target rides
+        // along as the safe suggestion for a caller that reads ambiguous scores.
+        let Some(verdict) = verdict
+            .filter(|v| v.is_valid() && !v.abstain && v.confidence >= self.config.min_confidence)
+        else {
+            return Classification::Ambiguous(vec![Score {
+                target: self.capable_target.clone(),
+                confidence: 0.0,
+            }]);
```
messiaen: should we just return an empty set of scores here?

messiaen (commented):
adding comments still reviewing. But looks good overal

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: I would imagine we should not touch decision here.

sabhatinas: If Im reading it right, decision is not mutable intentionally so.

crates/libsy/src/algorithms/llm_class.rs
sabhatinas: Is it different from returning confidence = 0?

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: ok

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: nvm, applying changes per your comment. passing in decision is enough.

sabhatinas: Done — `Event::Decision` now carries a mutable request, so processors can both bind state and rewrite the outbound request (e.g. tier system prompts) in a single pass. `Event::ModelRequest` is removed entirely.

crates/libsy/src/algorithms/llm_class.rs
sabhatinas: Yes, done — returns `Classification::Ambiguous(vec![])` now. The capable hint was speculative; nothing in the cascade calls `argmax(true)` on this.

crates/libsy/tests/stage_router.rs
```diff
@@ -0,0 +1,347 @@
...
+            .lock()
+            .iter()
+            .filter(|call| call.target != JUDGE)
+            .cloned()
+            .collect()
+    }
+}
+
+#[async_trait]
+impl RoutedLlmClient for RecordingClient {
```
messiaen: Looks like this is defined in `test/prompt` and in `test/stage_router`

Also can these test not go in the source file?

crates/libsy/Cargo.toml
```diff
@@ -31,5 +31,8 @@ tracing.workspace = true
 # SDK + in-memory exporter to assert what the observability layer records.
 opentelemetry_sdk = { version = "0.32", features = ["metrics", "testing"] }
 switchyard-llm-client.workspace = true
+# Decodes the wire-format fixtures the tool-signal tests are written against, so
+# they keep asserting on bodies a real client would send.
+switchyard-translation.workspace = true
```
messiaen: This is a little smelly... Why not just create the llm_request directly.

messiaen: really we don't want even a dev dep of translations

messiaen (approved):

crates/libsy/src/algorithms/llm_class.rs
sabhatinas: 👍

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: 👍

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: Done.

crates/libsy/tests/stage_router.rs
sabhatinas: Moved the stage_router integration tests into `stage.rs` `#[cfg(test)]` and deleted `tests/stage_router.rs`. No more duplicate `RecordingClient`.

crates/libsy/Cargo.toml
sabhatinas: Removed — rewrote the tool_signals test helpers to construct `LlmRequest` directly.

PR #172, opened 2026-07-28
feat(libsy): add EscalationRouter algorithm

linj-glitch:
Implements SWITCH-1054. Adds `EscalationRouter` to libsy: every session starts on the efficient model, an LLM judge reads the trajectory *before* each turn's model call, and a confirmed escalation latches the session to the capable model for the rest of the task. Judging ahead of the call means a turn costs one model call and the target's response — streamed or aggregated — reaches the caller untouched. A judge failure fails open to the efficient tier and never latches.

The router is a `FallThrough` over one classifier holding the whole policy, matching the `LlmTaskClassifier` shape. The confirmation streak lives in the composition's per-session `State` and doubles as the latch — it only reaches the threshold on the turn the judge escalated, and a confirmed session never consults the judge again to clear it:

| Streak | Decides | Tier |
| --- | --- | --- |
| `>= confirmations` | session already escalated — judge not called | strong |
| raised to `confirmations` this turn | judge verdicts have confirmed | strong |
| below `confirmations` | everything else | weak |

The last row is unconditional, which is what keeps a judge outage from failing the turn.

Because the streak is both the counter and the latch, there is no separate affinity component and no second per-session store. Escalation is session-scoped throughout; per-subagent escalation would need both the streak and the latch keyed by `(session, agent)` and is not attempted here.

The decision's tier comes from the classifier's `routing_tier`, and its reason from the composition's own `with_decision_reason`. A turn that escalated is distinguishable from one that was already latched by the judge consultation recorded before it in the trace, so no per-rule reason string is carried. Nothing in `core/` or `fall_through.rs` changes.

## Python default changes

This PR also flips five defaults in the Python escalation router. Four of them make Python match the Rust `EscalationJudgeConfig` added here, so the two implementations of the same router do not disagree on main:

| Setting | Before | After | Rust |
| --- | --- | --- | --- |
| `escalate_confirmations` | 1 | 2 | `confirmations: 2` |
| `recent_turn_window` | 14 | 28 | `recent_turn_window: 28` |
| `window_message_chars` | 300 | 500 | `window_message_chars: 500` |
| `max_request_chars` | 12 000 | 18 000 | `MAX_REQUEST_CHARS` |
| `judge_timeout_s` | 5.0 | 30.0 | no equivalent knob |

`judge_timeout_s` is the exception: Rust has no judge-timeout setting — the judge target's own `timeout_secs` governs there — so that value rests on the benchmark rather than on parity. All five are the benchmarked configuration.

**Existing Python users will see different routing after this merges.** `escalate_confirmations` 1→2 is the behavioural one: the judge must now confirm on a second judged turn before the strong latch fires, which on the benchmarked workload suppressed roughly two thirds of escalations at an equal-or-better solve rate.


crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,629 @@
...
+
+    fn pin(&self, session_id: &str) {
+        let mut pins = self.pins.lock();
+        if pins.len() >= MAX_PINS {
+            if let Some(evicted) = pins.keys().next().cloned() {
+                pins.remove(&evicted);
+            }
+        }
+        pins.insert(session_id.to_string(), ());
+    }
```
ayushag-nv: we should not be re-inventing pinning here. Try to use AffinityRouter stuff in here if possible.

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,629 @@
...
+        let prompt_schema = serde_json::to_string_pretty(prompt_schema).map_err(|error| {
+            LibsyError::AlgorithmError {
+                message: format!("escalation prompt schema could not be rendered: {error}"),
+            }
+        })?;
+        Ok(JudgeConfig {
+            system_prompt: PROMPT_TEMPLATE.replace("{{RESPONSE_SCHEMA}}", &prompt_schema),
+            response_schema: Some(response_schema),
+        })
+    }
```
ayushag-nv: All this seems to follow the same pattern and duplicate code from llm_class.rs -- can we uniffy ?? Only diff is in the run loop I believe how the latch behaviour work -- for that may be you should extend the AffinityRouter for EscalationAffinity ... ??

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,629 @@
...
+
+    fn reasoning(&self) -> Option<&str> {
+        Some(self.reason)
+    }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+}
```
ayushag-nv: why do we need this ? can't we use Decision directly ??

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,629 @@
...
+
+        let text = completion_text(&agg);
+        match parse_verdict(text.trim()) {
+            Some(v) => v.should_escalate,
+            None => {
+                tracing::warn!(
+                    target: "libsy",
+                    judge_model,
+                    "escalation judge verdict did not parse; skipping escalation"
+                );
```
ayushag-nv: The JudgeClassifier abstraction is generic enough to make a llm call using judge inputs and give u a output ? Why do we need to implement this again ??

ayushag-nv (changes requested):

eric-liu-nvidia (approved):

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Done — replaced the custom `pins` map with `AffinityRouter::new().with_latch_only([capable_name])`. Pinning now goes through the same mechanism as the rest of the codebase.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Done — extracted `load_judge_config(prompt, schema)` into `util/llm_judge.rs` so both classifiers share it. For the run-loop difference (the judge needs the efficient model's response, not just the original request), added minimal `EscalationJudge`/`EscalationPolicy` types that plug into `JudgeClassifier` — latch logic stays in `AffinityRouter`, judge plumbing in the existing abstraction.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Removed — `JudgeClassifier` uses its own internal `JudgeDecision`. No custom struct needed.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Agreed, removed. Added `EscalationJudge` (implements `Judge`, reads the efficient response from `state.extra`) and `EscalationPolicy` (implements `JudgePolicy`, maps verdict to classification). `JudgeClassifier` handles the call.

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,540 @@
...
+            .extra
+            .get(CANDIDATE_KEY)
+            .and_then(|v| {
+                if let StateValue::String(s) = v {
+                    Some(s.as_str())
+                } else {
+                    None
+                }
+            })
+            .unwrap_or("");
```
ayushag-nv: what does a candidate here refer to ?

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,540 @@
...
+                judge_target,
+                EscalationPolicy { capable: capable_name.clone() },
+            ),
+            affinity: AffinityRouter::new().with_latch_only([capable_name]),
+        })
+    }
+}
+
+#[async_trait]
+impl Algorithm<SharedState> for EscalationRouter {
```
ayushag-nv: @linj-glitch Can this be implemented the FallThrough pattern ? We are implementing all the algos using that

crates/libsy/src/algorithms/escalation.rs
linj-glitch: `candidate` is the efficient model's response text — stored in `state.extra` under `CANDIDATE_KEY` just before the judge call. `EscalationJudge::build_request` reads it to construct the prompt for the judge: "here's what the efficient model replied — was it sufficient?" Without it the judge would only see the original request and couldn't evaluate actual output quality.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Agreed it's the right long-term direction — Greg flagged the same thing. The blocker is that `FallThrough` currently has no `Event::Response`: processors only see `Event::Request` and `Event::Decision`, so there's no hook to feed the efficient model's reply into a classifier mid-run. Once `Processor` gets a response event, escalation folds in naturally. Greg said "we can do that later," so keeping it as a standalone algo for now and we can refactor when the event is there.

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,1150 @@
...
+const MAX_REQUEST_CHARS: usize = 18_000;
+
+/// Completion budget for one judge reply, covering any reasoning the judge model emits
+/// alongside the verdict.
+///
+/// A runaway guard, not a budget: output tokens cost what they generate, so a generous cap is
+/// nearly free, while a tight one truncates mid-reasoning into unparseable JSON and fails the
+/// judge open on every call. Sized well above the verdict itself — the benchmarked judge's
+/// `reason` ran ~40 tokens at the median and ~136 at its longest — leaving the remainder as
+/// reasoning headroom.
```
ayushag-nv: do we need so many lines of comments for 1 variable ? Can we just do it either 1 or 2 lines

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,1150 @@
...
+pub struct EscalationJudgeSettings {
+    /// Consecutive escalate verdicts required before the latch fires.
+    ///
+    /// `1` latches on the first verdict. Higher values filter one-shot eager verdicts — a
+    /// single failed command misread as a pattern — while keeping recall on real trouble,
+    /// which by definition persists across turns. On the benchmarked workload `2` suppressed
+    /// roughly two thirds of escalate verdicts, so this is the router's main cost dial.
+    ///
+    /// The streak is keyed on the session id and held per process. A request without a
+    /// session id cannot accumulate one, so at `2` or higher it can never escalate.
```
messiaen:
```suggestion
    /// Consecutive escalate verdicts required before the latch fires.
    /// `1` latches on the first verdict. 
    /// `2` or higher requires session_id
```

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,1150 @@
...
+        }
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+struct EscalationVerdict {
+    escalate: bool,
+    #[allow(dead_code)]
+    reason: String,
```
messiaen: can reason be removed?

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,1150 @@
...
+/// The tuning surface for the trajectory judge.
+///
+/// Deliberately small: these are the three settings an audit of the Python router's twenty-odd
+/// knobs found to actually change routing outcomes. Everything else it exposed is either a
+/// fixed invariant (see the `*_CHARS` constants above) or operational plumbing.
+///
+/// Defaults are the benchmarked configuration; omitted fields take those values.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
+pub struct EscalationJudgeSettings {
```
messiaen:
```suggestion
pub struct EscalationJudgeConfig {
```

crates/libsy/src/algorithms/escalation.rs
messiaen: @linj-glitch looks like it's still there?

I think we need a default Decision struct. `SimpleDecision` or similar for example

switchyard/lib/processors/escalation_judge_request_processor.py
messiaen, line 1: Why are we changing python files?

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -293,28 +294,8 @@ impl LlmTaskClassifier {
...
-        let prompt_schema = serde_json::to_string_pretty(prompt_schema).map_err(|error| {
-            LibsyError::AlgorithmError {
-                message: format!("capability prompt schema could not be rendered: {error}"),
-            }
-        })?;
-        Ok(JudgeConfig {
-            system_prompt: PROMPT_TEMPLATE.replace("{{RESPONSE_SCHEMA}}", &prompt_schema),
-            response_schema: Some(response_schema),
-        })
+        load_judge_config(PROMPT_TEMPLATE, SCHEMA_TEMPLATE)
```
messiaen, line 331: Not sure I follow why this change was made.

messiaen (commented):
Will continue review.  Here are some initial comments.

Overall I think it can be simpler.

switchyard/lib/processors/escalation_judge_request_processor.py
linj-glitch: These are the benchmarked defaults, and four of the five make Python match the Rust `EscalationJudgeConfig` this PR adds: `escalate_confirmations` 1→2, `recent_turn_window` 14→28, `window_message_chars` 300→500, `max_request_chars` 12k→18k. `system_chars` and `first_user_chars` already agreed.

Splitting them out would land the Rust router on main while the Python one still routes on different numbers — same router, two behaviours.

`judge_timeout_s` 5→30 is the exception: Rust has no judge-timeout setting (the judge target's own `timeout_secs` governs there), so that one rests on the benchmark rather than on parity.

Your underlying point stands though — the title doesn't advertise a Python behaviour change. I've added a section to the PR description calling it out.

crates/libsy/src/algorithms/llm_class.rs
linj-glitch: The escalation judge needs exactly this loading — render the schema into the prompt's `{{RESPONSE_SCHEMA}}` placeholder, keep the same object as the structured-output format — so the choice was to share it or copy those 20 lines into `util/escalation.rs`. The body moved verbatim to `util/llm_judge.rs::load_judge_config(prompt_template, schema_template)`, next to the `Judge` / `JudgeClassifier` / `JudgePolicy` traits both algorithms already share. `llm_class` keeps its one-line wrapper, so nothing about how it loads its own prompt changed.

It also retires the TODO on the line above:

> TODO: Move towards more generic loading and parsing of config when we multiple prompts and schemas to handle for same algorithm

A second algorithm with its own prompt and schema is what this PR adds, so that seemed like the moment to do it.

Only behavioural delta: the three error messages lost their `capability ` prefix, since the function now serves both judges — e.g. `capability response schema is invalid` → `response schema is invalid`. Construction-time only, and it fires solely on a malformed packaged asset.

Happy to revert it and give escalation its own copy if you'd rather keep the assets loaded per algorithm.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: `EscalationDecision` is gone as of the fall-through refactor (c7f27a0) — escalation no longer defines a `Decision` at all. The cascade publishes `FallThroughDecision`, and the tier/rationale that used to be hardcoded per branch now come from whichever classifier won the turn.

On the broader point: agreed, and the duplication is worse than one type. Four `Decision` impls in production today, ~73 lines, and three of them carry no payload at all:

| | Fields | Differs by |
| --- | --- | --- |
| `NoopDecision` | `model` | — |
| `PassthroughDecision` | `model_id` | nothing but the field name |
| `JudgeDecision` | `model` | `is_routed_call() == false`, static reason |
| `FallThroughDecision` | `model`, `tier`, `reasoning` | — |

`NoopDecision` and `PassthroughDecision` are the same struct twice. Eleven more files re-roll their own for tests and examples.

One type covers all four:

```rust
pub struct SimpleDecision { /* model, tier, reasoning, routed */ }

SimpleDecision::new("strong")
    .with_tier("strong")
    .with_reasoning("judge escalated the run")
SimpleDecision::new("judge").unrouted()   // JudgeDecision
```

Home would be `switchyard-protocol`, next to the `Decision` trait, re-exported from libsy. And the existing names become aliases —

```rust
pub type NoopDecision = SimpleDecision;
pub type PassthroughDecision = SimpleDecision;
pub type FallThroughDecision = SimpleDecision;
```

— so nothing public is removed and the `RandomDecision` downcast in `rand.rs` keeps working.

Net is comfortably negative in libsy, and consumers get one type to downcast to instead of four. Happy to land it here or as a follow-up — it touches four algorithms plus the protocol crate, so it may read cleaner as its own PR. Your call.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Removed

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Fixed, thanks

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -69,6 +106,20 @@ pub struct FallThrough<S = ()> {
...
+impl<S> FallThrough<S>
+where
+    S: Default + Send + 'static,
+{
+    /// Creates a router whose state is per-run: every turn starts from `S::default()`.
+    ///
+    /// For compositions typed over a state their components never read across turns —
+    /// because each keeps its own retention, as an affinity latch does. Nothing is held
+    /// per session id.
+    pub fn new_unretained(targets: LlmTargetSet) -> Self {
```
messiaen: Remove. This is not need

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -212,7 +262,7 @@ where
         let target = self.targets.get_target(&score.target)?;
         let decision: Arc<dyn Decision> = Arc::new(FallThroughDecision {
             selected_model: score.target.clone(),
-            reasoning: (self.decision_reason)(&self.name, &score),
+            reasoning: maybe_reason.unwrap_or_else(|| (self.decision_reason)(&self.name, &score)),
```
messiaen: ??

crates/libsy/src/core/classifier.rs
```diff
@@ -75,6 +75,16 @@ pub trait Classifier<S = ()>: Send + Sync {
...
+    /// Why this classifier picked `score`, when it can say more than the composition can.
+    ///
+    /// A cascade's decision reason is otherwise derived from the winning target alone, which
+    /// cannot distinguish two classifiers that select the same target for different reasons —
+    /// a session latched by a prior turn versus one escalated by this one. Returning `None`
+    /// (the default) leaves the composition's own wording in place.
+    fn reasoning(&self, _score: &Score) -> Option<String> {
+        None
+    }
+
```
messiaen: This isn't right. We could return a Decision from `Classification::score` perhaps within `Classification`

crates/libsy/src/algorithms/escalation.rs
```diff
@@ -0,0 +1,655 @@
...
+    }
+}
+
+/// Wires the cascade the wrapper drives.
+fn build_route(
+    efficient_target: LlmTarget,
+    capable_target: LlmTarget,
+    judge_target: LlmTarget,
+    config: EscalationJudgeConfig,
+) -> Result<FallThrough<State>> {
```
messiaen: Let's follow the pattern used by `LLmTaskClassifier`, a struct WRapper

crates/libsy/src/core/classifier.rs
linj-glitch: Agreed — dropped the trait method. `Classification` now carries a `Decided(Arc<dyn Decision>)` variant, so a classifier that knows which of its own rules fired returns the decision itself and `FallThrough` publishes it verbatim. Fixed in c2e7ce5.

crates/libsy/src/algorithms/fall_through.rs
linj-glitch: Removed. `new()` is back to its own body, and the escalation router builds with `new_with_state` like `LlmTaskClassifier` does. Fixed in c2e7ce5.

crates/libsy/src/algorithms/fall_through.rs
linj-glitch: Gone — that was the consumer side of the `Classifier::reasoning` hook. The cascade loop now takes the classifier's own `Decision` when it returns `Classification::Decided`, and only builds a `FallThroughDecision` otherwise:

```rust
let decision: Arc<dyn Decision> = match classification {
    Classification::Decided(decision) => decision,
    _ => Arc::new(FallThroughDecision {
        selected_model: score.target.clone(),
        reasoning: (self.decision_reason)(&self.name, &score),
        tier: classifier.routing_tier(&score.target),
    }),
};
```

Fixed in c2e7ce5.

crates/libsy/src/algorithms/escalation.rs
linj-glitch: Done. `build_route` is gone — the wiring is inline in `EscalationRouter::new()`, and the generic `Labelled<C>` decorator is replaced by one concrete `EscalationClassifier` that owns the affinity latch and the `ConfirmedJudge`, implements both roles, and applies the three ordered rules in `score()`. It publishes an `EscalationDecision` naming the tier and the rule that fired, which is what the `Classification::Decided` variant is for.

Two things fell out: `DefaultTarget` is no longer needed (removed, along with its export), and the escalation router now uses `new_with_state`.

Fixed in c2e7ce5.

linj-glitch:
Two commits on the escalation router, and how they were verified.

**`73c7318` — the judged turn was off by one.** `conversation_turn` returned `assistants + 1`, which was right when the judge ran *before* the turn's call. It now runs after, on a transcript with this turn's reply already appended, so every judge prompt was labelled `Conversation turn N+1`. The unit test missed it by exercising the helper directly instead of the appended-reply path; it now asserts the real contract.

**`184d7f8` — judges and reasoning models.** Targets a route consults as a judge (`judge_target`, `classifier_target`, or a `stage_router` classifier) get `chat_template_kwargs.enable_thinking = false` folded into their `extra_body`; serving targets are untouched. A verdict is a boolean the router parses, so thinking buys no routing quality while costing latency on the request path of every unlatched turn, and it keeps the judge's completion budget honest — a judge that spends its budget thinking and is cut off before answering has no verdict to read. No new config field: writing `extra_body.chat_template_kwargs` yourself takes the key back, so `{ enable_thinking = true }` opts a judge in.

This commit also **removes** the `reasoning_content` fallback added in `4f13352`. That fallback was chasing the wrong symptom. The empty-`content` replies that motivated it are `finish_reason: length` — the model still thinking when it hit the cap — and in that case `reasoning_content` holds prose with no verdict in it, so the fallback failed to parse and errored exactly as the direct parse does. Given headroom and a strict `response_format`, the JSON lands in `content` where it belongs, thinking on or off. The judge reads `content` only.

## Verified

Latch, live against Inference Hub — weak `nemotron-3-ultra`, strong `claude-opus-5`, judge `nemotron-nano-3.5-preview`, `confirmations = 1`, one session, four turns of a stuck-agent loop:

| Turn | Served | Duration | Upstream calls |
|---|---|---|---|
| 1 | nemotron-3-ultra | 4193 ms | efficient + judge |
| 2 | nemotron-3-ultra | 2552 ms | efficient + judge |
| 3 | **claude-opus-5** | 4610 ms | efficient + judge + capable |
| 4 | claude-opus-5 | **1941 ms** | capable only |

Turn 4 is the fastest of the run despite the most expensive model — one call, no efficient, no judge. `/metrics` confirms the shape: 3 efficient, 3 judge, 2 capable, and `switchyard_decisions_total` splits 2 weak / 2 strong with the tier labels attached.

Judge parsing was checked live with a judge opted into thinking: three turns, three parsed verdicts, zero `judge verdict unavailable` warnings — the verdict arrives in `content` regardless.

633 workspace tests pass; `cargo clippy --workspace --all-targets` and `cargo fmt --check` clean.

## Known gaps, not addressed here

**Inner model calls are unattributed.** The escalation classifier's efficient call and the shared judge call pass `Context::default()`, so they record `algorithm=""` while only the final routed call carries `algorithm="escalation"` — six of eight upstream calls in the run above. I had a commit threading the run's context onto `Driver`; it was reverted as the wrong shape, since context is threaded through the algorithm by design. The underlying constraint is that `Classifier::score` takes no `ctx`, so a classifier making its own call has nothing else to pass. Unchanged from before this PR.

**Judge spend is missing from cost accounting.** The judge appears in `switchyard_llm_calls_total` and the duration histogram but not in `switchyard_prompt_tokens_total` / `completion_tokens_total`, because token counters live in the server crate and are fed from `trace.last()` — inner calls never reach that layer. Python records judge usage into the classifier bucket to keep the escalation benchmark's cost math honest. Separately, `completion_tokens_total{tier="weak"}` includes the efficient completion discarded on an escalating turn.

Together these mean the router's measured cost is both understated and unattributable to the algorithm. Worth its own issue before benchmark numbers are read from these metrics.

crates/libsy/src/core/algorithm.rs
```diff
@@ -118,18 +118,30 @@ pub struct Driver {
...
-    /// [`run_stream`](Algorithm::run_stream).
-    pub(crate) fn new() -> Self {
+    /// [`run_stream`](Algorithm::run_stream), which stamps `context` first.
+    pub(crate) fn new(context: Context) -> Self {
         Self {
             driver: TypeErasedDriver::new(),
             routed_call: Arc::new(Mutex::new(None)),
+            context,
         }
     }
```
messiaen, line 131: This doesn't make sense. Context is passed through the algorithm by design. overriding it or storing it defeats the purpose

switchyard/lib/processors/escalation_judge_request_processor.py
ryan-lempka: Wait are we trying to change the algorithm and do the port at the same time?

Shouldn't we port the algo as is since that's what the benchmark results were generated against?

crates/libsy/src/prompts/escalation/schema.json
ayushag-nv, line 1: is this same as the python one ?

PR #173, opened 2026-07-28
feat(llm-client): add bounded upstream retries

nachiketb-nvidia:
## What

- Add configurable bounded retries to `switchyard-llm-client`.
- Expose `max_retries` in `switchyard-server` TOML with a default of `2` and a maximum of `10`.
- Retry buffered response-body transport failures while leaving streaming body failures outside the replay boundary.

## Why

Transient transport failures, timeouts, HTTP 408/429 responses, and upstream 5xx responses should not immediately fail a request. The retry budget and delay policy must remain explicit and bounded because retries can replay billable requests and extend total latency.

## How

- Retry only transport failures, timeouts, HTTP 408/429, and 5xx responses.
- Honor `Retry-After` seconds or HTTP dates with a 60-second cap; otherwise use bounded exponential backoff from 250 ms to 2 seconds.
- Return the final typed upstream error unchanged after exhausting the retry budget.
- Emit one `libsy.upstream_attempt` span per physical attempt with status, retry, and delay fields.
- Collect buffered bodies inside the retry boundary; return successful streaming responses immediately and never replay stream-body failures.
- Consolidate the repeated truncated-body cases into one table-driven test.

## What to review

- Retry classification and exhaustion behavior.
- Buffered versus streaming replay boundaries.
- `Retry-After` parsing and bounded backoff.
- Attempt metrics and tracing.
- Server retry defaults and maximum validation.

## Validation

- `cargo fmt --all --check`
- `cargo clippy -p switchyard-llm-client -p switchyard-server --all-targets -- -D warnings`
- `cargo test -p switchyard-llm-client -p switchyard-server`

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -5,9 +5,11 @@
 //! request, call the configured backend over HTTP, decode the neutral response.
 
 use std::collections::{BTreeMap, HashMap};
+use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
 use futures_util::StreamExt;
+use reqwest::header::{HeaderMap, RETRY_AFTER};
 use reqwest::RequestBuilder;
 use serde_json::{Map, Value};
 use switchyard_protocol::{
@@ -17,6 +19,7 @@ use switchyard_translation::{
     decode_aggregated_response, decode_request, decode_stream, encode_aggregated_response,
     encode_request, encode_stream, WireFormat,
 };
+use tracing::Instrument;
 
 use crate::backend::Backend;
 use crate::error::{LlmClientError, Result};
@@ -42,6 +45,10 @@ const RESERVED_HEADERS: &[&str] = &[
     "anthropic-version",
     "content-type",
 ];
+
+const INITIAL_RETRY_DELAY: Duration = Duration::from_millis(250);
+const MAX_RETRY_BACKOFF: Duration = Duration::from_secs(2);
+const MAX_RETRY_AFTER: Duration = Duration::from_secs(60);
 
 /// How one model is served: the `default_backend` used when the request does not
 /// pin a wire format, plus any `other_backends` reachable over additional formats.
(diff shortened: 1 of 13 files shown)


crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -334,6 +338,169 @@ impl FormatCodec for AnthropicMessagesCodec {
...
+        .into_iter()
+        .filter(|block| !is_unsigned_thinking_block(block))
+        .collect::<Vec<_>>();
+    let content = if kept.is_empty() {
+        Value::String(String::new())
+    } else {
+        Value::Array(kept)
+    };
+    message.insert("content".to_string(), content);
+    Value::Object(message)
```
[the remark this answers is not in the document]
nachiketb-nvidia: fixed

crates/libsy-llm-client/src/client.rs
```diff
@@ -162,46 +170,131 @@ impl TranslatingLlmClient {
...
+        }
+        let streaming = endpoint.allows_streaming()
+            && body.get("stream").and_then(Value::as_bool).unwrap_or(false);
+        let url = endpoint.url(backend);
+
+        let max_retries = u64::from(backend.max_retries());
+        let max_attempts = max_retries + 1;
+        let mut attempt = 0_u64;
+        loop {
+            let span = tracing::info_span!(
```
grahamking: I think this should be debug, otherwise on a normal request we would break the "one INFO line per request" goal (Apache HTTP server style).

crates/libsy-llm-client/src/client.rs
```diff
@@ -162,46 +170,131 @@ impl TranslatingLlmClient {
...
+                    }
+                    span.record("will_retry", will_retry);
+                    if !will_retry {
+                        return Err(failure.error);
+                    }
 
-        let builder = self.client.post(url).json(&body);
+                    let delay = retry_delay(attempt, failure.retry_after);
+                    span.record("retry_delay_ms", duration_millis(delay));
+                    drop(span);
```
grahamking, line 224: These two lines don't compute. Does `span.record` have a side effect we want?

crates/libsy-llm-client/src/client.rs
grahamking: > I agree with the existing review comment: these should be debug_span! to preserve the one-INFO-record-per-request logging contract when span lifecycle output is enabled

Had Codex review it, and will you look at that :grin:

grahamking:
I got this from Codex, looks legit. Could you check it?

> Retry metrics misclassify failures. `crates/libsy-llm-client/src/client.rs`:241 records an upstream attempt immediately after receiving headers. If a buffered 200 body is truncated, the new logic retries it, but the metric has already counted that failed attempt as success. 
>
> Additionally, `crates/libsy-llm-client/src/metrics.rs`:8 labels only 429/500/504 as retryable, while this PR retries 408 and every 5xx. This will understate failures in the direct-endpoint error-rate dashboards added on main. Record the metric after buffered-body collection and align its outcome classification with the retry policy.

This is likely because of the metrics stuff I just landed. Might need a rebase to see the issue.

nachiketb-nvidia:
> I got this from Codex, looks legit. Could you check it?
> 
> > Retry metrics misclassify failures. `crates/libsy-llm-client/src/client.rs`:241 records an upstream attempt immediately after receiving headers. If a buffered 200 body is truncated, the new logic retries it, but the metric has already counted that failed attempt as success.
> > Additionally, `crates/libsy-llm-client/src/metrics.rs`:8 labels only 429/500/504 as retryable, while this PR retries 408 and every 5xx. This will understate failures in the direct-endpoint error-rate dashboards added on main. Record the metric after buffered-body collection and align its outcome classification with the retry policy.
> 
> This is likely because of the metrics stuff I just landed. Might need a rebase to see the issue.

changed which codes are retryable!

crates/libsy-llm-client/src/client.rs
nachiketb-nvidia: yeup! I changed it to debug span, don't need too much otel

crates/libsy-llm-client/src/client.rs
nachiketb-nvidia: record updates the retained attempt span; dropping it before backoff keeps sleep time out of attempt latency.

unless we want to keep potential sleep time out of latency?

grahamking (approved):

crates/libsy-llm-client/src/client.rs
grahamking: `span.record` updates a global, so we do need it.

PR #174, opened 2026-07-28
feat(libsy): add prefill-probe classifier

anniesurla:
## Summary

- implement learned prefill-probe routing directly in `libsy` as a `Classifier`
- load and validate the v1 safetensors checkpoint, then run scaler, PCA-200, ensemble, and cost-aware inference
- define a transport-independent `PrefillProbe` boundary for hidden-state extraction
- bound successful task decisions in a process-randomized hash-keyed LRU cache
- route probe and inference failures to strong without caching the failure

## Context

This is the first of three smaller PRs replacing #140 in response to maintainer review:

1. `libsy` classifier and checkpoint inference (this PR)
2. vLLM hidden-state transport
3. server configuration and deployment documentation

This PR intentionally contains no HTTP client, filesystem hidden-state transport, Python binding, or server configuration.

## Implementation notes

- CPU inference runs through `tokio::task::spawn_blocking`.
- Raw task text is neither cached nor logged.
- Artifact metadata, tensor names, dtypes, shapes, and finite values are validated at construction.
- Routing uses an explicit weak/strong tier rather than a numeric decision threshold.

## Test plan

The live VPN-dependent test remains ignored by the workspace suite.

## Summary by CodeRabbit

* **New Features**
  * Added intelligent prefill routing that selects between strong and weak completion tiers based on prompt features and configurable cost trade-offs.
  * Added support for loading and validating routing artifacts, including checkpoint metadata and tensor data.
  * Added caching for routing decisions to improve repeated-task performance.
  * Added safe fallback to the strong completion tier when probing or inference fails.
  * Exposed configuration options for probe models, routing targets, costs, and cache capacity.

### DIFF
--- a/crates/libsy/src/algorithms/prefill_probe/policy.rs
+++ b/crates/libsy/src/algorithms/prefill_probe/policy.rs
@@ -1,0 +1,151 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Cost-aware selection between the learned router's weak and strong heads.
+
+use crate::{LibsyError, Result};
+
+/// Completion tier selected by the learned utility policy.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) enum PrefillTier {
+    Weak,
+    Strong,
+}
+
+/// Validated utility policy for two completion targets.
+#[derive(Clone, Copy, Debug)]
+pub(super) struct CostAwareRoutingPolicy {
+    lambda: f64,
+    normalized_weak_cost: f64,
+    normalized_strong_cost: f64,
+}
+
+impl CostAwareRoutingPolicy {
+    /// Validates the policy and min-max normalizes costs across weak and strong.
+    ///
+    /// With two targets, normalization makes routing depend on cost ordering,
+    /// not the magnitude of the difference between the configured costs.
+    pub(super) fn new(lambda: f64, weak_cost: f64, strong_cost: f64) -> Result<Self> {
+        if !lambda.is_finite() || !(0.0..=1.0).contains(&lambda) {
+            return Err(policy_error(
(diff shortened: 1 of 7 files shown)


ryan-lempka:
@anniesurla can you rebase?

crates/libsy/src/algorithms/prefill_probe/policy.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub(super) fn new(lambda: f64, weak_cost: f64, strong_cost: f64) -> Result<Self> {
+        if !lambda.is_finite() || !(0.0..=1.0).contains(&lambda) {
+            return Err(policy_error(
+                "routing policy lambda must be finite and in [0.0, 1.0]",
+            ));
+        }
+        validate_cost("weak_cost", weak_cost)?;
+        validate_cost("strong_cost", strong_cost)?;
+
+        let (normalized_weak_cost, normalized_strong_cost) = if weak_cost == strong_cost {
```
ryan-lempka, line 37: @anniesurla is there a longer term goal to take switchyards token / model pricing into account within the cost aware routing policy?

ryan-lempka (commented):
Have one question on longer term direction with cost ordering and how it relates to switchyard pricing. For this one it would be good to get a review from @sabhatinas given she has done work in a similar vein. 

Once you have wrapped the libsy staged implementation - can you take a look @sabhatinas?

crates/libsy/src/algorithms/prefill_probe/policy.rs
anniesurla: Yes. This is feasible. 

This PR currently establishes the transport-independent classifier and scoring rule. I will add the token-aware pricing integration in the probe-client and server follow-ups.

Switchyard can resolve the target models’ input/output rates, while the prefill probe supplies a routing-time prompt-token estimate and the checkpoint supplies per-target median output lengths and training normalization bounds. The integration layer can then compute (C_{k,q}) and pass it into the existing lambda policy.

nachiketb-nvidia:
closing as stale, keeping as reference!

PR #181, opened 2026-07-29
feat: switchyard rust server image

bhuvan002:
## What

This PR adds a separate Dockerfile that builds the Rust-based `switchyard-server`. It also raises the Axum request-body limit to 32 MiB (regression found during a terminal bench 2.1 run).

## Why

As we move from the Python frontend to the Rust frontend, we need an image capable of using the ruse server. The larger body limit supports provider-sized Claude Code requests.

Closes #

## How tested

- [x] `uv run pytest tests/ --ignore=tests/e2e` — 1,720 passed
- [x] `cargo fmt`, `cargo clippy`, and `cargo test -p switchyard-server`
- [x] Image smoke-tested as an arbitrary UID with a read-only root filesystem

## Checklist

- [x] No new Python classes or public symbols.
- [x] Unit test added for requests exceeding Axum's previous limit.

## Notes for reviewers

The Dockerfile is the exact Cloud Build path qualified end to end with scaled-evals.

### DIFF
--- a/benchmark/switchyard-rust-server.Dockerfile
+++ b/benchmark/switchyard-rust-server.Dockerfile
@@ -1,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+ARG RUST_VERSION=1.96.1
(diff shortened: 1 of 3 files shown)


grahamking (approved):

nachiketb-nvidia (approved):
is it usable with scaled evals and benchmarks directly now?

PR #183, opened 2026-07-29
refactor(translation): normalize Anthropic requests through IR

nachiketb-nvidia:
## What

- Normalize outbound Anthropic requests through the typed `LlmRequest` IR.
- Remove backend-owned JSON rewriting for system roles, reasoning, tool IDs, and thinking blocks.
- Scope provider extensions to their decoded source format so Anthropic-native fields do not leak across formats.
- Preserve exact native requests only when nested `cache_control` annotations require it.

## Why

Request semantics should be normalized once by source decoding and target encoding. The previous implementation moved a large `serde_json::Value` post-processor into the translation crate, but still duplicated behavior already represented by the IR and made provider ownership unclear.

## How

- Record the decoded `source_format` on `LlmRequest`.
- Decode Anthropic message-level system/developer turns into typed instructions.
- Decode OpenAI-style `reasoning_effort` and encode it as Anthropic `thinking` plus `output_config`.
- Re-emit top-level extensions only when their source format is Anthropic; retain native `context_management` and drop cross-provider extensions.
- Sanitize and collision-disambiguate tool IDs while encoding typed tool calls/results.
- Omit unsigned thinking and messages left empty after typed encoding.
- Keep the translation crate exact-round-trip contract unchanged; the native backend requests canonical encoding except for nested Anthropic cache markers that the current IR cannot represent.

## What to review

- `LlmRequest::source_format` and provider-extension ownership.
- Anthropic decode/encode behavior for instructions, reasoning, tools, and thinking.
- The narrow exact-replay exception for native `cache_control` blocks.
- Removal of the duplicate backend normalizer.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace`
- `cargo test -p switchyard-components --test adversarial_native_backends`

### DIFF
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -3,6 +3,8 @@
 
 //! Buffered codec for Anthropic Messages request and response JSON.
 
+use std::collections::BTreeMap;
+
 use serde_json::{json, Map, Value};
 
 use crate::codecs::common::{is_known_role_name, provider_extensions, text_from_blocks};
@@ -26,7 +28,7 @@ use crate::util::{
 use crate::util::{
     json_string, push_lossy, stable_id, string_value, validate_request_capabilities,
 };
-use crate::util::{normalize_anthropic_tool_use_ids, sanitize_anthropic_tool_use_id};
+use crate::util::{mapped_tool_id, sanitize_anthropic_tool_use_id};
 
 /// Format codec for Anthropic Messages payloads.
 pub struct AnthropicMessagesCodec;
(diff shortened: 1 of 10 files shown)


crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -334,6 +338,173 @@ impl FormatCodec for AnthropicMessagesCodec {
     }
 }
 
+// Makes both translated and exactly preserved requests valid for conservative
+// Anthropic Messages endpoints without dropping unknown provider extensions.
+fn normalize_outbound_request(mut body: Value) -> Value {
+    let Value::Object(object) = &mut body else {
+        return body;
+    };
+    object.remove("reasoning_effort");
```
grahamking: Why this?

grahamking (approved):

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
nachiketb-nvidia: moved towards using IR based resolution instead, so no removing anymore

PR #197, opened 2026-07-30
refactor(libsy): internalize fall-through routing

nachiketb-nvidia:
## What

- Make `FallThrough` and `DefaultTarget` crate-private implementation details.
- Keep `RandomDecision` available through its existing public alias.
- Remove external tests that assembled `FallThrough` directly.
- Remove the now-unused `with_component` builder and register dual-role test components explicitly.

## Why

Callers should use the assembled public algorithms (`Random`, `LlmTaskClassifier`, and `StageRouter`) instead of coupling to their shared cascade implementation.

## How

The algorithms module now re-exports the composition only within libsy. Tests that depended on the public implementation surface are deleted while the existing internal fall-through coverage remains.

## What to review

- The public API no longer exposes `FallThrough` or `DefaultTarget`.
- `RandomDecision` remains public and downcastable.
- No runtime routing behavior changes.

## Validation

```text
cargo test -p switchyard-libsy -p switchyard-server
```

## Summary by CodeRabbit

* **API Changes**
  * Removed the combined component-registration helper from fall-through routing.
  * Components that serve multiple routing roles must now be registered separately.
  * Limited visibility of fall-through routing types and removed one public export.

* **Tests**
  * Updated affinity-routing test setup for separate processor and classifier registration.
  * Removed outdated prompt-selection and stage-router integration tests.


crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -154,22 +154,6 @@ where
...
-    /// Registering the two separately is easy to half-wire: omit the processor and the
-    /// classifier silently never sees an assignment. This registers both at once.
-    pub fn with_component<T>(self, component: Arc<T>) -> Self
-    where
-        T: Processor<S> + Classifier<S> + 'static,
-    {
-        self.with_processor(component.clone())
-            .with_classifier(component)
-    }
-
```
ayushag-nv, line 188: claude was repeatedly adding this somehow. Where is that landed from ??

ayushag-nv (approved):

crates/libsy/src/algorithms/fall_through.rs
messiaen: I think it was part of the original implementation. Seems useful, no?

crates/libsy/tests/prompts.rs
messiaen, line 1: are these tests elsewhere?

messiaen (approved):

crates/libsy/src/algorithms/fall_through.rs
nachiketb-nvidia: just a wrapper function doing what would need to be done at a different level, I can bring it back if needed

grahamking (approved):

crates/libsy/tests/prompts.rs
nachiketb-nvidia: moved the test elsewhere!

PR #198, opened 2026-07-30
feat(libsy): remember context-window evictions for the session

ryan-lempka:
A target that overflows its context window is now remembered for the rest of the session, so later turns skip it instead of rediscovering the same overflow on every request.

An evicted target does not come back for the life of the session, so a conversation that compacts stays on the larger model. Re-admitting a model to the pool requires further work.

This brings the context window awareness in line with how we were handling such awareness in the Python stack before: https://github.com/NVIDIA-NeMo/Switchyard/blob/main/switchyard/lib/profiles/chain.py

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session-based fallback routing after context-window overflows.
  * Previously failed targets are avoided for the remainder of the session, reducing repeated failures.
  * Different sessions now maintain independent fallback behavior.
  * Routing continues to work after all eligible targets have been temporarily exhausted.

### DIFF
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -31,7 +31,7 @@ use crate::{
 type SessionStates<S> = Mutex<HashMap<String, Arc<AsyncMutex<S>>>>;
 
 /// Bounds the eviction cache on a long-running server.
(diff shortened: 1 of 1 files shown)


ayushag-nv (approved):
Looks good to me

PR #199, opened 2026-07-30
feat(libsy): expose task classifier and stage router in Python

nachiketb-nvidia:
## What

- Add `llm_task_classifier` and `stage_router` Python factories for libsy's
  `LlmTaskClassifier` and `StageRouter` algorithms.
- Expose every existing routing knob while preserving Rust defaults for optional settings.
- Export both factories through `switchyard.libsy.algorithms` with Python type declarations.

## Why

Python-hosted clients can currently run only the no-op and random libsy algorithms. These bindings make the two production routing algorithms available through the same minimal opaque `Algorithm` API.

## How

The PyO3 factories convert bound `LlmTarget` objects and keyword arguments directly into the existing Rust config structs. Construction and validation remain owned by libsy; no routing behavior is duplicated in Python.

## What to review

- Classifier target ordering and `TaskClassifierConfig` field mapping.
- Stage-router picker, notes, prompts, and optional classifier fallback mapping.
- Optional values retain their existing Rust defaults.

## Validation

- `cargo check -p switchyard-py`
- `uv run ruff check switchyard/libsy/algorithms.py switchyard_rust/libsy.py`
- Commit hooks: cargo clippy, mypy, and commitlint

No tests were added or run for this binding-only change.

### DIFF
--- a/switchyard/libsy/algorithms.py
+++ b/switchyard/libsy/algorithms.py
@@ -3,9 +3,9 @@
 
 """Factories for Rust-owned libsy algorithms."""
 
-from switchyard_rust.libsy import llm_classifier as llm_classifier
+from switchyard_rust.libsy import llm_task_classifier as llm_task_classifier
 from switchyard_rust.libsy import noop as noop
 from switchyard_rust.libsy import random as random
 from switchyard_rust.libsy import stage_router as stage_router
 
-__all__ = ["llm_classifier", "noop", "random", "stage_router"]
+__all__ = ["llm_task_classifier", "noop", "random", "stage_router"]
--- a/switchyard/libsy/algorithms.py
+++ b/switchyard/libsy/algorithms.py
@@ -3,7 +3,9 @@
 
 """Factories for Rust-owned libsy algorithms."""
 
+from switchyard_rust.libsy import llm_classifier as llm_classifier
 from switchyard_rust.libsy import noop as noop
 from switchyard_rust.libsy import random as random
+from switchyard_rust.libsy import stage_router as stage_router
 
-__all__ = ["noop", "random"]
+__all__ = ["llm_classifier", "noop", "random", "stage_router"]
--- a/switchyard_rust/libsy.py
+++ b/switchyard_rust/libsy.py
@@ -15,7 +15,7 @@ _EXPORTS = frozenset(
         "Algorithm",
         "LibsyError",
         "LlmTarget",
-        "llm_classifier",
+        "llm_task_classifier",
         "noop",
         "random",
         "stage_router",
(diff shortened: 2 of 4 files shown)


switchyard/libsy/algorithms.py
```diff
@@ -3,7 +3,9 @@
 
 """Factories for Rust-owned libsy algorithms."""
 
+from switchyard_rust.libsy import llm_classifier as llm_classifier
 from switchyard_rust.libsy import noop as noop
 from switchyard_rust.libsy import random as random
+from switchyard_rust.libsy import stage_router as stage_router
 
-__all__ = ["noop", "random"]
+__all__ = ["llm_classifier", "noop", "random", "stage_router"]
```
ayushag-nv: can u name it llmtaskclassifier or something -- the original name is that

switchyard_rust/libsy.py
```diff
@@ -51,6 +61,40 @@ def noop() -> Algorithm: ...
 
     def random(targets: Sequence[LlmTarget]) -> Algorithm: ...
 
+    def llm_classifier(
```
ayushag-nv: sam here: change the name to reflect "task"

ayushag-nv (approved):
looks good for classifier except the "naming"

switchyard_rust/libsy.py
nachiketb-nvidia: even when we can toggle session affinity to on or off?

switchyard_rust/libsy.py
```diff
@@ -51,6 +61,40 @@ def noop() -> Algorithm: ...
...
+        session_affinity: bool = False,
+        message_hash_fallback: bool = False,
+        recent_turn_window: int | None = None,
+    ) -> Algorithm: ...
+
+    def stage_router(
+        capable_target: LlmTarget,
+        efficient_target: LlmTarget,
+        *,
+        picker: str,
```
sabhatinas, line 100: can the LLM classifier confgi be composed instead of re-writing it here? stage router has LLM config inside it

switchyard_rust/libsy.py
```diff
@@ -51,6 +75,30 @@ def noop() -> Algorithm: ...
...
+        confidence_threshold: float,
+        recent_window: int | None = None,
+        escalation_note: str | None = None,
+        deescalation_note: str | None = None,
+        only_on_wrong_signal_escalation: bool = True,
+        capable_system_prompt: str | None = None,
+        efficient_system_prompt: str | None = None,
+        classifier_target: LlmTarget | None = None,
```
sabhatinas: why is this outside classifier_config?

PR #200, opened 2026-07-30
feat(server): report declared model capabilities on /v1/models

elyasmnvidian:
```toml
schema_version = 1

[llm_clients.primary]
format = "openai_chat"
base_url = "http://localhost:8000/v1"

[targets.deep]
id = "nvidia/deepseek-ai/deepseek-v4-pro"
llm_client = "primary"

[routes.smart]
id = "smart"
type = "passthrough"
target = "deep"
```

With this config, the route has no way to declare its context window or tool support. `GET /v1/models` returns this entry:

```json
{
  "id": "smart",
  "object": "model",
  "type": "model",
  "created": 0,
  "owned_by": "switchyard",
  "display_name": "smart",
  "capabilities": {
    "streaming": true,
    "tool_calling": null,
    "context_window": null,
    "supported_inbound_formats": [
      "openai-chat-completions",
      "openai-responses",
      "anthropic-messages"
    ]
  }
}
```

## Fix

`context_window` and `tool_calling` belong on the route because the route ID is the public model returned by `/v1/models`. Two routes can share a target and still advertise different settings.

```toml
[routes.smart]
id = "smart"
type = "passthrough"
target = "deep"
context_window = 1000000
tool_calling = true
```

The same endpoint now reports:

```json
{
  "streaming": true,
  "tool_calling": true,
  "context_window": 1000000,
  "supported_inbound_formats": [
    "openai-chat-completions",
    "openai-responses",
    "anthropic-messages"
  ]
}
```

These fields only describe the route; they do not enforce a request limit. Both are optional, so existing configs still return `null`.

The server rejects target-level declarations and a zero context window during config loading:

```text
unknown field `context_window`
route smart context_window must be greater than zero
```

Routing and request handling are unchanged.

## Test

```text
cargo test -p switchyard-server
```

The endpoint test gives three routes the same target. One route reports `true`, one reports `false`, and one leaves both fields `null`.

This is the operator-declared follow-up from #105.

### DIFF
--- a/crates/switchyard-server/src/config.rs
+++ b/crates/switchyard-server/src/config.rs
@@ -21,7 +21,7 @@ use switchyard_llm_client::{
 };
 use switchyard_protocol::RoutedLlmClient;
 
-use crate::{ServerError, ServerResult, ServerState};
+use crate::{ModelCapabilities, ServerError, ServerResult, ServerState};
 
 const SUPPORTED_SCHEMA_VERSION: u32 = 1;
 const MAX_CONFIGURED_RETRIES: u32 = 10;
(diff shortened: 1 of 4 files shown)


elyasmnvidian:
@coderabbitai review

grahamking:
@elyasmnvidian Should we declare this at the `route` level (synthetic model) rather than the target? That would make it obvious what happens in case of conflict.

```
[route.random
id = "nvidia/random"
type = "random"
targets = ["model_a", "model_b"]
context_window = 1000000
tool_calling = true
```

Maybe Slack thread?

elyasmnvidian:
@grahamking Done. I moved `context_window` and `tool_calling` to the route config and removed the target aggregation. `/v1/models` now reads these values from the route, so two routes can share a target and advertise different settings. I added a test for that case.

grahamking (approved):

PR #202, opened 2026-07-30
fix(examples): forward full ensemble requests and cap concurrent explorations

elyasmnvidian:
```rust
let request = LlmRequest {
    model: Some("auto".to_string()),
    messages: vec![
        Message::text(Role::System, "be terse"),
        Message::text(Role::User, "add 2 and 2"),
    ],
    tools: vec![ToolDefinition {
        name: "calc".to_string(),
        description: Some("do math".to_string()),
        parameters: serde_json::json!({"type": "object"}),
        strict: Some(true),
    }],
    tool_choice: Some(ToolChoice::Required),
    sampling: SamplingParams {
        temperature: Some(0.3),
        ..Default::default()
    },
    stream: true,
    ..Default::default()
};
```

## What happens

The ensemble example rebuilds each candidate request from the last user message. The candidates and the model chosen after exploration receive this smaller request:

```rust
LlmRequest {
    model: Some("auto".to_string()),
    messages: vec![Message::text(Role::User, "add 2 and 2")],
    tools: vec![],
    tool_choice: None,
    sampling: SamplingParams::default(),
    stream: false,
    ..Default::default()
}
```

The system message, tools, tool choice, sampling settings, and stream flag are missing, so the candidates answer a different request.

The example also counts an exploration only after the candidate calls finish. With `exploration_turns = 1`, two requests that arrive together can both explore:

```text
a/model:       2 candidate calls
b/model:       2 candidate calls
judge/haiku:   2 calls
```

## Fix

The example now sends the caller's complete request to every candidate and to the chosen model. It reads candidate streams concurrently before asking the judge to choose, so a slow stream does not delay reading the others.

A request claims an exploration slot before its first model call. Requests wait when all slots are claimed, and a dropped or failed turn returns its slot before a winner is recorded. The completed win counts determine when the example can choose one model.

The example still buffers candidate streams because it must read every candidate response before it can choose a winner.

## Evidence

With `exploration_turns = 1`, two simultaneous requests now make these calls when `b/model` wins the first request:

```text
a/model:       1 candidate call
b/model:       1 candidate call + 1 chosen-model call
judge/haiku:   1 call
```

The regression tests also show that:

- candidate and chosen-model calls receive the original messages, tools, tool choice, sampling settings, and stream flag;
- the judge reads the text from streamed candidate responses;
- a failed candidate stream removes only that candidate;
- a failed exploration returns its slot, so the next request does not hang.

This is a correctness fix in an example, not a performance change, so there is no benchmark.

## Testing

```text
cargo test -p switchyard-libsy --example ensemble

test result: ok. 14 passed; 0 failed
```

### DIFF
--- a/crates/libsy/examples/ensemble.rs
+++ b/crates/libsy/examples/ensemble.rs
@@ -13,10 +13,11 @@
 //!   NVIDIA_API_KEY=... cargo run -p switchyard-libsy --example ensemble
 //!
 //! Unlike the reference routers, this algorithm is **stateful**: the win tally,
-//! turn counter, and committed choice live behind a [`parking_lot::Mutex`] so one
-//! shared `&self` can serve a session's requests concurrently (see the
-//! `Algorithm` docs). In a proxy setup one `EnsembleOrchAlgo` is created per session,
-//! so this state is per-session.
+//! the claimed exploration count, and the committed choice live behind a
+//! [`parking_lot::Mutex`] so one shared `&self` can serve a session's
+//! requests concurrently (see the `Algorithm` docs); a [`tokio::sync::Notify`]
+//! wakes a request waiting for the exploration budget to finish. In a proxy setup
+//! one `EnsembleOrchAlgo` is created per session, so this state is per-session.
 
 use std::collections::BTreeMap;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


elyasmnvidian:
@coderabbitai review

grahamking (approved):

PR #203, opened 2026-07-30
refactor(libsy): classifier::score returns the response alongside the score

linj-glitch:
Extends the Classifier trait so score() returns (Classification, Option<Response>) instead of just Classification. The Option<Response> lets a classifier that already called a model hand back that response directly, so fall_through skips the outbound call and the turn is not paid for twice. All implementors updated. Part 2 of breaking up PR #172.

## Summary by CodeRabbit

- **New Features**
  - Classifiers can now provide an immediate response during routing, avoiding unnecessary model calls.
  - Added support for selecting the highest-confidence classification.
  - Existing routing, affinity, staging, and delegation behavior remains intact.

- **Bug Fixes**
  - Prevented duplicate model calls when a classifier has already produced a response.
  - Improved propagation of classifier decisions through routing pipelines.

### DIFF
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -239,6 +239,39 @@ where
         }
     }
 
+    fn eligible_targets(&self, ctx: &Context) -> usize {
+        self.targets
+            .targets()
+            .iter()
+            .filter(|t| !ctx.is_excluded(&t.semantic_name))
+            .count()
+    }
+
+    fn evicted_in_session(&self, session: Option<&str>) -> Vec<String> {
+        let Some(session) = session else {
+            return Vec::new();
+        };
+        self.session_evictions
+            .lock()
+            .get(session)
+            .map(|targets| targets.iter().cloned().collect())
+            .unwrap_or_default()
+    }
+
+    fn record_eviction(&self, session: Option<&str>, target: &str) {
+        let Some(session) = session else { return };
+        let mut sessions = self.session_evictions.lock();
+        if sessions.len() >= MAX_EVICTION_SESSIONS && !sessions.contains_key(session) {
+            if let Some(oldest) = sessions.keys().next().cloned() {
+                sessions.remove(&oldest);
+            }
+        }
+        sessions
+            .entry(session.to_string())
+            .or_default()
+            .insert(target.to_string());
+    }
+
     /// Calls `target`, falling back to the next eligible target whenever one overflows its
     /// context window, until a call succeeds or every target has been tried.
     ///
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -90,12 +90,15 @@ impl<S: Send> Classifier<S> for DefaultTarget {
         _state: &mut S,
         _request: &mut Request,
         _driver: Option<&Driver>,
-    ) -> Result<Classification> {
+    ) -> Result<(Classification, Option<Response>)> {
         // Zero confidence: this is a fallback, not a judgement.
-        Ok(Classification::Scores(vec![Score {
-            target: self.target.clone(),
-            confidence: 0.0,
-        }]))
+        Ok((
+            Classification::Scores(vec![Score {
+                target: self.target.clone(),
+                confidence: 0.0,
+            }]),
+            None,
+        ))
     }
 }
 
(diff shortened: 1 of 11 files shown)


eric-liu-nvidia (approved):

crates/libsy/src/algorithms/util/subagent.rs
```diff
@@ -47,21 +47,24 @@ where
...
             .is_some_and(Metadata::is_subagent_work);
-        Ok(Classification::Scores(if is_delegated_work {
-            vec![Score {
-                confidence: 1.0,
-                target: self.worker.clone(),
-            }]
-        } else {
-            Vec::new()
-        }))
+        Ok((
```
sabhatinas, line 57: can we isolate changes to llm classifier / fall through in this PR?

crates/libsy/src/algorithms/util/affinity.rs
```diff
@@ -180,18 +180,21 @@ where
         _state: &mut S,
         request: &mut Request,
         _driver: Option<&crate::Driver>,
-    ) -> crate::Result<Classification> {
+    ) -> crate::Result<(Classification, Option<crate::Response>)> {
```
sabhatinas, line 183: Why do we need this change ?

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -192,8 +195,18 @@ where
             }
         };
 
-        self.call_with_overflow_fallback(ctx, &driver, target, decision, request)
-            .await
+        // A classifier that already called a model — because deciding required one, and that
```
sabhatinas, line 221: when does this happen for escalation router? Is the strong/ weak model called as part of LLM judge ?

crates/libsy/src/core/classifier.rs
```diff
@@ -42,6 +42,17 @@ impl Classification {
             }
         }
     }
+
+    /// The top-scoring [`Score`]s, or an empty set when the classifier abstained (an empty set).
```
sabhatinas: we already have the cascaded score's argmax as part of fallthrough implementation.

crates/libsy/src/algorithms/util/affinity.rs
messiaen: Escalation calls the efficient model and judges the response.  if there is no escalation we can simply return the already generated response.

sabhatinas:
Escalation feature should be a configuration knob in LLM classifier to optionally latch to a strong tier.

Consider this trajectory: wwww(l)sssss

1. for turns < latch point marked as w, return resolved decision with score to route to weak model 
2. for turns > latch point marked as s, return resolved decision with score to route to strong model
3. Use session affinity helper to latch future turns by adding a processor post decision.

sabhatinas (changes requested):

sabhatinas:
can you add escalation as a flag/composable config inside llm classifier config? 
1. branch out in base implementation of llm classifier for classifier output 
2. llm classifier gives scores / decisions etc 
3. escalation router gives direct responses (with in-built latching)

crates/libsy/src/algorithms/fall_through.rs
sabhatinas: As per slack, efficient model creates a response inside classifier and need not be re-generated.

crates/libsy/src/algorithms/util/subagent.rs
linj-glitch: SubagentOverride implements Classifier<S>, so when the trait's score() signature changes, every implementation must update — the compiler requires it. The change is purely mechanical: wraps the existing return value in a tuple with None. No logic is altered.

crates/libsy/src/algorithms/util/affinity.rs
linj-glitch: Same reason — AffinityRouter implements Classifier<S>. The only change is wrapping the existing Classification in (_, None). Affinity never makes a model call, so it always returns None for the response.

crates/libsy/src/core/classifier.rs
linj-glitch: Agreed — max_classification() was anticipating PR 4 but doesn't belong here. Removed in the latest commit.

crates/libsy/src/algorithms/fall_through.rs
linj-glitch: Yes — the efficient model is called inside EscalationClassifier, and Option<Response> lets fall_through return that response directly when the judge says no escalation, so the turn isn't paid for twice.

sabhatinas (approved):

PR #205, opened 2026-07-30
feat(libsy): count classifier fail-open fallbacks on /metrics

elyasmnvidian:
```http
POST /v1/chat/completions
Content-Type: application/json

{"model":"switchyard/classifier","messages":[{"role":"user","content":"classify this"}]}
```

When the classifier's judge target returns HTTP 500, Switchyard still sends the request to the capable target. The request succeeds, but `/metrics` does not show that the classifier routed without a verdict:

```console
$ curl -s localhost:4000/metrics | grep classifier_fail_open
# no output
```

## Fix

A failed judge should not fail the caller's request. The classifier still routes without a verdict, and now increments `switchyard_classifier_fail_open_total` for each judge failure.

The counter has two labels: `judge_model` names the configured judge target, and `reason` is one of eight fixed categories. The code derives `reason` from the typed error and HTTP status only. It never uses request or response text as a label.

The non-5xx HTTP category is named `upstream_non_5xx`; an upstream error can carry any status below 500, not only a 4xx status.

## Before and after

Before this change, the request succeeds but the metric query returns nothing. After the change, the same judge failure produces:

```console
$ curl -s localhost:4000/metrics | grep classifier_fail_open
switchyard_classifier_fail_open_total{judge_model="judge-model",reason="upstream_5xx"} 1
```

A valid judge verdict does not increment the counter, even when it selects the same capable target as the fallback. Routing behavior and the caller's response are unchanged; this PR only adds the counter and its documentation. There is no performance claim.

## How tested

```console
cargo test -p switchyard-libsy client_errors_map_to_bounded_fail_open_reasons
cargo test -p switchyard-libsy --test observability classifier_fail_open_records_each_failure_stage
```

The first test covers all eight bounded reason categories, including `call_error`. The second drives the real `LlmTaskClassifier` through judge-call, stream-decoding, and verdict-parsing failures, then checks that a valid verdict is not counted.

### DIFF
--- a/crates/libsy/src/algorithms/util/llm_judge.rs
+++ b/crates/libsy/src/algorithms/util/llm_judge.rs
@@ -22,7 +22,7 @@ use crate::core::algorithm::{Driver, LlmTarget};
 use crate::core::classifier::{Classification, Classifier};
 use crate::core::state::State;
 use crate::{LibsyError, Result};
-use switchyard_protocol::{Context, Decision, Request, Response};
+use switchyard_protocol::{Context, Decision, LlmClientError, Request, Response};
 
 /// Builds the classifier-specific message view presented to a structured judge.
 pub(crate) trait ClassifierInput: Send + Sync {
@@ -215,7 +215,7 @@ where
     /// A judge is an optimization, not a dependency: failing the caller's request because the
     /// judge is down would be worse than routing without it, so every failure — transport,
     /// mid-stream, or unparseable reply — is logged and folded into `None` for the policy's
-    /// fail-closed branch. A closed driver stream is folded too; the algorithm's next driver
+    /// fallback branch. A closed driver stream is folded too; the algorithm's next driver
     /// call surfaces it, so nothing is masked.
     async fn verdict(
         &self,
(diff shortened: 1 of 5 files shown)


crates/libsy/src/algorithms/util/llm_judge.rs
```diff
@@ -102,21 +94,58 @@ where
...
+}
+
+/// Bounded fail-open reason for an upstream client error. Reads only the error
+/// kind and HTTP status, never any message or response body.
+fn client_error_reason(error: &LlmClientError) -> &'static str {
+    match error {
+        LlmClientError::Timeout { .. } => "timeout",
+        LlmClientError::Transport { .. } => "transport",
+        LlmClientError::UpstreamHttp { status, .. } if *status >= 500 => "upstream_5xx",
+        LlmClientError::UpstreamHttp { .. } => "upstream_4xx",
```
grahamking: That's not always a 4xx is it? All we know is status < 500.

elyasmnvidian: renamed the variable

grahamking (approved):

PR #206, opened 2026-07-30
feat(switchyard-server): Support /v1/stats endpoint

grahamking:
Returns same stats as Python.

All the new `stats/` folder is existing Rust code from
`switchyard-components`.

Reviewed-by: Claude:Opus 5 medium

## Summary by CodeRabbit

* **New Features**
  * Added a **`GET /v1/stats`** endpoint with accumulated request totals, errors, latency, token usage, per-model/per-tier performance, routing overhead, cache-eligibility metrics, and estimated costs.
  * Introduced optional **theoretical cache-hit tracking** via environment configuration.
  * Added **request-scoped run observations** for routed LLM calls and routing overhead, including timing and usage details.
* **Bug Fixes**
  * Improved stats reporting for streaming, buffered, successful/failed, and routed requests.
* **Revert**
  * Removed/disabled the legacy **`/v1/routing/stats`** endpoint.


grahamking:
@coderabbit review . Ignore code duplication between `switchyard-components` and `switchyard-server` as that is temporary.

crates/switchyard-server/src/stats.rs
```diff
@@ -0,0 +1,16 @@
...
+mod cache_eligibility;
+mod cost;
+
+pub(crate) use accumulator::{StatsAccumulator, StatsSnapshot, TokenUsage};
+pub(crate) use cache_eligibility::{prefix_probe, tracking_enabled_from_env};
+
+/// Round to six significant digits
+fn round6(value: f64) -> f64 {
+    (value * 1_000_000.0).round() / 1_000_000.0
+}
```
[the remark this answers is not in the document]
grahamking: Oops, thank you. Fixed.

crates/switchyard-server/src/lib.rs
```diff
@@ -216,6 +252,7 @@ pub fn build_switchyard_router(state: ServerState) -> Router {
         .route("/v1/responses", post(openai_responses))
         .route("/v1/messages/count_tokens", post(anthropic_count_tokens))
         .route("/v1/models", get(models))
+        .route("/v1/stats", get(get_stats))
```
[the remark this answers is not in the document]
grahamking: This is a demo server, security is out of scope.

crates/switchyard-server/src/stats/accumulator.rs
```diff
@@ -0,0 +1,621 @@
...
+    calls: u64,
+    errors: u64,
+    prompt_tokens: u64,
+    max_observed_context_tokens: u64,
+    completion_tokens: u64,
+    cached_tokens: u64,
+    cache_creation_tokens: u64,
+    cacheable_prompt_tokens: u64,
+    reasoning_tokens: u64,
+    seen_prefixes: HashSet<u64>,
```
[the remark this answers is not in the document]
grahamking: This is only used during evals, so OK to grow forever. I will add a comment.

grahamking:
@coderabbit review . Ignore code duplication between switchyard-components and switchyard-server as that is temporary. There have been lots of changes since last review.

crates/libsy/src/core/algorithm.rs
```diff
@@ -425,14 +480,24 @@ pub trait Algorithm: Send + Sync + 'static {
     /// Each [`Step::CallLlm`] is an offloaded model call the consumer must serve.
     /// The stream ends with a [`Step::ReturnToAgent`] on success, or an `Err` item on failure.
     fn run_stream(self: Arc<Self>, ctx: Context, request: Request) -> StepStream {
+        self.run_stream_observed(ctx, request, None)
+    }
+
+    /// Process a request to completion while reporting each model call to `observer`.
+    fn run_stream_observed(
```
nachiketb-nvidia: will this be exposed as public as well? might be okay if it is anyway

crates/switchyard-server/src/stats/accumulator.rs
```diff
@@ -0,0 +1,619 @@
...
+        stats.model_call_latency.record(backend_latency_ms);
+        if let Some(tier) = tier {
+            stats.tiers.insert(tier.to_string());
+            let tier_stats = inner.tier_stats_mut(tier, &model);
+            tier_stats.calls = tier_stats.calls.saturating_add(1);
+        }
+    }
+
+    /// Records one failed routed backend call.
+    pub(crate) fn record_error(&self, model: impl Into<String>, tier: Option<&str>) {
```
nachiketb-nvidia, line 56: are we keeping the "tier" wording for parity? do we need consensus?

crates/switchyard-server/src/stats/accumulator.rs
```diff
@@ -0,0 +1,619 @@
...
+        }
+    }
+
+    /// Records usage and terminal latency after a successful routed call.
+    pub(crate) fn record_usage(
+        &self,
+        model: impl Into<String>,
+        usage: TokenUsage,
+        total_latency_ms: f64,
+        tier: Option<&str>,
```
nachiketb-nvidia, line 75: same here

crates/switchyard-server/src/stats/accumulator.rs
```diff
@@ -0,0 +1,619 @@
...
+        .collect();
+    (models, totals)
+}
+
+fn build_classifier_snapshot(
+    models: &BTreeMap<String, ModelStats>,
+    total_requests: u64,
+    total_errors: u64,
+) -> ClassifierStatsSnapshot {
+    let (models, total_tokens) = build_model_snapshots(models, total_requests);
```
nachiketb-nvidia, line 434: this builds every time this function is called? as in every request?

crates/switchyard-server/src/stats/accumulator.rs
```diff
@@ -0,0 +1,619 @@
...
+    }
+}
+
+fn normalized_tier(tier: Option<&str>) -> Option<&str> {
+    tier.map(str::trim).filter(|tier| !tier.is_empty())
+}
+
+#[derive(Clone, Debug, Default)]
+struct StatsAccumulatorInner {
+    by_model: BTreeMap<String, ModelStats>,
```
nachiketb-nvidia, line 148: these are stored in memory right? should we use OTEL only in some way, or should we just get parity rn and then do cleanup later?

crates/switchyard-server/src/stats/cost.rs
```diff
@@ -0,0 +1,209 @@
...
+            input: 5.00,
+            output: 25.00,
+            cached: 0.50,
+            cache_write: 6.25,
+        },
+        "aws/anthropic/bedrock-claude-sonnet-4-6"
+        | "aws/anthropic/bedrock-claude-sonnet-4-5"
+        | "azure/anthropic/claude-sonnet-4-5"
+        | "claude-sonnet-4-6"
+        | "claude-sonnet-4-5" => ModelPrice {
```
nachiketb-nvidia: I guess we can have this for now for parity, need serious company wide solution for this ngl

crates/libsy/src/core/algorithm.rs
grahamking: Good point. I think it should only be `run_stream` with the optional observer. That would keep the API simple.

crates/switchyard-server/src/stats/accumulator.rs
grahamking: I tried to make it the same as Python for this PR. We can improve later.

crates/switchyard-server/src/stats/cost.rs
grahamking: Yes! On advice from @ayushag-nv I deleted cost. It's no longer in `/v1/stats`.

crates/switchyard-server/src/stats/accumulator.rs
grahamking: Yes. It's only called by the handler for `/v1/stats`.

In a high traffic setup we should cache that, but here even briefly outdated data would be annoying I think, and it's not expected to be a high traffic endpoint. In production (much later!) this would be an authenticated endpoint only for admins.

crates/switchyard-server/src/stats/accumulator.rs
grahamking: It's a bit of a pain. We have `/metrics` and `/v1/stats` that report many of the same values in different ways.

I think later we should try to modify the `/v1/stats` clients to read the Prometheus metrics, OR keep the `/v1/stats` endpoint but build it ourselves entirely from metrics.

For now I'm aiming at parity with Python, so we can remove that.

crates/libsy/src/core/algorithm.rs
grahamking: `run_stream_observed` is gone.

nachiketb-nvidia (approved):

PR #207, opened 2026-07-30
feat(libsy): decide api

messiaen:
## What

Added `decide` api

Still allows offloading, but add a `Driver::final_decision` method which `Algorithms` use to signal this model call is the selected model. 

So `run_decision_only_stream` functions the same as `run_steam` except the last model call is return as a `DecidedCall` instead of a `CallLlm` and then a `ReturnToAgent` 

In the streaming case `Driver::final_decision` calls `Driver::call_llm_target`.
In the decide case if simply returns a `DecidedCall` containing the `Decision`, `Request` and optionally the `Response` if the algo happens to compute it as part of the decision (e.g. escalation router)

```rust
async fn run(..)    -> Result<(Vec<Arc<dyn Decision>>, Response)>
async fn decide(..) -> Result<(Vec<Arc<dyn Decision>>, DecidedCall)>
fn run_stream(..)               -> StepStream              // Stream<Result<Step>>
fn run_decision_only_stream(..) -> DecisionOnlyStepStream  // Stream<Result<DecisionOnlyStep>>

type DecidedCall = (Arc<dyn Decision>, Box<Request>, Option<Box<Response>>);
```

changes `create_run_task` signature to support both decide and stream apis

```rust
async fn create_run_task(self: Arc<Self>, ctx: Context, driver: Driver, request: Request)
    -> Result<ResponseOrDecision>;

pub enum ResponseOrDecision {
    Response(Box<Response>),   // the call was served
    Decision(DecidedCall),     // the call is handed back
}
```

Algorithms don't change other than `create_run_task` signature and calling `final_decision`

`Driver` hold a `decide_only` flag which is set when the `Driver` is created in `decide`, `run`, `run_stream`, etc

## Why

Integrators want the ability to just get the decision without having libsy run the final model request.

Closes #

## How tested

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **New Features**
  * Added decision-only request handling, allowing routing decisions to be returned without serving the final model call.
  * Added support for returning an already-available response alongside routing decisions.
  * Added classification utilities for selecting the highest-confidence route.

* **Bug Fixes**
  * Prevented duplicate model calls when a classifier has already produced a response.
  * Improved fallback behavior when routing services are unavailable.

* **Documentation**
  * Expanded guidance for request execution modes, decision-only flows, and algorithm behavior.

### DIFF
--- a/crates/libsy/src/core/algorithm.rs
+++ b/crates/libsy/src/core/algorithm.rs
@@ -32,6 +32,21 @@ use crate::{DriverError, LibsyError, Result, observability};
 /// [`Algorithm::run_stream`]. Boxed so the trait method that produces it keeps
 /// `Arc<dyn Algorithm>` object-safe.
 pub type StepStream = Pin<Box<dyn Stream<Item = Result<Step>> + Send>>;
+/// A boxed, `Send` stream of [`DecisionOnlyStep`]s — the output of
+/// [`Algorithm::run_decision_only_stream`]. The decision-only counterpart of
+/// [`StepStream`]: it ends on a route to serve rather than on a served answer.
+pub type DecisionOnlyStepStream = Pin<Box<dyn Stream<Item = Result<DecisionOnlyStep>> + Send>>;
+
+/// Either step stream, as produced by [`Algorithm::run_stream_inner`].
+///
+/// The two run modes share one implementation, so the variant is fixed by the
+/// `decision_only` flag the run was started with rather than chosen per step.
+pub enum AnyStepStream {
+    /// A served run: every call is performed and the stream ends with an answer.
+    Step(StepStream),
+    /// A decision-only run: the stream ends with the route for the caller to serve.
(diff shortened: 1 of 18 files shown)


crates/libsy/src/core/algorithm.rs
```diff
@@ -598,35 +883,43 @@ pub trait Algorithm: Send + Sync + 'static {
     /// Each [`Step::CallLlm`] is an offloaded model call the consumer must serve.
     /// The stream ends with a [`Step::ReturnToAgent`] on success, or an `Err` item on failure.
     /// Report each model call to `observer`.
-    fn run_stream(
+    fn run_stream_inner(
```
messiaen, line 868: We should move the implementation to a private struct to keep the public api clean

grahamking:
Closing in favor of this refactor: https://github.com/NVIDIA-NeMo/Switchyard/issues/310

PR #210, opened 2026-07-30
feat(server): add native Python host with OpenTelemetry

nachiketb-nvidia:
## What

- Expose `switchyard_rust.server.Server(config, port=0)` as a context-managed native Rust server with `base_url`, `port`, and graceful shutdown.
- Refactor server startup around a reusable `BoundServer` so the binary and Python binding share TOML loading, socket binding, routing, and serving behavior.
- Move process-wide observability setup into `switchyard-server` and add optional OTLP/HTTP trace and metric export through standard OpenTelemetry environment variables.
- Preserve the existing stderr request log and Prometheus `/metrics` behavior, and flush pending telemetry during native server shutdown.

## Why

Python launchers need one straightforward binding that can host the Rust server in-process without duplicating server configuration or managing a separate subprocess. The embedded server must retain the same libsy metrics, traces, and request logging as the standalone binary.

## How

`Server` loads a TOML deployment, binds only to loopback, supports OS-selected ports, and runs Axum on PyO3’s shared Tokio runtime. Request handling stays on native runtime workers; `close()` releases the GIL while waiting for graceful shutdown and flushing telemetry.

Observability initializes once per process. OTLP export is enabled only when the standard endpoint environment variables are present, while terminal logging remains quiet by default with one structured event per LLM request. Launcher integration is intentionally deferred.

## What to review

- The `Server` lifecycle and Python API surface.
- The shared `BoundServer` abstraction used by both hosts.
- Process-wide OpenTelemetry initialization, filtering, and shutdown flushing.

## Validation

- `cargo fmt --all --check`
- `cargo check -p switchyard-server -p switchyard-py`
- `cargo clippy -p switchyard-server -p switchyard-py --all-targets -- -D warnings`
- `uv run maturin develop`
- `uv run ruff check .`
- `uv run mypy switchyard switchyard_rust`
- `uv run pytest tests/`: 1723 passed, 35 skipped
- Live provider E2E: `/health`, `/v1/models`, `/v1/chat/completions`, and `/metrics` returned 200; OTLP traces and metrics were received and shutdown flushed cleanly.

No tests were added in this change.

## Summary by CodeRabbit

* **New Features**
  * Added a Python-accessible Switchyard server that supports configurable ports, base URLs, graceful shutdown, and context-manager usage.
  * Added optional OpenTelemetry metrics and tracing export for improved observability.
  * Server startup and shutdown now provide more reliable lifecycle handling.

### DIFF
--- a/crates/switchyard-py/src/server_bindings.rs
+++ b/crates/switchyard-py/src/server_bindings.rs
@@ -110,11 +110,9 @@ impl PyServer {
 
 impl PyServer {
     fn close_inner(&mut self, py: Python<'_>, timeout_secs: f64) -> PyResult<()> {
-        if !timeout_secs.is_finite() || timeout_secs < 0.0 {
-            return Err(PyValueError::new_err(
-                "timeout_secs must be a finite non-negative number",
-            ));
-        }
+        let timeout = Duration::try_from_secs_f64(timeout_secs).map_err(|_| {
+            PyValueError::new_err("timeout_secs must be a supported finite non-negative number")
+        })?;
         if let Some(shutdown) = self.shutdown.take() {
             let _ = shutdown.send(());
         }
@@ -123,7 +121,7 @@ impl PyServer {
         };
         let task = self.task.take();
         let result = py.detach(move || {
-            let result = completion.recv_timeout(Duration::from_secs_f64(timeout_secs));
+            let result = completion.recv_timeout(timeout);
             if matches!(result, Err(RecvTimeoutError::Timeout)) {
                 if let Some(task) = task {
                     task.abort();
--- a/crates/switchyard-py/src/server_bindings.rs
+++ b/crates/switchyard-py/src/server_bindings.rs
@@ -1,0 +1,164 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Self-contained Python host for the Rust Switchyard server.
+
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::path::PathBuf;
+use std::sync::mpsc::{sync_channel, Receiver, RecvTimeoutError};
+use std::time::Duration;
+
+use pyo3::exceptions::{PyRuntimeError, PyValueError};
+use pyo3::prelude::*;
+use switchyard_server::config::load_server_state;
+use switchyard_server::{
+    flush_observability, initialize_observability, BoundServer, ServerResult, ServerRunOptions,
+    DEFAULT_LISTEN_BACKLOG,
+};
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+
+const DEFAULT_SHUTDOWN_TIMEOUT_SECS: f64 = 2.0;
+
+/// Running loopback server backed entirely by the native Rust implementation.
+#[pyclass(name = "Server", module = "switchyard_rust.server", unsendable)]
+struct PyServer {
+    addr: SocketAddr,
+    shutdown: Option<oneshot::Sender<()>>,
+    completion: Option<Receiver<ServerResult<()>>>,
+    task: Option<JoinHandle<()>>,
+}
+
+#[pymethods]
+impl PyServer {
+    /// Loads a TOML deployment and starts serving it on loopback.
+    #[new]
+    #[pyo3(signature = (config, *, port=0))]
+    fn new(config: PathBuf, port: u16) -> PyResult<Self> {
+        initialize_observability().map_err(server_error)?;
+        let state = load_server_state(config).map_err(server_error)?;
+        let runtime = pyo3_async_runtimes::tokio::get_runtime();
+        let server = {
+            let _guard = runtime.enter();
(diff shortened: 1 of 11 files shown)


crates/switchyard-py/src/server_bindings.rs
```diff
@@ -0,0 +1,164 @@
...
+struct PyServer {
+    addr: SocketAddr,
+    shutdown: Option<oneshot::Sender<()>>,
+    completion: Option<Receiver<ServerResult<()>>>,
+    task: Option<JoinHandle<()>>,
+}
+
+#[pymethods]
+impl PyServer {
+    /// Loads a TOML deployment and starts serving it on loopback.
```
messiaen, line 34: Why just on loopback?

crates/switchyard-py/src/server_bindings.rs
```diff
@@ -0,0 +1,164 @@
...
+                    backlog: DEFAULT_LISTEN_BACKLOG,
+                    dry_run: false,
+                    tls: None,
+                },
+            )
+            .map_err(server_error)?
+        };
+        let addr = server.local_addr();
+        let (shutdown, shutdown_receiver) = oneshot::channel();
+        let (completion_sender, completion) = sync_channel(1);
```
messiaen, line 56: We could use a oneshot here, right.  What am I missing?

crates/switchyard-server/src/observability.rs
messiaen, line 1: Does this duplicate the observability.rs in libsy?

nachiketb-nvidia: no its slightly different:

```
- crates/libsy/src/observability.rs:4 emits spans and metrics through global tracing/OTEL APIs. It intentionally does not
    configure exporters.

  - crates/switchyard-server/src/observability.rs:4 is the host setup: subscriber, OTLP exporter, resource, initialization, and
    flushing.
    ```

File: crates/switchyard-py/src/server_bindings.rs (hunk 1 above)
Comment (nachiketb-nvidia) [reply crates/switchyard-py/src/server_bindings.rs:34]: This binding is only for local launcher sidecars, the server binary supports external serving

File: crates/switchyard-py/src/server_bindings.rs (hunk 2 above)
Comment (nachiketb-nvidia) [reply crates/switchyard-py/src/server_bindings.rs:56]: Shutdown is a oneshot, completion needs recv_timeout so close can remain bounded

**Review by messiaen — APPROVED** [review]

File: crates/switchyard-server/src/observability.rs [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+    {
+        return false;
+    }
+    [
+        "OTEL_EXPORTER_OTLP_ENDPOINT",
+        &format!("OTEL_EXPORTER_OTLP_{signal}_ENDPOINT"),
+    ]
+    .into_iter()
+    .any(|name| env::var(name).is_ok_and(|value| !value.trim().is_empty()))
+}
```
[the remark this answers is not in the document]
nachiketb-nvidia: addressed

crates/switchyard-py/src/server_bindings.rs
```diff
@@ -0,0 +1,164 @@
...
+        match result {
+            Ok(result) => result.map_err(server_error),
+            Err(RecvTimeoutError::Timeout) => Err(PyRuntimeError::new_err(format!(
+                "server did not stop within {timeout_secs} seconds"
+            ))),
+            Err(RecvTimeoutError::Disconnected) => Err(PyRuntimeError::new_err(
+                "server task stopped without reporting a result",
+            )),
+        }
+    }
```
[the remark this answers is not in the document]
nachiketb-nvidia: addressed

crates/switchyard-server/src/observability.rs
```diff
@@ -0,0 +1,134 @@
...
+use opentelemetry_sdk::Resource;
+use tracing_subscriber::layer::SubscriberExt as _;
+use tracing_subscriber::util::SubscriberInitExt as _;
+use tracing_subscriber::{EnvFilter, Layer as _};
+
+use crate::{metrics, ServerError, ServerResult};
+
+const DEFAULT_LOG_FILTER: &str = "switchyard_server=info";
+const OPENTELEMETRY_LOG_FILTER: &str = "opentelemetry=warn";
+const OTEL_SPAN_FILTER: &str = "libsy=info";
```
grahamking, line 20: I think if you move `OTEL_SPAN_FILTER` into `DEFAULT_LOG_FILTER` (comma separated with the existing one), we can simplify.

See below...

crates/switchyard-server/src/observability.rs
```diff
@@ -0,0 +1,134 @@
...
+        .with_filter(filter);
+
+    if let Some(provider) = &tracer_provider {
+        let tracer = provider.tracer("switchyard");
+        tracing_subscriber::registry()
+            .with(format)
+            .with(
+                tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(EnvFilter::new(OTEL_SPAN_FILTER)),
```
grahamking, line 105: You can't override OTEL_SPAN_FILTER, it doesn't use the env var. I _think_ we should use `filter` from line 86 instead.

PR #211, opened 2026-07-30
feat(libsy): add escalation as config knob on LlmTaskClassifier

linj-glitch:
Draft of Ryan and Sabhatina's proposed design: escalation as a builder method on the existing capability classifier rather than a separate algorithm.

LlmTaskClassifier.with_escalation(EscalationJudgeConfig) swaps the inner classifier for an EscalationClassifier that calls the efficient model first, appends its reply to the transcript, consults the trajectory judge, and either returns the efficient response directly (on decline) or routes to capable (once the streak confirms). The session latch is the streak counter in State, replacing the AffinityRouter.

Targets PR 203 (the Option<Response> interface change) because EscalationClassifier relies on returning Some(efficient_response) to avoid a second model call on non-escalating turns.

## Summary by CodeRabbit

* **New Features**
  * Added intelligent escalation routing that can switch from an efficient model to a more capable model when repeated issues are detected.
  * Added configurable escalation thresholds and transcript context handling.
  * Added support for recording the selected model and optional routing rationale for one-off requests.
  * Added structured escalation verdict validation and guidance for common failure patterns.


crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -320,9 +468,52 @@ impl LlmTaskClassifier {
...
+            self.efficient_target.clone(),
+        ]);
+        Ok(Self {
+            route: FallThrough::<State>::new_with_state(targets)
+                .with_name(ALGORITHM_NAME)
+                .with_classifier(esc),
+            inner,
+            judge_target: self.judge_target,
+            efficient_target: self.efficient_target,
+            capable_target: self.capable_target,
```
messiaen: I don't love the override.  can we have `new` and `new_with_escalation`?

messiaen (commented):
leaving initial comment

crates/libsy/src/algorithms/llm_class.rs
linj-glitch: Done — replaced `with_escalation()` with `new_with_escalation()` and dropped the three stored target fields that were only needed to support the builder.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -263,10 +265,152 @@ struct TaskClassifier {
...
+        let efficient_response = Response {
+            llm_response: LlmResponse::Agg(agg),
+            metadata: efficient_response.metadata,
+        };
+
+        // Append the efficient reply so the judge reads this turn's completed trajectory.
+        let mut judge_request = request.clone();
+        judge_request.llm_request.messages.push(assistant_message(
+            efficient_response.llm_response.as_agg().unwrap(),
+        ));
```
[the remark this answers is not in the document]
linj-glitch: Fixed — the assistant message is now built from `agg` directly before it is moved into `efficient_response`, so the `.unwrap()` is gone.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -263,10 +265,149 @@ struct TaskClassifier {
...
+            .map_err(|e| LibsyError::AlgorithmError {
+                message: format!("failed to aggregate efficient response: {e}"),
+            })?;
+        // Append the efficient reply so the judge reads this turn's completed trajectory.
+        let mut judge_request = request.clone();
+        judge_request.llm_request.messages.push(assistant_message(&agg));
+        let efficient_response = Response {
+            llm_response: LlmResponse::Stream(agg.into_stream()),
+            metadata: efficient_response.metadata,
+        };
```
messiaen, line 375:
```suggestion
        let efficient_response = Response {
            llm_response: if judge_request.llm_request.stream {
                LlmResponse::Stream(agg.into_stream())
            } else {
                LlmResponse::Agg(agg)
            },
            metadata: efficient_response.metadata,
        };
```

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -263,10 +265,149 @@ struct TaskClassifier {
...
+                state
+                    .extra
+                    .insert(STREAK_KEY.to_string(), StateValue::Count(0));
+                Ok((decisive(&self.efficient.semantic_name), Some(efficient_response)))
+            }
+            None => {
+                // Judge unavailable: no evidence, leave streak intact, stay efficient.
+                Ok((decisive(&self.efficient.semantic_name), Some(efficient_response)))
+            }
+        }
```
messiaen: Can we do something like this?

```suggestion
        let held = streak(state);
        let (escalate, pending) = match classification {
            Classification::Scores(ref scores) if !scores.is_empty() => (true, held + 1),
            Classification::Scores(_) => (false, 0),
            Classification::Ambiguous(_) => (false, held),
        };
        state
            .extra
            .insert(PENDING_KEY.to_string(), StateValue::Count(pending));
            
        if escalate && pending >= self.confirmations {
            // The efficient reply is discarded: the capable model answers this turn.
            return Ok((decisive(&self.capable), None));
        }
        
        Ok((decisive(&self.efficient), efficient_response)
```

crates/protocol/src/stream.rs
```diff
@@ -88,6 +88,50 @@ impl LlmResponse {
     }
 }
 
+impl AggLlmResponse {
```
messiaen, line 91: needs a test

messiaen (changes requested):

crates/libsy/src/algorithms/llm_class.rs
linj-glitch: Done — added the stream check. Non-streaming clients still get `LlmResponse::Agg`.

crates/libsy/src/algorithms/llm_class.rs
linj-glitch: Applied. Also aligns with the existing `STREAK_KEY` naming.

crates/protocol/src/stream.rs
linj-glitch: Added `into_stream_round_trips_through_into_agg` — converts an `AggLlmResponse` to a stream and drives it back through `into_agg`, checking id, model, usage, stop reason, and content.

linj-glitch:
Terminal-Bench 2.0 smoke through the Rust libsy server (`switchyard-server --config`), escalation route, on Harbor-GKE. Tiers: weak `nvidia/nvidia/nemotron-3-ultra`, strong `azure/anthropic/claude-opus-5`, judge `nvidia/nvidia/nemotron-nano-3.5-preview`, `confirmations = 2`.

| Harness | Inbound | Served by Switchyard | Routing | Terminal cause |
|---|---|---|---|---|
| Codex (`rh-cli-0730223559`) | `openai_responses` | 9 × `200` | 8 weak, **1 strong** | task unsolved (`reward=0`) |
| Claude Code (`rh-cli-0730225003`) | `anthropic_messages` | 12 × `200` | 12 weak | `error_max_turns` (preset sets `max_turns=12`) |

## Observed

- **Escalation fired in a real task.** The Codex run served 8 turns weak, then routed one to `claude-opus-5`.
- **Both inbound wire formats work.** The Claude run drives Anthropic Messages against OpenAI-format tiers, so it exercises translation on top of the escalation path's buffer-and-re-emit — 12 streaming turns, no errors, no reconnects.
- **No Switchyard-side errors** in either run: every request `200`, no `judge verdict unavailable`.
- Claude never escalated: the judge declined for 12 turns on a task where the agent was making steady progress.

## Are the results expected?

Yes. Neither terminal state is a routing fault:

- Codex ran to completion and did not solve `adaptive-rejection-sampler` — a benchmark outcome for one task on a weak-tier model.
- Claude was cut off by the preset's own `max_turns=12` while working productively (writing R, running tests, debugging its own upper-hull construction). Harbor reports that exit as `NonZeroAgentExitCodeError`; the agent log shows `error_max_turns`, and `modelUsage` attributes 337k input / 8.9k output tokens to `switchyard`, so every turn round-tripped through the router.

**This validates plumbing and routing behaviour, not routing quality** — neither run passed its task, and solve rates need the full sweep.

messiaen:
@CodeRabbit review

crates/libsy/src/algorithms/util/escalation.rs
```diff
@@ -0,0 +1,514 @@
...
+    while text.chars().count() > MAX_REQUEST_CHARS && !window.is_empty() {
+        window.remove(0);
+        text = assemble(&window);
+    }
+    if text.chars().count() > MAX_REQUEST_CHARS {
+        let keep = MAX_REQUEST_CHARS.saturating_sub(TRUNCATION_SUFFIX.chars().count() + 1);
+        text = text.chars().take(keep).collect::<String>() + TRUNCATION_SUFFIX;
+    }
+    text
+}
```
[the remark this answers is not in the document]
linj-glitch: Fixed — moved `header` inside the `assemble` closure so it's recomputed from the current window slice on every call, including after `window.remove(0)`.

crates/protocol/src/stream.rs
```diff
@@ -88,6 +88,50 @@ impl LlmResponse {
...
+                        chunks.push(LlmResponseChunk::ReasoningDelta { index, text });
+                    }
+                    ContentBlock::ToolCall(tool) => {
+                        let args = serde_json::to_string(&tool.arguments).unwrap_or_default();
+                        chunks.push(LlmResponseChunk::ToolCallDelta {
+                            index,
+                            id: Some(tool.id),
+                            name: Some(tool.name),
+                            arguments_delta: Some(args),
+                        });
```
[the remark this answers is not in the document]
linj-glitch: Fixed — added a separate `tool_call_index` counter that increments per tool call, independent of the outer `output_index`.

crates/protocol/src/stream.rs
```diff
@@ -88,6 +88,50 @@ impl LlmResponse {
...
+                    }
+                    _ => {}
```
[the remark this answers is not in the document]
linj-glitch: Skipping for now — `LlmResponseChunk::ReasoningDelta` has no `signature` field, so preserving it requires a protocol change. The `_ => {}` arm is intentional: `Image`, `ToolResult`, `Refusal`, etc. don't appear in assistant outputs and have no chunk representation. Added a comment to that effect.

messiaen (approved):

PR #224, opened 2026-07-31
feat(observability): add GenAI semantic attributes

nachiketb-nvidia:
## What

Add OpenTelemetry GenAI semantic attributes to the existing Rust server and libsy spans while preserving the current telemetry surface.

- Name model-call spans `chat <model>` and mark them as client spans.
- Record the request model, streaming mode, sampling controls, output type, and reasoning level from the neutral IR.
- Record response IDs, served models, finish reasons, input/output/cache/reasoning token usage, and typed failures for buffered and streamed responses.
- Retain Switchyard route, algorithm, routing tier, session, and correlation attributes.
- Continue incoming W3C `traceparent` and `tracestate` at the Rust server boundary.

`gen_ai.provider.name` remains unset because the current configuration does not identify the provider authoritatively. Wire format, hostname, and model name are not reliable provider identities.

## Why

OTLP consumers need standard model-call data to reconstruct per-trace and per-session usage without an Intake-specific Rust sink. The existing spans already represent the correct request and model-call boundaries; this adds the portable GenAI projection and inbound trace continuity.

## How

- Extend `libsy.run` and `libsy.client_call` instead of creating duplicate GenAI spans.
- Populate request, response, and usage attributes from the neutral Switchyard IR.
- Keep the client span alive until a response stream completes, errors, or is abandoned.
- Write finish reasons directly to OpenTelemetry so they remain a typed string array.
- Let `switchyard-llm-client` add the resolved upstream server address and port.
- Validate exported spans with an in-memory OpenTelemetry exporter.

No prompt, completion, raw body, API key, or header content is recorded. Existing metrics, logs, and HTTP endpoints remain unchanged.

## What to review

- GenAI attribute names and numeric wire types.
- Cache-aware total input usage and finish-reason normalization.
- Streaming span completion, error, and cancellation behavior.
- W3C parent extraction at the server boundary.
- The intentional absence of inferred provider identity.

## Validation

```bash
cargo fmt --all -- --check
cargo clippy -p switchyard-libsy -p switchyard-llm-client --all-targets -- -D warnings
cargo test -p switchyard-libsy
```

Related: [SWITCH-1096](https://linear.app/nvidia/issue/SWITCH-1096/make-sure-switchyard-server-can-emit-routing-data-consumable-by-intake)

Follow-up: [SWITCH-1142](https://linear.app/nvidia/issue/SWITCH-1142/complete-genai-telemetry-metrics-propagation-and-openinference)

### DIFF
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -144,14 +144,6 @@ impl Backend {
         self.config().max_retries
     }
 
-    /// Stable GenAI semantic-convention provider name for this wire protocol.
-    pub(crate) fn provider_name(&self) -> &'static str {
-        match self {
-            Backend::OpenAiChat(_) | Backend::OpenAiResponses(_) => "openai",
-            Backend::Anthropic(_) => "anthropic",
-        }
-    }
-
     /// Whether this backend speaks the Anthropic Messages wire format — the only
     /// one with a `count_tokens` endpoint.
     pub fn is_anthropic(&self) -> bool {
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -144,6 +144,14 @@ impl Backend {
         self.config().max_retries
     }
 
+    /// Stable GenAI semantic-convention provider name for this wire protocol.
+    pub(crate) fn provider_name(&self) -> &'static str {
+        match self {
+            Backend::OpenAiChat(_) | Backend::OpenAiResponses(_) => "openai",
+            Backend::Anthropic(_) => "anthropic",
+        }
+    }
+
     /// Whether this backend speaks the Anthropic Messages wire format — the only
     /// one with a `count_tokens` endpoint.
     pub fn is_anthropic(&self) -> bool {
--- a/crates/libsy/src/core/algorithm.rs
+++ b/crates/libsy/src/core/algorithm.rs
@@ -618,7 +618,7 @@ pub trait Algorithm: Send + Sync + 'static {
         // One `libsy.run` span covers the whole algorithm task; the driver's
         // `libsy.llm_call` spans and decision logs nest inside it via `tracing`'s
         // contextual parenting.
-        let span = observability::run_span(self.name(), request.metadata.as_ref());
+        let span = observability::run_span(self.name(), &request);
         let observed_driver = task_driver.clone();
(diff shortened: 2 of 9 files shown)


crates/libsy-llm-client/src/backend.rs
```diff
@@ -144,6 +144,14 @@ impl Backend {
         self.config().max_retries
     }
 
+    /// Stable GenAI semantic-convention provider name for this wire protocol.
+    pub(crate) fn provider_name(&self) -> &'static str {
+        match self {
+            Backend::OpenAiChat(_) | Backend::OpenAiResponses(_) => "openai",
+            Backend::Anthropic(_) => "anthropic",
+        }
+    }
```
grahamking: Technically this is the wire format. 

Provider is https://openrouter.ai/providers for OpenRouter, and nvidia/openai/amazon/etc for Inference.

crates/libsy/src/core/algorithm.rs
```diff
@@ -684,12 +684,44 @@ pub trait Algorithm: Send + Sync + 'static {
...
+                gen_ai.response.model = tracing::field::Empty,
+                gen_ai.usage.input_tokens = tracing::field::Empty,
+                gen_ai.usage.output_tokens = tracing::field::Empty,
+                gen_ai.usage.cache_read.input_tokens = tracing::field::Empty,
+                gen_ai.usage.cache_creation.input_tokens = tracing::field::Empty,
+                gen_ai.usage.reasoning.output_tokens = tracing::field::Empty,
                 outcome = tracing::field::Empty,
+                otel.status_code = tracing::field::Empty,
+                error.type = tracing::field::Empty,
                 error = tracing::field::Empty,
```
grahamking, line 714: Do we need all the `Empty` fields? Will whatever reads this break if we omit them? Seems kinda noisy.

grahamking (approved):

BrianNewsom (approved):

crates/libsy/src/core/algorithm.rs
nachiketb-nvidia: yeah apparently we have to initialize things as empty here so later spans can start recording here

crates/libsy-llm-client/src/backend.rs
nachiketb-nvidia: fixed, making it unset since there;s no way to know in switchyard

PR #227, opened 2026-07-31
chore: rc prep and version bump

ryan-lempka:
Bumps the workspace and Python package to 0.2.0 ahead of the first release candidate.

- Crates inherit `version` from `[workspace.package]`, so future bumps are one line
- Publish workflow now skips prerelease tags (`!v*-*`)
- Drops the hardcoded version from the known-issues nav label

Without this, the release tag check fails, since it requires the tag to equal `v` + the pyproject version.

## Summary by CodeRabbit

* **New Release**
  * Updated the project version to 0.2.0 across supported packages and distributions.
  * Improved release handling so prerelease tags are not published automatically.

* **Documentation**
  * Simplified the navigation label for known issues by removing the outdated version reference.

### DIFF
--- a/crates/switchyard-skill-distillation/Cargo.toml
+++ b/crates/switchyard-skill-distillation/Cargo.toml
@@ -3,7 +3,7 @@
 
 [package]
 name = "switchyard-skill-distillation"
-version = "0.1.0"
+version.workspace = true
 description = "Source-neutral Rust contracts for Switchyard skill distillation"
 authors.workspace = true
 edition.workspace = true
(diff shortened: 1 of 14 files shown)


crates/switchyard-skill-distillation/Cargo.toml
ayushag-nv, line 1: do we need skill-distillation ??

ryan-lempka: Sounds like a different PR if we want to nix it

ryan-lempka: Need to check with @elyasmnvidian

nachiketb-nvidia (approved):

PR #229, opened 2026-07-31
feat(server): add static startup banner

nachiketb-nvidia:
## What

Adds a static SWITCHYARD
Screenshot from 2026-07-31 10-10-17
 startup banner to the Rust server startup output.

## Why

This gives the server a recognizable startup identity without adding a FIGlet/banner crate or any transitive runtime dependency.

## How

- Stores the generated banner as crates/switchyard-server/assets/startup_banner.txt.
- Includes it with include_str!, so the binary only carries the static banner bytes.
- Prepends the banner to the existing startup summary that prints the listening URL and configured routes.
- Prints the startup banner to stdout.
- Applies NVIDIA green (#76B900) ANSI color only when stdout is an interactive terminal; redirected output gets the same banner without ANSI escapes.

## Bloat

The checked-in banner asset is about 1.7 KB. No new crate dependency is added.

## Validation

- cargo fmt --all --check
- git diff --check
- Commit hook ran: cargo fmt, cargo clippy, commitlint

No test suite was run after the final visual banner revision, per request.


crates/switchyard-server/src/lib.rs
```diff
@@ -161,7 +163,7 @@ pub async fn run_server(state: ServerState, options: ServerRunOptions) -> Server
     }
 
     let server = BoundServer::bind(state, options)?;
-    eprintln!("{}", server.startup_banner());
+    eprintln!("{}", server.startup_banner(std::io::stderr().is_terminal()));
```
grahamking: Should this be `println`? I don't think startup banner should go to STDERR.

grahamking (approved):

PR #230, opened 2026-07-31
chore(docs): massive docs refactor for QA

ayushag-nv:
## What

Reworked docs with libsy and rust server

## Why

The motivation — what problem does this solve, or which ticket does it close?

Closes #

## How tested

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **Documentation**
  * Reorganized the README and documentation around launcher, standalone server, and library workflows.
  * Added comprehensive installation, configuration, startup, validation, routing, TOML schema, and CLI guidance.
  * Updated architecture and routing documentation for native Rust deployments and TOML configuration.
  * Documented context-overflow handling, subagent routing, telemetry opt-out, and supported protocol formats.
  * Removed outdated sticky-routing and legacy Python/YAML guidance.

* **Tests**
  * Expanded documentation checks for command coverage, section ordering, current workflows, and server options.

Relates to [SWITCH-1134](https://linear.app/nvidia/issue/SWITCH-1134/docs-cleanup-for-qa-v020-rc1)


docs/operations/context_window.md
```diff
@@ -1,36 +1,69 @@
 # Context-Window Handling
 
 When an upstream rejects a request because the prompt exceeds the model's
-context window, Switchyard evicts that target for the current request,
-reroutes to the configured fallback target, and retries once. If the fallback
-also overflows, the request fails with a 400 in the client's inbound wire
-format.
+context window, Switchyard drops that target and calls another target on the
```
ryan-lempka, line 4: We should add that this is session scoped - once a target is excluded for a session it is no longer eligible for the duration of the session.

docs/operations/context_window.md
```diff
@@ -1,36 +1,69 @@
...
+```text
+x-model-router-selected-model: openai/gpt-4o-mini
+x-model-router-rationale: openai/gpt-4o exceeded its context window; fell back to openai/gpt-4o-mini
+```
+
+## Evictions last for the session
+
+A conversation only grows, so a target that overflowed on one turn will overflow
+on the next. Switchyard remembers the overflow and skips that target for the
+rest of the session rather than paying for a call certain to fail.
```
ryan-lempka: Oh wait I see you call it out here. Maybe we don't need a full section for this call out?

linj-glitch:
**docs/routing_algorithms/escalation_router_routing.md** — one logical issue in the "When not to use" section:

> "Traffic without session identity. With confirmations above 1, the route cannot accumulate a streak and never latches."

This implies `confirmations = 1` works fine without a session. It doesn't. Without a session the streak is ephemeral (reset every request), so the latch never persists regardless of the `confirmations` value. With `confirmations = 1`, the route does escalate on any turn the judge fires on (within that single turn), but it re-evaluates efficient + judge every subsequent turn — there's no skip-judge benefit. The "when not to use" bullet should say this applies to any `confirmations` value, not just `> 1`.

Everything else I checked (default values, streak reset-on-decline / hold-on-failure, cost at the confirming turn, flowchart) matches the current implementation.

ryan-lempka (approved):

PR #234, opened 2026-07-31
feat(server): expand Rust startup banner

nachiketb-nvidia:
## What

Adds concise runtime guidance to the Rust server startup banner:

- the bound listening URL
- configured route model IDs
- the mounted endpoint list
- a copy-pasteable `curl` example for `/v1/chat/completions` using the first configured route

The optional `/v1/routing/session-stats` endpoint is only shown when routing-log support is enabled, matching router registration.

## Why

The existing banner only printed the server URL and route ids. On startup, users still had to know which endpoints are available and how to send a first request. This makes the Rust server self-describing without changing endpoint behavior.

## How

This stays isolated to `crates/switchyard-server/src/lib.rs` near the existing `startup_banner` helper. It derives the example request from `ServerState::models()` and uses a localhost-friendly request URL when the server binds to an unspecified address like `0.0.0.0`.

## Validation

- `cargo check -p switchyard-server`
- commit hooks ran `cargo fmt` and `cargo clippy`

Tests were not run per request.

## Summary by CodeRabbit

* **Enhancements**
  * Improved the server startup message with clearer listening and request URLs.
  * Added visibility into configured routes and available endpoints.
  * Included a sample `curl` request to help verify the server is running.
  * Request URLs now display accessible loopback addresses when the server binds to an unspecified host.
  * The routing-session statistics endpoint is shown when enabled.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -958,11 +958,23 @@ fn model_entry_json(model: &str) -> Value {
 
 fn startup_banner(options: &ServerRunOptions, state: &ServerState, color: bool) -> String {
     let scheme = if options.is_tls() { "https" } else { "http" };
+    let listen_url = url_for_addr(scheme, options.addr);
+    let request_url = request_url_for_addr(scheme, options.addr);
+    let routes = state.models().collect::<Vec<_>>();
+    let route_list = routes.join(", ");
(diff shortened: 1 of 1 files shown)


ayushag-nv (approved):

ryan-lempka (approved):
Love this

PR #248, opened 2026-08-03
fix(llm-client): strip Anthropic-incompatible fields on the Anthropic leg

sabhatinas:
## What

- The Rust server forwards `context_management` and `reasoning_effort` straight through to Anthropic Messages, which rejects both.
- `switchyard-components` already strips them; `libsy-llm-client` (what the Rust server uses) did not. This adds the same strip there.
- Runs before `merge_extra_body`, so a target can still set either field deliberately.

## Why

- Needed for Python/Rust parity — the Python path has handled this for a while and the Rust server quietly does not.
- Hit it with the stage router on TB-2.1: Claude Code sends `context_management: {clear_thinking_20251015}` every turn. Fine while the weak tier (GLM 5.2, OpenAI format) serves, but the first escalation to Opus 4.8 (Anthropic format) returns `400 clear_thinking_20251015 strategy requires thinking to be enabled or adaptive`.
- Killed every escalating task — 13 of 13 errored evaluations in one run, same 400 each time.
- Not really a stage-router bug: any route that switches wire formats mid-session hits it. Single-format routes are unaffected, which is why passthrough runs against each tier passed and only the mixed one failed.

## Testing

- New test `anthropic_requests_drop_openai_only_fields`, verified it fails without the strip and passes with it.
- `cargo test -p switchyard-llm-client` (44) and `-p switchyard-server` (38) pass.
- Rebuilt the image and reran the stage-router smoke that used to fail: passes with a real escalation (5 calls to GLM 5.2, 2 to Opus 4.8, reward 1.0). Before the fix Opus never completed a call.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -171,6 +171,11 @@ impl TranslatingLlmClient {
         // which keeps the caller's original `model`; force the resolved model so
         // the upstream always sees the target id.
         set_json_model(&mut body, model);
+        // Strip before `merge_extra_body` so a target can reinstate either field
+        // deliberately via `extra_body`.
+        if matches!(backend, Backend::Anthropic(_)) {
+            strip_anthropic_incompatible_fields(&mut body);
+        }
         merge_extra_body(&mut body, backend.extra_body());
         if matches!(backend, Backend::OpenAiChat(_)) {
             ensure_openai_stream_usage(&mut body);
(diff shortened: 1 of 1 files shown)


ayushag-nv (approved):

sabhatinas:
Two follow-ups I looked into after opening this.

## Should this live in the translation layer instead?

I don't think the enforcement point can, but the helper probably should.

- When the inbound format already matches the target format, `encode_request` restores the exact preserved source body verbatim (`PreservationPolicy` / `exact_preserved_request`). So anything the Anthropic codec dropped would just be reinstated. Our failing case is exactly Anthropic-in → Anthropic-out, so a translation-layer fix would not have fired at all.
- Conceptually these aren't translation concerns either. `UnknownFieldPolicy` (`Preserve` / `DropWithWarning` / `Reject`) is about fields unknown to the IR; `context_management` and `reasoning_effort` are known fields that one specific upstream rejects. That's a backend/provider question, which is where both implementations put it today.
- What does make sense: move the helper into `switchyard-translation/src/util.rs` next to `normalize_anthropic_tool_use_ids` and `sanitize_anthropic_tool_use_id`, and have both `switchyard-components` and `libsy-llm-client` call it. Same enforcement point, no duplicated logic. Happy to do that here if reviewers prefer.

## What else `switchyard-components` does that `libsy-llm-client` doesn't

`AnthropicNativeBackend::outbound_body` does four things; the Rust client now does one of them.

- [x] `strip_anthropic_incompatible_fields` — this PR
- [ ] `strip_unsigned_thinking_blocks` — "Anthropic requires signed thinking blocks on replay"
- [ ] `lift_message_level_system` — hoists message-level `role: "system"` turns to top-level `system` for older Bedrock/LiteLLM targets
- [ ] `normalize_anthropic_tool_use_ids` — already a shared `pub fn` in the translation crate, so this one is nearly free

The thinking-blocks gap looks like the next failure in the same scenario this PR fixes. GLM 5.2 emits thinking blocks with an empty signature — from a Hub 400 we captured:

```
[{'type': 'thinking', 'thinking': 'Let me start by reading the image...', 'signature': '', 'cache_control': {}}]
```

In a mixed-format session those unsigned blocks can be replayed on the Anthropic leg after an escalation, which is precisely what `strip_unsigned_thinking_blocks` exists to prevent. To be clear, I have not reproduced that on the Rust server — the patched benchmark run has not hit it so far — so this is a gap in parity rather than a confirmed bug.

Parity that is already fine: context-window 400 classification (`is_context_overflow`) and OpenAI stream-usage defaults both exist on the client side.

Worth deciding whether to close the remaining three here or track them separately.

PR #253, opened 2026-08-03
chore(libsy): better defaults for llm classifier

ayushag-nv:
## Why

Internal judge calls need enough output budget to return a complete verdict.

## What

Default judge calls to 4,096 output tokens and expose one top-level route override for both classifier modes.

## How

Copy max_output_tokens into the shared JudgeConfig once. Standard classification and escalation both build their judge requests from that value.

## Configuration

Classifier route:

~~~toml
[routes.classifier]
id = "switchyard/classifier"
type = "llm_classifier"
classifier_target = "judge"
weak_target = "weak"
strong_target = "strong"
base_threshold = 0.5
max_output_tokens = 4096
~~~

Escalation route:

~~~toml
[routes.escalation]
id = "switchyard/escalation"
type = "llm_classifier"
classifier_target = "judge"
weak_target = "weak"
strong_target = "strong"
base_threshold = 0.5
max_output_tokens = 4096
escalation = { confirmations = 2 }
~~~

## Where to Start Review

crates/libsy/src/algorithms/llm_class.rs, then crates/libsy/src/algorithms/util/escalation.rs

## Test Plan

- cargo fmt --all --check
- cargo clippy -p switchyard-libsy -p switchyard-server --all-targets -- -D warnings
- cargo test -p switchyard-server: 39 passed
- Qwen 3.5 35B reasoning-on passed standard and escalation routing
- DeepSeek V4 Flash passed with reasoning off; reasoning on remains degraded because the verdict is returned only in reasoning_content

## Relates to
- https://linear.app/nvidia/issue/SWITCH-1158/oss-020config-deepseek-v4-flash-emits-classifier-verdict-only-in

### DIFF
--- a/crates/libsy/examples/research_agent.rs
+++ b/crates/libsy/examples/research_agent.rs
@@ -109,11 +109,7 @@ async fn main() -> Result<()> {
         strong,
         TaskClassifierConfig {
             base_threshold: BASE_THRESHOLD,
-            min_confidence: 0.0,
-            capability_elevated_floor: None,
-            session_affinity: false,
-            message_hash_fallback: false,
-            recent_turn_window: None,
+            ..TaskClassifierConfig::default()
         },
     )?);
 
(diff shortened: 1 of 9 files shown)


nachiketb-nvidia (approved):
lgtm

PR #261, opened 2026-08-03
fix(switchyard-server): Remove redundant HTTP header name lower-case

grahamking:
`http::HeaderMap` already does that.

See:
https://docs.rs/http/latest/http/header/struct.HeaderName.html#method.as_str

AFAICT the only way we get HTTP headers is via axum / http crates, so we
don't need to re-do the normalization. This avoids an extra allocate and
copy of the headers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request header handling for routing logs and metadata processing.
  * Preserved original header-name casing during request processing.
  * Ignored empty or invalid header values when extracting routing context.


grahamking:
@coderabbit review

grahamking:
Closing this in favor of using `http::HeaderMap` more broadly.

PR #271, opened 2026-08-04
feat(llm-client): harden HTTP provider transport

bbednarski9:
## What

Hardens the reusable `switchyard-llm-client` HTTP transport used for provider
dispatch. The PR establishes five generic safety guarantees:

* configured provider requests do not follow redirects;
* connections and inactive response reads have safe, configurable timeouts;
* buffered success and retained error bodies have explicit memory bounds;
* credentials and provider error bodies are redacted from ordinary diagnostics; and
* individual SSE frames are bounded before line-oriented decoding can allocate
  without limit.

This is intentionally independent of the NeMo Relay dynamic-plugin implementation.
It changes the shared client and translation behavior for every caller rather than
placing Relay-specific protections in NVIDIA-NeMo/Switchyard#270.

#### 1\. Redirect rejection at the provider authority boundary

**Motivation:** A configured provider request can contain credentials and a full
prompt. Following an upstream redirect would allow a provider—or a compromised
endpoint—to move that request to a different authority. Redirects are therefore
returned as ordinary non-success HTTP responses instead of being followed.

* Implementation: [`TranslatingLlmClient` constructs a no-redirect shared client](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L116-L142>).
* Contract test: [`configured_provider_requests_do_not_follow_redirects`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L1158-L1196>) verifies the redirect target receives no request.

#### 2\. Configurable connection and idle-read timeouts

**Motivation:** Without transport timeouts, an unreachable or silent provider can
retain request resources indefinitely. The defaults are a 10-second connection
timeout and a 120-second idle-read timeout. The read timeout resets after every
successful read, so an active long-running stream is not limited to 120 seconds.
Callers serving providers that may legitimately remain silent longer can override
the values or disable the idle-read timeout.

* Public configuration: [`HttpTransportConfig`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L54-L76>) and [`new_with_transport_config`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L116-L142>).
* Error mapping: [reqwest timeouts remain the typed `LlmClientError::Timeout` variant](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L733-L747>).
* Contract test: [`routed_llm_client_exposes_timeout_variant`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L1824-L1858>) exercises a short timeout through the public configuration path.

#### 3\. Bounded buffered and error response bodies

**Motivation:** A provider controls both its response body and advertised
`Content-Length`. Relying on that metadata can reserve large amounts of memory
before any bytes arrive, while collecting an unbounded body allows concurrent
responses to exhaust the process. The client now grows buffers only from received
chunks, rejects buffered success bodies above 64 MiB, and retains at most 64 KiB
of an HTTP error body with an explicit truncation marker.

* Enforcement: [buffered success and error handling](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L310-L380>) and the shared [`collect_response_body`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L540-L580>) collector.
* Error-body test: [`upstream_error_body_is_bounded_and_redacted_from_display`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L1198-L1229>).
* Allocation regression: [`response_body_does_not_preallocate_advertised_content_length`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/client.rs#L1231-L1283>) advertises 64 MiB while yielding only two bytes.

#### 4\. Diagnostic redaction without weakening typed errors

**Motivation:** Provider credentials can be stored in static target headers, and
provider error bodies can echo prompts or secrets. Redacting only `Display` is
insufficient because structured tracing commonly records errors with `?error`,
which uses `Debug`. Header values and sensitive provider fields are now redacted
from both ordinary formatting paths. The bounded body/message fields remain
available to code that explicitly matches the typed error.

* Static headers: [`HttpBackendConfig::Debug` emits names but not values](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/backend.rs#L52-L62>), covered by [`debug_redacts_static_header_values`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/src/backend.rs#L285-L295>).
* Provider bodies: [`LlmClientError` display and debug formatting](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/protocol/src/client.rs#L73-L165>), covered by the [protocol redaction tests](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/protocol/src/client.rs#L259-L293>).

#### 5\. Bounded SSE frames before line buffering

**Motivation:** `AsyncBufReadExt::lines()` must accumulate a complete line before
returning it. A provider that never emits an SSE frame boundary could otherwise
grow that allocation without limit. The translation layer now checks raw transport
chunks before the line reader and rejects a frame above 8 MiB. The byte limiter and
parser share the same ASCII-whitespace boundary semantics so valid small frames do
not accumulate into a false oversized-frame failure.

* Enforcement: [`decode_stream_with_limit` and `SseFrameSizeTracker`](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/switchyard-translation/src/helpers.rs#L166-L304>).
* Tests: [chunk-spanning limit, oversized-frame, and whitespace-boundary coverage](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/switchyard-translation/src/helpers.rs#L682-L709>).

#### 6\. Public documentation

* [`switchyard-llm-client` README](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/crates/libsy-llm-client/README.md#L167-L216>) documents credentials, retries, timeout overrides, body limits, redaction, and the SSE limit.
* The [`Unreleased` changelog entry](<https://github.com/NVIDIA-NeMo/Switchyard/blob/889ce9553ceba1474c1569e6b25568b8513d2442/CHANGELOG.md#L7-L15>) records the behavior change for all client consumers.

## Why

`switchyard-llm-client` is the reusable provider-dispatch boundary for the
Switchyard server, libsy examples, and the external Relay plugin. It handles
credentials and untrusted network responses, so its defaults must protect:

* **authority integrity:** credentials and request bodies stay with the configured provider;
* **availability:** stalled connections and reads cannot retain resources indefinitely;
* **memory safety:** provider metadata, buffered responses, and SSE frames cannot cause unbounded allocation; and
* **credential/content privacy:** routine logs and spans do not copy target headers or provider-returned content.

These are generic client guarantees, not routing-algorithm behavior. PR NVIDIA-NeMo/Switchyard#270 can
build and function independently, while consuming these safer defaults once this
PR lands.

Relates to: NVIDIA-NeMo/Switchyard#270

## How tested

- [ ] `uv run ruff check .` clean (N/A: Rust-only change)
- [ ] `uv run mypy switchyard` clean (N/A: Rust-only change)
- [ ] `uv run pytest tests/` green (N/A: Rust-only change)
- [X] `cargo test -p switchyard-llm-client`
- [X] `cargo test -p switchyard-translation`
- [X] `cargo test -p switchyard-protocol -p switchyard-libsy`
- [X] `cargo clippy --workspace --all-targets -- -D warnings`
- [X] `cargo fmt --all -- --check`
- [ ] Manual smoke (not needed: behavior is covered with local HTTP fixtures)

## Checklist

- [X] One class per file; filename = `snake_case` of the primary class. (N/A: Rust-only change.)
- [X] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (N/A: no Python symbols.)
- [X] Unit tests added for new components / bug fixes.
- [X] README / `--help` updated if customer-facing surface changed.
- [X] Commits signed off (`Signed-off-by: Your Name <email>`) per the DCO.

## Notes for reviewers

* `TranslatingLlmClient::new` remains source-compatible and applies the safe defaults. The only new public surface is `HttpTransportConfig` plus `new_with_transport_config` for callers that need different timeout behavior.
* The read timeout is an **idle timeout**, not a total request deadline; every successful read resets it.
* Redirects are not followed. A provider 3xx is returned through the existing typed HTTP-error path.
* Oversized successful buffered responses fail. Oversized HTTP error bodies are truncated because callers still need bounded provider context for typed classification and diagnostics.
* Raw provider bodies remain available through explicit typed-error field access. Only implicit `Display` and `Debug` rendering is redacted.
* The 8 MiB SSE limit is enforced in `switchyard-translation`, before its line reader. This is the allocation boundary and protects every caller of the shared decoder.
* This PR adds no Relay plugin, manifest, Relay SDK dependency, routing behavior, or Relay-specific configuration.
* This PR is non-blocking for the Relay plugin integration. It captures hardening that could be made before production, but it should not prevent the integration from working.

## Summary by CodeRabbit

* **Security**
  * Sensitive header values are no longer exposed in diagnostic output.
  * HTTP error messages are sanitized to avoid displaying upstream response contents.
* **Reliability**
  * Added connection and read timeouts, redirect blocking, and response-size limits.
  * Oversized responses and streaming frames now fail safely with clear errors.
  * Large error responses are capped and marked as truncated.
* **Documentation**
  * Documented HTTP safety, timeout, redaction, and streaming limits.

### DIFF
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -52,14 +52,24 @@ pub struct HttpBackendConfig {
 impl fmt::Debug for HttpBackendConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let extra_header_names = self.extra_headers.keys().collect::<Vec<_>>();
+        let base_url = redacted_base_url(&self.base_url);
         f.debug_struct("HttpBackendConfig")
-            .field("base_url", &self.base_url)
+            .field("base_url", &base_url)
             .field("api_key", &self.api_key.as_ref().map(|_| "[REDACTED]"))
             .field("extra_header_names", &extra_header_names)
             .field("extra_body_keys", &self.extra_body.keys())
             .field("max_retries", &self.max_retries)
             .finish()
     }
+}
+
+fn redacted_base_url(base_url: &str) -> String {
+    let Ok(mut url) = reqwest::Url::parse(base_url) else {
+        return "[INVALID URL]".into();
+    };
+    let _ = url.set_username("");
+    let _ = url.set_password(None);
+    url.into()
 }
 
 /// A configured upstream backend, one variant per built-in wire format.
@@ -216,6 +226,21 @@ mod tests {
     fn openai_chat_url_joins_bare_v1() {
         let backend = Backend::OpenAiChat(config("https://api.openai.com/v1"));
         assert_eq!(backend.url(), "https://api.openai.com/v1/chat/completions");
+    }
+
+    #[test]
+    fn debug_redacts_base_url_userinfo() {
+        let debug = format!("{:?}", config("https://user:pass@provider.example/v1"));
+        assert!(debug.contains("provider.example/v1"));
+        assert!(!debug.contains("user"));
+        assert!(!debug.contains("pass"));
+    }
+
+    #[test]
+    fn debug_does_not_emit_an_invalid_base_url() {
+        let debug = format!("{:?}", config("not a valid url with a secret"));
+        assert!(debug.contains("[INVALID URL]"));
+        assert!(!debug.contains("secret"));
     }
 
     #[test]
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -51,10 +51,11 @@ pub struct HttpBackendConfig {
 
 impl fmt::Debug for HttpBackendConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let extra_header_names = self.extra_headers.keys().collect::<Vec<_>>();
         f.debug_struct("HttpBackendConfig")
             .field("base_url", &self.base_url)
             .field("api_key", &self.api_key.as_ref().map(|_| "[REDACTED]"))
-            .field("extra_headers", &self.extra_headers)
+            .field("extra_header_names", &extra_header_names)
             .field("extra_body_keys", &self.extra_body.keys())
             .field("max_retries", &self.max_retries)
             .finish()
@@ -279,6 +280,18 @@ mod tests {
         assert!(Backend::Anthropic(config("x")).is_anthropic());
         assert!(!Backend::OpenAiChat(config("x")).is_anthropic());
         assert!(!Backend::OpenAiResponses(config("x")).is_anthropic());
+    }
+
+    #[test]
+    fn debug_redacts_static_header_values() {
+        let mut config = config("https://provider.example/v1");
+        config
+            .extra_headers
+            .insert("authorization".into(), "Bearer provider-secret".into());
+
+        let debug = format!("{config:?}");
+        assert!(debug.contains("authorization"));
+        assert!(!debug.contains("provider-secret"));
     }
 
     #[test]
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -649,6 +649,16 @@ mod tests {
     }
 
     #[test]
+    fn decode_stream_accepts_data_fields_without_the_optional_space() -> Result<(), BoxError> {
+        let sse = b"data:{\"choices\":[{\"delta\":{\"content\":\"compact\"}}]}\n\ndata:[DONE]\n\n"
+            .to_vec();
+        let bytes = stream::once(async move { Ok::<Vec<u8>, LlmClientError>(sse) });
+        let chunks = decode_all(bytes, WireFormat::OpenAiChat)?;
+        assert_eq!(text_of(&chunks), "compact");
+        Ok(())
+    }
+
+    #[test]
     fn decode_stream_propagates_source_errors() -> Result<(), BoxError> {
         // A transport error mid-stream surfaces as an error item, not a panic.
         let bytes = stream::iter(vec![
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -26,6 +26,7 @@ static DEFAULT_TRANSLATION_POLICY: LazyLock<TranslationPolicy> =
     LazyLock::new(TranslationPolicy::default);
 static DEFAULT_TRANSLATION_ENGINE: LazyLock<TranslationEngine> =
     LazyLock::new(TranslationEngine::default);
+const MAX_SSE_FRAME_BYTES: usize = 8 * 1024 * 1024;
 
 /// Decodes a `wire_format` request body into the neutral IR.
 pub fn decode_request(wire_format: WireFormat, body: &Value) -> Result<LlmRequest> {
@@ -188,8 +189,21 @@ where
     // an `io::Error` item so `into_async_read`'s error bound resolves cleanly. The
     // source error is boxed intact rather than stringified, so
     // `llm_client_error_from_io` can recover its original variant on the way out.
+    // Validate frame growth before the line reader allocates an unbounded
+    // `String` for a provider event without a delimiter.
+    let bounded_bytes: Pin<
+        Box<dyn Stream<Item = std::result::Result<Vec<u8>, LlmClientError>> + Send>,
+    > = Box::pin(try_stream! {
+        let mut bytes = Box::pin(bytes);
+        let mut tracker = SseFrameSizeTracker::default();
+        while let Some(chunk) = bytes.next().await {
+            let chunk = chunk?;
+            tracker.observe(&chunk, MAX_SSE_FRAME_BYTES)?;
+            yield chunk;
+        }
+    });
     let io_bytes: Pin<Box<dyn Stream<Item = std::io::Result<Vec<u8>>> + Send>> =
-        Box::pin(bytes.map(|item| item.map_err(std::io::Error::other)));
+        Box::pin(bounded_bytes.map(|item| item.map_err(std::io::Error::other)));
     let lines = futures::io::BufReader::new(io_bytes.into_async_read()).lines();
 
     let mut state = StreamTranslationState {
@@ -239,6 +253,36 @@ where
     Ok(stream)
 }
 
+#[derive(Default)]
+struct SseFrameSizeTracker {
+    frame_bytes: usize,
+    current_line_has_content: bool,
+}
+
+impl SseFrameSizeTracker {
+    // Tracks blank-line frame boundaries across arbitrary transport chunks.
+    fn observe(&mut self, bytes: &[u8], limit: usize) -> std::result::Result<(), LlmClientError> {
+        for byte in bytes {
+            self.frame_bytes = self.frame_bytes.saturating_add(1);
+            match byte {
+                b'\n' if !self.current_line_has_content => self.frame_bytes = 0,
+                b'\n' => self.current_line_has_content = false,
+                b'\r' => {}
+                _ => self.current_line_has_content = true,
+            }
+            if self.frame_bytes > limit {
+                return Err(LlmClientError::InvalidResponse {
+                    source: Box::new(std::io::Error::new(
+                        std::io::ErrorKind::InvalidData,
+                        format!("provider SSE frame exceeded {limit} bytes"),
+                    )),
+                });
+            }
+        }
+        Ok(())
+    }
+}
+
 // Recover transport errors wrapped for `AsyncRead`; other reader failures are
 // invalid upstream responses.
 fn llm_client_error_from_io(error: std::io::Error) -> LlmClientError {
@@ -265,8 +309,8 @@ mod tests {
     };
 
     use super::{
-        decode_aggregated_response, decode_request, decode_stream, encode_aggregated_response,
-        encode_request, encode_stream, stamp_streamed_response_model,
+        SseFrameSizeTracker, decode_aggregated_response, decode_request, decode_stream,
+        encode_aggregated_response, encode_request, encode_stream, stamp_streamed_response_model,
     };
     use crate::{LlmResponseStream, WireFormat};
 
@@ -612,5 +656,17 @@ mod tests {
         };
         assert!(matches!(error, LlmClientError::ResponseTranslation(_)));
         Ok(())
+    }
+
+    #[test]
+    fn sse_frame_limit_spans_transport_chunks_and_resets_at_boundaries() {
+        let mut tracker = SseFrameSizeTracker::default();
+        tracker.observe(b"data: 1\n", 16).unwrap();
+        tracker.observe(b"\ndata: 2\n\n", 16).unwrap();
+
+        let mut oversized = SseFrameSizeTracker::default();
+        oversized.observe(b"data: ", 8).unwrap();
+        let error = oversized.observe(b"123", 8).unwrap_err();
+        assert!(matches!(error, LlmClientError::InvalidResponse { .. }));
     }
 }
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -204,9 +204,10 @@ where
     // `String` for a provider event without a delimiter.
     let bounded_bytes = try_stream! {
         let mut bytes = Box::pin(bytes);
+        let mut line_endings = SseLineEndingNormalizer::default();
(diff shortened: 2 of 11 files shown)


bbednarski9:
## `switchyard-rust-review` evidence

I reviewed PR #271 at [`2ef51d16`](https://github.com/NVIDIA-NeMo/Switchyard/commit/2ef51d1608ded9b5d280d27a7c18cc63768748c1) against `main` at [`a9c04b31`](https://github.com/NVIDIA-NeMo/Switchyard/commit/a9c04b31b8bdb5781fa0e0ebda26f68730917c33), following the repository's `switchyard-rust-review` skill. I completed separate passes for correctness, async/streaming behavior, memory allocation, HTTP and credential boundaries, public error contracts, comments/naming, and tests, followed by the required second pass over every changed hunk.

### Verdict: changes requested

#### High

1. **An untrusted `Content-Length` can reserve 64 MiB before any response bytes arrive.** [`collect_response_body`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/libsy-llm-client/src/client.rs#L526-L548) passes the provider-controlled `content_length()` directly to `Vec::with_capacity`, capped only by the 64 MiB body limit. A slow or malicious provider can therefore make every concurrent buffered call reserve 64 MiB while sending a tiny body; 256 calls can request 16 GiB of capacity before the limit protects anything. Start with an empty or small bounded capacity and grow only as bytes are received. Add a regression test with an inflated `Content-Length` and a minimal body. This independently confirms the existing CodeRabbit finding.

2. **Provider bodies are still exposed through `Debug`.** `LlmClientError` continues to derive `Debug` at [`protocol/src/client.rs:23-30`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/protocol/src/client.rs#L23-L30), so both retained fields at [`protocol/src/client.rs:73-89`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/protocol/src/client.rs#L73-L89) appear verbatim in `format!("{error:?}")` and structured `tracing` fields written with `?error`. Redacting only `Display` does not support the README claim that provider content is not copied into ordinary logs or spans. Implement a redacted `Debug` representation for these sensitive fields and test both `Display` and `Debug` for `ContextWindowExceeded` and `UpstreamHttp`.

#### Medium

3. **The new 120-second read timeout is a mandatory global behavior change.** The timeout constants at [`client.rs:54-58`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/libsy-llm-client/src/client.rs#L54-L58) are installed by the only public constructor at [`client.rs:98-113`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/libsy-llm-client/src/client.rs#L98-L113). Existing callers with a provider that legitimately emits no bytes for more than two minutes—particularly long reasoning requests—will now fail with no override. Keep a secure default, but expose a bounded client/transport configuration path and add a short-duration timeout test rather than baking one operational policy into every library consumer.

4. **The SSE limiter and parser disagree about whitespace-only frame boundaries.** `decode_stream` treats `line.trim_end().is_empty()` as a boundary at [`helpers.rs:214-238`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/switchyard-translation/src/helpers.rs#L214-L238), but [`SseFrameSizeTracker`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/switchyard-translation/src/helpers.rs#L256-L280) marks spaces/tabs as content and does not reset on `" \n"`. A sequence of individually valid small frames separated by whitespace-only lines is therefore counted cumulatively and can eventually be rejected as an oversized frame. Use exactly the parser's boundary semantics. The test at [`helpers.rs:661-670`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/switchyard-translation/src/helpers.rs#L661-L670) should exercise `decode_stream`, including a whitespace boundary, rather than only the private tracker. This independently confirms the existing CodeRabbit finding.

#### Low

5. **The stream limiter adds an avoidable trait-object allocation.** [`helpers.rs:194-206`](https://github.com/NVIDIA-NeMo/Switchyard/blob/2ef51d1608ded9b5d280d27a7c18cc63768748c1/crates/switchyard-translation/src/helpers.rs#L194-L206) boxes `bounded_bytes`, then immediately maps and boxes it again as `io_bytes`. Keep the `try_stream!` value concrete and erase the type only once at the `io_bytes` boundary.

### Positive checks

- The PR is now focused on the reusable HTTP client and contains no Relay/plugin coupling.
- Production changes add no `unwrap`/`expect`, blocking executor work, unbounded channels, or new dependencies.
- Redirect rejection, bounded body reads, source-error preservation, and static-header-value redaction are placed in the correct client/translation boundaries.
- `git diff --check` and the required second-pass hunk review were clean apart from the findings above.

### Validation performed at this head

- `cargo fmt --all -- --check` — passed.
- `cargo clippy --workspace --all-targets -- -D warnings` — passed.
- `cargo test -p switchyard-protocol -p switchyard-translation -p switchyard-libsy -p switchyard-llm-client` — passed (30 protocol, 25 translation unit plus integration suites, 222 libsy plus observability, and 48 LLM-client tests).
- `cargo test --workspace` — attempted; Rust compilation succeeded through the affected crates, but the local macOS link of `switchyard-py` failed because Python symbols were unavailable in this shell. GitHub's Rust and aggregate **CI Success** checks are green at `2ef51d16`, including the Ubuntu workspace validation.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -239,6 +253,36 @@ where
...
+impl SseFrameSizeTracker {
+    // Tracks blank-line frame boundaries across arbitrary transport chunks.
+    fn observe(&mut self, bytes: &[u8], limit: usize) -> std::result::Result<(), LlmClientError> {
+        for byte in bytes {
+            self.frame_bytes = self.frame_bytes.saturating_add(1);
+            match byte {
+                b'\n' if !self.current_line_has_content => self.frame_bytes = 0,
+                b'\n' => self.current_line_has_content = false,
+                b'\r' => {}
+                _ => self.current_line_has_content = true,
```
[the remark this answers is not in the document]
bbednarski9: Fixed in e8081d13414c5526335a2bf38e2b8afc5d93fadd. SseFrameSizeTracker now treats ASCII whitespace-only lines as frame boundaries, matching the decoder for the reported space-newline case. The regression uses two frames separated by a whitespace-only line under a tight byte limit, so it fails if the second frame inherits the first frame byte count. Translation tests and workspace Clippy pass.

bbednarski9:
Comment#1 above addressed in e8081d1

bbednarski9:
Comment#2 addressed in a61a981

bbednarski9:
comment#3 addressed in b63db0f
comment#4 addressed in ca0c1aa

bbednarski9:
comment#5 addressed in 4fb0a2a

bbednarski9:
Some small (potential) regressions to follow-up on:
SSE bare-CR line endings fail. The standard permits CRLF, LF, or CR line endings. [sse_regressions.rs (line 67)](/private/tmp/switchyard-pr271-probes/tests/sse_regressions.rs:67) reproduces this. [WHATWG SSE rules](https://html.spec.whatwg.org/multipage/server-sent-events.html).

Valid data:{...} fields without the optional space are silently ignored. [sse_regressions.rs (line 75)](/private/tmp/switchyard-pr271-probes/tests/sse_regressions.rs:75) reproduces this.

Backend debug output leaks credentials embedded in URL userinfo. [redaction.rs (line 31)](/private/tmp/switchyard-pr271-probes/tests/redaction.rs:31) reproduces this. Static header and typed provider-error redaction otherwise passed.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -201,8 +227,8 @@ where
         futures::pin_mut!(lines);
         while let Some(line) = lines.next().await {
             let line = line.map_err(llm_client_error_from_io)?;
-            // A blank line (allowing a bare CR for CRLF streams) ends the frame.
-            if line.trim_end().is_empty() {
+            // An ASCII-whitespace-only line ends the frame.
+            if is_sse_frame_boundary(&line) {
                 let parsed = sse::parse_json_sse_frame(&frame, marker)
```
mnajafian-nv, line 233: Tested on head 889ce95. `data:{...}` is dropped on this path because the parser still only accepts `data: ` with a space. SSE allows both forms, so this should accept `data:` too.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -239,6 +265,44 @@ where
     Ok(stream)
 }
 
+fn is_sse_whitespace(byte: &u8) -> bool {
+    byte.is_ascii_whitespace()
+}
+
+fn is_sse_frame_boundary(line: &str) -> bool {
+    line.as_bytes().iter().all(is_sse_whitespace)
```
mnajafian-nv, line 274: Confirmed on head 889ce95 with a `\r\r`-delimited SSE stream. This still fails with `ResponseTranslation(...)` instead of decoding the frame, so bare-CR line endings are not handled on this path yet.

crates/libsy-llm-client/src/backend.rs
```diff
@@ -51,10 +51,11 @@ pub struct HttpBackendConfig {
 
 impl fmt::Debug for HttpBackendConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let extra_header_names = self.extra_headers.keys().collect::<Vec<_>>();
         f.debug_struct("HttpBackendConfig")
             .field("base_url", &self.base_url)
```
mnajafian-nv: One redaction gap from a focused test on head 889ce95: with `https://user:pass@provider.example/v1`, `Debug` prints `base_url` verbatim, so URL userinfo can leak into ordinary diagnostics.

gburachas:
This overlaps something I had queued, so flagging it here rather than opening a competing change.

I have a total per-request timeout (`timeout_secs` on `[llm_clients.<name>]`, tracked in #277). It is complementary to the two timeouts here rather than an alternative: a connection timeout bounds setup, and an idle-read timeout bounds the gap between reads, but neither bounds total duration. An upstream that emits one token every 100 seconds keeps resetting the idle-read timer while holding the request open indefinitely.

If this PR lands first I will add it as a third field on `HttpTransportConfig` rather than as a separate mechanism. Happy to do that as a follow-up, or to fold it in here if you would prefer it in one change.

PR #273, opened 2026-08-04
fix: resolve routing tier for fallback and affinity-reuse decisions

### LINKED ISSUE #276: [bug] llm_classifier: fallback and affinity-reuse decisions log without a routing tier
## Symptom

A routing decision made by the cascade fallback or served from an affinity pin logs without its routing tier, while a judged decision to the same target logs with one. Both lines come from the same site, so a log consumer sees the same model sometimes tiered and sometimes not:

```
fall-through selected ok-strong (confidence 1.000); routing tier: strong
fall-through selected ft-strong (confidence 0.000)
```

## Reproduction

Two tests in the linked PR, both failing without its `fall_through.rs` change:

- `judge_unavailable_fallback_logs_the_routing_tier`: the judge call fails, `TaskClassifier` fails open, `DefaultTarget` decides, and the decision logs with no tier.
- `affinity_reuse_logs_the_routing_tier`: the second turn of a pinned session is decided by `AffinityRouter` and likewise logs with no tier.

## Expected vs. actual

- **Expected:** a decision names the tier of the model actually selected, regardless of which classifier in the cascade decided it. #338 states that routing-tier and fallback details are preserved in the decision reasoning, and #413 moved that reasoning into the decision log.
- **Actual:** `with_routing_tier` asks only the deciding classifier for the tier (`crates/libsy/src/algorithms/fall_through.rs:281`). `DefaultTarget` (`crates/libsy/src/algorithms/fall_through.rs:71`) and `AffinityRouter` (`crates/libsy/src/algorithms/util/affinity.rs`) do not implement `Classifier::routing_tier`, so the trait default returns `None` (`crates/libsy/src/core/classifier.rs:75`) and the tier is dropped.

## Environment

- Switchyard version / commit: current `main`, verified at `61ee1b31`
- OS / arch: macOS ARM64, source build
- Inbound format: Chat Completions
- Backend: OpenAI-compatible (observed live via OpenRouter)

## History

Filed originally against the pre-#338 metric shape, where this same gap dropped the `tier` attribute from `switchyard.requests` and split one model across a labeled and an unlabeled series (observed live: 1 request labeled `tier="weak"`, 45 on the strong model with no tier). #338 removed that attribute from the counter deliberately and moved routing-tier detail into the decision reasoning, and #413 moved the reasoning into the log. The cause is unchanged, so this issue now describes the log-side gap.

## Additional context

Related: #205 counts how often the classifier failed open and why; this issue is about which tier actually served, so the two compose. #272 covers why agentic sessions hit the fallback so often.

thossullivan:
Closes #276.

## Why

A routing decision made by the cascade fallback, or served from an affinity pin, logs without its routing tier while a judged decision to the same target logs with one:

```
fall-through selected ok-strong (confidence 1.000); routing tier: strong
fall-through selected ft-strong (confidence 0.000)
```

#338 preserves routing-tier and fallback details in the decision reasoning and #413 moved that reasoning into the log, so a decision that omits the tier loses detail the log is meant to carry.

## What

`with_routing_tier` asks only the deciding classifier for the tier. Neither `DefaultTarget` nor `AffinityRouter` implements `Classifier::routing_tier`, so the trait default returns `None` and the tier is dropped. This resolves the tier across the cascade when the deciding classifier defines none.

## How

One lookup in `crates/libsy/src/algorithms/fall_through.rs`: fall back to the first classifier in the cascade that defines a tier for the selected target. `TaskClassifier` sits in the cascade in both cases, so the fallback and affinity paths resolve the same `weak`/`strong` labels a judged decision would produce.

## Test Plan

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test -p switchyard-protocol -p switchyard-libsy -p switchyard-llm-client -p switchyard-server`
- [x] `git diff --check`
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/ -m "not integration"`

Two tests in `crates/libsy-llm-client/tests/observability.rs` cover the fallback and affinity-reuse paths. Both fail without the `fall_through.rs` change.

## Note on this branch

This was rebased at maintainer request and became a rewrite rather than a replay. The original commits patched `FallThroughDecision.tier` and asserted a `tier` attribute on `switchyard.requests`; #338 removed both, so there was nothing left for them to apply to. The defect they described survives in the decision log, and this branch addresses it there.

### DIFF
--- a/crates/libsy-llm-client/tests/observability.rs
+++ b/crates/libsy-llm-client/tests/observability.rs
@@ -1436,3 +1436,39 @@ async fn classifier_fail_open_records_each_failure_stage() -> switchyard_libsy::
     }
     Ok(())
 }
+
+/// A decision made by the cascade fallback still names its routing tier.
+///
+/// The judge call fails, so `TaskClassifier` fails open and `DefaultTarget`
+/// decides. That decider defines no tier of its own, so without the cascade
+/// lookup the selection logs `tier: None` while a judged decision to the same
+/// target logs its tier. `AffinityRouter` reuse takes the same path.
+#[tokio::test]
+async fn fallback_decision_logs_the_routing_tier() -> switchyard_libsy::Result<()> {
+    let _guard = serialize_test().lock().await;
+    let (store, _, _, _, _) = telemetry();
+
+    let client = Arc::new(JudgeClient {
+        judge_model: "ft-judge".into(),
+        outcome: JudgeOutcome::CallFailure,
+    }) as Arc<dyn RoutedLlmClient>;
+    run(
+        classifier_router("ft-judge", "ft-weak", "ft-strong")?,
+        client,
+        classifier_request(),
+    )
+    .await?;
+
+    let events = store.events();
+    assert!(
+        events.iter().any(|event| {
+            event.fields.get("target").map(String::as_str) == Some("ft-strong")
+                && event
+                    .fields
+                    .get("tier")
+                    .is_some_and(|tier| tier.contains("strong"))
+        }),
+        "fallback selection logged without its routing tier: {events:?}"
+    );
+    Ok(())
+}
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -264,7 +264,13 @@ where
         // 3. Resolve the target and log the choice.
         algorithm::ensure_model_is_target(&self.targets, &score.target)?;
         let target = score.target.clone();
-        let tier = deciding.routing_tier(&target);
+        // A fallback or affinity-reuse decider carries no tier of its own, so
+        // resolve it across the cascade rather than logging it as None.
+        let tier = deciding.routing_tier(&target).or_else(|| {
+            self.classifiers
+                .iter()
+                .find_map(|c| c.routing_tier(&target))
+        });
         tracing::info!(algorithm=self.name, target=%score.target, confidence=score.confidence, tier = ?tier, "Model selected");
 
         // 4. Post-decision replay: every processor sees the selection so stateful ones


crates/libsy/tests/observability.rs
```diff
@@ -1121,3 +1121,139 @@ async fn classifier_metrics_count_only_the_final_routed_call() -> switchyard_lib
...
+            .run(
+                Context::default(),
+                request_with_metadata("affinity-tier-session", "affinity-tier-corr"),
+            )
+            .await?;
+        assert_eq!(
+            trace.last().and_then(|decision| decision.routing_tier()),
+            Some("weak")
+        );
+    }
```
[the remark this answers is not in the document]
thossullivan: Good catch, the test could have passed via a second classification rather than the reuse path. Added a judge-call counter to ClassifierClient and asserted the second request adds none, so the affinity pin is now what the test proves. Fixed in 842aa15.

ayushag-nv:
@thossullivan Thanks for putting up the PR. Can you put up a Issue and link a PR to that.

thossullivan:
Filed #276 for this and linked it in the description. Thanks for the look!

ayushag-nv:
@thossullivan Thanks for putting up the PR. Can you please rebase over main once

thossullivan:
@ayushag-nv sorry for the delay, bad storms knocked out our power for about a week! 

Rebased. It became a rewrite rather than a replay.

The original commits patched `FallThroughDecision.tier` and asserted a `tier` attribute on `switchyard.requests`. #338 removed both, so there was nothing left to replay. You were right that the `tier` field is gone (like you mentioned in the issue).

The gap survives in the decision log, where #338 moved the detail and #413 turned it into a log message. `with_routing_tier` asks only the deciding classifier, and neither `DefaultTarget` nor `AffinityRouter` implements `routing_tier`, so on 61ee1b31:

​```
fall-through selected ok-strong (confidence 1.000); routing tier: strong
fall-through selected ft-strong (confidence 0.000)
​```

The branch is now one commit that resolves the tier across the cascade when the deciding classifier defines none, plus two tests covering the fallback and affinity-reuse paths that fail without it.

If you would rather the tier stay out of this path entirely, I am still happy to close or work on the cleanup instead!

crates/libsy-llm-client/tests/observability.rs
```diff
@@ -1357,3 +1357,92 @@ async fn classifier_fail_open_records_each_failure_stage() -> switchyard_libsy::
...
 }
+
+/// A decision made by the cascade fallback still names its routing tier.
+///
+/// The judge call fails, so `TaskClassifier` fails open and `DefaultTarget`
+/// decides. That decider defines no tier of its own, so before the cascade
+/// lookup the decision logged without one while a judged decision logged with
+/// one.
+#[tokio::test]
+async fn judge_unavailable_fallback_logs_the_routing_tier() -> switchyard_libsy::Result<()> {
```
nachiketb-nvidia: cam we reduce to one targetted test?

nachiketb-nvidia (approved):
looks good! there's a conflict i see

quick rebase oughtta fix it

Thanks!

Sorry about the rebases, we were doing a core library refactor, making it nice for integrators

crates/libsy-llm-client/tests/observability.rs
thossullivan: `fallback_decision_logs_the_routing_tier` is the one I kept and it still fails without the fix. I dropped the `affinity-reuse` test rather than folding it in, so that path is covered by the same lookup but no longer has a test of its own. Happy to add it back if you would rather have the coverage!

thossullivan:
@nachiketb-nvidia No problem at all with the rebases! Love the progress on the repo :) 

Rebased onto main and reduced to the single test as requested. The refactor actually made this even simpler. 

With the tier now a structured field on the `Model selected` event rather than text appended to the message, the change is one lookup (i.e., ask the cascade when the deciding classifier defines none). The old tests asserted on message text that no longer exists, so they needed rewriting regardless.

PR #294, opened 2026-08-04
fix(server): handle graceful process shutdown

nachiketb-nvidia:
## What

- handle Ctrl+C and Unix SIGTERM in the Rust server binary
- apply one configurable drain deadline to HTTP and TLS servers
- expose `--shutdown-timeout`, defaulting to 30 seconds

## Why

Docker and Kubernetes terminate processes with SIGTERM. The server previously handled only Ctrl+C, plain HTTP could drain forever, and TLS allowed only two seconds. That could either drop telemetry and active requests immediately or block shutdown indefinitely.

## How

Both transports now use the existing `axum-server` graceful-shutdown handle. The signal future remains a small Tokio wrapper because SIGTERM is Unix-specific; embedded callers can continue supplying their own shutdown future and deadline.

## What to review

- signal behavior in `shutdown_signal`
- the shared bounded drain path in `BoundServer::serve`
- the 30-second CLI default versus the preserved two-second Python binding contract

## Validation

- `cargo check -p switchyard-py`
- live SIGTERM run: server logged the drain and exited with status 0

## Summary by CodeRabbit

* **New Features**
  * Added graceful server shutdown with configurable request-draining time.
  * Added `--shutdown-timeout`, defaulting to 30 seconds.
  * Shutdown now responds to Ctrl+C and Unix `SIGTERM`, allowing active requests to complete before stopping.

* **Bug Fixes**
  * Added enforcement of shutdown deadlines so requests do not block server termination indefinitely.

* **Documentation**
  * Documented graceful shutdown behavior and timeout configuration.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -309,18 +309,13 @@ async fn serve_tls(
     let config = RustlsConfig::from_pem_file(tls.cert, tls.key)
         .await
         .map_err(server_io_error)?;
-    let handle = axum_server::Handle::new();
-    let shutdown_task = schedule_shutdown(handle.clone(), shutdown_timeout, shutdown);
-
     let std_listener = listener.into_std().map_err(server_io_error)?;
-    let result = axum_server::from_tcp_rustls(std_listener, config)
-        .map_err(server_io_error)?
-        .handle(handle)
-        .serve(router.into_make_service())
-        .await
-        .map_err(server_io_error);
-    shutdown_task.abort();
-    result
+    let server = axum_server::from_tcp_rustls(std_listener, config).map_err(server_io_error)?;
+    let handle = axum_server::Handle::new();
+    let server = server
+        .handle(handle.clone())
+        .serve(router.into_make_service());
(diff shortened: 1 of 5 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -302,31 +310,51 @@ async fn serve_tls(
...
+    shutdown: impl Future<Output = ()> + Send + 'static,
+) -> task::JoinHandle<()> {
+    tokio::spawn(async move {
+        shutdown.await;
+        tracing::info!(
+            ?timeout,
+            "shutdown signal received; draining active requests"
+        );
+        handle.graceful_shutdown(Some(timeout));
+    })
```
grahamking: Instead of the two `task` and sharing the handle, you could use ` tokio::select!`. You await _both_ the server and the shutdown signal, and act on the first one.

grahamking (approved):

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: fixed!

PR #298, opened 2026-08-05
fix(routing): scope context-overflow history by session and agent

elyasmnvidian:
```toml
schema_version = 1

[llm_clients.mock]
format = "openai_chat"
base_url = "http://127.0.0.1:<port>/v1"
max_retries = 0

[targets.weak]
id = "model/weak"
llm_client = "mock"

[targets.strong]
id = "model/strong"
llm_client = "mock"

[routes.random]
id = "switchyard/random"
type = "random"
targets = ["weak", "strong"]
weights = [1, 0]
```

With a local provider where `model/weak` rejects `overflow` for context length and
`model/strong` succeeds, send these requests under one session:

```text
child-a (agent=child-a): input=overflow
parent (root):           input=fits
child-b (agent=child-b): input=fits
```

## The bug

One oversized child request marked `weak` overflowed for the whole session, so the parent
and `child-b` then skipped `weak` even though their shorter requests fit:

```text
child-a calls: weak, strong; selected strong
parent calls:  strong        # wrong: weak was never tried
child-b calls: strong        # wrong: weak was never tried
```

The overflow history keyed every request by session ID alone. A child's overflow was
therefore recorded against the session and excluded `weak` for the parent and every sibling.

## The fix

`SessionEvictions` now uses the same crate-private routing identity as `AffinityRouter`:

- a root request uses its session ID;
- a child request uses its session and agent IDs.

Both paths build that identity with `RoutingIdentity::from_request`. A child missing either ID
keeps no overflow history instead of sharing the parent's. When the host marks the session
final, Switchyard clears the root and every child's history.

`random`, `llm_classifier`, and `stage_router` route their final model call through
`FallThrough`, so all three receive the fix. `passthrough` and `noop` are unchanged.

## After

```text
child-a calls:     weak, strong; selected strong
child-a next turn: strong
parent calls:      weak; selected weak
child-b calls:     weak; selected weak
```

## Proof

The same production-path regression test was applied to both revisions without copying the
production fix:

```text
origin/main 31e0afb2: FAIL
left:  Some("model/strong")
right: Some("model/weak")

PR 7dd20f81: PASS
```

The test starts a local HTTP provider and drives the real header parser, route, client error
mapper, session-final cleanup, and missing-agent behavior. It does not call a live provider.

## Checks

```text
cargo fmt --all
cargo fmt --all --check
cargo test -p switchyard-libsy
cargo test -p switchyard-server --test server overflow_history_is_scoped_to_agent_and_session_lifetime -- --exact
cargo test --workspace --exclude switchyard-py
uv run ruff check .
uv run mypy switchyard
uv run pytest tests/ -m "not integration"
```

The full `cargo test --workspace` command reaches the known local macOS PyO3 link failure for
`switchyard-py`; the workspace run excluding only that extension crate passes.

## Scope

An earlier version of this PR also failed over to the next target on an unavailable target.
That is a separate change and will come in its own PR. This PR only fixes context-overflow
history isolation.

### DIFF
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -15,6 +15,7 @@
 //!
 //! Every composition retains one thing regardless: a target that overflows its context window is
 //! remembered for the rest of its session and skipped on later turns.
+//! An unavailable target is skipped only for the current request.
 
 use std::{
     collections::HashMap,
@@ -32,7 +33,7 @@ use crate::core::algorithm::{
 use crate::core::classifier::{Classification, Classifier, Score};
 use crate::core::processor::{Event, Processor};
 use crate::{LibsyError, Result};
-use switchyard_protocol::{Context, Decision, Request, Response};
+use switchyard_protocol::{Context, Decision, Request, Response, RoutingFallbackReason};
 
 struct SessionState<S> {
     state: Arc<AsyncMutex<S>>,
@@ -57,6 +58,7 @@ pub struct FallThroughDecision {
     /// Human-readable explanation of the selection.
     pub reasoning: String,
     tier: Option<&'static str>,
+    fallback_reason: Option<RoutingFallbackReason>,
 }
 
 impl Decision for FallThroughDecision {
@@ -66,6 +68,10 @@ impl Decision for FallThroughDecision {
 
     fn routing_tier(&self) -> Option<&str> {
         self.tier
+    }
+
+    fn fallback_reason(&self) -> Option<RoutingFallbackReason> {
+        self.fallback_reason
     }
 
     fn reasoning(&self) -> Option<&str> {
@@ -237,7 +243,7 @@ where
             identity.as_ref(),
         );
         let session_state = self.session_state(&request);
-        let (target, decision, served) = match session_state {
+        let (target, decision, served, deciding) = match session_state {
             Some(state) => {
                 let mut state = state.lock().await;
                 self.route(&mut state, &ctx, &driver, &mut request).await?
@@ -256,7 +262,7 @@ where
         match served {
             Some(response) => Ok(response),
             None => {
-                algorithm::call_llm_with_overflow_fallback(
+                algorithm::call_llm_with_fallback(
                     ctx,
                     &driver,
                     &self.targets,
@@ -265,7 +271,12 @@ where
                     request,
                     identity.as_ref(),
                     &self.session_evictions,
-                    |from, to| self.fallback_decision(from, to),
+                    |request, target| {
+                        for classifier in &self.classifiers {
+                            classifier.target_unavailable(request, target);
+                        }
+                    },
+                    |from, to, reason| self.fallback_decision(deciding.as_ref(), from, to, reason),
                 )
                 .await
             }
@@ -280,18 +291,26 @@ where
         self.session_evictions.remove_session(session);
     }
 
-    /// The decision published when an overflow sends the request to a different target.
-    fn fallback_decision(&self, from: &LlmTarget, to: &LlmTarget) -> Arc<dyn Decision> {
+    /// The decision published when a route-level failure selects a different target.
+    fn fallback_decision(
+        &self,
+        deciding: &dyn Classifier<S>,
+        from: &LlmTarget,
+        to: &LlmTarget,
+        reason: RoutingFallbackReason,
+    ) -> Arc<dyn Decision> {
+        let failure = match reason {
+            RoutingFallbackReason::ContextWindow => "exceeded its context window",
+            RoutingFallbackReason::Unavailable => "was unavailable",
+        };
         Arc::new(FallThroughDecision {
             selected_model: to.semantic_name.clone(),
             reasoning: format!(
-                "{} exceeded its context window; fell back to {}",
-                from.semantic_name, to.semantic_name
+                "{} {failure}; fell back to {}",
+                from.semantic_name, to.semantic_name,
             ),
-            tier: self
-                .classifiers
-                .iter()
-                .find_map(|c| c.routing_tier(&to.semantic_name)),
+            tier: deciding.routing_tier(&to.semantic_name),
+            fallback_reason: Some(reason),
         })
     }
 
@@ -315,7 +334,12 @@ where
         ctx: &Context,
         driver: &Driver,
         request: &mut Request,
-    ) -> Result<(LlmTarget, Arc<dyn Decision>, Option<Response>)> {
+    ) -> Result<(
+        LlmTarget,
+        Arc<dyn Decision>,
+        Option<Response>,
+        Arc<dyn Classifier<S>>,
+    )> {
         // 1. Processor chain accumulates request-side facts into the composition's state.
         for processor in &self.processors {
             processor.process(state, Event::Request(request)).await?;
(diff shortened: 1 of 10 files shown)


elyasmnvidian:
@coderabbitai review

crates/libsy/src/core/algorithm.rs
```diff
@@ -407,126 +404,28 @@ impl LlmTargetSet {
...
-/// A conversation only grows, so a target that could not fit one turn will not fit a
-/// later one; remembering it lets the next turn skip a call certain to fail. Requests
-/// without a session id are not tracked — there is nothing to remember them by.
-#[derive(Default)]
-pub(crate) struct SessionEvictions {
-    sessions: Mutex<HashMap<String, HashSet<String>>>,
-}
-
-impl SessionEvictions {
-    /// Forgets overflow history for a completed session.
```
ayushag-nv, line 425: may be we should not remove it and fix this one to handle the sub agent case here

elyasmnvidian: Agreed. After splitting unavailable-target failover into its own change, this PR keeps SessionEvictions in core::algorithm and only changes its key from the session ID to the same session-or-subagent identity used by affinity.

crates/switchyard-server/tests/server.rs
```diff
@@ -513,8 +534,49 @@ fn load_test_config(toml: &str) -> TestResult<ServerState> {
     Ok(load_server_state(config.path())?)
 }
 
+fn ordered_random_state(base_url: &str, first: &str, second: &str) -> TestResult<ServerState> {
```
nachiketb-nvidia: one regression test should be enough

crates/libsy/src/algorithms/util/affinity.rs
```diff
@@ -24,24 +24,19 @@ use async_trait::async_trait;
...
-/// A sub-agent request is keyed by `session + agent` and a root request by session alone,
-/// so a sub-agent's assignment is scoped within — but distinct from — its session's.
+/// Map key for a retained model assignment.
 #[derive(Clone, Hash, PartialEq, Eq)]
 enum AffinityKey {
-    /// One model per session, for root-agent traffic.
-    Session(String),
-    /// One model per identified child agent within a session.
-    Subagent { session: String, agent: String },
+    Routing(RoutingIdentity),
```
nachiketb-nvidia: the bug here is probably that incorrect Session strings or Subagent enum variants are being passed. There's probably an ALAP AMAN fix we can make, targetted, on the server side specifically

nachiketb-nvidia (changes requested):

crates/switchyard-server/tests/server.rs
elyasmnvidian: Yes — down to one test.

The only new test is `overflow_history_is_scoped_to_agent_and_session_lifetime`. It runs one session straight through the whole behavior, in order:

- a child sends an oversized request, the weak model overflows, and the request falls back to the strong model;
- the parent and a sibling child then send short requests and still get the weak model;
- the session ends, which clears the history;
- a child that says it is a subagent but sends no agent id keeps no history.

One test covers all four. The two tests about a target being unavailable moved out with that change to a separate PR.

crates/libsy/src/algorithms/util/affinity.rs
elyasmnvidian: No — the fix is in libsy, not on the server side. The agent values coming in are already correct, so there is nothing to fix there.

The bug shows up even when the request sends the right agent headers. Three different agents share one session, and only the first request is oversized:

```
child-a  agent-id=child-a  is-subagent=true   prompt: long   -> overflows the weak model
parent   agent-id=root     is-subagent=false  prompt: short
child-b  agent-id=child-b  is-subagent=true   prompt: short
```

Selected model per request today:

```
child-a: strong   # correct: weak overflowed, so it fell back to strong
parent:  strong   # wrong: the short prompt fits, should be weak
child-b: strong   # wrong: should be weak
```

The parent and sibling never even call the weak model. `SessionEvictions` in `crates/libsy/src/core/algorithm.rs` keys the overflow history by session id alone, so one child's overflow drops the weak model for every agent in the session. The affinity code right next to it already keys a child by session + agent; the overflow history did not.

The fix keys the overflow history the same way affinity does: a root request by session, a child by session + agent. A child with no agent id keeps no history instead of sharing the root's, and the whole session's history clears when the session ends.

Same three requests after the fix:

```
child-a: strong   parent: weak   child-b: weak
```

That is the exact sequence the bug report (SWYD-106) expects. Verified with `cargo test -p switchyard-server --test server overflow_history_is_scoped_to_agent_and_session_lifetime` and by running the bug's own stage_router repro.

If you have seen a real subagent that is not detected at all — wrong is_subagent or agent id coming from a harness — that is a separate problem from this one, and I am happy to file it.

crates/libsy/src/core/algorithm.rs
```diff
@@ -407,51 +407,89 @@ impl LlmTargetSet {
...
 }
 
-/// Bounds process-local overflow history. Dropping a live session's entry costs one
-/// rediscovered overflow, so the victim choice does not need to be exact.
-const MAX_EVICTION_SESSIONS: usize = 1_024;
+/// Key for overflow history: a root request by its session, a child request by its session
+/// and agent. Keying a child finer than its session keeps one child's overflow from evicting
+/// a target for the parent or a sibling sharing the session.
+#[derive(Clone, Hash, PartialEq, Eq)]
+pub(crate) enum RoutingIdentity {
```
nachiketb-nvidia, line 415: why this dup enum from AffinityKey?

ayushag-nv:
Core overflow history should stay in `SessionEvictions`, but it does not need a second identity enum. The narrow design is a private tuple alias:

```rust
type OverflowIdentity = (String, Option<String>);
```

The key semantics are:

- root request: `(session_id, None)`
- identified child: `(session_id, Some(agent_id))`
- child missing either ID: no key, so no overflow history is retained

`FallThrough` should compute this key once before routing and use the same value for both `exclude_evicted` and `record`. `remove_session` can remove every tuple whose session matches. This keeps the behavior in this PR: an overflow from one child does not exclude a target for the root or a sibling, while `session_final` still clears the whole session.

I would not reuse `AffinityKey` from core. It is private to the affinity component, and its construction includes affinity-only policy such as subagent-only mode and message-hash fallback. Making core overflow handling depend on it would couple two separate policies. If more components later need a canonical metadata identity, we can extract a neutral shared type in a separate change; this fix does not require one.

I tested the tuple-alias version locally. The focused server regression, `cargo fmt --all --check`, strict workspace Clippy, and `cargo test --workspace` all pass. The existing server regression remains sufficient.

crates/libsy/src/core/algorithm.rs
elyasmnvidian: Good catch. There was no reason to keep a second enum. AffinityRouter and overflow history now use the same crate-private RoutingIdentity and RoutingIdentity::from_request constructor. Affinity subagent-only and message-hash behavior remains local to AffinityRouter. The focused server regression, libsy tests, workspace Clippy, and repository checks pass.

elyasmnvidian:
@ayushag-nv Thanks. The current update keeps `SessionEvictions` in `core::algorithm` and keeps affinity-only policy inside `AffinityRouter`, as you suggested.

Nachiket also flagged that the tuple/enum shape would duplicate the identity already used by affinity. I addressed both points with a neutral crate-private `RoutingIdentity` and `RoutingIdentity::from_request` in `core::algorithm`. That type only represents request metadata: root is session, identified child is session plus agent, and an incomplete child has no identity. `subagents_only` and message-hash fallback remain private to affinity.

This avoids two constructors drifting while preserving the same narrow behavior and ownership. The production regression fails on current `main` when the parent selects strong, passes on this head, and all local checks listed in the PR body pass.

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -198,9 +200,7 @@ where
             .as_ref()
             .and_then(|metadata| metadata.session_final)
             == Some(true);
-        let result = self
-            .execute_session(ctx, driver, request, session.as_deref())
-            .await;
+        let result = self.execute_session(ctx, driver, request).await;
```
ayushag-nv, line 209: why this change here ? request instead of session ?

ayushag-nv (approved):

nachiketb-nvidia (approved):

PR #300, opened 2026-08-05
fix(translation): report Responses max-token truncation as incomplete

ryan-lempka:
**Bug**
Currently we hardcode the status on the OpenAI Responses path to `completed`. Per the OpenAI Responses API spec, a response that stopped because it hit the max token limit should be `incomplete` with an `incomplete_details.reason` of `max_output_tokens`.

**Fix**
This PR detects that case and reports it on both the buffered and streaming paths. The buffered codec matches `StopReason::MaxTokens` off the decoded response. The streaming codec matches the raw upstream stop reason (`length` from OpenAI, `max_tokens` from Anthropic). In both cases the status is set to `incomplete`, `incomplete_details` is added, and mark the truncated message output item `incomplete`.

**Refs**
OpenAPI spec ref: https://github.com/openai/openai-openapi/blob/dc708bbe9a149bc35132c567ef3a3fdd7a24ab49/openapi.yaml#L56902-L56905

### DIFF
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -272,6 +272,12 @@ impl FormatCodec for OpenAiResponsesCodec {
                 .and_then(|output| output.stop_reason),
             Some(StopReason::MaxTokens)
         );
+        let status = if is_truncated {
+            "incomplete"
+        } else {
+            "completed"
+        };
+        let incomplete_details = is_truncated.then(|| json!({ "reason": "max_output_tokens" }));
         Ok(EncodedResponse {
             body: embed_preservation(
                 json!({
@@ -279,8 +285,8 @@ impl FormatCodec for OpenAiResponsesCodec {
                     "object": "response",
                     "created_at": 0,
                     "model": response.model.clone().unwrap_or_else(|| "unknown".to_string()),
-                    "status": if is_truncated { "incomplete" } else { "completed" },
-                    "incomplete_details": is_truncated.then(|| json!({ "reason": "max_output_tokens" })),
+                    "status": status,
+                    "incomplete_details": incomplete_details,
                     "output": encode_responses_output(&response.outputs),
                     "usage": encode_responses_usage(&response.usage),
                     "parallel_tool_calls": true,
@@ -1110,6 +1116,11 @@ fn encode_responses_output(outputs: &[ResponseOutput]) -> Value {
                     .any(|block| matches!(block, ContentBlock::ToolCall(_)));
                 let text = text_from_blocks(&output.content, "");
                 let reasoning = reasoning_text_from_blocks(&output.content, "\n");
+                let status = if matches!(output.stop_reason, Some(StopReason::MaxTokens)) {
+                    "incomplete"
+                } else {
+                    "completed"
+                };
                 let mut items = Vec::new();
 
                 if !reasoning.is_empty() {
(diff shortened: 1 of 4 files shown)


crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -256,14 +256,21 @@ impl FormatCodec for OpenAiResponsesCodec {
...
         Ok(EncodedResponse {
             body: embed_preservation(
                 json!({
                     "id": response.id.clone().unwrap_or_else(|| "resp_switchyard".to_string()),
                     "object": "response",
                     "created_at": 0,
                     "model": response.model.clone().unwrap_or_else(|| "unknown".to_string()),
-                    "status": "completed",
+                    "status": if is_truncated { "incomplete" } else { "completed" },
+                    "incomplete_details": is_truncated.then(|| json!({ "reason": "max_output_tokens" })),
```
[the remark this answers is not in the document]
ryan-lempka: You are correct - if the provider speaks Responses `StopReason::MaxTokens` would not be there to find. The original bug was reported against provider using Chat Completions.

The response is now parsed and checked for the truncation signal - and we set `stop_reason` to `StopReason::MaxTokens` based on `incomplete_details.reason` on the response body.

crates/switchyard-translation/src/codecs/responses/stream.rs
```diff
@@ -284,11 +297,12 @@ fn finish_responses_stream(state: &mut StreamTranslationState) -> Vec<Value> {
         .collect::<Vec<_>>();
 
     out.push(json!({
-        "type": "response.completed",
+        "type": if is_truncated { "response.incomplete" } else { "response.completed" },
```
nachiketb-nvidia: nit: can we pull this out

crates/switchyard-translation/src/codecs/responses/stream.rs
```diff
@@ -284,11 +297,12 @@ fn finish_responses_stream(state: &mut StreamTranslationState) -> Vec<Value> {
...
         "response": {
             "id": responses_id(state),
             "object": "response",
-            "status": "completed",
+            "status": if is_truncated { "incomplete" } else { "completed" },
```
nachiketb-nvidia: nit: same thing

crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -256,14 +266,21 @@ impl FormatCodec for OpenAiResponsesCodec {
...
+        );
         Ok(EncodedResponse {
             body: embed_preservation(
                 json!({
                     "id": response.id.clone().unwrap_or_else(|| "resp_switchyard".to_string()),
                     "object": "response",
                     "created_at": 0,
                     "model": response.model.clone().unwrap_or_else(|| "unknown".to_string()),
-                    "status": "completed",
+                    "status": if is_truncated { "incomplete" } else { "completed" },
```
nachiketb-nvidia: nit: here too

nachiketb-nvidia (approved):

crates/switchyard-translation/src/codecs/responses/stream.rs
ryan-lempka: Done

crates/switchyard-translation/src/codecs/responses/buffered.rs
ryan-lempka: Done

crates/switchyard-translation/src/codecs/responses/stream.rs
ryan-lempka: Done

PR #304, opened 2026-08-05
fix(server): return Codex metadata from /v1/models

elyasmnvidian:
```json
{
  "object": "list",
  "data": [{"id": "switchyard", "object": "model"}],
  "model_pool": ["switchyard"]
}
```

## What happens

Codex direct-provider discovery decodes this response as a Codex model catalog, not only as an OpenAI model list.

```text
failed to decode models response: missing field `models`
Unknown model switchyard is used. This will use fallback model metadata.
```

## Why

`GET /v1/models` currently returns only the OpenAI-compatible `data` array. Codex expects a top-level `models` array whose entries include its model metadata fields, so it discards the route's declared context window and tool support.

The endpoint now returns both arrays from the same route registry. Existing `data`, `model_pool`, and default-model fields remain unchanged. Each Codex entry reflects the route's declared context window, tool support, and reasoning, which the route registry owns and feeds to both arrays. The constant Codex card shape is a static mirror of Codex's `ModelInfo`; the launcher builds the same card in `switchyard/cli/launchers/codex_model_catalog.py`, and the two are kept in step by hand.

This is an additive response change. Existing OpenAI-compatible model discovery keeps the same fields and values.

## Proof

The same local route now returns both catalogs:

```text
keys: data, default_model, first_id, has_more, last_id, model_pool, models, object
models[0].slug: switchyard
models[0].context_window: 128000
models[0].shell_type: shell_command
```

The installed Codex CLI then completes the direct-provider request without the decode or fallback warning:

```text
codex
pong
```

The proof used a local OpenAI-compatible upstream; it did not contact a model provider.

## How tested

```text
cargo test -p switchyard-server --test server models_endpoint_reports_declared_route_capabilities_and_null_when_undeclared
```

The regression test checks the OpenAI and Codex catalogs together, including declared, restricted, and undeclared route capabilities.

## Summary by CodeRabbit

* **New Features**
  * Added Codex-compatible model metadata to the models endpoint.
  * Model entries now include context-window details, shell support, patch tool information, priority, and reasoning capabilities.
* **Bug Fixes**
  * Improved model metadata coverage for declared, restricted, and undeclared routes.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -1045,6 +1045,11 @@ fn model_list_payload<'a>(
     json!({
         "object": "list",
         "data": entries.iter().map(|(model, caps)| model_entry_json(model, *caps)).collect::<Vec<_>>(),
+        "models": entries
+            .iter()
+            .enumerate()
+            .map(|(priority, (model, caps))| codex_model_entry_json(model, *caps, priority))
+            .collect::<Vec<_>>(),
         "first_id": first_id,
         "last_id": last_id,
         "has_more": false,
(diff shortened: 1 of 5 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -999,6 +1009,64 @@ fn model_entry_json(model: &str, capabilities: ModelCapabilities) -> Value {
...
+
+// The reasoning-effort presets Codex offers for a reasoning-capable route. Kept in
+// step with the launcher template in `codex_model_catalog.py`.
+fn reasoning_levels() -> Value {
+    json!([
+        {"effort": "low", "description": "Fast responses with lighter reasoning"},
+        {"effort": "medium", "description": "Balances speed and reasoning depth"},
+        {"effort": "high", "description": "Greater reasoning depth"},
+        {"effort": "xhigh", "description": "Extra high reasoning depth"},
+    ])
```
grahamking, line 1155: There's a lot of assumptions in here for reasoning levels (they are all over the place in real models), and in `codex_model_entry_json`.

Is there a way (likely later) we can load this from the backend?We're a proxy after all, we should re-publish what the backend has.

grahamking (approved):

crates/switchyard-server/src/lib.rs
elyasmnvidian: Good call — I added a `TODO` in `codex_model_entry_json` for this (81f2d135). Only `context_window`, `tool_calling`, and `reasoning` are model facts a backend can publish, so those should come from the backend, with the route's config value as the fallback. The rest of the card — `shell_type`, `apply_patch_tool_type`, `base_instructions`, the reasoning-level presets — are Codex client conventions that no backend returns, so they stay constant.

It only works where the backend publishes the data. OpenRouter's `/api/v1/models` returns `context_length` and `supported_parameters`, so the packaged OpenRouter deployment can load all three. The NVIDIA gateway returns id-only models and blocks `/model/info` for our key, so config stays the fallback there. I'll file the follow-up.

PR #307, opened 2026-08-05
feat(libsy): expose algorithm metrics through driver

nachiketb-nvidia:
## What

- Add typed `Driver::record_counter` and `Driver::record_histogram` APIs for algorithm-defined metrics.
- Emit each metric once to OpenTelemetry and as `RunObservation::AlgorithmMetric` for host-side aggregation.
- Move the existing classifier fail-open counter onto the Driver path.

## Why

Algorithms need to publish domain-specific measurements without receiving or depending on a server accumulator. This keeps algorithms server-independent while allowing both exported telemetry and local consumers to derive the same rollups.

Implements [SWITCH-1227](https://linear.app/nvidia/issue/SWITCH-1227/expose-per-algorithm-counters-and-histograms-through-driver).

## How

The Driver captures the algorithm name once per run, adds it to every custom metric, records the metric through the global OpenTelemetry provider, and forwards the same owned event to the existing `RunObserver`. Metric names and attribute keys are static; attribute values must remain bounded and low-cardinality. Non-finite histogram samples are dropped from both outputs.

The Rust server intentionally ignores the new observation variant for now. Endpoint integration remains separate work.

## Aggregate proof

One comprehensive integration test emits two batches through the same observer and reconstructs the accumulator-style summary from raw events:

- counter total: `5`
- histogram: count `4`, total `100`, min `10`, max `40`, average `25`, p50 `30`, p99 `40`
- OpenTelemetry independently reports the same counter total and histogram count/sum
- non-finite samples appear in neither output

This demonstrates that a host can derive the Python accumulator rollups without coupling algorithms to accumulator storage.

## How tested

- [x] No live network calls

## Checklist

- [x] New public Rust symbols are exported from `switchyard-libsy`
- [x] One comprehensive integration test covers the new contract
- [x] Commit is DCO signed off

## Notes for reviewers

Please focus on the typed metric API, the guarantee that OpenTelemetry and `RunObserver` see the same event, and the intentional absence of server accumulator integration.

## Summary by CodeRabbit

- **New Features**
  - Added support for algorithm-defined counters and histograms.
  - Algorithm metrics now include bounded attributes and are available through observers and OpenTelemetry.
  - Added public metric types and streaming support for integrations.

- **Bug Fixes**
  - Improved fail-open metric reporting for judge transport, response, and parsing failures.
  - Invalid histogram samples no longer affect recorded aggregates.

### DIFF
--- a/crates/libsy/src/algorithms/util/stage.rs
+++ b/crates/libsy/src/algorithms/util/stage.rs
@@ -23,7 +23,7 @@ use serde::Deserialize;
 use super::prompts;
 use super::tool_signals::ToolSignals;
 use crate::Result;
-use crate::core::algorithm::Driver;
+use crate::core::algorithm::{Driver, MetricAttribute};
 use crate::core::classifier::{Classification, Classifier, Score};
 use crate::core::state::{State, StateValue};
 use switchyard_protocol::Request;
@@ -43,6 +43,9 @@ const HARD_SEVERITY: f64 = 0.7;
 const SIGNAL_UNIT: f64 = 0.10;
 /// Critical severity forces the capable tier regardless of the scorer.
 const SEVERITY_CRITICAL: f32 = 1.0;
+const SCORE_METRIC: &str = "switchyard.stage_router.score";
+const CONFIDENCE_METRIC: &str = "switchyard.stage_router.confidence";
+const DIMENSION_METRIC: &str = "switchyard.stage_router.dimension";
 
 /// The two tiers a turn can route to.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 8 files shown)


crates/libsy/src/core/algorithm.rs
```diff
@@ -185,6 +230,63 @@ impl Driver {
...
+        name: &'static str,
+        value: AlgorithmMetricValue,
+        attributes: Vec<MetricAttribute>,
+    ) {
+        let observation = AlgorithmMetricObservation {
+            algorithm: self.algorithm.to_string(),
+            name,
+            value,
+            attributes,
+        };
```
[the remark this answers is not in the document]
nachiketb-nvidia: fixed

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -340,6 +343,37 @@ pub fn pick_tier(signal: &ToolSignals, mode: PickerMode, confidence_threshold: f
...
+    driver.record_histogram(CONFIDENCE_METRIC, confidence, []);
+
+    let dimensions = dimensions_from_signal(signal);
+    for (name, value) in [
+        ("severity", dimensions.severity),
+        ("spinning", dimensions.spinning),
+        ("exploring", dimensions.exploring),
+        ("production_intensity", dimensions.production_intensity),
+    ] {
+        driver.record_histogram(
```
sabhatinas, line 369: This is great! We can stick to reporting raw scores. can you update the routing algorithms page for stage router where Im talking about decision sources ?

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -340,6 +343,37 @@ pub fn pick_tier(signal: &ToolSignals, mode: PickerMode, confidence_threshold: f
...
+        } => (*score, confidence.unwrap_or(score.abs())),
+        PickOutcome::ConsultClassifier {
+            score, confidence, ..
+        } => (*score, *confidence),
+    };
+    driver.record_histogram(SCORE_METRIC, score, []);
+    driver.record_histogram(CONFIDENCE_METRIC, confidence, []);
+
+    let dimensions = dimensions_from_signal(signal);
+    for (name, value) in [
```
sabhatinas, line 363: can we generalize this a bit? if we add a new signal, Ill have to manually update the strings here.

nachiketb-nvidia: that's probably best practice since that would be a meaningful update, can probably pull this up to a const variable would that be preferable?

nachiketb-nvidia:
Superseded by #342, which keeps metrics owned by the stage router and emits them directly through OpenTelemetry without adding Driver or RunObservation metric APIs.

Issue #301, opened 2026-08-05
Session affinity keys on an empty session id, and is silent when it cannot key at all


joelagnel:
## Scope clarification

The observed 9-judge-call TB-Lite failure was specifically **Harbor running Hermes** with `session_affinity = true` and `message_hash_fallback` omitted/false. Hermes did not send a native session header. Harbor did generate a stable UUID per trial attempt, but sent it as `proxy_x_session_id`, which Switchyard consumed for routing statistics but not affinity.

If `message_hash_fallback = true`, affinity was already best-effort functional by hashing the first user message. PR #308 therefore adds a fourth fix that recognizes Harbor's exact per-attempt ID: it makes Harbor/Hermes sticky without the fallback and supersedes the heuristic when the fallback is enabled. The empty-ID safeguard and diagnostic changes are general rather than Harbor/Hermes-specific.

Correction to section 1 below: current `main` already trims empty bare HTTP header values after #269. The empty-ID libsy safeguard still protects structured metadata containing an empty nested session value and direct callers that construct `Metadata { session_id: Some("") }`.

## Summary

Session affinity has two failure modes that are silent: it can key on an identity that
isn't one, and it can key on nothing at all. Both surface as a route that reports itself
as configured while behaving as if affinity were off -- or, in the first case, worse than
off. We hit the second while benchmarking TB-Lite and lost a while to it, then found the
first by reading the surrounding code.

## 1. An empty session id is treated as a real identity

`AffinityRouter::affinity_key` accepts whatever `metadata.session_id` holds:

```rust
// crates/libsy/src/algorithms/util/affinity.rs
if let Some(metadata) = request.metadata.as_ref()
    && let Some(session) = metadata.session_id.clone()
{
    ...
    return Some(AffinityKey::Session(session));
}
```

Header parsing does not normalise empty values — `resolve_path` in
`crates/protocol/src/metadata.rs` returns `Some(raw.clone())` verbatim — so a harness that
sends `x-switchyard-session-id:` with an empty value yields `Some("")`. Every request in
every task then shares `AffinityKey::Session("")`, the first task's model latches, and the
whole run inherits it. The message-hash fallback cannot save this: it lives in the `else`
branch that is now unreachable.

This is cross-task contamination that produces plausible-looking numbers, not an error.

The sibling consumer of the same field already guards it, which is what makes this look
like an oversight rather than intent:

```rust
// crates/libsy/src/algorithms/fall_through.rs
[...]

PR #308, opened 2026-08-05
feat(libsy): warn when affinity cannot derive a key

joelagnel:
## What

- Emit a one-time warning when session affinity cannot derive a usable request identity.
- Keep intentional `subagents_only` root abstention silent.
- Preserve existing message-hash fallback and child-identity behavior.

## Why

`session_affinity = true` can be silently inert when a request supplies no supported session metadata and `message_hash_fallback` is disabled. Every turn is then classified again even though affinity is configured.

Benchmark clients without session metadata can enable `message_hash_fallback` as a best-effort task key. Production clients should send a generic supported session header such as `x-switchyard-session-id`.

Addresses the warning portion of #301.

## How tested

- [x] `uv run ruff check .`
- [x] `uv run pytest tests/`

## Scope

This PR does not recognize benchmark-private proxy headers, change server request-log formatting, or change configuration/public APIs.

### DIFF
--- a/crates/libsy/src/algorithms/util/affinity.rs
+++ b/crates/libsy/src/algorithms/util/affinity.rs
@@ -19,6 +19,7 @@
 
 use std::collections::{HashMap, HashSet, hash_map::DefaultHasher};
 use std::hash::{Hash, Hasher};
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use async_trait::async_trait;
 use parking_lot::Mutex;
@@ -55,6 +56,8 @@ pub struct AffinityRouter {
     /// Held on the instance so the two roles share one process-local map through a
     /// single registered [`Arc`](std::sync::Arc); bounded by [`MAX_ASSIGNMENTS`].
     assignments: Mutex<HashMap<RoutingIdentity, String>>,
+    /// Whether the "no identity to key on" warning has already been emitted.
+    unkeyed_warning_emitted: AtomicBool,
 }
 
 impl AffinityRouter {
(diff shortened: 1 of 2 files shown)


crates/protocol/src/metadata.rs
```diff
@@ -57,6 +57,9 @@ const CLAUDE_PARENT_AGENT_ID_HEADER: &str = "x-claude-code-parent-agent-id";
 // OpenCode session header — used for session_id correlation only (not a routing signal).
 const OPENCODE_SESSION_ID_HEADER: &str = "x-session-id";
 
+// Harbor benchmark proxy session header.
+const HARBOR_PROXY_SESSION_ID_HEADER: &str = "proxy_x_session_id";
```
ayushag-nv: we dont need internal proxy headers here. Please work with message_hash_fallback for the identifier

crates/libsy/src/algorithms/util/affinity.rs
```diff
@@ -95,26 +98,43 @@ impl AffinityRouter {
...
                 })
-            })
-            .flatten()
+            }
+            (None, false) => None,
+        };
+        // Affinity that never keys anything is silent otherwise: the route reports itself as
+        // configured while every turn is classified afresh. Say so once rather than per turn.
+        if key.is_none() && self.should_warn_unkeyed() {
+            tracing::warn!(
```
ayushag-nv, line 124: This warning will not show up with the server default log filter. Without an explicit target, tracing uses `switchyard_libsy::algorithms::util::affinity`, while the server enables `libsy=info`. I reproduced this with the exact filter: the implicit-target event was dropped, while `target: "libsy"` was emitted. Please add that target here and cover the emitted event rather than only the atomic flag.

crates/libsy/src/algorithms/util/affinity.rs
```diff
@@ -95,26 +98,43 @@ impl AffinityRouter {
...
+                is_subagent,
+                message_hash_fallback = self.message_hash_fallback,
+                "affinity is enabled but this request carries no usable identity, so no \
```
ayushag-nv, line 128: This advice is misleading when `message_hash_fallback` is already enabled but the first user message has no text, for example an image-only request. `first_user_message_hash` still returns `None`, yet this tells the operator to enable the fallback. Could we say that the fallback also requires usable first-user text?

ayushag-nv (approved):

PR #309, opened 2026-08-05
fix(libsy): keep classifier history from orphaning a tool result

eric-liu-nvidia:
## Issue

`LlmClassifier` trims conversation history by message count, so the window can begin on a tool result and leave out the assistant tool call that introduced its id. The classifier subrequest then carries an orphan tool result, which violates the tool-call protocol and can be rejected by the classifier backend or scored on structurally incomplete context.

## Solution

The window start now moves back to the nearest index that keeps every tool call and its result together. If no earlier start can pair a result, the counted start is kept, so an unanswerable result cannot widen the window to the whole conversation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Conversation history trimming now preserves complete tool-call and tool-result pairs when preparing input for classification.
  * Plain conversation history continues to use the requested trailing window without changes.

### DIFF
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -3,7 +3,7 @@
 
 //! Judge-backed capability, escalation, and custom-policy routing.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -89,10 +89,41 @@ fn trim_messages(messages: &[Message], recent_turn_window: usize) -> Vec<Message
...
+/// result answering it, leaving the judge a result whose call id was never introduced.
+/// The start therefore moves back to the nearest one that keeps every tool pair whole.
+/// A result that no earlier start can pair — its call sits before the opening task, which
+/// this function never reaches — falls back to the counted start, so an unanswerable
+/// result cannot widen the window to the entire conversation.
+fn window_start(tail: &[&Message], recent_turn_window: usize) -> usize {
+    let counted = tail.len().saturating_sub(recent_turn_window);
+    (0..=counted)
+        .rev()
+        .find(|start| !has_orphan_tool_result(&tail[*start..]))
```
elyasmnvidian:
```text
tail = [Text, ..., Text, ToolResult("missing")]
recent_turn_window = 1
```

With no matching call, `.find` checks every earlier start. The call-set pass and result pass examine `2 * (1 + ... + n)` blocks and allocate `n` sets for a tail of `n` messages before returning `counted`. This work runs before the classifier calls its judge and grows quadratically with the request. Please scan once from newest to oldest, track unmatched result IDs in a `HashSet`, and return `counted` if the set never empties.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -89,10 +89,41 @@ fn trim_messages(messages: &[Message], recent_turn_window: usize) -> Vec<Message
...
+fn has_orphan_tool_result(window: &[&Message]) -> bool {
+    let blocks = || window.iter().flat_map(|message| &message.content);
+    let calls: BTreeSet<&str> = blocks()
+        .filter_map(|block| match block {
+            ContentBlock::ToolCall(call) => Some(call.id.as_str()),
+            _ => None,
+        })
+        .collect();
+    blocks().any(|block| match block {
+        ContentBlock::ToolResult(result) => !calls.contains(result.tool_call_id.as_str()),
```
elyasmnvidian:
```text
tail = [
    ToolCall("x"),
    ToolResult("x"),
    Text("later"),
    ToolCall("x"),
    ToolResult("x"),
]
recent_turn_window = 4
```

The counted window starts at the first result. Because `calls` also contains the later `ToolCall("x")`, this check returns `false` and sends:

```text
[ToolResult("x"), Text("later"), ToolCall("x"), ToolResult("x")]
```

The first result still has no preceding call. Switchyard does not reject repeated IDs before this helper. Track unmatched result IDs while scanning backward so only an earlier call can satisfy a result, and add this sequence as a regression test.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -1444,6 +1476,99 @@ mod tests {
...
+            Message::text(Role::Assistant, "recent 1"),
+            Message::text(Role::User, "recent 2"),
+            Message::text(Role::Assistant, "recent 3"),
+            Message::text(Role::User, "recent 4"),
+        ];
+
+        // The five-message tail begins exactly on the tool result.
+        let kept = trim_messages(&messages, 5);
+
+        assert_eq!(orphan_tool_results(&kept), Vec::<String>::new());
```
elyasmnvidian:
```rust
let kept = vec![
    Message::text(Role::System, "client instructions"),
    Message::text(Role::User, "initial task"),
];
assert_eq!(orphan_tool_results(&kept), Vec::<String>::new());
```

This assertion stays green when trimming drops both the call and result, and `orphan_tool_results` repeats the same order-insensitive set check as production. Assert the exact retained message sequence so the test requires both blocks in the right order. `trimming_without_tool_calls_keeps_only_the_window` repeats `a_window_widens_the_judge_to_the_surrounding_conversation`; replace it with the documented case where the call is before the opening user message and its result is after it.

elyasmnvidian (changes requested):
The current scan revisits a long classifier history for every candidate start, and it accepts a tool result when a matching call appears later. The regression test also stays green if trimming drops the pair. I ran `cargo fmt --all --check`, workspace clippy with warnings denied, and `cargo test -p switchyard-libsy`; all passed.

elyasmnvidian (approved):
please fix merge conflicts

PR #312, opened 2026-08-05
fix(libsy): put judge system prompt in instructions, not messages

linj-glitch:
PR #204 refactored `CapabilityJudge` to a `StructuredJudge` type alias, but `StructuredJudge::build_request` inserts the system prompt into `messages` as a `Role::System` entry instead of `LlmRequest.instructions`. This silently reverted the fix from PR #195 and removed its test.

For OpenAI-compatible backends the wire output happens to be identical, but for Anthropic backends a `Role::System` message in `messages` serializes as a user turn — the `system` field ends up empty. Fix `StructuredJudge::build_request` to use `instructions` (as `InstructionBlock`), and update the tests in `llm_class.rs` and `escalation.rs` that were asserting the old messages-based shape.

### DIFF
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -916,7 +916,8 @@ mod tests {
 
     use super::*;
     use switchyard_protocol::{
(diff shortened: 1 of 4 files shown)


ryan-lempka (approved):
LGTM

PR #313, opened 2026-08-05
fix: Cleanup FallThrough sessions

grahamking:
- Automatically when they finish (`session_final`).
- After one hour of inactivity, to catch those that never finish
  cleanly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session lifecycle management to remove completed and stale sessions automatically.
  - Preserved active or recently accessed sessions during cleanup.
  - Ensured session cleanup occurs even when execution encounters errors.
  - Removed session eviction data after completion to prevent lingering state.

- **Tests**
  - Added coverage for final-session cleanup, activity timestamp updates, and inactive-session removal.

### DIFF
--- a/crates/libsy/src/algorithms/fall_through.rs
+++ b/crates/libsy/src/algorithms/fall_through.rs
@@ -16,7 +16,11 @@
 //! Every composition retains one thing regardless: a target that overflows its context window is
 //! remembered for the rest of its session and skipped on later turns.
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Once, Weak},
+    time::{Duration, Instant},
+};
 
 use async_trait::async_trait;
 use parking_lot::Mutex;
@@ -28,7 +32,15 @@ use crate::core::processor::{Event, Processor};
 use crate::{LibsyError, Result};
 use switchyard_protocol::{Context, Decision, Request, Response, RoutedLlmClient};
 
-type SessionStates<S> = Mutex<HashMap<String, Arc<AsyncMutex<S>>>>;
+struct SessionState<S> {
+    state: Arc<AsyncMutex<S>>,
+    last_accessed: Instant,
+}
+
+type SessionStates<S> = Mutex<HashMap<String, SessionState<S>>>;
+
+const SESSION_STATE_TTL: Duration = Duration::from_secs(24 * 60 * 60);
+const SESSION_CLEANUP_INTERVAL: Duration = Duration::from_secs(60 * 60);
 
 /// The decision a fall-through run produces: the selected model plus a human-readable reason.
 pub struct FallThroughDecision {
@@ -104,7 +116,8 @@ pub struct FallThrough<S = ()> {
     processors: Vec<Arc<dyn Processor<S>>>,
     classifiers: Vec<Arc<dyn Classifier<S>>>,
     targets: LlmTargetSet,
-    session_states: Option<SessionStates<S>>,
+    session_states: Option<Arc<SessionStates<S>>>,
+    cleanup_started: Once,
     session_evictions: SessionEvictions,
 }
 
@@ -118,6 +131,7 @@ impl FallThrough<()> {
             classifiers: Vec::new(),
             targets,
             session_states: None,
+            cleanup_started: Once::new(),
             session_evictions: SessionEvictions::default(),
         }
     }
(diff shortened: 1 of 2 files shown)


grahamking:
@coderabbit review

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -28,7 +32,15 @@ use crate::core::processor::{Event, Processor};
...
-type SessionStates<S> = Mutex<HashMap<String, Arc<AsyncMutex<S>>>>;
+struct SessionState<S> {
+    state: Arc<AsyncMutex<S>>,
+    last_accessed: Instant,
+}
+
+type SessionStates<S> = Mutex<HashMap<String, SessionState<S>>>;
+
+const SESSION_STATE_TTL: Duration = Duration::from_secs(24 * 60 * 60);
+const SESSION_CLEANUP_INTERVAL: Duration = Duration::from_secs(60 * 60);
```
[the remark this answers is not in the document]
grahamking: Good catch. Fixed.

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -1178,9 +1267,97 @@ mod tests {
...
+            .expect("test timestamp is representable");
+        let states = router
+            .session_states
+            .as_ref()
+            .expect("stateful router has a session registry");
+        for session in states.lock().values_mut() {
+            session.last_accessed = stale;
+        }
+
+        remove_inactive_sessions(states, now, SESSION_STATE_TTL);
```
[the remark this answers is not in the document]
grahamking: Nice!

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -169,18 +190,46 @@ where
...
+        self.start_cleanup_task();
+        let session = session_id(&request);
+        let session_final = request
+            .metadata
+            .as_ref()
+            .and_then(|metadata| metadata.session_final)
+            == Some(true);
+        let result = self
+            .execute_session(ctx, driver, request, session.as_deref())
+            .await;
```
elyasmnvidian, line 203: `AbortOnDrop(handle.abort_handle())` aborts this future when the response stream is dropped, so `remove_session` does not run for a `session_final` request. A small type whose `Drop` implementation calls `remove_session` keeps cleanup on this cancellation case:

```rust
let _cleanup = session
    .as_deref()
    .filter(|_| session_final)
    .map(|session| FinalSessionCleanup::new(self, session));

self.execute_session(ctx, driver, request, session.as_deref()).await
```

crates/libsy/src/algorithms/fall_through.rs
```diff
@@ -1178,9 +1273,91 @@ mod tests {
...
+        let first_accessed = router
+            .session_states
+            .as_ref()
+            .expect("stateful router has a session registry")
+            .lock()
+            .get("session-1")
+            .expect("session state was inserted")
+            .last_accessed;
+
+        tokio::time::sleep(std::time::Duration::from_millis(1)).await;
```
elyasmnvidian: This test checks `Instant::now()` rather than cleanup behavior. Replace it with a recent idle session case in `cleanup_removes_only_inactive_idle_sessions`:

```rust
states
    .lock()
    .get_mut("recent")
    .expect("recent session exists")
    .last_accessed = now;

remove_inactive_sessions(states, now, SESSION_STATE_TTL);
assert!(states.lock().contains_key("recent"));
```

elyasmnvidian (commented):
A cancelled final request can keep its session data, and one test checks a timestamp rather than cleanup behavior. The suggested changes are inline.

nachiketb-nvidia (approved):

crates/libsy/src/algorithms/fall_through.rs
grahamking: There's an hourly cleanup job to catch those, and any others that don't finish cleanly.

grahamking: I do like the `Drop` suggestion. Later probably.

crates/libsy/src/algorithms/fall_through.rs
grahamking: Good point. `cleanup_removes_only_inactive_idle_sessions` already tests this case. Removed the test.

elyasmnvidian (approved):

PR #314, opened 2026-08-05
fix(overflow): add hub GLM phrase to context-overflow detector

linj-glitch:
The hub's GLM model returns "Input length N exceeds the maximum allowed input length of M tokens" with code "400" rather than "context_length_exceeded", so is_context_overflow returned false and the fallback in #302 was never consulted. Adds "exceeds the maximum allowed input length" to the OpenAI phrase list in both crates, with a test asserting the exact GLM error body shape.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of input-length limit errors from OpenAI-compatible providers.
  * Added support for recognizing overflow errors returned through Hub and LiteLLM-wrapped GLM responses, including non-standard error codes.
  * These errors are now handled consistently as context overflow conditions.

### DIFF
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -24,6 +24,7 @@ const OPENAI_OVERFLOW_PHRASES: &[&str] = &[
     "context window",
     "context length is only",
     "please reduce the length of the input",
+    "exceeds the maximum allowed input length",
 ];
 
 // Anthropic has no structured `error.code`, so detection is phrase-based only.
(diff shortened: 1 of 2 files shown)


linj-glitch:
**Why phrase matching and not error code matching**

A reviewer might ask why we're extending the phrase list rather than adding a structured code check, since the code path already short-circuits on `error.code == "context_length_exceeded"` before falling back to phrases.

The answer is that GLM (via the hub's LiteLLM proxy) doesn't emit a semantic error code. The actual response body is:

```json
{"error": {"message": "Input length 877338 exceeds the maximum allowed input length of 639968 tokens", "code": "400"}}
```

`code` here is just the HTTP status stringified, not an OpenAI-style error code. There's nothing structured to match on — phrase detection is the only hook this provider gives us.

The existing comment in the source already covers the safety argument: a false positive on phrase matching triggers one bounded evict-and-retry, not an infinite loop, so erring toward matching the message is the right trade-off.

eric-liu-nvidia (approved):

PR #322, opened 2026-08-06
fix(libsy): keep tier prompts and handoff notes on a same-format hop

sabhatinas:
Tier system prompts and handoff notes never reached the model on a same-format hop.

**Why.** A codec stores the verbatim inbound body at decode, the algorithm then mutates the canonical IR only, and at encode the codec replays that stored body whenever one exists for the target format — returning a snapshot taken before the addition, without ever reading the IR. All three buffered codecs do this, so Anthropic→Anthropic and Responses→Responses dropped it too. Cross-format usually misses the lookup, but not always: preservation carries bodies embedded by earlier hops, so a multi-hop request can hit the same short-circuit.

**Fix.** Drop the preserved bodies once something has been added, which sends the codec down its normal encode path. Two call sites and one helper. The whole map is cleared, not just the inbound entry — the addition is missing from every stored body equally.

**Verified** against the ticket's reproduction with a stub upstream: `<NO SYSTEM MESSAGE>` on `main`, the configured prompt on this branch. No new tests — the existing fixtures now seed a preserved body, so the tests already covering each mutation assert what happens to it, and fail on unpatched code.

Fixes SWITCH-1224

### DIFF
--- a/crates/libsy/src/algorithms/util/prompts.rs
+++ b/crates/libsy/src/algorithms/util/prompts.rs
@@ -14,6 +14,14 @@
 //!
 //! Which text, and when, is the caller's policy; this module only knows how to
 //! place it so the provider accepts it and the prompt cache survives.
+//!
+//! **Anything added here must call [`drop_exact_replay`].** Both shapes above
+//! mutate the normalized request, and a codec asked to encode for the format the
+//! request arrived in replays the body captured at decode instead of reading that
+//! request — so an addition that leaves exact replay in place never reaches the
+//! model. This is not enforced: a future processor that mutates the request and
+//! forgets the call reintroduces SWITCH-1224, silently and without a failing
+//! test.
 
 use std::collections::BTreeMap;
 
@@ -41,6 +49,25 @@ pub fn append_note(request: &mut Request, note: &str) {
             .messages
             .push(Message::text(Role::User, note)),
     }
+    drop_exact_replay(request);
+}
+
+/// Gives up exact same-format replay for this turn.
+///
+/// A codec replays the preserved inbound body verbatim when the target format
+/// matches the source, which is what keeps a same-format hop lossless. That body
+/// predates anything added here, so leaving it in place would encode the request
+/// as it arrived and silently drop the addition. Dropping it sends the codec down
+/// its normal path, which encodes from the request itself.
+///
+/// Every stored body goes, not just the one for the inbound format: preservation
+/// also carries bodies embedded by earlier hops, and the addition is missing from
+/// all of them equally.
+///
+/// Call this from any new code that mutates the request. Nothing checks that you
+/// have.
+fn drop_exact_replay(request: &mut Request) {
+    request.llm_request.preservation.requests.clear();
 }
 
 /// System prompts keyed by routing target. A target left unset is routed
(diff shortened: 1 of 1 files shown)


crates/protocol/src/llm.rs
```diff
@@ -326,6 +335,49 @@ pub struct LlmRequest {
...
+        self.preservation.request_seal = None;
+        self.preservation.request_seal = Some(self.shape_fingerprint());
+    }
+
+    /// Whether the preserved bodies still describe this request.
+    ///
+    /// Returns `false` once anything has changed since [`Self::seal_preservation`]
+    /// — a routing algorithm inserting a tier system prompt, appending a handoff
+    /// note, rewriting the model — which is what stops a same-format hop from
+    /// replaying a body that predates the change.
```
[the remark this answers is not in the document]
sabhatinas: Good catch — that was a real contradiction: the freshness contract listed a model rewrite as invalidating, while `shape_fingerprint` deliberately excludes `model` so that `send_encoded`'s rewrite-then-stamp path keeps exact replay. Corrected in 3787e0e: the list now covers only tier prompts and handoff notes, with the `model` exception called out and pointed at `shape_fingerprint` for the reason.

crates/switchyard-translation/tests/request_translation.rs
```diff
@@ -1490,3 +1491,110 @@ fn anthropic_thinking_is_dropped_from_responses_input() -> TestResult {
...
 }
+
+/// A same-format hop replays the exact inbound body, which is what makes it
+/// lossless — but only while the IR still matches that body. Once a router has
+/// added an instruction the body predates, replay would silently drop it, so
+/// every codec must fall back to encoding from normalized fields.
+///
+/// Regression test for SWITCH-1224.
+#[test]
+fn same_format_encoding_drops_exact_replay_once_the_ir_gains_an_instruction() -> TestResult {
```
nachiketb-nvidia: one regression test, or change in existing tests should be enough

crates/protocol/src/llm.rs
```diff
@@ -326,6 +335,53 @@ pub struct LlmRequest {
     pub preservation: PreservationMetadata,
 }
 
+impl LlmRequest {
+    /// Records the current shape of this request against its preserved bodies.
+    ///
+    /// Codecs call this once decoding is complete. Until it is called the
+    /// preserved bodies are treated as stale, so an un-sealed request always
+    /// re-encodes from normalized fields.
+    pub fn seal_preservation(&mut self) {
```
nachiketb-nvidia: the problem is likely that two things are not happening one after another or happening in the wrong order. Some small refactors, coupling functions should be fine. But there is probably a more AMAN ALAP solution to this

sabhatinas: Agreed, and the seal is gone.

crates/switchyard-translation/tests/request_translation.rs
sabhatinas: Done 

- `each_target_gets_its_own_prompt` — replay is given up once a tier prompt is added
- `a_note_leaves_the_rest_of_the_conversation_untouched` — same for the handoff note
- `an_unconfigured_target_is_left_untouched` — replay is kept when nothing was added, so the fix does not cost losslessness

nachiketb-nvidia (approved):

PR #334, opened 2026-08-08
fix(server): do not append [DONE] after an upstream in-band stream error

enwaiax:
## Problem

When an upstream emits an SSE error event mid-stream, `/v1/chat/completions`
forwards that error frame and then still appends `data: [DONE]` — the OpenAI Chat
success sentinel. An SDK client that stops at `[DONE]` treats the truncated,
failed turn as a normally completed answer.

Captured against the release binary (loopback upstream, one content frame followed
by one error frame):

```
data: {"choices":[{"delta":{"content":"partial",...}}],...}

data: {"error":{"code":"stream_failed","message":"upstream failed after stream start",...}}

data: [DONE]      <-- the defect
```

`/v1/messages` and `/v1/responses` are unaffected — neither uses a `[DONE]`
sentinel, and both already terminate on the error event.

## Root cause

`crates/switchyard-server/src/sse.rs` tracks a local `failed` flag, but only a
framing/JSON error or a transport-level stream error sets it. An upstream in-band
error event is a well-formed JSON event, so it flows through `frame_event()` as
ordinary data and leaves `failed == false`:

```rust
if !failed && target_format == WireFormat::OpenAiChat {
    yield Ok(Event::default().data("[DONE]"));   // runs even for a failed stream
}
```

The translation layer already knows the stream failed: `encode_stream()` checks
`state.errored` and returns early, deliberately skipping `codec.finish()`. That
outcome was simply never communicated to the serving layer, which cannot otherwise
tell a clean EOF from an early stop.

This also matches the contract already stated in
`crates/switchyard-translation/src/helpers.rs`: *"An in-band error is terminal for
every target format: the encoder emits the pre-error content and the error, then
drops any later chunk."*

## Fix

Expose the outcome across the layer boundary and consult it before emitting the
sentinel. `[DONE]` stays in the serving layer — framing is the serving layer's job
per the `RawEventStream` doc contract, and `[DONE]` is not a JSON event object, so
it does not belong in a codec.

- new `StreamOutcome` (a shared `AtomicBool`) set where `encode_stream` already
  returns early on `state.errored`
- new `encode_stream_with_outcome()`; **`encode_stream()` keeps its signature** and
  delegates to it, so existing callers (e.g. `libsy-llm-client`) are untouched
- `frame_stream()` skips the sentinel when the outcome reports an in-band error

## Tests

Two new unit tests in `sse.rs`:

- `upstream_in_band_error_suppresses_the_done_marker` — the reported defect
- `clean_stream_still_emits_the_done_marker` — guards the OpenAI Chat contract for
  the success path

The pre-existing `stream_error_terminates_without_done_marker` still passes.

## Verification

```
cargo fmt --check                                       -> clean
cargo clippy --workspace --all-targets -- -D warnings   -> 0 warnings
cargo test --workspace                                  -> 21 test binaries, 0 failed
```

End-to-end A/B against the same loopback upstream, same test, only the binary
swapped:

```
patched binary  -> saw_done=False   (no [DONE] after the error frame)
v0.2.0-rc2      -> saw_done=True    (defect reproduces)
```

## Note

The Python `switchyard serve` entry point has the same defect
(`switchyard/lib/endpoints/sse_helpers.py` yields `[DONE]` unless an exception is
raised, and an upstream error event is not an exception). It has no equivalent
`errored` signal to read, so it needs its own fix rather than a port of this one —
not included here to keep this PR to a single component.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming error handling so failed in-band responses no longer include a misleading completion marker.
  * Preserved correct completion markers for successfully completed streams.
  * Continued suppressing completion markers when stream framing encounters an error.

### DIFF
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -75,6 +75,16 @@ pub type RawEventStream = Pin<
     >,
 >;
 
+/// Shared flag reporting whether a [`RawEventStream`] ended on an in-band error.
+///
+/// An in-band error is terminal: the encoder emits the error event and stops
+/// without calling `finish`, so no terminal event follows. The serving layer
+/// cannot tell that apart from a clean end-of-stream, yet it must not append a
+/// success sentinel (OpenAI Chat's `[DONE]`) to a failed stream. This flag
+/// carries that outcome across the layer boundary; `false` until the encoder
+/// observes an error.
+pub type StreamOutcome = std::sync::Arc<std::sync::atomic::AtomicBool>;
+
 /// Encodes a stream of IR chunks into a stream of target-format wire events.
 ///
 /// `served_model` is exposed as the response model (via the stream state's
@@ -86,6 +96,16 @@ pub fn encode_stream(
     target: WireFormat,
     served_model: Option<String>,
 ) -> std::result::Result<RawEventStream, LlmClientError> {
+    encode_stream_with_outcome(chunks, target, served_model).map(|(stream, _)| stream)
+}
+
+/// Same as [`encode_stream`], plus a [`StreamOutcome`] the caller can read once
+/// the stream ends to tell a clean finish from an in-band error termination.
+pub fn encode_stream_with_outcome(
+    chunks: LlmResponseStream,
+    target: WireFormat,
+    served_model: Option<String>,
+) -> std::result::Result<(RawEventStream, StreamOutcome), LlmClientError> {
     let target_format: FormatId = target.into();
     // The target is always a built-in wire format, so this lookup cannot fail; a
     // failure returns as an `Err` rather than a panic.
@@ -103,6 +123,8 @@ pub fn encode_stream(
         ..Default::default()
     };
     let mut chunks = chunks;
+    let outcome: StreamOutcome = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false));
+    let outcome_for_stream = std::sync::Arc::clone(&outcome);
 
     let events = try_stream! {
         while let Some(item) = chunks.next().await {
@@ -118,6 +140,9 @@ pub fn encode_stream(
                 yield value;
             }
             if state.errored {
+                // Terminal: report the outcome so the serving layer skips any
+                // success sentinel, and stop without calling `finish`.
+                outcome_for_stream.store(true, std::sync::atomic::Ordering::Release);
                 return;
             }
         }
@@ -131,7 +156,7 @@ pub fn encode_stream(
         }
     };
 
-    Ok(Box::pin(events))
+    Ok((Box::pin(events), outcome))
 }
 
 // The raw-response helper promises that the caller sees the model that served the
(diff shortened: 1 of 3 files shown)


nachiketb-nvidia (approved):
Thanks for the catch, seems a rebase off main is in order.

LGTM otherwise!

nachiketb-nvidia (changes requested):
Retracting my approval for now while this still needs follow-up before merge.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -75,6 +75,16 @@ pub type RawEventStream = Pin<
...
 
+/// Shared flag reporting whether a [`RawEventStream`] ended on an in-band error.
+///
+/// An in-band error is terminal: the encoder emits the error event and stops
+/// without calling `finish`, so no terminal event follows. The serving layer
+/// cannot tell that apart from a clean end-of-stream, yet it must not append a
+/// success sentinel (OpenAI Chat's `[DONE]`) to a failed stream. This flag
+/// carries that outcome across the layer boundary; `false` until the encoder
+/// observes an error.
+pub type StreamOutcome = std::sync::Arc<std::sync::atomic::AtomicBool>;
```
nachiketb-nvidia, line 86: think we can solve this more locally without adding `StreamOutcome`.

  The `Arc<AtomicBool>` works, but it feels like the wrong abstraction here. The server is the layer appending `[DONE]`, so the server should also own the decision to suppress it. In this case the signal is already present in the stream item: an OpenAI Chat error frame like `{"error": ...}` is terminal and
  should not be followed by `[DONE]`.

  Could we keep this entirely inside `sse.rs` with a small private enum instead?

  ```rust
  enum FrameOutcome {
      Continue(Event),
      TerminalError(Event),
  }

  fn frame_event(value: Value, target_format: WireFormat) -> Result<FrameOutcome, BoxError> {
      let is_terminal_error =
          target_format == WireFormat::OpenAiChat && value.get("error").is_some();

      let event = Event::default().data(serde_json::to_string(&value)?);

      Ok(if is_terminal_error {
          FrameOutcome::TerminalError(event)
      } else {
          FrameOutcome::Continue(event)
      })
  }

  Then frame_stream can keep local state:

  let mut failed = false;

  while let Some(item) = stream.next().await {
      match item {
          Ok(value) => match frame_event(value, target_format) {
              Ok(FrameOutcome::Continue(event)) => yield Ok(event),
              Ok(FrameOutcome::TerminalError(event)) => {
                  failed = true;
                  yield Ok(event);
                  break;
              }
              Err(error) => {
                  failed = true;
                  yield Ok(error_event(error));
                  break;
              }
          },
          Err(error) => {
              failed = true;
              yield Ok(error_event(error));
              break;
          }
      }
  }

  if !failed && target_format == WireFormat::OpenAiChat {
      yield Ok(Event::default().data("[DONE]"));
  }
```

  That keeps the fix targeted:

  - no translation API change
  - no public StreamOutcome
  - no Arc<AtomicBool> / memory ordering discussion for a single stream loop
  - the [DONE] decision stays next to the code that emits [DONE]

  Longer term, if we want stream termination to be a first-class contract, I’d rather see a typed translated stream item with an explicit terminal status. But for this PR, keeping it private and local to SSE framing seems cleaner.

nachiketb-nvidia (changes requested):

bhuvan002:

Issue #310, opened 2026-08-05
Proposal: libsy refactor


grahamking:
READ THE NEW VERSION: https://github.com/NVIDIA-NeMo/Switchyard/issues/310#issuecomment-5246443798

---

# A simpler core for libsy

*Switchyard · crates/libsy*

`run_stream` only. Three kinds of `Step`. No clients, no timeouts, no type erasure. About 1,000 fewer lines.

---

## 1. The idea

`run_stream` becomes the only way to use libsy. You get a stream of steps. You answer the ones that need answering. You don't have to respond.

Changes are mostly described in the example doc comments.

---

## 2. The three steps

```rust
/// One item on an algorithm's step stream.
pub enum Step {
    /// Please make this model call, then `respond` with the result.
    CallLlm(Box<CallLlm>),
    /// A routing choice, published as it is made.
    Decision(Decision),
    /// The final answer. libsy sends this when the algorithm returns.
    Done(Box<Response>),
}
```

A decision is a plain struct. There is no trait and nothing to downcast.

```rust
#[derive(Clone, Debug)]
pub struct Decision {
    /// The model to call.
    pub selected_model: String,
    /// Why, for logs and traces.
    pub reasoning: Option<String>,
    /// True for the call that answers the request. False for the
    /// classifier and judge calls made along the way.
    pub is_answer_call: bool,
}
```

A call the `libsy` user should make. The fields are public.

```rust
pub struct CallLlm {
    pub ctx:      Context,
    pub request:  Request,
    pub decision: Decision,   // decision.selected_model is the model_id this should go to.
    reply: oneshot::Sender<Result<Response>>,
}

impl CallLlm {
    /// Give the algorithm your result.
    pub fn respond(self, result: Result<Response>);

    /// Or take the contents and don't answer at all - aka "decision-only".
    pub fn into_parts(self) -> (Context, Request, Decision);
}
```

---

## 3. Writing an algorithm

An algorithm gets a `Driver` to work with. The driver has two methods.

```rust
impl Driver {
    /// Ask for a model call and wait for the answer.
    pub async fn call(&self, request: Request, decision: Decision) -> Result<Response>;

    /// Publish a routing choice.
    pub async fn decide(&self, decision: Decision) -> Result<()>;
}
```

Algorithm implementers must implement this trait which has two methods.

```rust
#[async_trait]
pub trait Algorithm: Send + Sync + 'static {
    fn name(&self) -> &str;

[...]

PR #338, opened 2026-08-10
refactor(libsy): make Decision a struct

ayushag-nv:
## Why

The `Decision` trait makes algorithms define separate concrete decision types even though callers need the same routing data from each one. [SWITCH-1240](https://linear.app/nvidia/issue/SWITCH-1240/make-decision-a-struct-instead-of-a-trait) replaces that type-erased contract with one explicit representation.

## What

- Replace the `Decision` trait with a cloneable struct containing private `selected_model_id`, `reasoning`, and `is_answer_call` fields.
- Construct decisions through `Decision::new(...)` and expose their values through read-only accessors.
- Update libsy algorithms, the LLM client, server integration, and Python bindings to use `Arc<Decision>`.
- Remove algorithm-specific decision types and exports, including `SimpleDecision`, `NoopDecision`, `PassthroughDecision`, and `RandomDecision`.
- Store the semantic name of each target in `selected_model_id`, and preserve routing-tier and fallback details in `reasoning`.
- Mark classifier and judge calls as non-answer calls while routed generation calls are answer calls.

This is an intentional breaking Rust API change: implementations now construct the shared `Decision` type, and the removed decision types and trait methods have no compatibility aliases.

## How

Algorithms create the shared type with `Decision::new(...)`. Consumers use `selected_model_id()`, `reasoning()`, and `is_answer_call()` instead of accessing fields directly, dispatching through a trait, or downcasting. `selected_model_id` identifies the configured target by its semantic name; provider-model resolution remains owned by the routed client.

This PR deliberately retains `Arc<Decision>` and the current driver shape. Moving decisions by value belongs to the broader [libsy refactor in #310](https://github.com/NVIDIA-NeMo/Switchyard/issues/310).

## Where to Start Review

1. `crates/protocol/src/client.rs` for the concrete type, constructor, and accessors.
2. `crates/libsy/src/algorithms/fall_through.rs` for target selection and reasoning preservation.
3. `crates/libsy/src/core/algorithm.rs` and `crates/libsy-llm-client/src/run.rs` for driver and call handling.
4. `crates/switchyard-server/src/lib.rs` and `crates/switchyard-py/src/libsy_bindings.rs` for downstream behavior.

## Test Plan

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test -p switchyard-protocol -p switchyard-libsy -p switchyard-llm-client -p switchyard-server`
- [x] `uv run pytest tests/test_libsy_minimal_bindings.py -q`
- [x] `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Routing decisions now expose the selected target, optional reasoning, and whether a call produced the final answer.
  - Python bindings and algorithm outputs provide the updated decision metadata.
  - Fallback explanations are included in decision reasoning and relevant response headers.

- **Observability**
  - Metrics and traces distinguish answer calls from routing and evaluation calls.
  - Usage and latency reporting now focuses on the selected target without routing-tier fields.

- **Refactor**
  - Unified routing behavior around a shared decision representation across algorithms and clients.

### DIFF
--- a/crates/protocol/src/client.rs
+++ b/crates/protocol/src/client.rs
@@ -132,18 +132,31 @@ impl RoutingFallbackReason {
 /// A routing choice produced by an algorithm.
 #[derive(Clone, Debug)]
 pub struct Decision {
-    /// The target to call.
-    pub selected_target_id: String,
+    /// The model identifier selected for the call.
+    selected_model_id: String,
     /// Why, for logs and traces.
-    pub reasoning: Option<String>,
+    reasoning: Option<String>,
     /// True for an answer-generating call. False for classifier and judge calls.
-    pub is_answer_call: bool,
+    is_answer_call: bool,
 }
 
 impl Decision {
-    /// The target to call.
-    pub fn selected_target_id(&self) -> &str {
-        self.selected_target_id.as_str()
+    /// Creates a decision and records whether its call produces the answer.
+    pub fn new(
+        selected_model_id: impl Into<String>,
+        reasoning: Option<String>,
+        is_answer_call: bool,
+    ) -> Self {
+        Self {
+            selected_model_id: selected_model_id.into(),
+            reasoning,
+            is_answer_call,
+        }
+    }
+
+    /// The model identifier selected for the call.
+    pub fn selected_model_id(&self) -> &str {
+        self.selected_model_id.as_str()
     }
 
     /// Why this decision was made.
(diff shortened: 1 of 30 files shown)


ayushag-nv:
@coderabbit full review

crates/protocol/src/client.rs
```diff
@@ -129,54 +129,31 @@ impl RoutingFallbackReason {
...
-        &self.selected_model
+impl Decision {
+    /// The target to call.
+    pub fn selected_target_id(&self) -> &str {
+        self.selected_target_id.as_str()
     }
 
-    fn reasoning(&self) -> Option<&str> {
+    /// Why this decision was made.
+    pub fn reasoning(&self) -> Option<&str> {
```
nachiketb-nvidia, line 163: if everything inside the struct is pub, we dont need these anymore

ayushag-nv: made the things private in struct and keep the functions for readibility.

nachiketb-nvidia (approved):

grahamking (approved):

Issue #277, opened 2026-08-04
feat: bound how long a routed request may take


gburachas:
## Summary

A routed request has no upper bound on how long it can take. There are two places a bound is needed, and this issue tracks both. Supersedes #346.

## 1. A total timeout on the upstream request

`crates/switchyard-server/src/config.rs` has no per-client timeout setting. An upstream that accepts a request and then responds slowly holds the connection for as long as it likes.

**Relationship to #271.** That PR adds a connection timeout and an idle-read timeout to `TranslatingLlmClient`. Both are needed and neither bounds total duration: an idle-read timeout resets on every successful read, so an upstream that emits one token every 100 seconds never trips it while keeping the request open indefinitely.

The proposal here is a total per-request bound, configurable per client:

```toml
[llm_clients.weak]
timeout_secs = 600
```

If #271 lands first this should be added to its `HttpTransportConfig` rather than as a separate mechanism. I have commented on #271 to that effect.

One consequence to note: since #316 moved token counting to an inherent method that still goes through `send_encoded`, a per-client timeout would also bound `count_tokens`.

## 2. A deadline on the judge consultation

For `llm_classifier` routes, the judge call runs before the routed call. `JudgeClassifier::verdict` in `crates/libsy/src/algorithms/util/llm_judge.rs` has no time limit, so a slow judge delays a request that the user is waiting on for content the judge does not produce.

Proposal: `judge_deadline_ms` on the route. On expiry the judge is treated as unavailable and the route falls back through the existing path used for any other judge failure, so no new behaviour is introduced. Unset by default.

Two details:

- The bound must cover the whole consultation, not only the HTTP call. A judge that returns headers promptly and then stops mid-stream would otherwise hold the turn for the same length of time.
- Expiry should be reported through the existing fail-open reporting (`report_fail_open` in the same file), so it appears in the `switchyard.classifier_fail_open` metric alongside the other reasons rather than as an untracked path.

`judge_deadline_ms = 0` should be rejected at load time rather than treated as "no deadline".

## Sequencing

These are separate changes and will be submitted as separate pull requests. The judge deadline has no external dependency. The client timeout waits on #271.

gburachas:
Refreshing this against what has landed since, so the scope is unambiguous.

[...]

PR #351, opened 2026-08-11
fix(libsy): harden the classifier judge path

### LINKED ISSUE #278: fix(libsy): redact upstream error content from judge warning logs
## Problem — conversation content is written to logs today

`JudgeClassifier::verdict` logs every unavailable verdict with the error's raw `Display`:

```rust
tracing::warn!(target: "libsy", judge_model, error = %error, "judge verdict unavailable; ...");
```

That is not safe here, and the leak is concrete rather than theoretical:

- `LibsyError::ClientCall` renders as `"client call to target {target:?} failed: {source}"`;
- its source `LlmClientError::UpstreamHttp` renders as `"upstream returned HTTP {status}: {body}"`.

`body` is the upstream error body **verbatim**. Providers routinely quote the offending request back
in it, so any 4xx from a judge target writes conversation text into the operator's logs. Boxed
transport, decode, and FFI sources can carry the same.

This runs on a per-turn path, in the component most likely to be pointed at a self-hosted endpoint
with unfamiliar error formats.

## Proposal

A `robustness` module with `safe_error_summary` / `safe_client_error`, used at the judge's warn site.

Each case contributes only vetted operational detail — the failure's class, the target or model it
concerns, and a status code where one exists:

| Error | Logged as |
|---|---|
| `UpstreamHttp { status, body }` | `upstream HTTP 400` — body dropped |
| `ContextWindowExceeded { model, message }` | `context window exceeded for model weak` — message dropped |
| `Transport` / `InvalidResponse` / `Timeout` / `Ffi` | class only; boxed source dropped |
| `External { operation, source }` | `{operation} failed` — static label kept, source dropped |
| `AlgorithmError { message }` | **verbatim** — libsy builds it from static text |

libsy's own `AlgorithmError` is reported in full deliberately: it is built from a static string plus a
serde position and is the one message an operator needs to diagnose a bad judge reply. This mirrors
the `safe_types` idea from the Python implementation this is ported from.

## Design note — the exhaustive match earned its keep

The match over `LibsyError` is deliberately exhaustive rather than `_ => error.to_string()`, so a new
variant cannot start leaking by omission. That paid off immediately: the compiler caught
`External { operation, source }`, which pairs a safe static label with an unvetted boxed source from a
user extension, and which a catch-all would have silently rendered in full.

## Test approach

Each test asserts **both** that the summary omits the sensitive text **and** that the error's
`Display` still contains it — so the tests fail if the redaction is ever bypassed, not merely if it
changes shape.

## Scope

3 files, +177/−8. New `crates/libsy/src/algorithms/util/robustness.rs` (with 6 tests), its module
declaration, and 20 lines at the judge's log site. No public API change, no config surface.

## Why this is its own PR

Independent of the other two classifier changes: it is a data-hygiene defect on the logging path,
with no overlap in code path or surface. It is also the most clearly severity-driven of the three,
so it should not be queued behind a prompt-tuning discussion or a breaking-config debate.

## Validation

| Gate | Result |
|---|---|
| `cargo test --workspace` | 646 passed (baseline 640 + 6 new) |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --check` | clean |
| `uv run pytest tests/` | unchanged from baseline |

## Follow-up worth raising separately

Only the judge path is fixed here. Other `%error` log sites in the workspace may have the same
exposure and are worth an audit — deliberately out of scope so this stays reviewable.

### LINKED ISSUE #279: fix(libsy): cap the windowed judge payload and restate the routing instruction last
## Summary

When an `llm_classifier` route sets `recent_turn_window`, the judge is sent recent conversation in addition to the routing task. Two problems follow from how that payload is built, both in `crates/libsy/src/algorithms/llm_class.rs`.

Supersedes #349.

## Problem 1: the payload has no size limit

`trim_messages` (`llm_class.rs:84`) selects the messages the judge sees and returns them unchanged. The window is counted in **turns**, and turn size varies widely: a window of 4 turns is a few hundred characters of conversation, or tens of thousands if one turn carries a large tool result. Judge cost and latency vary accordingly for a single fixed configuration, and a large tool result can crowd out the request being judged.

Switchyard already solves this for its other judge. `crates/libsy/src/algorithms/util/escalation.rs`, which builds the payload for the escalation router, has:

- `MAX_REQUEST_CHARS = 18_000`, a cap on the whole payload
- `SYSTEM_CHARS = 1_000` and `FIRST_USER_CHARS = 2_000`, per-message caps on the parts worth protecting
- `truncate_middle` and `TRUNCATION_SUFFIX = "...<truncated>"`, so clipping is visible to the model
- a loop in `summarize_for_judge` (`escalation.rs:293`) that drops the oldest window entry until the payload fits

The capability and custom classifier paths have none of this. The same kind of window is bounded for one judge and unbounded for another.

## Problem 2: the routing instruction ends up buried

The judge is asked to route the user's request and answer with a routing JSON object. With a window set, that instruction is followed by assistant turns and tool output, and the judge tends to respond about the trailing content instead.

Example: a request of "check my account balance" followed by a window containing a long stack trace from a failed tool call. The verdict describes the stack trace rather than the balance lookup.

Restating the instruction as the **last** message fixes it. This is prompt ordering only; the text is a single sentence:

```
Route the conversation above. Output ONLY the routing JSON object, nothing else.
```

(An earlier revision of this issue called that "reinforcement", which invites confusion with reinforcement learning. It is unrelated.)

## Proposal

Two changes, to be submitted as separate pull requests:

1. **Restate the routing instruction after windowed content.** Only on the windowed path: with `recent_turn_window` unset, `task_messages` (`llm_class.rs:150`) sends the opening task and the latest user follow-up and no assistant or tool content, so there is nothing to crowd it out. The test should assert the instruction is the last message, not merely present, since position is the whole point.

2. **Cap the payload**, reusing the shape of `summarize_for_judge`, and expose `MAX_REQUEST_CHARS` as a route-level `judge_char_budget`. Default it to the current value so no existing deployment changes behaviour.

## Relationship to #487

#487 made `classify_trigger = "user_turn"` the natural setting for interactive agents, which is where a window is most likely to be configured. Both profiles added in `benchmark/routing-profiles/` set `recent_turn_window = 6`.

### LINKED ISSUE #346: feat(libsy): bound judge consultations with a configurable deadline
## Problem

The classifier judge call sits **in front of** the routed call. A judge that stalls stalls the turn behind it, and that consultation has no bound of its own — the route waits as long as the upstream is willing to hold the connection.

This is the one LLM call in the request path that the user is not waiting on for content. Its entire value is deciding which tier answers, and that value decays to zero the moment it costs more latency than the routing saves. There is no configuration that expresses "consult the judge, but not for longer than this".

## Proposal

`judge_deadline_ms` on the `llm_classifier` route:

```toml
[routes.assistant]
type = "llm_classifier"
judge_deadline_ms = 1500
```

On expiry the judge counts as unavailable and routing falls back through **exactly the same fail-open path** as any other judge failure — no new routing behaviour, no new failure mode to reason about. `None` (the default) leaves it unbounded, so existing deployments are unaffected.

Two details that matter:

- **The whole consultation is bounded, not just the HTTP call.** A judge that returns its headers promptly and then stalls mid-stream would otherwise hold the turn open just as long. The timeout therefore wraps both the call and the response aggregation.
- **Expiry is reported through the existing fail-open taxonomy** (`reason = "deadline_exceeded"`), so it lands on `/metrics` via `record_classifier_fail_open` alongside the other reasons rather than as an untracked silent path. The message is libsy-authored text plus the configured duration, so it carries no upstream content.

A zero deadline is rejected at load time rather than treated as "no deadline" — that is a config mistake with a silent and expensive failure mode.

## Relationship to #277

Different layer, and they compose. #277 bounds an individual upstream request at the client. This bounds the *routing decision* regardless of how many requests or how much streaming it involves.

Rebased onto current `main` and green.

gburachas:
Closes #278, closes #346, closes #279.

Three related fixes to the `llm_classifier` judge path, kept in one PR because they touch the same two functions and would conflict as separate branches. Each is one reviewable commit.

## 1. Redact upstream error content from judge warning logs (#278)

`report_fail_open` logs `error = %error`, and `LlmClientError::UpstreamHttp`'s `Display` interpolates the raw upstream body — on a warning that fires on **every** turn whose judge is unavailable. Providers routinely echo the offending request back in a 400, and the judge's request is a condensed view of the user's conversation.

This keeps everything #205 built — `report_fail_open`, the bounded `reason` taxonomy, `record_classifier_fail_open` — and changes one parameter: it now takes an already-redacted summary produced by an exhaustive match over `LibsyError`/`LlmClientError`.

Two properties worth preserving in review:

- the match is **deliberately exhaustive** rather than defaulting to `to_string()`, so a new error variant cannot start leaking by omission. Resolving a future compile error there with a `_ =>` arm would discard the whole guarantee.
- the test asserts both that the summary omits the secret **and** that `error.to_string()` still contains it. The second assertion documents that the hazard is still live in `Display`.

Not covered by #271, which redacts `Debug` and caps body size; the `Display` path is untouched by it.

## 2. Bound judge consultations with a configurable deadline (#346)

`judge_deadline_ms` on the route. The judge call sits in front of the routed call, so a stalled judge stalls the turn with no bound of its own.

The **whole consultation** is bounded, not just the HTTP call — a judge that returns headers promptly then stalls mid-stream would otherwise hold the turn just as long. Expiry folds into the existing fail-open path with `reason = "deadline_exceeded"`, so it appears on `/metrics` alongside the other reasons rather than as an untracked silent path. Default `None` preserves current behaviour exactly.

## 3. Reinforce the routing task after windowed conversation content (#279)

With `recent_turn_window` set, the judge sees conversation content after the task, and tends to assess the content rather than the request. This appends a short reinforcement line **end-positioned**, which is the only position where it works — the test asserts `contents.last()`, not merely presence.

Only the windowed path is affected. The default path carries user task messages only (the anchor and latest follow-up, per #309), so there is nothing there to outrank.

## How tested

- [x] `cargo test --workspace`, and again with `--test-threads=1` (this path registers process-global OTel instruments)
- [x] `uv run pytest tests/ -m "not integration"` — 871 passed, 9 skipped
- [x] `uv run ruff check .`
- [x] `uv run --only-group docs mkdocs build --strict`
- [x] Commits signed off per the DCO

Rebased onto current `main`, which required re-siting all three against #309 (`build_messages` rewritten), #312 (judge system prompt moved to `instructions`), #332/#338/#340 (`verdict()` rewritten) and #337 (`LibsyError::MissingClient` removed).

One rebase detail worth flagging, since it is the kind of thing that passes review while being wrong: a test here asserted `messages.len() == 2`, which held only while the judge system prompt lived in `messages[0]`. After #312 it lives in `instructions`. The assertion is now `== 1` **plus** an assertion that `instructions` is non-empty — decrementing the number alone would have silently stopped checking that the rubric reaches the model at all.

## Summary by CodeRabbit

* **New Features**
  * Added optional time limits for LLM judge consultations, with automatic fallback when a judge times out.
  * Added end-position reinforcement for windowed classifier prompts.
  * Exposed `judge_deadline_ms` configuration through Python and server configuration.

* **Bug Fixes**
  * Improved error handling with sanitized diagnostic summaries that omit sensitive response content.

* **Documentation**
  * Documented judge timeout configuration and fallback behavior.

### DIFF
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -33,6 +33,17 @@ const SCHEMA_TEMPLATE: &str = include_str!("../prompts/capability-classifier/sch
 /// Telemetry label for this algorithm's spans, metrics, and logs.
 const ALGORITHM_NAME: &str = "llm_task_classifier";
 
+/// Restates the task after the conversation the judge is asked to route.
+///
+/// The rubric leads the request, which works while the payload is a single short
+/// message. Once `recent_turn_window` includes real conversation, those instructions
+/// sit far from the generation point and the judge sometimes *answers* the
+/// conversation instead of classifying it — the reply then fails to parse, the
+/// verdict is unavailable, and the turn silently falls back. Repeating the task at
+/// the end is what reliably prevents that.
+const END_POSITION_REINFORCEMENT: &str =
+    "Route the conversation above. Output ONLY the routing JSON object, nothing else.";
+
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
 struct TaskClassifierVerdict {
@@ -156,10 +167,20 @@ impl ClassifierInput for TaskInput {
     fn build_messages(&self, _state: &State, request: &Request) -> Vec<Message> {
         // The default preserves the whole-task anchor and latest user update. A
         // configured window widens that to the surrounding conversation.
-        match self.recent_turn_window {
+        let mut messages = match self.recent_turn_window {
             Some(window) => trim_messages(&request.llm_request.messages, window),
             None => task_messages(&request.llm_request.messages),
+        };
+        // Only the windowed path carries assistant turns and tool traffic for the judge
+        // to be distracted by. The default path is user task messages only — the anchor
+        // and the latest follow-up — so there is nothing there to outrank.
+        if self.recent_turn_window.is_some() {
+            messages.push(Message::text(
+                Role::User,
+                END_POSITION_REINFORCEMENT.to_string(),
+            ));
         }
+        messages
     }
 }
 
@@ -1613,6 +1634,50 @@ mod tests {
                 Message::text(Role::User, "recent 2"),
             ]
         );
+    }
+
+    #[test]
+    fn a_window_reinforces_the_task_after_the_conversation() -> Result<()> {
+        let contents = judged_contents(2)?;
+
+        // Last, not merely present: the reinforcement only works end-positioned,
+        // after the conversation content it is meant to outrank.
+        assert_eq!(
+            contents.last().map(String::as_str),
+            Some(END_POSITION_REINFORCEMENT)
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn the_single_message_path_is_left_unreinforced() -> Result<()> {
+        // No window means no conversation to be distracted by, so the default
+        // request shape stays exactly as it was.
+        let judge = capability_judge(None)?;
+        let request = Request {
+            llm_request: LlmRequest {
+                messages: vec![Message::text(Role::User, "the task")],
+                ..LlmRequest::default()
+            },
+            raw_request: None,
+            metadata: None,
+        };
+
+        let built = judge.build_request(&State::default(), &request);
+
+        // The task message alone; the rubric reaches the judge as an instruction block
+        // rather than a message, so nothing was dropped by it not being counted here.
+        assert_eq!(built.llm_request.messages.len(), 1);
+        assert!(!built.llm_request.instructions.is_empty());
+        assert!(
+            !built
+                .llm_request
+                .messages
+                .iter()
+                .filter_map(|message| message.text_content("\n"))
+                .any(|text| text.contains(END_POSITION_REINFORCEMENT))
+        );
+        Ok(())
     }
 
     #[test]
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -5,6 +5,7 @@
 
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
 use serde::{Deserialize, Deserializer};
@@ -244,6 +245,13 @@ pub struct TaskClassifierConfig {
     pub contract: ClassifierContractConfig,
     /// Maximum completion tokens available to the classifier verdict.
     pub max_output_tokens: u64,
+    /// Bounds one judge consultation, in milliseconds.
+    ///
+    /// The judge call sits in front of the routed call, so a stalled judge stalls the
+    /// turn. On expiry the judge counts as unavailable and routing falls back exactly
+    /// as it does for any other judge failure. `None` (the default) leaves it
+    /// unbounded, preserving existing behaviour.
+    pub judge_deadline_ms: Option<u64>,
 }
 
 /// Flat serialized shape that maps prompt settings into the runtime contract.
@@ -263,6 +271,8 @@ struct TaskClassifierConfigWire {
     prompt: Option<String>,
     #[serde(default = "default_judge_max_output_tokens")]
     max_output_tokens: u64,
+    #[serde(default)]
+    judge_deadline_ms: Option<u64>,
 }
 
 impl<'de> Deserialize<'de> for TaskClassifierConfig {
@@ -283,6 +293,7 @@ impl<'de> Deserialize<'de> for TaskClassifierConfig {
             recent_turn_window: wire.recent_turn_window,
             contract,
             max_output_tokens: wire.max_output_tokens,
+            judge_deadline_ms: wire.judge_deadline_ms,
         })
     }
 }
@@ -301,6 +312,7 @@ impl Default for TaskClassifierConfig {
             recent_turn_window: None,
             contract: ClassifierContractConfig::default(),
             max_output_tokens: DEFAULT_JUDGE_MAX_OUTPUT_TOKENS,
+            judge_deadline_ms: None,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ayushag-nv:
@gburachas can you address coderabbit comments first, then I can review.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -32,6 +33,17 @@ const SCHEMA_TEMPLATE: &str = include_str!("../prompts/capability-classifier/sch
...
+/// Restates the task after the conversation the judge is asked to route.
+///
+/// The rubric leads the request, which works while the payload is a single short
+/// message. Once `recent_turn_window` includes real conversation, those instructions
+/// sit far from the generation point and the judge sometimes *answers* the
+/// conversation instead of classifying it — the reply then fails to parse, the
+/// verdict is unavailable, and the turn silently falls back. Repeating the task at
+/// the end is what reliably prevents that.
+const END_POSITION_REINFORCEMENT: &str =
+    "Route the conversation above. Output ONLY the routing JSON object, nothing else.";
```
ayushag-nv, line 45: This looks like some new feature add, may be we should discuss this once. Can you break down into smaller directed PRs for ease of review. Looks like it solves three issues simultaneously

nachiketb-nvidia:
closing due to inactivity

gburachas:
@ayushag-nv apologies for the silence on this, and thanks for the feedback. You were right that it bundled three unrelated changes.

It has been split. The first piece is #497, which contains only the judge log redaction and closes #278. The CodeRabbit comments raised here are addressed in it: `AlgorithmError::message` and `LlmClientError::Configuration::message` are no longer logged verbatim, and `LlmClientError::General` has an explicit arm.

The other two follow separately once that lands, since all three touched the same helper module:

- the judge deadline (#277), including the point raised here that `judge_deadline_ms` was accepted on escalation and custom routes but only applied in capability mode
- the windowed-payload change (#279)

PR #352, opened 2026-08-11
feat(libsy-llm-client): harden upstream request handling

### LINKED ISSUE #260: feat(libsy-llm-client): report rejected extra_body keys as a configuration error
### Problem

Per-target `extra_body` keys are merged into the outbound request body in
`crates/libsy-llm-client/src/client.rs` at a point that sits **outside** the retry loop, so the merge
happens once, before any attempt.

When an upstream rejects the request *because of* one of those keys — a 400/422 naming an unknown or
unexpected parameter — every retry re-sends the same rejected key. The retry cannot succeed, and
there is no way for a caller to say "drop this knob and try again".

The case we hit in practice: reasoning-control knobs (`chat_template_kwargs`, `enable_thinking`) that
some OpenAI-compatible servers accept and others reject outright. A call carrying the knob fails
permanently against a rejecting server, even though the identical call without it would succeed.

### Proposal

Move the `extra_body` merge inside the retry loop and allow named keys to be dropped after a
parameter rejection. Roughly 20 lines.

Deliberately narrow:

- Only a **400/422** whose body names the parameter, or an unknown/extra-parameter condition,
  qualifies. A 429 or 5xx must surface as-is — silently stripping a knob and re-firing on a
  rate limit would double the load and mask the real failure.
- **Retry exactly once** without the knob. No escalation ladder, no second strip.

### Scope

`crates/libsy-llm-client/` only. Benefits every route that configures `extra_body`, which is why it
seems worth doing as its own change rather than inside a feature PR.

### Notes

Happy to open the PR. Wanted to check the direction first, since it moves an existing statement
across the retry boundary rather than adding alongside it — if you'd prefer a different shape (e.g.
an explicit per-target "droppable keys" list instead of inferring from the reject), that's easy to
adjust.

### LINKED ISSUE #277: feat: bound how long a routed request may take

### LINKED ISSUE #345: feat(libsy): evict targets that reject a request as unservable
## Problem

The fallback ladder in `call_model_with_fallback` recognises two reasons a target can fail a route:

```rust
RoutingFallbackReason::ContextWindow => evictions.record(identity, failed),  // session-sticky
RoutingFallbackReason::Unavailable   => target_unavailable(&request, failed), // request-local
```

There is a third case it does not model: a target that rejects the request as one it **cannot serve at all** — an unsupported modality, a tool schema it does not accept, a feature the model lacks. Providers return these as a 400, so today they fall through `classify_fallback`'s `_ => return None` and surface to the caller as a hard error. The route has another target that could have served the turn and never tries it.

## Proposal

A `CapabilityRejected { model, message }` variant on `LlmClientError`, detected at the same place `ContextWindowExceeded` is (a 400 whose body matches a backend-supplied predicate), and a third `RoutingFallbackReason::Capability`.

**The routing question is which of the two existing behaviours it should share, and the answer is not the obvious one.** A capability reject looks like an availability problem — the target refused the request — but it is a permanent property of that target for this conversation. `Unavailable` is request-local, so mapping it there re-probes the rejecting target on every subsequent turn, paying a guaranteed failure per turn for the rest of the session. It belongs with `ContextWindow`:

```rust
RoutingFallbackReason::ContextWindow | RoutingFallbackReason::Capability => {
    evictions.record(identity, failed)
}
```

Both are "this target cannot serve this conversation", which is exactly what `SessionEvictions` records.

This distinction compiles either way and is invisible in a single-turn test, so it is worth pinning with a test that asserts the rejecting target is probed **exactly once across three turns**. I have verified that test fails if the variant is routed to `Unavailable` instead.

## Scope

- `crates/protocol/src/client.rs` — new `LlmClientError` variant and `RoutingFallbackReason::Capability` (`as_str()` → `"capability"`). This is a public-API addition, which is the main thing worth a design opinion before a PR.
- `crates/libsy-llm-client/src/{backend,client}.rs` — detection, ordered after the overflow check because an oversized request is recoverable on the same target and a capability reject never is.
- `crates/libsy/src/core/algorithm.rs` — `classify_fallback` mapping and the dispatch above.
- `crates/switchyard-server/src/lib.rs` — error mapping.

Related: #298 introduced the reason-based ladder this extends; #273 touches the same `fall_through.rs` region.

Implementation is rebased onto current `main` and passing `fmt`/`clippy -D warnings`/`test --workspace` (including `--test-threads=1`). Raising it as an issue first rather than opening the PR, since the `protocol` addition deserves a design call.

gburachas:
**Draft — the third commit carries an open design question (#345) that I would like a view on before this is landable. The first two are ready.**

Closes #277, closes #260, closes #345. Refs #271.

## 1. Bound upstream requests with a configurable timeout (#277)

`timeout_secs` on `[llm_clients.<name>]`, plus a guard treating an unusable value as no timeout rather than panicking.

Complementary to #271, not overlapping: that PR adds connect and *idle-read* timeouts at the reqwest-builder level; this is the **total per-request** bound, exposed in the deployment TOML. An upstream that emits a token every 100 s never trips an idle-read timeout. If #271 lands first I will wire this to its `HttpTransportConfig` rather than keeping a parallel mechanism.

Since #316 moved token counting to an inherent method that still routes through `send_encoded`, this also bounds `count_tokens` — desirable, but new, so worth stating.

## 2. Retry once without injected `extra_body` on a parameter reject (#260)

`merge_extra_body` only inserts keys the request omits, so it now returns the keys it actually injected. On a parameter reject those exact keys are stripped from the encoded body and the request is retried once.

The strip is **not** a budgeted retry — it does not consume `max_retries` — because it is a correction of our own injection rather than a hopeful repeat of the same request. A persistent reject surfaces rather than looping; there is a test for that.

## 3. Evict targets that reject a request as unservable (#345) — the part needing a view

Adds `LlmClientError::CapabilityRejected` and `RoutingFallbackReason::Capability`, so a target that rejects a request as one it cannot serve at all falls forward instead of failing the turn.

**The question is which existing behaviour it should share.** A capability reject looks like an availability problem, but it is a permanent property of that target for this conversation, so this maps it alongside `ContextWindow`:

```rust
RoutingFallbackReason::ContextWindow | RoutingFallbackReason::Capability => {
    evictions.record(identity, failed)
}
RoutingFallbackReason::Unavailable => target_unavailable(&request, failed),
```

Mapping it to `Unavailable` instead compiles cleanly and is invisible in a single-turn test — it just silently re-probes the rejecting target every turn for the rest of the session. `a_capability_reject_falls_forward_and_evicts_like_an_overflow` pins it by asserting the target is probed **exactly once across three turns**; I verified that test fails under the `Unavailable` mapping.

This is a public-API addition to `protocol`, which is why it is in draft.

## How tested

- [x] `cargo test --workspace`, and again with `--test-threads=1`
- [x] `uv run pytest tests/ -m "not integration"` — 871 passed, 9 skipped
- [x] `uv run ruff check .`, `mkdocs build --strict`
- [x] Commits signed off per the DCO

Each commit builds and tests independently, so the first two can be split out if the third needs longer.

Rebased onto current `main`. The `fall_through.rs` test harness needed a full rewrite for #337/#338/#340 (`LlmTarget` lost `llm_client`, `Arc<dyn Decision>` → `Arc<Decision>`, `Serve` closures replaced client structs); it is modelled on the existing `overflowing` helper.

### DIFF
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -35,6 +35,19 @@ const ANTHROPIC_OVERFLOW_PHRASES: &[&str] = &[
     "context length",
 ];
 
+// Phrases marking a request the model cannot serve at all, whatever its size:
+// multimodal content sent to a text-only deployment, or a server started without
+// its multimodal projector. The wording comes from the serving stack rather than
+// a provider error envelope, and no provider assigns it a structured `error.code`,
+// so one phrase list covers every backend variant.
+const CAPABILITY_REJECT_PHRASES: &[&str] = &[
+    "mmproj",
+    "image input",
+    "does not support image",
+    "does not support multimodal",
+    "no multimodal support",
+];
+
 /// Shared HTTP configuration for one upstream backend.
 #[derive(Clone)]
 pub struct HttpBackendConfig {
@@ -195,6 +208,16 @@ impl Backend {
             ),
             Backend::Anthropic(_) => is_overflow_body(body, |_| false, ANTHROPIC_OVERFLOW_PHRASES),
         }
+    }
+
+    /// Whether an upstream 400 `body` says the model cannot serve this request at
+    /// all — as opposed to it merely being too large.
+    ///
+    /// Provider-independent: the rejection is emitted by the serving stack, so the
+    /// same phrase list applies to every backend variant and there is no structured
+    /// check to short-circuit on.
+    pub(crate) fn is_capability_reject(&self, body: &str) -> bool {
+        is_overflow_body(body, |_| false, CAPABILITY_REJECT_PHRASES)
     }
 }
 
@@ -369,5 +392,37 @@ mod tests {
             )
         );
         assert!(!backend.is_context_overflow(r#"{"error":{"message":"overloaded"}}"#));
+    }
+
+    #[test]
+    fn detects_capability_reject_across_backends() {
+        // Provider-independent: the serving stack emits it, so every variant matches.
+        for backend in [
+            Backend::OpenAiChat(config("x")),
+            Backend::OpenAiResponses(config("x")),
+            Backend::Anthropic(config("x")),
+        ] {
+            assert!(backend.is_capability_reject(
+                r#"{"error":{"message":"image input is not supported by this model"}}"#
+            ));
+            assert!(backend.is_capability_reject(
+                r#"{"error":{"message":"server was started without an mmproj file"}}"#
+            ));
+            // Plain-text bodies from proxies still classify.
+            assert!(backend.is_capability_reject("this model does not support image content"));
+        }
+    }
+
+    #[test]
+    fn capability_reject_and_overflow_do_not_overlap() {
+        let backend = Backend::OpenAiChat(config("x"));
+        // An overflow is not a capability reject: a smaller request can still succeed.
+        let overflow = r#"{"error":{"code":"context_length_exceeded","message":"too long"}}"#;
+        assert!(backend.is_context_overflow(overflow));
+        assert!(!backend.is_capability_reject(overflow));
+        // And unrelated failures are neither.
+        let rate_limit = r#"{"error":{"message":"rate limit exceeded"}}"#;
+        assert!(!backend.is_context_overflow(rate_limit));
+        assert!(!backend.is_capability_reject(rate_limit));
     }
 }
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -3,7 +3,7 @@
 
 //! Per-provider backend configuration: wire format, upstream URL, and auth.
 
-use std::{collections::BTreeMap, fmt};
+use std::{collections::BTreeMap, fmt, time::Duration};
 
 use reqwest::RequestBuilder;
 use serde_json::Value;
@@ -48,6 +48,8 @@ pub struct HttpBackendConfig {
     pub extra_body: BTreeMap<String, Value>,
     /// Additional attempts after the initial upstream request.
     pub max_retries: u32,
+    /// Per-attempt request timeout in seconds. `None` leaves the request unbounded.
+    pub timeout_secs: Option<f64>,
 }
 
 impl fmt::Debug for HttpBackendConfig {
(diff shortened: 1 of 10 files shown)


gburachas:
Closing this and splitting it, following the review feedback on #351 that a PR should address one issue rather than several.

This PR covered three: a total request timeout (#277), an `extra_body` retry (#260), and a capability-reject fallback (#345). They will be resubmitted separately:

- **#260 (`extra_body` retry)** will be opened on its own. It touches only `crates/libsy-llm-client/src/client.rs`.
- **#277 (request timeout)** overlaps #271, which adds connection and idle-read timeouts to the same client. I have commented there rather than opening a competing change.
- **#345 (capability reject)** adds two public variants to the `protocol` crate and has not had a design response yet. It stays open as an issue until it does.

The branch is preserved; no work is lost.

PR #355, opened 2026-08-11
fix: misleading errors when rejecting unsupported optional payloads

andrewwhitecdw:
**Summary**

The rejection helpers for optional request payloads (`instructions`, `extensions`, `preservation`) currently call the shared `_sequence` / `_mapping` validators to decide whether to raise `ValueError("... is not supported")`. When the caller passes an unexpected scalar type, those validators raise `"must be a sequence"` or `"must be a mapping"` instead of the intended unsupported-field message, which is confusing for users debugging their request.

**Root cause**

In `examples/experimental/litellm/src/switchyard_litellm/client.py`:

- `_reject_sequence_payload` only rejects truthy sequences, but `_sequence(value, field)` raises `"must be a sequence"` for scalars or strings before the unsupported message can be emitted.
- `_reject_extensions_payload` and `_reject_preservation_payload` call `_mapping(...)`, which raises `"must be a mapping"` for non-mapping values instead of the unsupported message.

**Fix**

Replace the helper calls with direct `isinstance` shape checks that always raise `"... is not supported"` for any non-trivial, unsupported value while preserving the existing no-op behavior for empty/default values.

**Testing**

```bash
cd examples/experimental/litellm
.venv/bin/python -m pytest tests/test_client.py -q
```

Result: `33 passed`.

**Contributor guidelines**
- Commit includes DCO `Signed-off-by` trailer.
- Single-commit branch.
- Changes limited to one file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for optional instruction, extension, and preservation settings.
  * Empty instruction lists are now accepted.
  * Unsupported nested values are rejected more consistently while preserving existing error messages.

### DIFF
--- a/examples/experimental/litellm/src/switchyard_litellm/client.py
+++ b/examples/experimental/litellm/src/switchyard_litellm/client.py
@@ -39,7 +39,11 @@ def _reject_sequence_payload(
     value = request.get(field)
     if value is None:
         return
-    if _sequence(value, field):
+    if (
+        isinstance(value, Sequence)
+        and not isinstance(value, (str, bytes, bytearray))
+        and value
+    ):
         raise ValueError(f"{field} is not supported")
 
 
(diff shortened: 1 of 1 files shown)


ayushag-nv:
@andrewwhitecdw Thanks for the contrib. It's still in early development. Things might change here. Also the current change only rejects empty sequences, there needs to more checks that should be added. I would say, not required right now.

PR #356, opened 2026-08-11
fix: simplify identical source/target loop in payload builder

andrewwhitecdw:
fix: simplify identical source/target loop in payload builder

### Root cause
The loop copied sampling values with a source-to-target mapping, but the keys (`temperature`, `top_p`) were identical on both sides, so the mapping added noise without adding flexibility.

### Fix
Iterate over the keys directly. Behavior is preserved: values are still copied only when present.

```diff
-    for source, target in (("temperature", "temperature"), ("top_p", "top_p")):
-        value = sampling.get(source)
-        if value is not None:
-            payload[target] = value
+    for key in ("temperature", "top_p"):
+        value = sampling.get(key)
+        if value is not None:
+            payload[key] = value
```

### Testing
- `pytest tests/test_client.py tests/test_payload_sampling.py -v`: 34 passed
- Full `pytest` has one pre-existing failure unrelated to this change: `test_stage_router_drives_both_litellm_targets`.
- `mypy` reports one pre-existing `no-redef` error on line 127 (also present on unmodified `main`).

### Why existing tests missed it
No functional change; the new regression test explicitly verifies `temperature`/`top_p` are forwarded through the payload builder.

### Contributor guidelines
- DCO sign-off: included.
- Squash: one commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured temperature and top-p sampling settings are correctly forwarded in generated requests.

* **Tests**
  * Added coverage verifying that both sampling values are preserved in request payloads.

### DIFF
--- a/examples/experimental/litellm/src/switchyard_litellm/client.py
+++ b/examples/experimental/litellm/src/switchyard_litellm/client.py
@@ -236,10 +236,10 @@ def _payload(request: Mapping[str, object], model: str) -> dict[str, Any]:
     tool_choice = _tool_choice(request)
     if tool_choice is not None:
         payload["tool_choice"] = tool_choice
-    for source, target in (("temperature", "temperature"), ("top_p", "top_p")):
-        value = sampling.get(source)
+    for key in ("temperature", "top_p"):
+        value = sampling.get(key)
         if value is not None:
-            payload[target] = value
+            payload[key] = value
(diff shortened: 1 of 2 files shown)


ayushag-nv:
@andrewwhitecdw the target and source being same is just a case here, I would keep it as it is. Not hurtful to have a clear separation. Will be useful if there is a case in future for source and target to be different.

PR #363, opened 2026-08-11
docs(stage-router): mark capable_first experimental, and warn at startup

sabhatinas:
Every published threshold and routing result comes from `efficient_first` runs. `capable_first` works and the server accepts it, but it has not been benchmarked — so there is no calibrated `confidence_threshold` for it and no measured accuracy or cost to set expectations against.

Docs now lead with `efficient_first` and carry a warning admonition; the TOML schema table marks `capable_first` experimental. The server also logs a warning at startup when a `stage_router` route selects it, so the caveat reaches operators who never read the page:

```
WARN switchyard_server::config: stage_router route stage uses picker "capable_first",
which is experimental: published thresholds and routing results all come from
"efficient_first", so there is no calibrated confidence_threshold for it and no
measured accuracy or cost. Use "efficient_first" unless you are running your own
calibration.
```

Verified the warning fires for `capable_first` and stays silent for `efficient_first` (`--dry-run` on both). `cargo test -p switchyard-server`, clippy, `cargo fmt --check` and `mkdocs build --strict` all pass.

## Summary by CodeRabbit

* **New Features**
  * Added a server startup warning when the experimental `capable_first` routing picker is selected.

* **Documentation**
  * Clarified that `efficient_first` is the recommended picker.
  * Documented that published thresholds and routing results are based on `efficient_first`.
  * Noted that `capable_first` is unbenchmarked and lacks calibrated metrics.

### DIFF
--- a/crates/switchyard-server/src/config.rs
+++ b/crates/switchyard-server/src/config.rs
@@ -928,6 +928,11 @@ fn build_algorithm(
             classifier,
             ..
         } => {
+            if matches!(picker, PickerMode::CapableFirst) {
+                tracing::warn!(
+                    "stage_router route {route_name} uses picker \"capable_first\", which is experimental: published thresholds and routing results all come from \"efficient_first\", so there is no calibrated confidence_threshold for it and no measured accuracy or cost. Use \"efficient_first\" unless you are running your own calibration."
+                );
+            }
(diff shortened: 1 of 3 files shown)


ayushag-nv (approved):
looks good

PR #364, opened 2026-08-11
chore: Simplify `run_stream`, rename `Driver::info` to `Driver::decide`.

grahamking:
Part of https://github.com/NVIDIA-NeMo/Switchyard/issues/310

- `run_stream` simplification. In Claude's words:

> Gone: futures::stream::select, the stream::once tail future, the filter_map, and two of the three Box::pins

- Rename `Driver::info` to `Driver::decide`. This is how an algorithm
  says it has decided to call an LLM for a user. This does not mean it
  is the final call. Escalation router in particular can make two
  `decide` calls. And `FallThrough` makes one per attempt.

- Remove a now-unused error.

Assisted-by: Claude:Opus 5 medium

## Summary by CodeRabbit

* **Improvements**
  * Routing and algorithm decisions are now consistently recorded as decision events for clearer observability.
  * Stream processing now reports `MissingFinalResponse` when an algorithm terminates unexpectedly, providing more consistent error handling.
  * Completion handling has been streamlined so final results are published directly before streams close.
* **Breaking Changes**
  * The `Driver::info` method has been renamed to `Driver::decide`.
  * The `AlgorithmTask` error variant has been removed.

### DIFF
--- a/crates/libsy/src/core/algorithm.rs
+++ b/crates/libsy/src/core/algorithm.rs
@@ -8,13 +8,14 @@
 use std::{
     collections::{HashMap, HashSet},
     future::Future,
+    panic::AssertUnwindSafe,
     pin::Pin,
     sync::Arc,
     time::Instant,
 };
 
 use async_trait::async_trait;
-use futures::{Stream, StreamExt};
+use futures::{FutureExt, Stream, StreamExt};
 use parking_lot::Mutex;
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
@@ -184,7 +185,7 @@ impl Driver {
     /// Publish a routing [`Decision`] as a [`Step::Decision`] on the stream.
     /// Each successfully published decision is counted and logged with its
     /// reasoning; a decision the stream never accepted is not recorded.
-    pub async fn info(&self, decision: Decision) -> Result<()> {
+    pub async fn decide(&self, decision: Decision) -> Result<()> {
         self.step_tx
             .send(Ok(Step::Decision(decision.clone())))
             .await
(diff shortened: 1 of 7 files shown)


grahamking:
@coderabbit review

crates/libsy/src/core/algorithm.rs
```diff
@@ -546,50 +556,45 @@ pub trait Algorithm: Send + Sync + 'static {
...
-                    Err(source) => Err(LibsyError::AlgorithmTask { source }),
-                };
-                finish_driver.finish(result).await
-            })
-            .filter_map(|finish_result| async move { finish_result.err().map(Err) }),
-        );
-
-        let stream: StepStream = Box::pin(stream);
-        Box::pin(futures::stream::select(stream, tail).map(move |step| {
+        Box::pin(ReceiverStream::new(step_rx).map(move |step| {
```
ayushag-nv, line 597: does RECEIVERSTream thing encapsulates all of the above which you deleted ??"

grahamking: Good question, but no.

It's in tokio. It turns a channel receiver (`step_rx`) into a `futures::Stream`. Then that `map` (which is from `StreamExt` not the stdlib) replaces every item in the stream by that little function, which keeps the abort handle alive and passes on the `Step`.

The parts I deleted were mostly keeping the `JoinHandle` alive, which was only necessary if the algorithm panics. That's covered by the `catch_unwind` stuff instead.

ayushag-nv: Nice thanks for the explanation

ayushag-nv (approved):

PR #368, opened 2026-08-11
docs: add a self-hosted (local vLLM) deployment example

edrethardo:
Every deployment example in the repo points at a hosted provider (`docs/reference/toml_schema.md`, `docs/getting_started.md`, `switchyard/cli/defaults/openrouter.toml` all use OpenRouter). There is no reference for pointing a client at a local OpenAI-compatible server, which is the case the local-model story depends on. `docs/vllm-serve-hidden-state.md` covers hidden-state extraction, not using vLLM as a routing backend.

This adds `examples/local-vllm/` with a hybrid deployment — local weak tier and local classifier, hosted strong tier — plus a short README.

Two things that cost me time and are now written down:

- **`id` must match `/v1/models` on the local server**, not the checkpoint path. With vLLM that is `--served-model-name`; unset, it defaults to the full path passed to `vllm serve`.
- **`api_key_env` is omitted for the local client.** Per `toml_schema.md` omitting it sends no authentication, which is what a server started without `--api-key` expects. Worth demonstrating, since every existing example sets a key.

Docs only, no code changes.

## Verification

- `switchyard-server` accepts the config; the launcher starts a proxy against it and an agent runs through it.
- Confirmed the env-var check fires as documented when `OPENROUTER_API_KEY` is unset:
  `invalid server config ...: llm client openrouter could not read api_key_env OPENROUTER_API_KEY: environment variable not found`

I did not run `pre-commit` (it needs the full dev env sync); happy to fix anything CI flags.

## Summary by CodeRabbit

- **New Features**
  - Added an example configuration for routing requests through a self-hosted, OpenAI-compatible model.
  - Supports local-first classification with hosted handling for more complex requests.
  - Includes session affinity, threshold-based routing, and message-hash fallback options.

- **Documentation**
  - Added setup and usage instructions for starting the local model, deploying routes, running dry runs, and testing requests.
  - Documented model IDs, authentication, and local-versus-hosted classification settings.

### DIFF
--- a/examples/local-vllm/README.md
+++ b/examples/local-vllm/README.md
@@ -1,0 +1,49 @@
+# Routing to a self-hosted model
+
+Every deployment example in this repository points at a hosted provider. This one
+puts the **weak tier on a local OpenAI-compatible server** — vLLM, NIM, Ollama or
+llama.cpp — and keeps a hosted model for the hard turns, so routine steps never
+leave the machine.
+
+## Files
+
+| File | Purpose |
+|---|---|
+| `routes.toml` | Hybrid deployment: local weak + local classifier, hosted strong. |
+
+## Run it
+
+Start the local server, giving it a stable name to route to:
+
+```bash
+vllm serve <checkpoint> --served-model-name my-local-model --port 8000
+```
+
+Point the deployment at it and check the wiring without sending traffic:
+
+```bash
+export OPENROUTER_API_KEY=...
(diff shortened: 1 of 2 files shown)


ayushag-nv:
@edrethardo I won't add this here. Deploying vllm is another art itself, so I won't refer that in switchyard. The route toml files are self explanatory to put the base-url stuff depending on the deployment. I suggest to close this one.

PR #370, opened 2026-08-11
fix(translation): report content filter stops as Anthropic refusal

### LINKED ISSUE #369: [bug] OpenAI content_filter is silently translated to Anthropic end_turn
## Symptom

When an Anthropic Messages request is served by an OpenAI-compatible backend, an upstream `finish_reason` of `content_filter` is translated to an Anthropic `stop_reason` of `end_turn`.

A response blocked or truncated by content filtering is therefore presented as a normal completion. The translation also emits no diagnostic indicating that the stop reason was lost.

## Reproduction

Translate an OpenAI Chat response to Anthropic Messages with the following values:

- Response ID: `chatcmpl-probe`
- Model: `gpt-4o`
- Assistant content: `partial answer before the block`
- Finish reason: `content_filter`
- Prompt tokens: 10
- Completion tokens: 5
- Total tokens: 15

Call `translate_response` with `WireFormat::OpenAiChat` as the source, `WireFormat::AnthropicMessages` as the target, and the default translation policy.

The translated response reports `stop_reason: "end_turn"`.

A control response with `finish_reason: "stop"` also reports `stop_reason: "end_turn"`, making the filtered response indistinguishable from a normal completion.

## Expected vs. actual

- **Expected:** Translation should not silently make a content-filter stop indistinguishable from a normal completion. It should preserve the distinction, emit a lossy-conversion diagnostic and honor the configured lossy-conversion policy, or translate the event as a failure.
- **Actual:** The response contains `stop_reason: "end_turn"` and no translation diagnostic.

## Environment

- Switchyard commit SHA: `fb3fc30`
- Python: `python3 --version` reports 3.9.6, though this path is Rust-only
- OS / arch: macOS 26.2, Darwin 25.2.0, arm64
- Install path: source build
- Inbound format: Anthropic Messages
- Backend: OpenAI or another OpenAI-compatible upstream

## Additional context

The buffered OpenAI decoder preserves `content_filter` as `StopReason::ContentFilter` in `crates/switchyard-translation/src/codecs/openai_chat/buffered.rs:1133`. The OpenAI encoder maps that normalized variant back to `content_filter` at line 1148.

The loss occurs in the Anthropic buffered encoder at `crates/switchyard-translation/src/codecs/anthropic/buffered.rs:986`, where `StopReason::ContentFilter` is mapped to `end_turn`.

The streaming path has the same behavior. The OpenAI stream decoder copies the upstream finish reason into the normalized stop event at `crates/switchyard-translation/src/codecs/openai_chat/stream.rs:142`. The Anthropic stream encoder stores that reason at `crates/switchyard-translation/src/codecs/anthropic/stream.rs:206`, then maps unrecognized values, including `content_filter`, to `end_turn` at line 571.

The stream accumulator also recognizes `content_filter` as `StopReason::ContentFilter` at `crates/protocol/src/stream.rs:432`.

Because the Anthropic vocabulary handled by this codec has no direct content-filter equivalent, the wire behavior needs a design decision. Protocol-safe options include:

1. Emit a lossy-conversion diagnostic and honor `LossyConversionPolicy`.
2. Translate a filtered generation into an error response or event.
3. Preserve the original reason in an explicitly supported extension, if one exists or is introduced.

I can provide the local probe as a regression test once the desired behavior is chosen.

Atharva-Kanherkar:
## What

- Translate OpenAI `content_filter` stops to Anthropic `refusal` in buffered and streaming responses.
- Emit the required Anthropic `stop_details` object for synthesized refusals.
- Decode Anthropic `refusal` back to the normalized content-filter reason.
- Normalize streamed Anthropic refusals as `StopReason::ContentFilter` when responses are accumulated.
- Translate streamed Anthropic refusals to OpenAI `content_filter`.
- Add buffered, streaming, and accumulator regression coverage.

## Why

An OpenAI-compatible backend could stop generation because of content filtering, but Anthropic clients received `end_turn`. That made a filtered or truncated response indistinguishable from a normal completion.

The normalized response already preserves this condition as `StopReason::ContentFilter`. The loss occurred when the Anthropic codecs encoded the stop reason. Anthropic documents `refusal` for safety refusals and requires a `stop_details` object, so the translated response now preserves the relevant meaning with a valid Anthropic shape.

`Error` and `Unknown` remain mapped to `end_turn` and are outside this change.

Closes #369

## Validation

- `cargo fmt --all --check`
- `cargo clippy --ignore-rust-version -p switchyard-protocol -p switchyard-translation --all-targets -- -D warnings`
- `cargo test --ignore-rust-version -p switchyard-protocol`: 18 unit tests and 2 documentation tests passed
- `cargo test --ignore-rust-version -p switchyard-translation`: 119 tests passed
- Commitlint across all three PR commits: 0 problems and 0 warnings

The full workspace test run is blocked locally while linking the existing `switchyard-py` PyO3 cdylib because of a macOS deployment-target mismatch in `aws-lc-sys`. The same linker failure reproduces on a clean tree. CI runs the workspace suite on Linux.

The local Rust commands use `--ignore-rust-version` because the available Homebrew compiler is Rust 1.96.0 while the repository declares Rust 1.96.1.

### DIFF
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -433,7 +433,7 @@ fn stop_reason_from_str(reason: Option<&str>) -> StopReason {
     match reason {
         Some("length" | "max_tokens") => StopReason::MaxTokens,
         Some("tool_calls" | "function_call" | "tool_use") => StopReason::ToolUse,
-        Some("content_filter" | "refusal") => StopReason::ContentFilter,
+        Some("content_filter") => StopReason::ContentFilter,
         Some("stop" | "end_turn" | "stop_sequence") | None => StopReason::EndTurn,
         Some(_) => StopReason::Unknown,
     }
@@ -488,15 +488,6 @@ mod tests {
                 text: "Hello".to_string()
             }]
         );
-    }
-
-    #[test]
-    fn folds_anthropic_refusal_as_content_filter() {
-        let agg = fold(vec![LlmResponseChunk::MessageStop {
-            reason: Some("refusal".to_string()),
-        }]);
-
-        assert_eq!(agg.outputs[0].stop_reason, Some(StopReason::ContentFilter));
     }
 
     #[test]
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -433,7 +433,7 @@ fn stop_reason_from_str(reason: Option<&str>) -> StopReason {
     match reason {
         Some("length" | "max_tokens") => StopReason::MaxTokens,
         Some("tool_calls" | "function_call" | "tool_use") => StopReason::ToolUse,
-        Some("content_filter") => StopReason::ContentFilter,
+        Some("content_filter" | "refusal") => StopReason::ContentFilter,
         Some("stop" | "end_turn" | "stop_sequence") | None => StopReason::EndTurn,
         Some(_) => StopReason::Unknown,
     }
@@ -488,6 +488,15 @@ mod tests {
                 text: "Hello".to_string()
             }]
         );
+    }
+
+    #[test]
+    fn folds_anthropic_refusal_as_content_filter() {
+        let agg = fold(vec![LlmResponseChunk::MessageStop {
+            reason: Some("refusal".to_string()),
+        }]);
+
+        assert_eq!(agg.outputs[0].stop_reason, Some(StopReason::ContentFilter));
     }
 
     #[test]
--- a/crates/switchyard-translation/src/codecs/anthropic/stream.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/stream.rs
@@ -572,9 +572,12 @@ fn anthropic_stop_reason(reason: Option<&str>) -> String {
     match reason {
         Some("length") => "max_tokens".to_string(),
         Some("tool_calls") | Some("function_call") => "tool_use".to_string(),
-        Some("end_turn") | Some("max_tokens") | Some("tool_use") | Some("stop_sequence") => {
-            reason.unwrap_or("end_turn").to_string()
-        }
+        Some("content_filter") => "refusal".to_string(),
+        Some("end_turn")
+        | Some("max_tokens")
+        | Some("tool_use")
+        | Some("stop_sequence")
+        | Some("refusal") => reason.unwrap_or("end_turn").to_string(),
         _ => "end_turn".to_string(),
     }
 }
--- a/crates/switchyard-translation/src/codecs/anthropic/stream.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/stream.rs
@@ -117,6 +117,12 @@ fn decode_anthropic_stream(
                 // Remember the provider stop reason: Anthropic delivers it here, on
                 // `message_delta`, while the terminal `message_stop` carries none of its own.
                 state.stop_reason = Some(stop_reason.to_string());
+                state.stop_details = object
+                    .get("delta")
+                    .and_then(Value::as_object)
+                    .and_then(|delta| delta.get("stop_details"))
+                    .filter(|details| !details.is_null())
+                    .cloned();
                 out.push(LlmResponseChunk::MessageStop {
                     reason: Some(stop_reason.to_string()),
                 });
@@ -265,7 +271,10 @@ fn finish_anthropic_stream(state: &mut StreamTranslationState) -> Vec<Value> {
             "delta": {
                 "stop_reason": anthropic_stop_reason(state.stop_reason.as_deref()),
                 "stop_sequence": Value::Null,
-                "stop_details": anthropic_stop_details(state.stop_reason.as_deref()),
+                "stop_details": anthropic_stop_details(
+                    state.stop_reason.as_deref(),
+                    state.stop_details.as_ref(),
+                ),
             },
             "usage": anthropic_stream_usage(state),
         }));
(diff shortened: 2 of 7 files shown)


Atharva-Kanherkar:
hi @nachiketb-nvidia, i have loved using switchyard. i ran into this bug and put together a small fix. if you have time, could you please take a look? thank you!

crates/protocol/src/stream.rs
```diff
@@ -433,7 +433,7 @@ fn stop_reason_from_str(reason: Option<&str>) -> StopReason {
     match reason {
         Some("length" | "max_tokens") => StopReason::MaxTokens,
         Some("tool_calls" | "function_call" | "tool_use") => StopReason::ToolUse,
-        Some("content_filter") => StopReason::ContentFilter,
+        Some("content_filter" | "refusal") => StopReason::ContentFilter,
```
nachiketb-nvidia: nit: i dont think we need this extra "refusal". This is an IR so its provider neutral, so codecs can account for "refusal", onus is on specific codecs

crates/protocol/src/stream.rs
```diff
@@ -490,6 +490,15 @@ mod tests {
         );
     }
 
+    #[test]
+    fn folds_anthropic_refusal_as_content_filter() {
```
nachiketb-nvidia: this test could be removed after last comment

crates/switchyard-translation/tests/response_translation.rs
```diff
@@ -600,3 +600,64 @@ fn incomplete_responses_source_survives_translation() -> TestResult {
     assert_eq!(output["choices"][0]["finish_reason"], "length");
     Ok(())
 }
+
+// Verifies a moderation stop reaches Anthropic clients as `refusal` rather than
+// being reported as a normal `end_turn`.
+#[test]
+fn openai_content_filter_translates_to_anthropic_refusal() -> TestResult {
+    let engine = TranslationEngine::default();
```
nachiketb-nvidia, line 608: nit: pls reduce tests to one, or even better, change an existing test. with previous comments, some of these tests will become redundant.

nachiketb-nvidia (changes requested):

crates/switchyard-translation/src/codecs/anthropic/stream.rs
```diff
@@ -572,13 +573,28 @@ fn anthropic_stop_reason(reason: Option<&str>) -> String {
...
         Some("tool_calls") | Some("function_call") => "tool_use".to_string(),
-        Some("end_turn") | Some("max_tokens") | Some("tool_use") | Some("stop_sequence") => {
-            reason.unwrap_or("end_turn").to_string()
-        }
+        Some("content_filter") => "refusal".to_string(),
+        Some("end_turn")
+        | Some("max_tokens")
+        | Some("tool_use")
+        | Some("stop_sequence")
+        | Some("refusal") => reason.unwrap_or("end_turn").to_string(),
```
nachiketb-nvidia: "content_filter" -> "refusal" but "refusal" is "end_turn"?

crates/switchyard-translation/src/codecs/anthropic/stream.rs
```diff
@@ -572,13 +573,28 @@ fn anthropic_stop_reason(reason: Option<&str>) -> String {
...
     }
 }
 
+// Emits the metadata object required by Anthropic refusal events.
+fn anthropic_stop_details(reason: Option<&str>) -> Value {
+    match reason {
+        Some("content_filter" | "refusal") => json!({
+            "type": "refusal",
+            "category": Value::Null,
+            "explanation": Value::Null,
```
nachiketb-nvidia: lets add category and explanation in, in line with whatever anthropic API does

crates/protocol/src/stream.rs
Atharva-Kanherkar: removed, it only covered the ir folding that just went away.

crates/switchyard-translation/tests/response_translation.rs
Atharva-Kanherkar: cut down to one test per file, each covering both directions instead of a test per direction. four tests down to two.

crates/switchyard-translation/src/codecs/anthropic/stream.rs
Atharva-Kanherkar: it did return refusal, the unwrap_or("end_turn") just made it look otherwise. folded refusal in with content_filter and bound the passthrough arm by pattern so there is no unwrap_or left to misread. behaviour is unchanged.

crates/switchyard-translation/src/codecs/anthropic/stream.rs
Atharva-Kanherkar: looked this up, the docs say both fields are null when the refusal maps to no named category, and that null is a normal permanent value rather than a placeholder. openai reports no category, and the documented ones are all anthropic policy specific, so i did not want to invent one. the actual gap was the other direction, a real anthropic stop_details with a category was getting overwritten with nulls. both codecs now replay the source object and only fall back to nulls when synthesising.

crates/protocol/src/stream.rs
Atharva-Kanherkar: agreed, reverted it back to just content_filter. the anthropic codec already maps refusal in map_anthropic_stop_reason, so the ir stays provider neutral.

Atharva-Kanherkar:
i also mutation tested the new preservation asserts to confirm they actually fail without the fix.

nachiketb-nvidia (approved):
thank you!

PR #371, opened 2026-08-11
feat(libsy): advisor_gate review-gate algorithm

eric-liu-nvidia:
**Stack 1/3** — this PR now contains only the core algorithm (libsy), per review. The rest of the original diff moved up the stack:

image

| Stack | PR | Content |
|---|---|---|
| 1/3 | this PR | `AdvisorGate` libsy algorithm + unit tests |
| 2/3 | #382 | `type = "advisor"` server route config |
| 3/3 | #383 | `/v1/stats` projection + server e2e tests |

Adds the **advisor review gate** to the Rust server as a new route type: `type = "advisor"`.

## Benchmark results (Terminal-Bench 2.1)

| Configuration | Accuracy |
|---|---|
| Nemotron 3 Ultra (baseline) | 43.8% |
| Nemotron 3 Ultra + Opus 4.8 as advisor review gate | **54.7% ± 0.7 pp** |

With $64.76 ± $4.95 additional Opus 4.8 cost, the advisor review gate lifts Nemotron 3 Ultra accuracy by almost 25%.

## How it works

There are two models: an **executor** (the model doing the work) and an **advisor** (a stronger model that checks the work).

1. The client talks to the executor as usual. While the executor is still working — its replies contain tool calls — everything passes straight through. The advisor is not involved.
2. When the executor produces a reply with **no tool calls**, that means it thinks it is done (or it is proposing a plan). The proxy holds that reply back instead of returning it.
3. The proxy sends the conversation and the held-back reply to the advisor and asks one question: is this work complete and correct? The advisor answers **APPROVE** or **REDO**.
4. **APPROVE** → the held-back reply is released to the client unchanged.
   **REDO** → the client never sees that reply. The proxy puts the advisor's feedback into the conversation ("not done yet — here is what is missing") and calls the executor again, so it keeps working.
5. The review happens **once per session** (`max_reviews`, default 1). After that, every call is a plain passthrough with zero overhead.

If the advisor is down or replies with something unparseable, the gate fails open: the executor's reply goes to the client as if approved. Advisor problems never block the executor.

Why this shape: giving advice up front made the executor trust the plan and skip its own test-and-iterate loop. Checking only the final "I'm done" claim leaves the executor's normal behavior untouched and catches exactly one failure mode — stopping too early.

## Example

```toml
[routes.gated]
id = "switchyard-advisor"
type = "advisor"
executor_target = "executor"
advisor_target = "advisor"   # judge-only; set max_retries = 0 on its llm_client
```

Note for reviewers: this adds `regex` as a production dependency of libsy (for the configurable review trigger pattern).


ayushag-nv:
@eric-liu-nvidia can you address coderabbit first

nachiketb-nvidia (dismissed):
Let's break this up into 2 or 3 MRs

- one for core logic
- one more for potential telemetry 
- one more for server side config changes

also, if we have any prompts, let's make sure they're also knobs

crates/switchyard-server/tests/server.rs
```diff
@@ -2138,3 +2169,365 @@ async fn request_and_upstream_errors_use_the_inbound_wire_format() -> TestResult
...
+
+fn advisor_chat_body(prompt: &str) -> Value {
+    json!({
+        "model": "switchyard/advisor",
+        "messages": [{"role": "user", "content": prompt}]
+    })
+}
+
+#[tokio::test]
+async fn advisor_route_approve_flow_and_stats() -> TestResult {
```
grahamking: Can you wrap all of your tests in a `#[cfg(test)] mod tests {`? That will ensure they don't get compiled unless you are running tests. Grep for those strings in most other files for an example.

crates/libsy/src/algorithms/advisor_gate.rs
ayushag-nv, line 1: 2k lines of code for one file is huge. May be try to move some utilities in the util folder and reduce the scope. Prompt should not live in the code file, we have prompts folder for that

ayushag-nv (dismissed):

crates/libsy/src/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,2127 @@
...
+    if text.is_empty() {
+        return None;
+    }
+    let joined = text.join("\n");
+    if joined.is_empty() {
+        None
+    } else {
+        Some(joined)
+    }
+}
```
ayushag-nv: This can be util and removed from the main algo file

crates/libsy/src/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,2127 @@
...
+
+/// Assistant turns already in the request — the stall checkpoint's clock.
+fn assistant_turns(messages: &[Message]) -> u32 {
+    let count = messages
+        .iter()
+        .filter(|message| message.role == Role::Assistant)
+        .count();
+    u32::try_from(count).unwrap_or(u32::MAX)
+}
```
ayushag-nv: These too can move to util or some other file

eric-liu-nvidia:
@nachiketb-nvidia done — split into a 3-PR stack along exactly those lines:

| Stack | PR | Content |
|---|---|---|
| 1/3 | this PR | core logic: `AdvisorGate` libsy algorithm + unit tests |
| 2/3 | #382 | server-side config: `type = "advisor"` route |
| 3/3 | #383 | telemetry: `/v1/stats` projection + server e2e tests |

#382 targets this PR's branch and #383 targets #382's, so each diff shows only its own layer; they merge bottom-up.

On prompts as knobs: they already are — `reviewer_system_prompt` and `redo_feedback_prefix` are TOML fields on the route (in #382); the constants in libsy are only the defaults.

@ayushag-nv CodeRabbit is addressed: the doc-comment suggestion is applied in #383, and the "does not compile" finding is incorrect — `StatsAccumulator::reset` takes `&self` (interior mutability via the inner mutex, crates/switchyard-server/src/stats/accumulator.rs:140), same pattern as the existing stage_router projection test; the test builds and passes in CI.

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Done in 97ec7bd4 — the file is split into focused submodules: `advisor_gate.rs` (config, budget ledger, gate flow, ~640 lines), `advisor_gate/turn.rs` (turn buffering/inspection/replay), `advisor_gate/transcript.rs` (transcript serialization + verdict parsing), `advisor_gate/telemetry.rs` (metrics + audit lines), `advisor_gate/tests.rs` (the test suite). Prompts moved out of code into `crates/libsy/src/prompts/advisor-gate/*.md` via `include_str!`, same as capability-classifier. I used an `advisor_gate/` submodule rather than `algorithms/util/` because these helpers are gate-specific with a single consumer — `util/` holds cross-algorithm shared code; happy to move any that become shared.

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Moved to `advisor_gate/turn.rs` in 97ec7bd4 (turn-inspection helpers live together there).

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Moved to `advisor_gate/turn.rs` in 97ec7bd4.

crates/switchyard-server/tests/server.rs
eric-liu-nvidia: These live in `tests/server.rs`, which is a Cargo integration-test target — files under `tests/` are compiled only by `cargo test`, so a `#[cfg(test)]` wrapper there is a no-op (the existing ~40 tests in this file are bare for the same reason; the `#[cfg(test)] mod tests` pattern you're pointing at applies to unit tests inside `src/`, which the libsy tests in this stack do use — `advisor_gate/tests.rs` is declared `#[cfg(test)] mod tests;`). Happy to wrap if you'd still prefer it for consistency.

crates/switchyard-server/src/stats/algorithms.rs
```diff
@@ -3,26 +3,32 @@
...
     registry: Registry,
+    advisor_gate_baseline: Option<AdvisorGateCumulative>,
     stage_router_baseline: Option<StageRouterCumulative>,
 }
 
 /// Curated algorithm-specific data included in the JSON stats response.
 #[derive(Clone, Debug, Default, PartialEq, Serialize)]
 pub(crate) struct AlgorithmStatsSnapshot {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub advisor_gate: Option<AdvisorGateStatsSnapshot>,
```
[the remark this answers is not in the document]
eric-liu-nvidia: Applied — `AlgorithmStatsSnapshot` now documents that each block is present only when the deployment runs the matching algorithm (in the stats-projection commit on #383).

crates/switchyard-server/src/stats/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,257 @@
...
+    #[test]
+    fn advisor_gate_projection_preserves_reviews_discards_and_reset_baseline() {
+        let registry = Registry::new();
+        let exporter = opentelemetry_prometheus::exporter()
+            .with_registry(registry.clone())
+            .build()
+            .unwrap_or_else(|error| panic!("failed to build metrics exporter: {error}"));
+        let provider = SdkMeterProvider::builder().with_reader(exporter).build();
+        let meter = provider.meter("switchyard");
+        let stats = StatsAccumulator::new(registry, ["advisor_gate"]);
```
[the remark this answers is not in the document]
eric-liu-nvidia: Not a real finding: `StatsAccumulator::reset` takes `&self` (interior mutability via the inner mutex — crates/switchyard-server/src/stats/accumulator.rs:140), same as the existing stage_router projection test. The test compiles and passes in CI.

crates/libsy/src/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,650 @@
...
+//! through unreviewed; the first *terminal* turn — no tool calls (or a text
+//! match under the `pattern` trigger) — is buffered and shown to a stronger
+//! advisor model together with the full transcript. `APPROVE` releases the
+//! buffered turn unchanged; `REDO` appends the discarded turn's text and the
+//! advisor's plan as feedback, then re-invokes the executor so it keeps
+//! working. Each budget scope (one benchmark evaluation, one session, or the
+//! whole instance — see [`budget_scope`]) is reviewed at most `max_reviews`
+//! times; afterwards every call is a pure passthrough.
+//!
+//! This design is a near-superset of solo executor behavior: identical until
```
sabhatinas, line 16: Do you have stats on when the algo claims executor to be "planning" ? How many turns and which turn is detected as needing advisor tool call?

crates/libsy/src/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,650 @@
...
+            advisor_max_tokens: 2048,
+            advisor_temperature: None,
+            transcript_max_chars: 200_000,
+            fail_open: true,
+        }
+    }
+}
+
+/// The trigger with its pattern compiled once at construction.
+enum CompiledTrigger {
```
sabhatinas, line 153: This should move to tool_signal.rs and can you use the existing signal buckets in addition to your patterns to detect planning mode? I have seen almost all turns using a tool call in TB 2.1 style tasks, curious what patterns you are detecting.

ting-hong-shieh:
## Prompt contract review

**[P1] Preserve the requested deliverable on REDO** — `reviewer-system-prompt.md:1-6`, `redo-feedback-prefix.md:1`, and `advisor_gate.rs:403-432`.

The reviewer prompt covers both a proposed plan and a completion claim, but every REDO tells the executor that the task is “NOT yet complete” and to “keep working until it is genuinely done.” `AdvisorGate::redo()` appends that text as a new user message before invoking the executor. For a plan-only request, this later instruction can be read as authorization to move from revising the plan to implementing it. Please make the prefix defer to the original request—for example, “Revise the plan or continue the implementation, as required by the original request”—and add a plan-only test that asserts REDO feedback does not expand the requested deliverable.

**[P2] Tell the reviewer when the transcript is truncated** — `reviewer-system-prompt.md:1` and `transcript.rs:31-62`.

The system prompt says the advisor receives “the full transcript” and “every action,” but `review_transcript()` drops the middle when the serialized messages exceed `transcript_max_chars` (200,000 by default). The marker exposes the omission in the transcript text, but the higher-priority system prompt still asserts that nothing is missing. This can make the advisor treat absent evidence as evidence that an action did not happen. Please either remove the completeness claim and state that long transcripts may omit the middle, or avoid truncating input that the prompt describes as complete. A test should cover the over-cap path and the reviewer instruction together.

linj-glitch (approved):

crates/libsy/src/algorithms/advisor_gate/transcript.rs
```diff
@@ -0,0 +1,96 @@
...
+) -> String {
+    let text = serde_json::to_string(messages).unwrap_or_default();
+    let text = middle_drop(text, cap);
+    format!(
+        "Conversation so far (JSON):\n\n{text}\n\nThe executor's latest turn (a plan, or its claim the task is done):\n{}",
+        review_tail.unwrap_or(NO_TEXT_PLACEHOLDER)
+    )
+}
+
+/// Keeps the first `cap / 4` and last `cap - cap / 4` characters of an
```
sabhatinas, line 48: what is cap here?

crates/libsy/src/algorithms/advisor_gate/transcript.rs
```diff
@@ -0,0 +1,96 @@
...
+    Redo { plan: String },
+}
+
+// ── Transcript and verdict ──────────────────────────────────────────────────
+
+/// Serializes the conversation for the advisor. The JSON body is capped with
+/// a middle drop — the head keeps the task statement, the tail keeps the
+/// recent evidence a completeness review is about — while the terminal turn
+/// is appended uncapped.
+pub(super) fn review_transcript(
```
sabhatinas, line 35: cant seem to find where `review_transcript` is called.

crates/libsy/src/algorithms/advisor_gate.rs
```diff
@@ -0,0 +1,651 @@
...
+                .await;
+        }
+
+        // Gated phase: generate the turn once, fully buffered, so the gate
+        // can inspect it before the client sees anything.
+        driver.decide(self.executor_decision()).await?;
+        let response = driver
+            .call_model(request.clone(), vec![self.executor.clone()], true)
+            .await?;
+        let turn = buffer_turn(self.executor.as_str(), response).await?;
```
sabhatinas, line 345: does this handle streaming responses?

sabhatinas:
## Algorithm structure

Here is how the pieces should fit together, for alignment before diving into specifics.

---

### 1. Signal processor (runs on every request)

A `Processor<State>` impl that reads the incoming request and writes an `AdvisorSignals` struct onto the shared `State`. It computes:

- `tool_result_count` — total tool results accumulated in the conversation
- `turn_depth` — number of assistant turns in the conversation (message-count proxy)
- `last_turn_has_tool_call` — whether the most recent assistant turn contained a tool-use block
- `last_turn_visible_text` — the visible text of the most recent assistant turn (for pattern matching)

This runs before the executor and before any trigger decision. It is stateless — it reads from the request, not from any gate-internal counter.

---

### 2. Trigger classifier (reads state, declares fire/no-fire)

A `Classifier<State>` impl that reads `state.advisor_signals` and determines whether this turn should be reviewed. It has two modes:

- **NoToolCall**: fires when `!last_turn_has_tool_call && tool_result_count >= gate_min_tool_results`
- **Pattern**: fires when `last_turn_visible_text` matches the configured regex
- **Stall**: fires when `turn_depth >= gate_stall_turns` (derived from the same signals, not a separate latch)

In the fallthrough algorithm, this classifier always returns `confidence: 1.0` for the executor — the classifier's job is purely to tell the gate whether a review is warranted, not to pick a different model.

---

### 3. Budget processor (runs after routing decision)

A `Processor<State>` impl that manages per-scope review counts. Scope resolution precedence: `proxy_x_session_id` header → session id → instance-wide.

- Reads `config.max_reviews` as the per-scope budget
- On each request: checks the scope's counter against the budget
  - If budget spent: stamps `state.advisor_budget_exhausted = true`; all subsequent turns pass through without buffering
  - If budget available: increments the counter, stamps `state.advisor_budget_reserved = true`
- Budget tracking is the only mutable shared state; it lives here, not inside the gate algorithm

---

### 4. Gate algorithm (thin orchestrator)

With signals and budget already on `State`, `route_inner` becomes a flat sequence:

```
call executor → buffer full response

if not triggered (from state.advisor_signals) OR budget exhausted (from state):
    release buffer → done

call consult():
    serialize conversation transcript (with middle-drop if over cap)
    call advisor model with reviewer system prompt
    parse verdict (APPROVE / REDO)

on APPROVE:
    release buffer → done

on REDO:
    discard buffer
    append executor turn echo + advisor plan as user message
    call executor again as live passthrough → done

on consult failure:
    release buffer → done
```

`consult()` is a self-contained function: it takes the request, the buffered turn text, and the config, and returns a verdict. It has no dependency on gate-internal mutable state.

---

### 5. Feedback injection (on REDO)

The discarded executor turn's visible text (or reasoning if no visible text, or a placeholder if neither) is echoed back as an assistant message. The advisor's plan is prepended with `redo_feedback_prefix` and appended as a user message. The executor is then re-invoked as a live passthrough — the client sees the recovery response, never the discarded draft.

---

This keeps each concern in one place: signals are extracted once per request in the processor, trigger logic is testable in isolation in the classifier, budget is a processor that can be swapped or extended, and the gate algorithm itself is a flat readable sequence with no inline state mutations.

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Measured on TB-2.1 with the Claude Code harness (89 tasks, Nemotron executor,
benchmarked knobs `max_reviews=3, gate_stall_turns=30, gate_min_tool_results=3`):

- The gate found a reviewable turn on **87/89 sessions** (REDO on 81, APPROVE on
  the rest). The 2 never-fired members were timeout grinders that never
  produced a tool-less turn — the 30-turn stall checkpoint was added for
  exactly that shape and reached 4 of the 6 known grinder tasks in the
  stall-variant arm.
- *Which* turn: the `no_tool_call` trigger fires at the executor's completion
  declaration — typically the session's final turn, so its index varies with
  task length. The stall checkpoint fires at assistant-turn 30 when nothing
  triggered earlier. `gate_min_tool_results=3` skips the 1–2 early commentary
  turns chatty executors emit before real work starts.
- One clarification: there is no advisor *tool* here and no "planning"
  classifier — `review_gate` is proxy-triggered. It keys on the shape of the
  turn the executor just generated (no `tool_use` blocks = it intends to stop),
  not on classifying intent. The tool-call strategy you may be thinking of is
  the sibling python backend, not this PR.
- Going forward this is observable per deployment: #383 exposes
  `advisor_gate.reviews` split by verdict × trigger at `/v1/stats`.

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Two different inputs are in play, which is why this doesn't fold into
`tool_signals.rs` as-is:

- `ToolSignals::from_request` reads the **request** — accumulated history.
- The gate's trigger reads the turn the executor **just generated**
  (`has_tool_use(&turn.agg)` on the buffered response). That can't be a
  request-side signal: when the executor answers without a tool call the
  harness stops, so there is no next request on which a request-side bucket
  could ever see that turn. The gate has to inspect the generated turn before
  releasing it — that's the entire interception point.

On "what patterns": the default trigger has none — it is purely the absence of
`tool_use` in the generated turn. You're right that in TB-2.1 nearly every
turn calls tools; that's exactly what makes the first tool-less turn a
high-precision stop signal (measured: 87/89 sessions produced one). The
optional `pattern` trigger exists for text-protocol harnesses (e.g. terminus
declares completion with `task_complete: true` in plain text — every turn is
tool-less there).

The two request-side guards (`count_tool_results`, `assistant_turns`) *are*
ToolSignals-shaped, and I'm happy to reuse those buckets for them in a
follow-up so the counting lives in one place. The response-side trigger has to
stay fused to the buffer-and-release step.

crates/libsy/src/algorithms/advisor_gate/transcript.rs
eric-liu-nvidia: `cap` is the route's `transcript_max_chars` — a character budget on the
serialized JSON conversation (default 200,000 chars ≈ 50k tokens at ~4
chars/token). Head keeps `cap/4` (the task statement), tail keeps the rest
(the recent evidence), with the truncation marker spliced between; the
terminal turn under review is appended uncapped. Added a doc comment saying
this at the definition (`9f858f9b`).

crates/libsy/src/algorithms/advisor_gate/transcript.rs
eric-liu-nvidia: It's called from `AdvisorGate::consult()` —
`crates/libsy/src/algorithms/advisor_gate.rs:458` on the current head — which
builds the consult request from its output. Easy to miss since the submodule
split: `transcript.rs` holds the pure serialization helpers, the only caller
is the parent module.

crates/libsy/src/algorithms/advisor_gate.rs
eric-liu-nvidia: Yes. `buffer_turn` consumes a streamed executor response to completion,
retaining the provider events (preservation included). On APPROVE the buffered
events replay to the client verbatim as a stream — covered by
`streamed_approval_replays_preserved_events_verbatim` and
`streamed_tool_call_turn_replays_without_review`; mid-stream upstream errors
propagate while buffering (`mid_stream_error_propagates_while_buffering`).
The REDO continuation and the budget-spent passthrough are served as live
streams. The one inherent cost: for a gated turn, the first client byte waits
until the turn finishes generating — review-before-release can't stream ahead
of the verdict.

eric-liu-nvidia:
@ting-hong-shieh Both points taken, both addressed:

**[P1]** Accepted — the default REDO prefix no longer asserts "the task is NOT
yet complete / keep working until done" unconditionally. It now defers to the
original deliverable: *"…continuing with the deliverable the request asked
for — revise the plan if a plan was requested, or keep working with your
tools, acting rather than replying with prose, if implementation was…"*
(`f6704cc5`). New test `redo_feedback_defers_to_the_requested_deliverable`
drives a plan-only REDO and asserts the injected feedback carries the deferral
clause and not the old completion claim.

**[P2]** The completeness claim was dropped in `b1481c4b` when the baked
prompts were replaced with the benchmark-hardened versions: the reviewer
contract now states the transcript "may be truncated in the middle", names the
exact marker, and gives a cannot-judge channel (REDO-with-restate) for
over-truncated cases — so absent middle evidence is never treated as evidence
of absence. `9f858f9b` adds the joint test you asked for:
`over_cap_consult_flags_truncation_to_the_reviewer` runs an over-cap
conversation and asserts the consult carries the marker in the transcript AND
the instruction describing it in the system prompt.

eric-liu-nvidia:
@sabhatinas Thanks for laying this out — big overlap with the current shape, one
structural point where I think the proposal breaks the mechanism, and one
where the invariants argue for keeping state in the algorithm.

**Where we already agree.** `consult()` is exactly the self-contained function
you describe: request + buffered turn + config in, verdict out, no gate state.
`route_inner` is the flat sequence in your §4 (passthrough → buffer →
trigger? → consult → APPROVE replay / REDO feedback → reinvoke). §5 matches
the implementation verbatim (echo visible text, else reasoning, else
placeholder; prefix + plan as user turn; live-stream recovery).

**Signal processor (§1–2): the trigger can't be request-side.** A
`Processor<State>` that runs "before the executor" can only see conversation
history. But the turn that matters is the one the executor is about to
produce: when it answers without a tool call, the harness stops — there is no
subsequent request on which `last_turn_has_tool_call` could fire. A
request-side trigger reviews one turn late for mid-task turns and never for
the terminal turn, which is the premature-"done" moment the gate exists to
catch. So `no_tool_call`/`pattern` must read the buffered response, fused to
the release decision. The two genuinely request-side guards
(`tool_result_count`, `turn_depth`) I'm glad to source from the existing
`ToolSignals` buckets in a follow-up so the counting is shared with the stage
router. (Also why stall is a separate latch rather than derived per-turn: a
stall consumed by a simultaneous trigger deliberately doesn't latch, so the
checkpoint can still fire later if that review gets refunded.)

**Budget processor (§3): reservation can't be split from the consult
outcome.** Two invariants couple them: failed consults *refund* the budget
(a pre-routing reservation leaks budget on advisor outages — the refund
decision is only knowable after the consult), and unparseable verdicts refund
and approve. Under concurrency, two same-scope requests race one review slot —
`concurrent_same_scope_requests_consult_once` pins reserve-or-pass atomicity
with the trigger decision; splitting reservation (processor) from consumption
(algorithm) reintroduces exactly that TOCTOU. Budget is deliberately the
gate's only mutable state, behind one lock, next to the code that refunds it.

**Proposal.** I extract an `AdvisorSignals` struct computed at the top of
`route_inner` (readability, and it reuses `ToolSignals` for the request-side
guards), keep trigger + budget fused to the buffered turn, and if the team
wants the full `Processor`/`Classifier` decomposition we do it as a follow-up
PR — so the measured behavior (k=3 Nemotron, k=2 GLM-parity runs) ships
unchanged here.

eric-liu-nvidia:
> ## Algorithm structure
> Here is how the pieces should fit together, for alignment before diving into specifics.
> 
> ### 1. Signal processor (runs on every request)
> A `Processor<State>` impl that reads the incoming request and writes an `AdvisorSignals` struct onto the shared `State`. It computes:
> 
> * `tool_result_count` — total tool results accumulated in the conversation
> * `turn_depth` — number of assistant turns in the conversation (message-count proxy)
> * `last_turn_has_tool_call` — whether the most recent assistant turn contained a tool-use block
> * `last_turn_visible_text` — the visible text of the most recent assistant turn (for pattern matching)
> 
> This runs before the executor and before any trigger decision. It is stateless — it reads from the request, not from any gate-internal counter.
> 
> ### 2. Trigger classifier (reads state, declares fire/no-fire)
> A `Classifier<State>` impl that reads `state.advisor_signals` and determines whether this turn should be reviewed. It has two modes:
> 
> * **NoToolCall**: fires when `!last_turn_has_tool_call && tool_result_count >= gate_min_tool_results`
> * **Pattern**: fires when `last_turn_visible_text` matches the configured regex
> * **Stall**: fires when `turn_depth >= gate_stall_turns` (derived from the same signals, not a separate latch)
> 
> In the fallthrough algorithm, this classifier always returns `confidence: 1.0` for the executor — the classifier's job is purely to tell the gate whether a review is warranted, not to pick a different model.
> 
> ### 3. Budget processor (runs after routing decision)
> A `Processor<State>` impl that manages per-scope review counts. Scope resolution precedence: `proxy_x_session_id` header → session id → instance-wide.
> 
> * Reads `config.max_reviews` as the per-scope budget
> * On each request: checks the scope's counter against the budget
>   
>   * If budget spent: stamps `state.advisor_budget_exhausted = true`; all subsequent turns pass through without buffering
>   * If budget available: increments the counter, stamps `state.advisor_budget_reserved = true`
> * Budget tracking is the only mutable shared state; it lives here, not inside the gate algorithm
> 
> ### 4. Gate algorithm (thin orchestrator)
> With signals and budget already on `State`, `route_inner` becomes a flat sequence:
> 
> ```
> call executor → buffer full response
> 
> if not triggered (from state.advisor_signals) OR budget exhausted (from state):
>     release buffer → done
> 
> call consult():
>     serialize conversation transcript (with middle-drop if over cap)
>     call advisor model with reviewer system prompt
>     parse verdict (APPROVE / REDO)
> 
> on APPROVE:
>     release buffer → done
> 
> on REDO:
>     discard buffer
>     append executor turn echo + advisor plan as user message
>     call executor again as live passthrough → done
> 
> on consult failure:
>     release buffer → done
> ```
> 
> `consult()` is a self-contained function: it takes the request, the buffered turn text, and the config, and returns a verdict. It has no dependency on gate-internal mutable state.
> 
> ### 5. Feedback injection (on REDO)
> The discarded executor turn's visible text (or reasoning if no visible text, or a placeholder if neither) is echoed back as an assistant message. The advisor's plan is prepended with `redo_feedback_prefix` and appended as a user message. The executor is then re-invoked as a live passthrough — the client sees the recovery response, never the discarded draft.
> 
> This keeps each concern in one place: signals are extracted once per request in the processor, trigger logic is testable in isolation in the classifier, budget is a processor that can be swapped or extended, and the gate algorithm itself is a flat readable sequence with no inline state mutations.

@sabhatinas 
I only refactored my code partially based on your feedback. The remaining piece either breaks my current advisor review gate design which provided good accuracy and cost or can be shipped as another independent new advisor review variant. Feel free to work on it. While you are working on it, make sure to not regressing my advisor review gate routing algorithm.

eric-liu-nvidia:
@sabhatinas Converting the structure discussion above into tracked follow-ups (filed by Zengyuan Li, posted via the Linear connector).

One technical addition since the last exchange that reconciles most of your proposal with the terminal-turn constraint: the Processor trait already has an Event::ModelResponse variant (crates/libsy/src/core/processor.rs). The decomposition can keep your exact shape if signal extraction is split by event — tool_result_count and turn_depth from Event::Request, and last_turn_has_tool_call and last_turn_visible_text from Event::ModelResponse on the buffered executor turn — with the trigger classifier evaluated after the executor call. That preserves "signals extracted once, trigger testable in isolation" without reviewing one turn late (or never, on the terminal turn — the premature-"done" moment the gate exists to catch).

Follow-ups filed under the Advisor Routing Strategy Linear project:

\- SWITCH-1298: the Processor/Classifier decomposition itself, with the event-split signals, a reserve/refund budget interface (refund on failed consult or unparseable verdict, the failed-consult cap, and the reserve-before-await atomicity that concurrent_same_scope_requests_consult_once pins), and a must-not-lose checklist (drop_exact_replay after REDO surgery, decide() before every executor call, discarded-turn accounting and audits) as acceptance criteria. Assigned to you per "feel free to work on it" — reassign freely. https://linear.app/nvidia/issue/SWITCH-1298

\- SWITCH-1299: source the request-side guards (tool_result_count, turn_depth) from the shared ToolSignals buckets, as offered in-thread. https://linear.app/nvidia/issue/SWITCH-1299

\- SWITCH-1300: the stall-semantics decision (latch vs derived). Derived-only re-fires every turn past the threshold, which burns budget under max_reviews > 1 and re-fires after refunded consults; if it becomes derived it needs an explicit "at most one stall-attributed review per conversation" rule. https://linear.app/nvidia/issue/SWITCH-1300

\- SWITCH-1301: redo_action escalate parity with the Python gate (f2287fe4). One design constraint it puts on the decomposition: the "classifier always returns confidence 1.0 for the executor — never picks a different model" framing must not be baked in, since escalate is exactly a model-switch outcome of the review. https://linear.app/nvidia/issue/SWITCH-1301

If there's anything else from your comment you want tracked separately, feel free to add it to the project.

PR #379, opened 2026-08-12
fix(translation): drop empty assistant messages in OpenAI Chat encoding

Pana:
## What

Use this project with moonshot kimi-k3 model, will encounter error: Invalid request: the message at position 5 with role 'assistant' must not be empty

## Why

So this PR is used to improve compatibility to kimi API

Closes #

## How tested

## Checklist

## Notes for reviewers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAI Chat request translation by omitting empty assistant messages that contain no tool calls.
  * Preserved assistant messages that contain tool calls, even when their text content is null.
  * Corrected handling of assistant turns containing only private thinking content while retaining surrounding user messages.

### DIFF
--- a/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
@@ -205,6 +205,12 @@ impl FormatCodec for OpenAiChatCodec {
         for message in &request.messages {
             messages.extend(encode_message_to_openai(message, &mut diagnostics, policy)?);
         }
+        // Some upstreams (e.g. Moonshot/Kimi) reject an assistant message whose
+        // `content` is an empty string, which normalized turns with no text and
+        // no tool calls (e.g. a replayed reasoning-only turn) would otherwise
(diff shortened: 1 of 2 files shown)


ayushag-nv:
@Pana Thanks for adding the fix. Can you please add the reproduction steps for the issue, and before the fix snapshot with error and after the fix snapshot.

Pana:
I used the moonshot platform API and key to start the switchyard server, and then switched the codex to the switchyard server's URL through the CC switch. When conversing with the same codex, I encountered the error: Invalid request: the message at position 5 with role 'assistant' must not be empty.

nachiketb-nvidia (commented):
Thanks for catching this!

The thinking-only case makes sense to drop once translation leaves the assistant turn with no usable content. However, this check is now too broad: an assistant message can have empty content while still carrying reasoning or reasoning_details, so this would drop meaningful data. Could we only remove
  messages whose serialized object contains exactly role: "assistant" and empty content, with no other payload fields?

bhuvan002:

PR #381, opened 2026-08-12
ci: Build static server binaries in CI

ebendler:
## What

This PR modifies the release action to produce and upload static binaries for the `switchyard-server` tool. It also updates all gh actions to their latest versions to ensure node24 compat for CI runners.

## Why

Static binaries are needed so that switchyard can be packaged and used in places like agent plugins without having to haul in the whole Rust SDK.

## Summary by CodeRabbit

- **New Features**
  - Added automated builds and release packaging for server binaries across Linux, macOS, and Windows.
  - Published binaries are smoke-tested, checksummed, and attached to releases alongside Python distributions.

- **Chores**
  - Updated automation tooling used for testing, documentation, packaging, performance checks, and releases.
  - Pinned cross-compilation toolchain versions for more consistent builds.
  - Updated the JSON Schema dependency to a newer compatible version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
     if: needs.changes.outputs.full_ci == 'true'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - uses: astral-sh/setup-uv@v6
         with:
           python-version: "3.12"
@@ -67,7 +67,7 @@ jobs:
     if: needs.changes.outputs.full_ci == 'true'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - name: Check SPDX headers
         run: |
           set -u
@@ -105,7 +105,7 @@ jobs:
     if: needs.changes.outputs.full_ci == 'true'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - name: Install Rust components
         run: rustup component add rustfmt clippy
       - name: cargo fmt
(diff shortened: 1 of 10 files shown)


grahamking (changes requested):
Could you explain this sentence?

> without having to haul in the whole Rust SDK.

There is no such thing as a dynamically linked Rust SDK. Here are the dynamic dependencies today:

```
$ ldd target/release/switchyard-server
        linux-vdso.so.1
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
        /lib64/ld-linux-x86-64.so.2
```

ebendler:
> Could you explain this sentence?
> 
> > without having to haul in the whole Rust SDK.
> 
> There is no such thing as a dynamically linked Rust SDK. Here are the dynamic dependencies today:
> 
> ```
> $ ldd target/release/switchyard-server
>         linux-vdso.so.1
>         libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
>         libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
>         libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
>         /lib64/ld-linux-x86-64.so.2
> ```

Correct. But there is no binary distribution of this product, so in order to ship a plugin that needs `switchyard-server` to be present, the user would have to:

install rustup -> install rust -> `cargo install switchyard-server`, which downloads a ton of crates and takes a few minutes to build.

Or we can post static musl binaries that run everywhere and get it with curl.

grahamking:
Thanks for the explanation. I understand now.

`switchyard-server` is a demo server for `libsy`. It isn't something people would include in a plugin at this stage. We're only at version 0.2.0, which is too early for us to ship binaries. Closing for now. Thanks again though.

PR #384, opened 2026-08-12
fix: preserve Codex MCP namespaces through translation

bgrins:
## What

This is a fix for Codex wrapping MCP tools into namespaces (https://github.com/openai/codex/issues/33263) in a way that makes MCP tools unavailable to other backends.

Codex sends MCP tools inside `{"type": "namespace", ...}` container. Chat-only servers accept `{"type": "function"}` and drop the container, so the model never sees any MCP tool.

The request codex now flattens the container into its child function tools, and the response path re-attaches the namespace to each returned `function_call` — `{"type": "function_call", "name": "search", "namespace": "mcp__docs__"}` — which is what Codex needs to dispatch back to the right MCP server. Covers the buffered body and Responses streaming events.

## Why

Pointing Codex at any Chat backend through Switchyard silently loses all MCP tools.

## How tested

Rust-only change

- [x] `uv run pytest tests/` green — 134 passed, 2 deselected
- [x] `cargo fmt --all --check` / `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test` green for affected crates — 236 tests
- [x] Manual smoke: Codex 0.147.0 → switchyard-server → chat-only upstream, with a live stdio MCP server. Without the fix Codex answers `NO_TOOL_AVAILABLE` and the MCP log stops at `tools/list`; with it, `tools/call` fires and the tool's token reaches the answer. Same result on Ollama (`gemma4`) and OpenRouter (`google/gemini-3.5-flash`).

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. — n/a, Rust only
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. — n/a, no Python API change
- [x] README / `--help` updated if customer-facing surface changed. — n/a, transparent to existing configs

## Notes for reviewers

I don't know where a `namespace` feature is heading in terms of provider support or clients outside of Codex, so I might be missing something. I don't see any prior reference to it in this project, though.

Be aware that colliding tool names will be flattened here (if two MCP servers expose the same tool name, neither call gets a namespace and one would disappear). This could be fixed by adding a prefix or similar, but that might change behavior if someone was prompting by the tool name, so I left that alone.

## Summary by CodeRabbit

- **New Features**
  - Added support for preserving MCP tool namespaces in OpenAI Responses requests and responses.
  - Supports namespace restoration for both buffered and streaming function-call output.
  - Added recursive handling for nested namespace tool containers.

- **Bug Fixes**
  - Prevents loss of tool namespace information during request translation.
  - Safely avoids restoring namespaces when tool names are ambiguous.

- **Tests**
  - Added coverage for namespace restoration, nested tools, and duplicate tool names.


bgrins:
Here's an e2e reproduction against local Ollama (default) or OpenRouter — both reproduce identically.

<details>
<summary><b>repro-codex-mcp-namespace.sh</b></summary>

```bash
#!/usr/bin/env bash
#
# Reproduction for openai/codex#33263.
#
# Codex sends MCP tools inside a non-standard {"type":"namespace",...} container.
# Chat-only upstreams accept only {"type":"function"} and drop it, so the model
# never sees any MCP tool.
#
# Builds a proxy (Switchyard) without and with the fix, points Codex at each
# through a chat-only upstream, and runs one prompt against a live MCP server.
#
# Expected: unfixed answers NO_TOOL_AVAILABLE and the MCP log stops at
# tools/list; fixed dispatches tools/call and returns the token A7F3-2C91.
#
# Backends: BACKEND=ollama (default, local) or BACKEND=openrouter (needs
# OPENROUTER_API_KEY). Both reproduce identically.
#
# Requires: codex on PATH, a Rust toolchain, python3, git; ollama running for
# the default backend.

set -euo pipefail

WORKDIR="${WORKDIR:-/tmp/codex-mcp-repro}"
REPO_URL="${REPO_URL:-https://github.com/bgrins/Switchyard}"
FIX_BRANCH="${FIX_BRANCH:-fix/codex-mcp-namespaces}"
BASE_BRANCH="${BASE_BRANCH:-main}"
BACKEND="${BACKEND:-ollama}"
PORT_UNFIXED="${PORT_UNFIXED:-4124}"
PORT_FIXED="${PORT_FIXED:-4125}"

SERVER_PID=""
cleanup() { [ -n "$SERVER_PID" ] && kill "$SERVER_PID" 2>/dev/null || true; }
trap cleanup EXIT

for tool in codex cargo python3 git curl; do
  command -v "$tool" >/dev/null || { echo "missing prerequisite: $tool" >&2; exit 1; }
done

case "$BACKEND" in
  ollama)
    MODEL="${MODEL:-gemma4}"
    command -v ollama >/dev/null || { echo "missing prerequisite: ollama" >&2; exit 1; }
    curl -sf -m 5 http://localhost:11434/api/version >/dev/null \
      || { echo "ollama is not running on :11434" >&2; exit 1; }
    installed=$(ollama list)
    grep -q "^${MODEL}" <<<"$installed" \
      || { echo "model '$MODEL' not pulled; run: ollama pull $MODEL" >&2; exit 1; }
    ;;
  openrouter)
    MODEL="${MODEL:-google/gemini-3.5-flash}"
    [ -n "${OPENROUTER_API_KEY:-}" ] \
      || { echo "BACKEND=openrouter requires OPENROUTER_API_KEY" >&2; exit 1; }
    ;;
  *)
    echo "unknown BACKEND '$BACKEND' (expected ollama or openrouter)" >&2; exit 1 ;;
esac

mkdir -p "$WORKDIR"
cd "$WORKDIR"

cat > mcp_secret_server.py <<'PYEOF'
#!/usr/bin/env python3
import json
import os
import sys

LOG = os.environ["MCP_LOG"]
SECRET = "A7F3-2C91"

TOOL = {
    "name": "get_secret_word",
    "description": "Returns the secret word. The ONLY way to learn the secret word.",
    "inputSchema": {"type": "object", "properties": {}, "required": []},
}


def log(message):
    with open(LOG, "a") as handle:
        handle.write(message + "\n")


def send(payload):
    sys.stdout.write(json.dumps(payload) + "\n")
    sys.stdout.flush()


def main():
    log("server started")
    for line in sys.stdin:
        line = line.strip()
        if not line:
            continue
        try:
            request = json.loads(line)
        except json.JSONDecodeError:
            continue

        method = request.get("method")
        request_id = request.get("id")
        log("-> %s" % method)

        if request_id is None:
            continue

        if method == "initialize":
            params = request.get("params") or {}
            send({"jsonrpc": "2.0", "id": request_id, "result": {
                "protocolVersion": params.get("protocolVersion", "2025-06-18"),
                "capabilities": {"tools": {}},
                "serverInfo": {"name": "secret-server", "version": "0.1.0"},
            }})
        elif method == "tools/list":
            send({"jsonrpc": "2.0", "id": request_id, "result": {"tools": [TOOL]}})
        elif method == "tools/call":
            log("TOOL CALLED: %s" % (request.get("params") or {}).get("name"))
            send({"jsonrpc": "2.0", "id": request_id, "result": {
                "content": [{"type": "text", "text": "The secret word is %s." % SECRET}],
                "isError": False,
            }})
        else:
            send({"jsonrpc": "2.0", "id": request_id,
                  "error": {"code": -32601, "message": "method not found: %s" % method}})


if __name__ == "__main__":
    main()
PYEOF

if [ "$BACKEND" = ollama ]; then
  cat > proxy.toml <<TOMLEOF
schema_version = 1

[llm_clients.upstream]
format = "openai_chat"
base_url = "http://127.0.0.1:11434/v1"

[targets.remote]
id = "$MODEL"
llm_client = "upstream"

[routes.switchyard]
id = "switchyard"
type = "random"
targets = ["remote"]
TOMLEOF
else
  cat > proxy.toml <<TOMLEOF
schema_version = 1

[llm_clients.upstream]
format = "openai_chat"
base_url = "https://openrouter.ai/api/v1"
api_key_env = "OPENROUTER_API_KEY"

[targets.remote]
id = "$MODEL"
llm_client = "upstream"

[routes.switchyard]
id = "switchyard"
type = "random"
targets = ["remote"]
TOMLEOF
fi

[ -d switchyard ] || git clone --quiet "$REPO_URL" switchyard
pushd switchyard >/dev/null
git fetch --quiet origin "$BASE_BRANCH" "$FIX_BRANCH"

echo "building unfixed proxy ($BASE_BRANCH)..."
git checkout --quiet "origin/$BASE_BRANCH"
cargo build --quiet -p switchyard-server --bin switchyard-server
cp target/debug/switchyard-server "$WORKDIR/switchyard-server-unfixed"

echo "building fixed proxy ($FIX_BRANCH)..."
git checkout --quiet "origin/$FIX_BRANCH"
cargo build --quiet -p switchyard-server --bin switchyard-server
cp target/debug/switchyard-server "$WORKDIR/switchyard-server-fixed"
popd >/dev/null

# The early exit keeps the failing case from hunting for a tool it cannot see.
PROMPT='Do you have a tool named get_secret_word? If it is NOT in your available tools, reply exactly NO_TOOL_AVAILABLE and stop immediately - do not search the filesystem, do not run any commands. If it IS available, call it once and report the word.'

run_case() {
  local binary="$1" port="$2" label="$3"

  rm -f "$WORKDIR/mcp-calls.log"
  "$WORKDIR/$binary" --config "$WORKDIR/proxy.toml" --port "$port" \
    > "$WORKDIR/server-$label.log" 2>&1 &
  SERVER_PID=$!
  sleep 3

  echo
  echo "=================== $BACKEND / $label ==================="
  MCP_LOG="$WORKDIR/mcp-calls.log" OPENAI_API_KEY=switchyard codex exec \
    -c model_provider="switchyard" \
    -c model_providers.switchyard.name="switchyard" \
    -c model_providers.switchyard.base_url="http://127.0.0.1:$port/v1" \
    -c model_providers.switchyard.wire_api="responses" \
    -c model_providers.switchyard.env_key="OPENAI_API_KEY" \
    -c model_providers.switchyard.requires_openai_auth=false \
    -c mcp_servers.secret.command="python3" \
    -c mcp_servers.secret.args="[\"$WORKDIR/mcp_secret_server.py\"]" \
    -c mcp_servers.secret.env.MCP_LOG="$WORKDIR/mcp-calls.log" \
    --dangerously-bypass-approvals-and-sandbox \
    --skip-git-repo-check \
    -m switchyard \
    "$PROMPT" 2>&1 | tail -12

  echo "---- MCP server log ----"
  cat "$WORKDIR/mcp-calls.log"

  kill "$SERVER_PID" 2>/dev/null || true
  wait "$SERVER_PID" 2>/dev/null || true
  SERVER_PID=""
}

run_case switchyard-server-unfixed "$PORT_UNFIXED" unfixed
run_case switchyard-server-fixed   "$PORT_FIXED"   fixed
```

</details>

## Output

`BACKEND=openrouter`, `google/gemini-3.5-flash`. Codex's per-run noise (session id, token counts, the repeated final message) trimmed:

```
building unfixed proxy (main)...
building fixed proxy (fix/codex-mcp-namespaces)...

=================== openrouter / unfixed ===================
user
Do you have a tool named get_secret_word? If it is NOT in your available tools, reply exactly NO_TOOL_AVAILABLE and stop immediately - do not search the filesystem, do not run any commands. If it IS available, call it once and report the word.
codex
NO_TOOL_AVAILABLE
---- MCP server log ----
server started
-> initialize
-> notifications/initialized
-> tools/list

=================== openrouter / fixed ===================
user
Do you have a tool named get_secret_word? If it is NOT in your available tools, reply exactly NO_TOOL_AVAILABLE and stop immediately - do not search the filesystem, do not run any commands. If it IS available, call it once and report the word.
codex
I have the `get_secret_word` tool available, so I will now call it to retrieve the secret word for you.
mcp: secret/get_secret_word started
mcp: secret/get_secret_word (completed)
codex
The secret word is `A7F3-2C91`.
---- MCP server log ----
server started
-> initialize
-> notifications/initialized
-> tools/list
-> tools/call
TOOL CALLED: get_secret_word
```

The unfixed run answers `NO_TOOL_AVAILABLE` and the MCP log stops at `tools/list` — the tool was never offered to the model. The fixed run dispatches `tools/call` and the token reaches the answer.

bgrins:
Fixed the coderabbit issues

nachiketb-nvidia (changes requested):
There's a mechanism by which we can add extra stuff from APIs, like MCP namespaces into the IR.

Its called `ProviderExtensions`. We should make a use of that so as to save this between translations.

Correct me if this is not the intention.


This might solve issues on the ambiguous MCP names:
mcp__a__.search
  mcp__b__.search
lead ^ to same search tool.


We can also keep all the changes within translation if we have something like 

```
pub struct ToolDefinition {
      pub name: String,
      ...
      pub provider: ProviderExtensions,
  }
```
so as to give ability for more varied tool types

thoughts?

bgrins:
> There's a mechanism by which we can add extra stuff from APIs, like MCP namespaces into the IR.

> Its called ProviderExtensions. We should make a use of that so as to save this between translations.

> Correct me if this is not the intention.

> This might solve issues on the ambiguous MCP names:
> mcp__a__.search
> mcp__b__.search
> lead ^ to same search tool.

Those are good points @nachiketb-nvidia. I have a branch that builds this as you suggest here https://github.com/NVIDIA-NeMo/Switchyard/compare/main...bgrins:experiment/provider-extensions-namespaces?expand=1. Doing some more testing with that and if it looks good then will force push that into the PR branch.

bgrins:
@nachiketb-nvidia I've implemented your suggestion as I understand it (sending namespaces through IR, recording mapping in ProviderExtensions, and then splitting it back in the translation response). It does seem better, and fixes the ambiguous case - thanks. Let me know what you think.

nachiketb-nvidia (changes requested):
Can we collapse most of this test matrix into one high-level contract test?

  The behavior we actually need to protect is end-to-end

  I think one server integration test in `crates/switchyard-server/tests/server.rs` would cover this better than the current spread of helper/unit tests.

More test is more code to maintain as well :) 


as codex says:

Rough shape:

  - Send a `/v1/responses` request with two namespace containers:
    - `mcp__a.search`
    - `mcp__b.search`
  - Include a prior `function_call` in history for `search` under `mcp__b`.
  - Include a `tool_choice` forcing `search` under `mcp__b`.
  - Assert the mocked Chat upstream sees:
    - tool definitions named `mcp__a__search` and `mcp__b__search`
    - the historical call rewritten as `mcp__b__search`
    - the forced tool choice rewritten as `mcp__b__search`
  - Have the mock upstream stream back a tool call named `mcp__b__search`.
  - Assert the Responses stream sent back to Codex has `name: "search"` and `namespace: "mcp__b"` on the output item and completed response.

nachiketb-nvidia:
Thank you for using and catching this! Pls also let us know is if the translation UX is working well.

bgrins:
I simplified the testing as you suggested to a contract test in server.py. That also caught a bug (fixed in buffered.rs) where a tool was reaching upstream without the prefix and didn't match the tools array.

I kept 3 unit tests in crates/switchyard-translation/src/codex_namespaces.rs because it has behavior that's not observable to end to end tests (eg logic to "not" do mapping in certain cases). Happy to drop them if you'd rather just have the one test.

nachiketb-nvidia (commented):
This is close enough architecturally for me as a targeted compatibility fix. I don't think we need to block this PR on a larger IR redesign for tool identity; that can come later.

  The main thing I would like cleaned up before merge is the public boundary:

  - Please avoid making `encode_aggregated_response` / `encode_stream` source-breaking for downstream users. Either keep the existing signatures as wrappers using default extensions, or add separate `*_with_extensions` helpers for this path.
  - `codex_namespaces` looks like translation internals, so I don't think it should be a public module.
  - Same for the `qualified_tool_origins` / `restore_qualified_tool_names` re-exports from `switchyard-translation`; let's keep those internal unless there is a concrete external API need.
  - There are a few dangling comments left in `codex_namespaces.rs` from the tests that were removed; please clean those up too.

  After that, I think this is fine to treat as a tactical Codex MCP compatibility fix. A more structured IR representation for tool identity can be a separate follow-up which I will take up

bgrins:
Updated in https://github.com/NVIDIA-NeMo/Switchyard/pull/384/changes/f4c7a0339e3eb9cd3a83d2c0282e411fbe7bc4e2

nachiketb-nvidia (approved):

bgrins:
Thank you for the help @nachiketb-nvidia! Can I ask, when will https://pypi.org/project/nemo-switchyard/ be updated next? Would like to pull this in for a project via pip without rebuilding the wheel.

nachiketb-nvidia:
@bgrins our next release planned to be on Sept 21, so expect it near then.

Side: what is the most enticing part of using the client/translation part of switchyard? Why not just use something like litellm for your use case? Asking out of curiousity, and maybe even guide engineering direction.

bgrins:
@nachiketb-nvidia I did try litellm proxy but hit this same error, so figured I'd try to patch Switchyard. The usecase is eval'ing a few different models across a few different harnesses for security scans. I do have one more thing I've run into (PR in #553) but it's generally been working well.

PR #385, opened 2026-08-12
fix(metrics): extend LLM latency histogram buckets

### LINKED ISSUE #262: [bug] model call latency histogram buckets top out at 10s, so p99 is silently capped
## Symptom

`switchyard_model_call_latency_ms` and `switchyard_total_latency_ms` use the OpenTelemetry SDK's default explicit bucket boundaries, whose largest finite bucket is 10000 ms. Any model call slower than 10s falls into +Inf, and `histogram_quantile` can then only return 10000. A p99 of 10s, 40s, and 5 minutes are indistinguishable on `/metrics`.

## Reproduction

```bash
./target/release/switchyard-server --config routes.toml --host 127.0.0.1 --port 4000

curl -s http://127.0.0.1:4000/v1/chat/completions \
  -d '{"model":"switchyard/split","messages":[{"role":"user","content":"say hi"}],"max_tokens":48}'

curl -s http://127.0.0.1:4000/metrics | grep model_call_latency_ms_bucket
```

Observed boundaries:

​```
le="0" le="5" le="10" le="25" le="50" le="75" le="100" le="250" le="500"
le="750" le="1000" le="2500" le="5000" le="7500" le="10000" le="+Inf"
​```

`switchyard_routing_overhead_ms` is unaffected because `crates/switchyard-server/src/metrics.rs` installs a view for it specifically (`routing_overhead_buckets`, matching on `instrument.name() != "switchyard.routing_overhead_ms"`). Every other histogram falls through to the SDK defaults.

## Expected vs. actual

- **Expected:** latency quantiles remain meaningful across the range LLM calls actually occupy.
- **Actual:** histogram_quantile(0.99, ...) clamps to exactly 10000 whenever the quantile lands in the +Inf bucket, producing a flat line at 10s that hides the real tail.

## Why it matters

The default boundaries are a general-purpose set tuned for sub-second RPC. LLM serving routinely exceeds 10s on long generations, reasoning models, and large contexts, and the tail is precisely what routing decisions are evaluated on. A p99 that cannot exceed 10s makes the latency families unusable for the comparison Switchyard exists to support: is the capable tier worth its extra latency?

Alerting is affected the same way, since a threshold above 10s can never fire.

## Environment

- Switchyard version (or commit SHA): switchyard_build_info{version="0.2.0"}
- Python version (`python --version`): Rust switchyard-server path
- OS / arch: macOS, aarch64
- Install path (`uv sync`, `pip install nemo-switchyard`, source build, etc.): source build, cargo build --release -p switchyard-server
- Inbound format (Chat Completions / Anthropic Messages / Responses): Chat Completions
- Backend (OpenAI / Anthropic / NVIDIA Inference Hub / other): NV Inference Hub

y4ho0:
## What

- add dedicated bucket boundaries for model-call and end-to-end LLM latency histograms
- preserve the OpenTelemetry SDK boundaries through 10 seconds and extend them to 15 seconds, 30 seconds, 60 seconds, 2 minutes, and 5 minutes
- verify that both Prometheus histograms expose the five-minute bucket

## Why

The OpenTelemetry SDK defaults stop at 10 seconds, so slower model calls fall into the `+Inf` bucket. That makes Prometheus quantiles unable to distinguish a 10-second response from a multi-minute LLM generation.

The dedicated view keeps the existing short-latency resolution while making the long tail observable.

Closes #262

## How tested

- [x] `cargo test -p switchyard-server` — 56 tests passed
- [x] `cargo test --workspace --exclude switchyard-py` — 490 tests and doctests passed

A full workspace run was also attempted. It reached the unchanged `switchyard-py` cdylib and stopped on local macOS CPython symbol linking; all non-PyO3 workspace tests passed.

## Checklist

- [x] Regression coverage added for the changed metrics behavior
- [x] No public API or metric label changes
- [x] Commit signed off per the DCO

## Notes for reviewers

The boundaries are fixed to keep this bug fix focused, matching the existing routing-overhead view. Making them configurable can remain a separate change if operators need deployment-specific ranges.

## Summary by CodeRabbit

* **New Features**
  * Improved latency metrics to capture model-call and end-to-end response times across a broader range, including delays up to five minutes.
  * Added minimum and maximum latency measurements for more complete performance monitoring.

* **Tests**
  * Added coverage confirming five-minute latency buckets are reported correctly.

### DIFF
--- a/crates/switchyard-server/src/metrics.rs
+++ b/crates/switchyard-server/src/metrics.rs
@@ -19,6 +19,13 @@ const ROUTING_OVERHEAD_BUCKETS_MS: &[f64] = &[
     0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 25.0, 50.0, 100.0, 250.0, 500.0, 1000.0, 2500.0, 5000.0,
 ];
 
+/// Bucket boundaries for model-call and end-to-end LLM latency histograms.
+/// Retains the SDK defaults through 10 seconds and extends them for long generations.
+const LLM_LATENCY_BUCKETS_MS: &[f64] = &[
+    0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0,
+    10_000.0, 15_000.0, 30_000.0, 60_000.0, 120_000.0, 300_000.0,
+];
+
 struct Metrics {
     registry: Registry,
     provider: SdkMeterProvider,
@@ -43,6 +50,7 @@ fn initialize() -> Result<Metrics, String> {
     let mut builder = SdkMeterProvider::builder()
         .with_reader(exporter)
         .with_view(routing_overhead_buckets)
+        .with_view(llm_latency_buckets)
         .with_resource(crate::observability::resource());
     if crate::observability::otlp_enabled("METRICS") {
         let exporter = opentelemetry_otlp::MetricExporter::builder()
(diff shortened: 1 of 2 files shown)


ayushag-nv:
@grahamking Can you review this one ?

nachiketb-nvidia (approved):
seems fine to me

grahamking (approved):

grahamking:
Thanks @y4ho0 !

PR #386, opened 2026-08-12
fix(cli): preserve Claude Code background model override

### LINKED ISSUE #258: [Bug] Launcher sends Claude Code's background requests to the agent route, running it twice per question
## Environment
- Repo: `NVIDIA-NeMo/Switchyard` @ `main`
- Claude Code 2.1.220, macOS
- Route bundle: `examples/gym_biology_agent/routes.yaml`

## Description

Claude Code makes two API calls per question: one to answer it, one to name the
session. `switchyard launch claude` points both at whatever route you pass to
`--model`. When that route can reach an agent, the naming request runs the agent
too — the title request contains the user's own text, so the classifier reads it
and routes it exactly where the real question went.

The result is that every question runs the agent twice.

## Steps to Reproduce

1. `bash examples/gym_biology_agent/run_agent_stack.sh`
2. `switchyard launch claude --model demo/biomni-agent --config examples/gym_biology_agent/routes.yaml`
3. Ask one biology question, e.g. `annotate the scFv of keytruda`
4. Read `/v1/stats` on the proxy port from the launch banner

## Actual Result

Two agent runs for one question, both successful:

```
classifier requests: 2
gym/biomni-agent: calls=2  errors=0
```

The stack log shows the title run finishing on its own:

```
BioMni completed; trace_steps=18 tokens=169774
```

It ignored "write a title" and produced a full antibody annotation — 18 steps and
169,774 tokens to name a chat.

## Expected Result

One agent run per question. Background work goes somewhere small.

## Suggested Fix

`_claude_env` sets `ANTHROPIC_SMALL_FAST_MODEL` to the same value as
`ANTHROPIC_MODEL`, and because the launcher's overrides are merged on top of the
environment, that wins even if the user sets the variable themselves. Honouring an
existing value lets background traffic be pointed elsewhere.

Worth noting for anyone fixing this: simply dropping the variable does not work.
Claude Code would fall back to its own default (`claude-haiku-4-5`), and route
tables match on exact ids with no catch-all, so that request fails with
`model_not_found`. A bundle needs a small passthrough route for background work,
and the launcher needs to stop forcing the variable so it can be selected.

ting-hong-shieh:
## What

Preserve an existing `ANTHROPIC_SMALL_FAST_MODEL` value when building the
Claude Code launcher environment. When the variable is unset, continue using
the active `--model` value as the fallback.

Add a regression test that keeps the primary agent route separate from an
explicit background route.

## Why

Claude Code uses its small/fast model setting for background requests such as
session-title generation. `_claude_env` previously assigned the launcher model
to both `ANTHROPIC_MODEL` and `ANTHROPIC_SMALL_FAST_MODEL`, overwriting any
background route configured by the user.

When the primary route can reach an agent, this sends the background request to
the same agent and can run it twice for one user question.

Leaving `ANTHROPIC_SMALL_FAST_MODEL` unset is not sufficient. Claude Code then
uses its own default model ID, which may not exist in a route table that matches
exact IDs. Preserving an explicit override lets deployments route background
traffic separately while retaining the current fallback behavior.

Closes #258

## How tested

- [x] `uv run ruff check .`
- [x] `uv run pytest -s tests/test_launchers.py -q` — 8 passed
- [x] `uv run pytest -s tests/ -v` — 135 passed, 2 skipped
- [x] `uv run switchyard launch claude --help`

## Checklist

- [x] No class or filename changes.
- [x] No new public symbols.
- [x] Regression test added for the bug fix.
- [x] README and `--help` changes are not required; the CLI surface is unchanged.
- [x] Commit signed off per the DCO.

## Notes for reviewers

The launcher still falls back to `--model` when
`ANTHROPIC_SMALL_FAST_MODEL` is not set, so existing launcher behavior remains
unchanged unless the user supplies a background-model override.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved custom `ANTHROPIC_SMALL_FAST_MODEL` environment settings when launching Claude.
  * Continued to set `ANTHROPIC_MODEL` to the selected agent model.
* **Tests**
  * Added regression coverage to verify environment overrides remain intact.

### DIFF
--- a/switchyard/cli/launchers/claude_code_launcher.py
+++ b/switchyard/cli/launchers/claude_code_launcher.py
@@ -66,8 +66,9 @@ def _claude_env(port: int, model: str) -> dict[str, str]:
     * ``ANTHROPIC_BASE_URL`` — our proxy URL.
     * ``ANTHROPIC_AUTH_TOKEN`` — opaque token; skips Console OAuth.
     * ``ANTHROPIC_API_KEY=""`` — silences the auth-conflict warning.
-    * ``ANTHROPIC_MODEL`` / ``ANTHROPIC_SMALL_FAST_MODEL`` — initial
-      active model for the session.
+    * ``ANTHROPIC_MODEL`` — initial active model for the session.
+    * ``ANTHROPIC_SMALL_FAST_MODEL`` — existing background-model override,
+      falling back to the active model.
     * ``ANTHROPIC_CUSTOM_MODEL_OPTION`` — registers ``model`` as a custom
       slot in ``/model`` so the user can come back to it after toggling
       to a builtin.
@@ -79,7 +80,7 @@ def _claude_env(port: int, model: str) -> dict[str, str]:
         "ANTHROPIC_AUTH_TOKEN": "switchyard",
         "ANTHROPIC_API_KEY": "",
         "ANTHROPIC_MODEL": model,
-        "ANTHROPIC_SMALL_FAST_MODEL": model,
+        "ANTHROPIC_SMALL_FAST_MODEL": os.environ.get("ANTHROPIC_SMALL_FAST_MODEL", model),
         "ANTHROPIC_CUSTOM_MODEL_OPTION": model,
         "CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY": "1",
     }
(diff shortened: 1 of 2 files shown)


ayushag-nv (approved):
Looks good. Approving. Thanks for the PR

PR #387, opened 2026-08-12
fix(translation): preserve reasoning order in mixed stream chunks

### LINKED ISSUE #242: [bug] Mixed reasoning/content stream chunks are reordered for Anthropic clients
## Environment

- Repo / commit: `NVIDIA-NeMo/Switchyard@3acf3d86f10fead9f434b858dbbcf01a660b8db0` (`main` on 2026-08-02)
- Python: 3.13.5
- Install path: source checkout with `uv`
- Inbound format: Anthropic Messages streaming (Claude Code)
- Backend: NVIDIA Inference Hub, OpenAI Chat Completions streaming, `nvidia/nvidia/nemotron-3-ultra-nvfp4`

## Description

When an OpenAI-compatible provider returns `reasoning_content` and final `content` in the same stream delta, Switchyard emits the visible text before the reasoning. Translating that sequence to Anthropic closes the text block and opens a new thinking block, so Claude Code displays punctuation from the end of the reasoning trace as a later `∴` section in the middle of the final answer.

This affects reasoning-capable OpenAI backends used through the Anthropic Messages endpoint. The model's non-streaming response remains correctly separated into `reasoning_content` and `content`; the corruption is introduced by stream-event ordering.

## Steps to Reproduce

1. Create an OpenAI Chat-to-Anthropic stream translation state.
2. Translate this valid provider boundary chunk, which was observed from Nemotron 3 Ultra:

```json
{
  "id": "chatcmpl-test",
  "object": "chat.completion.chunk",
  "model": "nvidia/nvidia/nemotron-3-ultra-nvfp4",
  "choices": [{
    "index": 0,
    "delta": {
      "reasoning_content": ".",
      "content": "Hello"
    },
    "finish_reason": null
  }]
}
```

3. Inspect the generated Anthropic content-block events, or run Claude Code through this route and send `Hello!`.

## Actual Result

`decode_openai_chat_stream` appends `TextDelta` before `ReasoningDelta`. The Anthropic encoder consequently produces this logical order:

```text
text_delta("Hello")
thinking_delta(".")
```

In Claude Code this can render like:

```text
Hey
∴ .
! I see...
```

## Expected Result

Reasoning must precede visible text when both fields occur in the same provider chunk:

```text
thinking_delta(".")
text_delta("Hello")
```

The translated Anthropic stream should contain one thinking block followed by one text block, without reopening thinking after visible output begins.

## Suggested Fix

Decode `reasoning_content` / `reasoning` before `content` within each OpenAI Chat stream delta. Add a regression test using a single chunk containing both fields and assert the Anthropic event order is thinking block, reasoning delta, thinking close, text block, text delta.

Blakeolson21:
## What

Preserve reasoning-before-content order when an OpenAI Chat stream chunk contains both
`reasoning_content` and visible `content`.

Add a focused OpenAI Chat to Anthropic Messages stream regression test using the one-chunk
reproducer from Issue 242. The test asserts the complete thinking block, thinking close, and text
block event sequence.

## Why

Reasoning models can return reasoning and final content in the same stream delta. Emitting visible
content first makes the Anthropic translation close the text block and open a later thinking block,
which can interleave reasoning with the final answer. We hit this while streaming reasoning models
through the translation layer.

Closes #242

## How tested

- [x] `uv run pytest tests/` green (`134 passed`, `2 deselected` with integration tests excluded)
- [ ] Manual smoke (not run; the issue's exact chunk is covered by the cross-format regression test)
- [x] `cargo test --workspace` green
- [x] Focused regression test demonstrated red before the decoder change and green after it

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. No new classes or files.
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. No new public symbols.
- [x] README / `--help` updated if customer-facing surface changed. No customer-facing API surface changed.

## Notes for reviewers

The production diff only moves existing `content` delta emission below the existing reasoning-field
loop. Parsing, field precedence, tool calls, finish reasons, and error behavior are unchanged.

The regression test asserts the exact Anthropic event order from the reported reproducer: thinking
block start, reasoning delta, signature delta, thinking block stop, text block start, and text delta.
Chunks containing only reasoning or only content retain their existing output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming responses now consistently deliver reasoning content before text when both arrive together.
  * Text extraction behavior remains unchanged.

* **Tests**
  * Added coverage to verify the correct ordering of reasoning and text content in translated streaming responses.

### DIFF
--- a/crates/switchyard-translation/tests/stream_translation.rs
+++ b/crates/switchyard-translation/tests/stream_translation.rs
@@ -5,8 +5,10 @@
 
 pub mod common;
 
+use std::collections::HashMap;
+
 use pretty_assertions::assert_eq;
-use serde_json::json;
+use serde_json::{Value, json};
 use switchyard_protocol::{LlmResponseStreamEvent, ResponseAccumulator, StopReason};
 use switchyard_translation::{
     LlmResponseChunk, StreamTranslationState, TranslationEngine, WireFormat, decode_stream_event,
@@ -15,6 +17,48 @@ use switchyard_translation::{
 use common::{REASONING_MODEL, text_and_encrypted_reasoning_details};
 
 type TestResult = std::result::Result<(), Box<dyn std::error::Error + Send + Sync>>;
+
+// Reduces Anthropic stream events to ordered labels (`<block>_start`, `<delta>`,
+// `<block>_stop`) so ordering assertions stay readable without restating each payload.
+fn event_labels(events: &[Value]) -> Vec<String> {
+    let mut block_types: HashMap<u64, String> = HashMap::new();
+    events
+        .iter()
+        .map(|event| {
+            let index = event
+                .get("index")
+                .and_then(Value::as_u64)
+                .unwrap_or_default();
+            match event
+                .get("type")
+                .and_then(Value::as_str)
+                .unwrap_or_default()
+            {
+                "content_block_start" => {
+                    let block_type = event
+                        .get("content_block")
+                        .and_then(|block| block.get("type"))
+                        .and_then(Value::as_str)
+                        .unwrap_or_default()
+                        .to_string();
+                    block_types.insert(index, block_type.clone());
+                    format!("{block_type}_start")
+                }
+                "content_block_delta" => event
+                    .get("delta")
+                    .and_then(|delta| delta.get("type"))
+                    .and_then(Value::as_str)
+                    .unwrap_or_default()
+                    .to_string(),
+                "content_block_stop" => {
+                    let block_type = block_types.get(&index).cloned().unwrap_or_default();
+                    format!("{block_type}_stop")
+                }
+                other => other.to_string(),
+            }
+        })
+        .collect()
+}
 
 // Same-format replay returns the same parsed JSON value, including provider-specific fields.
 #[test]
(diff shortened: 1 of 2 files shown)


Blakeolson21:
Thanks for the review. `25da6fe` removes the redundant test doc comment, and `f9cb482` merges current `main` without conflicts. The full local Rust and Python gates are green.

crates/switchyard-translation/tests/stream_translation.rs
```diff
@@ -303,6 +303,84 @@ fn openai_chat_stream_event_translates_to_anthropic_message_events() -> TestResu
...
+
+    let events = engine.translate_event(
+        &mut state,
+        WireFormat::OpenAiChat,
+        WireFormat::AnthropicMessages,
+        &chunk,
+    )?;
+
+    assert_eq!(
+        events,
```
nachiketb-nvidia: we're sort of duplicating test behaviour here, more test code is also more code to maintain.

Lets assert labels like

```
assert_eq!(
      event_labels(&events),
      vec![
          "message_start",
          "thinking_start",
          "thinking_delta",
          "signature_delta",
          "thinking_stop",
          "text_start",
          "text_delta",
      ]
  );
```

nachiketb-nvidia (changes requested):
just one nit, otherwise looks good to me!

Also needs a rebase :)

crates/switchyard-translation/tests/stream_translation.rs
Blakeolson21: Done. Added an event_labels helper that reduces the stream to ordered labels and asserted the label sequence, so the test verifies reasoning-first ordering without restating each event payload. Also rebased onto the latest main.

nachiketb-nvidia (approved):
lgtm! needs a conflict resolution/rebase off main

then I'll run the workflow

Thank you!

PR #389, opened 2026-08-12
fix(translation): preserve multimodal tool results

### LINKED ISSUE #380: [bug] image and document blocks in tool_result are serialized into a text string when translating to OpenAI Chat
## Symptom

An Anthropic `tool_result` whose content is a list of blocks (for example, a text block plus an image block) is translated to an OpenAI Chat `tool` message as a single string, with every non-text block dumped as raw JSON. The downstream model therefore cannot consume the image or document as multimodal input. No error is raised and the request can still return 200.

Anthropic explicitly allows `text`, `image`, `document`, and `search_result` blocks inside `tool_result.content`. However, OpenAI Chat tool messages are text-only: the official generated `ChatCompletionToolMessageParam` schema accepts a string or an array of text parts, not `image_url` or file parts:

https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_tool_message_param.py

Therefore, emitting an `image_url` directly inside `role: "tool"` would not be a valid OpenAI Chat representation. The bug is that Switchyard silently destroys the structured modality instead of performing a valid lowering or reporting the lossy/unsupported conversion.

## Reproduction

Inbound format: Anthropic Messages. Route: passthrough to an `openai_chat` target. A small mock backend is used only to record the exact bytes Switchyard forwards.

`routes.toml`:

```toml
schema_version = 1
[llm_clients.local]
format = "openai_chat"
base_url = "http://127.0.0.1:9999/v1"
authorization = "Bearer x"
[targets.cap]
id = "captured-model"
llm_client = "local"
[routes.primary]
id = "captured-model"
type = "passthrough"
target = "cap"
```

```bash
switchyard-server --config routes.toml --port 9000

curl -s http://localhost:9000/v1/messages \
  -d '{
    "model": "captured-model",
    "max_tokens": 100,
    "tools": [{"name":"screenshot","description":"take a screenshot","input_schema":{"type":"object","properties":{"t":{"type":"string"}},"required":["t"]}}],
    "messages": [
      {"role":"user","content":"screenshot the page"},
      {"role":"assistant","content":[{"type":"tool_use","id":"toolu_1","name":"screenshot","input":{"t":"page"}}]},
      {"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_1","content":[
        {"type":"text","text":"here it is:"},
        {"type":"image","source":{"type":"base64","media_type":"image/png","data":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="}}
      ]}]}
    ]
  }'
```

## Expected vs. actual

- **Expected:** Switchyard should preserve the modality using a representation valid for OpenAI Chat—for example, emit the textual tool result as the required `tool` message and lower the image/document into a subsequent multimodal `user` message, while preserving ordering for parallel tool calls. If a faithful lowering is not supported, the conversion should be rejected or produce a clear lossy-conversion/unsupported-modality diagnostic.
- **Actual:** the forwarded `tool` message contains one plain string with the image block serialized as JSON. There is no multimodal content part and no diagnostic.

```
role: "tool"
content: "here it is: {\"source\":{\"data\":\"iVBORw0KGgoAAA...==\",\"media_type\":\"image/png\",\"type\":\"base64\"},\"type\":\"image\"}"
```

A `document` block inside `tool_result` behaves the same way: it is serialized into the text string rather than preserved or explicitly rejected.

## Impact

This affects real agent workflows in which tools return images or documents, including browser automation, screenshot tools, visual inspection, OCR, and document readers. The tool appears to succeed, but the downstream vision-capable model is blind to the result.

For realistic screenshots or documents, serializing base64 into literal text can also consume a large number of tokens or cause a context-limit failure, in addition to losing the modality.

## Environment

- Switchyard version (or commit SHA): 2bef154 (2bef154970d23cacf9c83b4fe9c1cd90212623e8), release build from source
- Python version: 3.9.6
- OS / arch: macOS arm64
- Install path: source build (`cargo build --release -p switchyard-server`)
- Inbound format: Anthropic Messages
- Backend: OpenAI-compatible (a local mock was used to capture the forwarded request)

## Additional context

Anthropic documents multimodal tool results here:

https://platform.claude.com/docs/en/agents-and-tools/tool-use/handle-tool-calls

Streaming tool-call translation looked correct in the same setup; this report concerns non-text blocks nested inside buffered `tool_result` content.

Possibly related to #152 (filtering unsupported input modalities per target), though here the block is neither filtered nor rejected—it is silently stringified.

ting-hong-shieh:
## What changed

- Preserve `text`, `image`, and `document` blocks nested in Anthropic `tool_result.content` when decoding into the neutral IR.
- Keep textual output in the required OpenAI Chat `tool` messages and lower non-text output into a following multimodal `user` message.
- Emit all parallel tool-result messages before the lowered multimodal content so their ordering remains valid.
- Report the role lowering through the existing lossy-conversion policy; strict policy rejects it.

## Why

The Anthropic decoder previously flattened every non-text block inside `tool_result.content` into JSON text. By the time the OpenAI Chat encoder ran, the image or document structure had already been lost, so the downstream model received base64 or metadata as ordinary text.

OpenAI Chat tool messages accept text content but not image or file parts. User messages accept those multimodal parts, so the translation now splits the content at that protocol boundary instead of silently stringifying it.

Closes #380.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace`

Regression coverage includes base64 images, base64 documents, parallel tool-result ordering, diagnostics, and strict rejection policy.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of document content, including file IDs, base64 data, and optional filenames.
  * Preserved text, images, documents, and other content types in tool results instead of flattening them into text.
  * Improved conversion of multimodal tool results to OpenAI Chat messages while preserving content order.

* **Bug Fixes**
  * Added diagnostics for conversions that move non-text content or may lose information.
  * Added an option to reject lossy conversions with a clear error.

### DIFF
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -568,6 +568,9 @@ fn decode_anthropic_content_block(
         Some("input_file") | Some("file") => vec![ContentBlock::File {
             source: decode_file_source(block),
         }],
+        Some("document") => vec![ContentBlock::File {
+            source: decode_anthropic_file_source(block),
+        }],
         _ => vec![ContentBlock::Unknown {
             provider: WireFormat::AnthropicMessages.into(),
             raw: Value::Object(block.clone()),
@@ -575,30 +578,40 @@ fn decode_anthropic_content_block(
     })
 }
 
-// Converts Anthropic tool-result content into text-like IR blocks.
+// Preserves supported Anthropic tool-result blocks in the neutral IR.
 fn decode_tool_result_content(value: &Value) -> Vec<ContentBlock> {
     match value {
         Value::String(text) => vec![ContentBlock::Text { text: text.clone() }],
         Value::Array(blocks) => {
-            let mut text = Vec::new();
+            let mut content = Vec::new();
             for block in blocks {
                 if let Some(block) = block.as_object() {
-                    if block.get("type").and_then(Value::as_str) == Some("text") {
-                        text.push(
-                            block
+                    match block.get("type").and_then(Value::as_str) {
+                        Some("text") => content.push(ContentBlock::Text {
+                            text: block
                                 .get("text")
                                 .and_then(Value::as_str)
                                 .unwrap_or_default()
                                 .to_string(),
-                        );
-                    } else {
-                        text.push(json_string(&Value::Object(block.clone())));
+                        }),
+                        Some("image") => content.push(ContentBlock::Image {
+                            source: block
+                                .get("source")
+                                .cloned()
+                                .map(ImageSource::Raw)
+                                .unwrap_or_else(|| ImageSource::Raw(Value::Object(block.clone()))),
+                        }),
+                        Some("document") => content.push(ContentBlock::File {
+                            source: decode_anthropic_file_source(block),
+                        }),
+                        _ => content.push(ContentBlock::Unknown {
+                            provider: WireFormat::AnthropicMessages.into(),
+                            raw: Value::Object(block.clone()),
+                        }),
                     }
                 }
             }
-            vec![ContentBlock::Text {
-                text: text.join(" "),
-            }]
+            content
         }
         Value::Null => vec![ContentBlock::Text {
             text: String::new(),
@@ -607,6 +620,31 @@ fn decode_tool_result_content(value: &Value) -> Vec<ContentBlock> {
             text: json_string(other),
         }],
     }
+}
+
+// Decodes Anthropic document sources into normalized file sources.
+fn decode_anthropic_file_source(block: &Map<String, Value>) -> FileSource {
+    let Some(source) = block.get("source").and_then(Value::as_object) else {
+        return FileSource::Raw(Value::Object(block.clone()));
+    };
+    if source.get("type").and_then(Value::as_str) == Some("file")
+        && let Some(file_id) = source.get("file_id").and_then(Value::as_str)
+    {
+        return FileSource::FileId(file_id.to_string());
+    }
+    if source.get("type").and_then(Value::as_str) == Some("base64")
+        && let Some(data) = source.get("data").and_then(Value::as_str)
+    {
+        return FileSource::FileData {
+            data: data.to_string(),
+            filename: block
+                .get("title")
+                .or_else(|| source.get("filename"))
+                .and_then(Value::as_str)
+                .map(ToOwned::to_owned),
+        };
+    }
+    FileSource::Raw(Value::Object(source.clone()))
 }
 
 // Decodes Anthropic tool definitions into normalized tool definitions.
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -554,14 +554,9 @@ fn decode_anthropic_content_block(
             content: decode_tool_result_content(block.get("content").unwrap_or(&Value::Null)),
             is_error: block.get("is_error").and_then(Value::as_bool),
         })],
-        Some("image") => {
-            let source = block
-                .get("source")
-                .cloned()
-                .map(ImageSource::Raw)
-                .unwrap_or_else(|| ImageSource::Raw(Value::Object(block.clone())));
-            vec![ContentBlock::Image { source }]
-        }
+        Some("image") => vec![ContentBlock::Image {
+            source: ImageSource::Raw(Value::Object(block.clone())),
+        }],
         Some("input_image") | Some("image_url") => decode_image_source(block)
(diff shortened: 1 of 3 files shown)


crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -609,6 +622,31 @@ fn decode_tool_result_content(value: &Value) -> Vec<ContentBlock> {
...
+            data: data.to_string(),
+            filename: block
+                .get("title")
+                .or_else(|| source.get("filename"))
+                .and_then(Value::as_str)
+                .map(ToOwned::to_owned),
+        };
+    }
+    FileSource::Raw(Value::Object(source.clone()))
+}
```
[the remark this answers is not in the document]
ting-hong-shieh: Fixed in bd4aea4d. The fallback now preserves the complete document block, and the regression test disables preservation to exercise the Anthropic decode → IR → encode path directly. `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p switchyard-translation` pass locally.

ayushag-nv:
@ting-hong-shieh Thanks for putting this PR. Can you please attach a snapshot of the output or trace before this fix and after the fix. Will be easy to understand and review with that. Thanks

ting-hong-shieh:
Thanks, @ayushag-nv. I ran the same buffered Anthropic → OpenAI Chat translation at the PR base (`48b3b71d`) and the current head (`bd4aea4d`). No provider call was involved.

Input `tool_result.content`:

```json
[
  {"type": "text", "text": "here it is:"},
  {
    "type": "image",
    "source": {
      "type": "base64",
      "media_type": "image/png",
      "data": "iVBORw0KGgo="
    }
  }
]
```

Before (`48b3b71d`):

```json
{
  "messages": [
    {
      "role": "tool",
      "tool_call_id": "toolu_1",
      "content": "here it is: {\"source\":{\"data\":\"iVBORw0KGgo=\",\"media_type\":\"image/png\",\"type\":\"base64\"},\"type\":\"image\"}"
    }
  ],
  "diagnostics": []
}
```

The image block is serialized into ordinary tool-message text.

After (`bd4aea4d`):

```json
{
  "messages": [
    {
      "role": "tool",
      "tool_call_id": "toolu_1",
      "content": "here it is:"
    },
    {
      "role": "user",
      "content": [
        {
          "type": "image_url",
          "image_url": {
            "url": "data:image/png;base64,iVBORw0KGgo="
          }
        }
      ]
    }
  ],
  "diagnostics": [
    {
      "code": "lossy_conversion",
      "severity": "Warning",
      "message": "OpenAI Chat tool messages only support text; non-text tool-result content was moved to a user message",
      "source": "anthropic_messages",
      "target": "openai_chat"
    }
  ]
}
```

Text now remains in the tool message, while the image is preserved as multimodal content in the following user message. The role lowering is also reported through the existing lossy-conversion diagnostic.

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -609,6 +622,31 @@ fn decode_tool_result_content(value: &Value) -> Vec<ContentBlock> {
...
+    };
+    if source.get("type").and_then(Value::as_str) == Some("file")
+        && let Some(file_id) = source.get("file_id").and_then(Value::as_str)
+    {
+        return FileSource::FileId(file_id.to_string());
+    }
+    if source.get("type").and_then(Value::as_str) == Some("base64")
+        && let Some(data) = source.get("data").and_then(Value::as_str)
+    {
+        return FileSource::FileData {
```
ayushag-nv: This normalization loses `source.media_type` and moves the document’s top-level `title` into `source.filename`. With preservation disabled, a valid base64 Anthropic document therefore re-encodes without the required `media_type` and with `title` in the wrong place, which can make the upstream request invalid. Could we retain the Anthropic fields here (or preserve the raw block) and add a round-trip test for a known base64 document?

ting-hong-shieh: Addressed in `d7fcd6d`. Anthropic image and document blocks now keep their complete provider wrappers in the IR, so preservation-disabled round trips retain `source.media_type`, top-level `title`, and the remaining document fields. OpenAI Chat conversion maps only portable base64 data plus `title`; Anthropic-managed `file_id` values use the existing lossy/reject path instead of being forwarded as OpenAI IDs.

Added round-trip and cross-provider regression coverage. `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` pass locally.

ayushag-nv (approved):
lgtm

PR #390, opened 2026-08-12
fix(libsy): fall back on efficient stream transport errors

### LINKED ISSUE #388: [bug] Escalation router returns HTTP 500 instead of falling through on efficient stream transport error
## Symptom

An escalation-mode `llm_classifier` route returns HTTP 500 when the efficient target's streaming response body terminates with a transport error while `into_agg()` is buffering it for the judge. The route does not fall through to the capable target, so a transient weak-backend transport failure becomes an end-to-end agent failure.

## Reproduction

Configure an escalation classifier with an OpenAI-compatible streaming efficient target and a capable fallback:

```toml
[targets.efficient]
id = "efficient-model"
llm_client = "efficient_backend"

[targets.capable]
id = "capable-model"
llm_client = "capable_backend"

[targets.judge]
id = "judge-model"
llm_client = "judge_backend"

[routes.router]
id = "router-model"
type = "llm_classifier"
mode = "escalation"
classifier_target = "judge"
strong_target = "capable"
weak_target = "efficient"
max_output_tokens = 4096
escalation = { confirmations = 2, recent_turn_window = 28, window_message_chars = 500 }
```

Start the server and send a streaming Chat Completions request:

```bash
switchyard-server --config profiles.toml --host 0.0.0.0 --port 4143

curl -N http://localhost:4143/v1/chat/completions \
  -d '{
    "model":"router-model",
    "stream":true,
    "messages":[{"role":"user","content":"Create a Python script called weather.py that fetches weather data for San Francisco using the wttr.in API and prints a summary."}]
  }'
```

Have the efficient backend return a streaming response whose HTTP body is interrupted before the SSE stream completes (for example, an upstream gateway closes an idle/incomplete body). In our PinchBench reproduction, the first efficient turn completed and its tool executed successfully. The next efficient turn's response body was terminated upstream after about 30 seconds.

This reproduced in three independent task trials. Hermes retried each failed request three times; the nine failed attempts consistently ended after approximately 30–31 seconds.

## Expected vs. actual

- **Expected:** A transport failure while consuming/aggregating the efficient response should be recoverable by the escalation route—ideally by falling through to the capable target, analogous to the existing `ContextWindowExceeded` fallback. At minimum, the transport error should remain a structured client-call error so routing/retry policy can handle it rather than becoming an opaque algorithm error.
- **Actual:** `EscalationClassifier::score()` successfully receives `Response`, then `efficient_response.llm_response.into_agg()` encounters the body-stream transport error. The error is converted directly to `LibsyError::AlgorithmError`, the route returns HTTP 500, and the capable target is never called.

```text
WARN ... algorithm="llm_task_classifier" outcome="error"
  error=failed to aggregate efficient response: upstream transport error: error decoding response body

ERROR ... LLM request failed wire_format=openai_chat status=500
  requested_model="router-model"
  streaming=true
  handling_duration_ms=30469.808085
  error="failed to aggregate efficient response: upstream transport error: error decoding response body"
```

The relevant path on current `main` is:

```rust
let agg = efficient_response
    .llm_response
    .into_agg()
    .await
    .map_err(|e| LibsyError::AlgorithmError {
        message: format!("failed to aggregate efficient response: {e}"),
    })?;
```

The initial call path only special-cases `LlmClientError::ContextWindowExceeded`; transport errors raised later while consuming the stream bypass that fallback entirely.

## Environment

- Switchyard version (or commit SHA): `switchyard-server 0.2.0`; observed on `2d47895f4d50928d89113a86320085ac42f4c2c5`; the same aggregation/error-mapping path is present on `main` at `48b3b71d3cc629aa9eb011852f5a7da90957ba22`
- Python version (`python --version`): Python 3.12.3
- OS / arch: Ubuntu 24.04, Linux 6.17.0, x86_64
- Install path (`uv sync`, `pip install nemo-switchyard`, source build, etc.): source release build (`target/release/switchyard-server`)
- Inbound format (Chat Completions / Anthropic Messages / Responses): OpenAI Chat Completions, streaming
- Backend (OpenAI / Anthropic / NVIDIA Inference Hub / other): OpenAI-compatible NVIDIA-hosted efficient backend; NVIDIA Inference Hub for capable and judge targets

## Additional context

This is distinct from an agent timeout and from the task's external API call. The external tool call completed successfully. The failure occurred on the subsequent LLM response stream while Switchyard buffered the efficient answer for classification.

In a 21-task evaluation, the same infrastructure failure produced a zero on this task in all three repetitions, lowering the router mean by 4.76 percentage points. Fixed-model runs completed the task successfully. Increasing the outer agent timeout does not address the issue because the upstream HTTP body itself is being terminated.

A regression test could provide an efficient `LlmResponse::Stream` that emits one or more chunks and then yields `LlmClientError::Transport`, assert that the capable target is called, and assert that the client receives the capable response rather than HTTP 500.

ting-hong-shieh:
## What

- Fall back to the capable target when the efficient response stream raises a transport error during aggregation.
- Preserve other aggregation failures as typed client-call errors.
- Add regression tests for the transport fallback and the non-transport error boundary.

## Why

The efficient client call completes as soon as it returns a stream handle. A transport error can surface later, while `into_agg()` consumes that stream for the escalation judge. That error was converted to `AlgorithmError`, so the route returned HTTP 500 without calling the capable target.

This change treats only a mid-stream transport failure as recoverable. Other aggregation errors keep their original client error type instead of silently changing targets.

Closes #388

## How tested

## Checklist

- [x] Unit tests cover the bug fix and its non-transport boundary.
- [x] No public API or customer-facing documentation changed.
- [x] Commit includes the required DCO sign-off.

## Notes for reviewers

The fallback is intentionally limited to `LlmClientError::Transport`; invalid or malformed responses remain structured client-call failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved escalation routing reliability by falling back to a capable target when streamed responses encounter transport failures.
  * Preserved specific client errors for other response aggregation failures.

### DIFF
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -531,13 +531,17 @@ impl Classifier<State> for EscalationClassifier {
             }) => return Ok((decisive(&self.capable), None)),
             Err(e) => return Err(e),
         };
-        let agg = efficient_response
-            .llm_response
-            .into_agg()
-            .await
-            .map_err(|e| LibsyError::AlgorithmError {
-                message: format!("failed to aggregate efficient response: {e}"),
-            })?;
+        // The call resolves when its stream handle arrives; transport can still fail while
+        // buffering. Fall back only for that availability failure and keep other errors typed.
+        let agg = match efficient_response.llm_response.into_agg().await {
+            Ok(agg) => agg,
+            Err(LlmClientError::Transport { .. }) => {
+                return Ok((decisive(&self.capable), None));
+            }
+            Err(source) => {
+                return Err(LibsyError::client_call(self.efficient.clone(), source));
+            }
+        };
         // Append the efficient reply so the judge reads this turn's completed trajectory.
         let mut judge_request = request.clone();
         judge_request
@@ -938,8 +942,8 @@ mod tests {
 
     use super::*;
     use switchyard_protocol::{
-        ContentBlock, InstructionBlock, LlmClientError, LlmRequest, Metadata, ToolCall, ToolResult,
-        completion_text, text_request, text_response,
+        ContentBlock, InstructionBlock, LlmClientError, LlmRequest, LlmResponseChunk, Metadata,
+        ToolCall, ToolResult, completion_text, text_request, text_response,
     };
 
     use crate::algorithms::util::llm_judge::Judge;
@@ -1771,6 +1775,21 @@ mod tests {
         }
     }
 
+    /// Returns a stream that emits partial content before failing during aggregation.
+    fn streamed_then_error(error: LlmClientError) -> Response {
+        Response {
+            llm_response: LlmResponse::Stream(Box::pin(futures::stream::iter([
+                Ok(LlmResponseChunk::TextDelta {
+                    index: 0,
+                    text: "partial".to_string(),
+                }
+                .into()),
+                Err(error),
+            ]))),
+            metadata: None,
+        }
+    }
+
     /// Builds a router with escalation enabled (`confirmations=1` latches on the first verdict).
     fn escalation_router() -> Result<Arc<LlmTaskClassifier>> {
         Ok(Arc::new(LlmTaskClassifier::new(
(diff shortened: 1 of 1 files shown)


ayushag-nv (approved):
Nice Catch. Thanks for the contribution. Approving

PR #391, opened 2026-08-12
build(python): support Python 3.10

### LINKED ISSUE #366: Proposal: lower the wheel's abi3 floor to cp310
## Proposal

`crates/switchyard-py` currently builds with pyo3's `abi3-py312`, so the published wheel is `cp312-abi3` and `requires-python = ">=3.12"`. Lowering the feature to `abi3-py310` produces a single `cp310-abi3` wheel that serves CPython 3.10 through 3.14+ — same binary, no per-version builds.

## Why

Many production hosts that would embed Switchyard as a routing library (rather than run the proxy) are still on 3.10/3.11 — long-lived backends where the Python upgrade is a bigger project than the routing integration. The Rust core compiles fine against the 3.10 stable ABI (we verified: only the pyo3 feature line changes; the crate uses no 3.12-only C-API surface).

The pure-Python layer needs only two mechanical accommodations, both verified on 3.10:

- `datetime.UTC` → `timezone.utc` (3 call sites)
- `from typing import Self` → `typing_extensions.Self` fallback (5 call sites, marker dep `typing-extensions>=4; python_version < '3.11'`)

## Trade-offs

- Loses 3.12-only stdlib/typing conveniences in new code unless guarded.
- CI matrix ideally adds a 3.10 leg for the binding tests.

We're running this configuration in production at Gumloop (fork: gumloop/Switchyard) and are happy to send the PR if the floor change is acceptable in principle.

nachiketb-nvidia:
## What

Lower the supported Python floor for `nemo-switchyard` and its native bindings from 3.12 to 3.10.

## Why

This makes Switchyard usable by Python 3.10 and 3.11 consumers while retaining one abi3 wheel per platform. Closes #366.

## How

- set `requires-python` and Python tooling targets to 3.10
- build the PyO3 extension with `abi3-py310`
- test Python 3.10 through 3.14 in CI and release workflows
- build and smoke-test the minimum-compatible wheel with CPython 3.10
- update installation guidance and regenerate `uv.lock`
- keep the experimental Harbor dependency gated to Python 3.12+

## Validation

- `uv run pytest tests/ -v -m "not integration"` (`134 passed`, `2 deselected`)
- `uv run ruff check .`
- `uv run mypy switchyard switchyard_rust`
- `cargo fmt --all --check`
- `cargo test --workspace`
- built `nemo_switchyard-0.2.0-cp310-abi3-manylinux_2_34_x86_64.whl`
- verified wheel metadata reports `Requires-Python: >=3.10` and imports on CPython 3.10 and 3.12

## Summary by CodeRabbit

- **New Features**
  - Added support for Python 3.10 and newer, including Python 3.10 and 3.11 compatibility coverage.
  - Updated package compatibility to support the Python 3.10 ABI.

- **Documentation**
  - Updated installation instructions, requirements, and release guidance to reference Python 3.10+.

- **Tests**
  - Expanded validation across Python 3.10–3.14, including package builds, installation checks, and CLI smoke tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
       - uses: actions/checkout@v4
       - uses: astral-sh/setup-uv@v6
@@ -179,7 +179,7 @@ jobs:
       - name: Verify slim install (no extras) succeeds
         working-directory: /tmp
         run: |
-          uv run --isolated --no-project --python 3.12 \
+          uv run --isolated --no-project --python 3.10 \
             --with "${SWITCHYARD_DEFAULT_PACKAGE}" \
             python -c "import switchyard; print('import OK, version:', switchyard.__version__)"
           uv run --isolated --no-project --python 3.14 \
@@ -188,7 +188,7 @@ jobs:
       - name: Assert heavy packages absent from slim install
         working-directory: /tmp
         run: |
-          uv run --isolated --no-project --python 3.12 \
+          uv run --isolated --no-project --python 3.10 \
             --with "${SWITCHYARD_DEFAULT_PACKAGE}" \
             python - <<'PY'
           import importlib.util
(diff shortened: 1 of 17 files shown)


bhuvan002 (approved):
Benchmarking scripts LGTM!

ayushag-nv (approved):

PR #393, opened 2026-08-12
fix: Stamp the selected ModelId on LlmRequest much earlier

grahamking:
Previously `libsy` would give `CallModel` a request with the incoming model
ID (e.g. `switchyard/random`), and a `Decision` with the correct id (e.g. `openai/gpt-oss-120b`)
as chosen by the routing algorithm. `libsy-llm-client` would stamp the later onto 
the former before making the actual call.

Now `Driver::call_model` stamps the correct model ID onto `LlmRequest`,
so everything after that can use `LlmRequest` directly.

This was confusing me for a while. Seems easier to understand like this.

This also means `CallModel` can read the `selected_model_id()` off the
request now, which gives us options for changing `Decision` later.

The `RoutedLlmClient::call` method took a `Decision` purely to extract it's 
model_id and stamp it on the request, so we also simplify that to `call(Request)`.

Claude's review:
 > The net effect is simpler client code and a cleaner contract on CallModel.

Assisted-by: Codex:GPT 5.6 Sol high

## Summary by CodeRabbit

* **Improvements**
  * Requests now consistently use the model selected by routing.
  * Random-routing behavior now preserves the selected model in outgoing client requests.
  * Model selection and request handling are more consistent across token counting, rewrites, and standard calls.

* **Documentation**
  * Clarified how selected models are represented and propagated through requests and routed clients.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -13,7 +13,7 @@ use reqwest::RequestBuilder;
 use reqwest::header::{HeaderMap, RETRY_AFTER};
 use serde_json::{Map, Value};
 use switchyard_protocol::{
-    Decision, LlmRequest, LlmResponse, Metadata, ModelId, Request, Response, RoutedLlmClient,
+    LlmRequest, LlmResponse, Metadata, ModelId, Request, Response, RoutedLlmClient,
 };
 use switchyard_translation::{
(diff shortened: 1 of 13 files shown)


grahamking:
@coderabbit review

ayushag-nv (approved):
Nice Catch. Looks good.

PR #394, opened 2026-08-13
fix(libsy): warn when the scorer cannot reach a picker's non-default tier

### LINKED ISSUE #264: [Bug] stage_router: capable_first can't offload at the default confidence_threshold=0.5
When benchmarking stage routing on TB-Lite with a frontier strong tier and a local small model weak tier, I reached for `capable_first` and left the threshold at its default.

The runs came back perfectly healthy: every request 200, no errors. But nothing ever went to the weak tier. Not a single turn.

After digging into the scorer I think this is a property of the shipped defaults rather than anything specific to our setup, so it seemed worth writing up in case others hit the same thing.

## The short version

`picker: capable_first` (`switchyard/lib/profiles/stage_router_config.py:76`) and `confidence_threshold: 0.5` (`:81`) are both defaults, and in combination the scorer can never select the efficient tier. You get there by omitting both knobs, which we suspect is the common path.

## Where it comes from

In `crates/libsy/src/algorithms/util/stage.rs` (`SIGNAL_UNIT=0.10`, `SCORE_GAIN=5.0`, `HARD_SEVERITY=0.7`):

```rust
let raw = SIGNAL_UNIT
let score = (SCORE_GAIN * raw).tanh();
// confidence = score.abs();  gate is `if scored.confidence >= confidence_threshold`  (stage.rs:319)
```

Each maxed signal contributes one unit, so with `k` agreeing signals the confidence lands on a discrete ladder:

```
confidence = tanh(SCORE_GAIN × SIGNAL_UNIT × k) = tanh(0.5k)

k=1 → 0.4621     k=2 → 0.7616     k=3 → 0.9051
```

`0.5` happens to sit in the gap between the k=1 and k=2 rungs.

That's fine in the *escalation* direction, because three terms push toward capable (`severity`, `spinning`, `exploring`) and any two of them reach k=2 = 0.7616. It's the de-escalation direction where it gets stuck: only **one** term is negative — `production_intensity` — and since it comes from `ratio(...)` it's bounded to `[0, 1]`. There's no second negative signal to stack with it, so the efficient side can never climb past the first rung:

```
max |confidence_negative| = |tanh(5.0 × 0.10 × (−1))| = tanh(0.5) = 0.4621171573
```

And `capable_first` only leaves its default tier when the score is confidently negative. So the 3-vs-1 asymmetry between the two directions is really what's going on here — which is also why `efficient_first@0.5` behaves exactly as documented and only `capable_first@0.5` gets stuck.

There's a near-miss that makes it easy to miss: before the squash, one maxed signal is exactly `5.0 × 0.10 = 0.5`, which *would* satisfy the inclusive `>=` gate. The `tanh` compression pulls it down to `0.4621` — about 7.6% short.

## What we measured

Two seeds of TB-Lite (20 tasks each, Hermes agent) at the default `capable_first@0.5`:

| seed | tasks | weak-tier turns | weak tokens |
|---|---|---|---|
| 1 | 20 | **0 / 217** | 0.0% |
| 2 | 20 | **0 / 222** | 0.0% |

439 consecutive turns, none routed to the efficient tier, with zero upstream errors in either run.

Changing only the threshold to `0.45` (same profile, same tasks) -- offload becomes **13.9%** (token-weighted), and `efficient_first@0.45` reaches 58.0%. So the tier plumbing is working fine; it's specifically the `capable_first` + `>0.4621` combination that can't fire.

Worth noting the hard de-escalate shortcut (`tests_passed && recent write/edit >= 1`) didn't rescue it either — that path bypasses the scorer entirely, but across those 439 turns it never fired, so what you observe is a complete no-op rather than just a dead scorer branch.

## The tricky part for users

The cliff sits at `tanh(0.5) = 0.46211715726`:

| threshold | 0.40 | 0.45 | 0.46 | 0.4621 | 0.4622 | 0.47 | 0.48 | 0.5 |
| scorer offload fires? | yes | yes | yes | yes | **no** | **no** | **no** | **no** |

Someone wondering "why isn't anything offloading?" would very reasonably nudge `0.5 → 0.48 → 0.47`, still get nothing, and conclude the feature doesn't work. Nothing in the logs or in `routing_decisions.stage_router` separates "the signals were weak this turn" from "this branch cannot fire at all" -- every turn just shows up as `no_signal` / `fall_open`.

## The docs get very close to this

`docs/routing_algorithms/stage_router_routing.md:109` already spells out the key fact — one full signal scores ~`0.46`, "just under `0.5`" — and draws exactly the right conclusion for *escalation*: a decisive escalation needs a strong signal plus corroboration. That reasoning is sound in the positive direction.

The subtlety is that the same sentence doesn't carry over to de-escalation, because there's only the one negative signal available to corroborate with. A few nearby places read a little differently once that's taken into account:

- **:86** — "only drops to the efficient tier when the signals say 'efficient' and clear the threshold" — accurate in general, but at the recommended `0.5` that path can't be taken.
- **:90** — "Raise it and only strong, decisive signals drop a turn to efficient" — reads as a continuous dial; above `0.4621` the rate isn't lower so much as fixed at zero.
- **:101** — "Set `0.5` explicitly. It's the recommended starting point."
- **:175** — "a `0.5` threshold takes ~1.5 signals of agreement — a policy that escalates ~20% of tasks maps roughly to `confidence_threshold: 0.5` with `capable_first`" — with `capable_first` the scorer-driven action is de-escalation, so we weren't sure how to square this one with the above. Quite possibly it was calibrated with a classifier attached, in which case it may just need a note.

## Possible fixes

Any one of these would have saved us the detour:

1. Lower the default `confidence_threshold` to `0.45`.
2. Warn at config load when `picker == capable_first && confidence_threshold > tanh(0.5)`.
3. Add a second negative signal so the efficient direction can reach the k=2 rung.

There's already a test pinning the constant — `one_signal_scores_below_half` (`stage.rs:570`) asserts `0.4621 < 0.5` — so the value is known and deliberate; it just doesn't currently flag the interaction with the default picker.

Happy to put up a PR for whichever direction you prefer, and happy to run any repro that would be useful. Thanks.

WUKUNTAI-0211:
## What

Warns at route construction when a `stage_router` route pairs a picker with a `confidence_threshold` the scorer can never reach in the direction that leaves the default tier. No routing behaviour changes.

```rust
pub fn max_efficient_confidence() -> f64 { confidence_for_units(1.0) }  // 0.4621…
pub fn max_capable_confidence()   -> f64 { confidence_for_units(2.0) }  // 0.7616…

pub fn scorer_cannot_leave_default(mode, confidence_threshold) -> bool
```

## Why

Closes #264

`score_signal` sums one positive axis against one negative dimension, then squashes with `tanh`. Each direction has a hard ceiling below `1.0`:

| direction | units available | ceiling |
|---|---|---|
| efficient | `production_intensity` only, bounded to `1.0` by `ratio` | `0.4621` |
| capable | `severity` at its `HARD_SEVERITY` cap, plus one of `spinning` / `exploring` (they partition the not-producing case) | `0.7616` |

Above its ceiling the gate never opens, so the scorer can only ever confirm the default tier. The reporter measured 439 consecutive turns at `capable_first` + `0.5` with zero turns routed to the efficient tier and zero upstream errors. Nothing in the logs or in `routing_decisions.stage_router` separates "signals were weak this turn" from "this branch cannot fire at all".

`efficient_first` + `1.0` is the mirror case, and the `[0.0, 1.0]` range check accepts it today.

## How tested

- [x] `cargo test --workspace` green (23 test binaries, 0 failed)
- [x] `cargo clippy --workspace --all-targets` clean
- [x] `uv run pytest tests/ --ignore=tests/e2e` green (135 passed)
- [x] Manual smoke: captured the log for all four combinations — `capable_first` + `0.5` and `efficient_first` + `1.0` warn; `0.45` and `0.5` respectively stay silent.

`tests/e2e/test_closed_book_proxy_integration.py` fails locally, but it fails identically on a clean checkout of `main` — a pre-existing network-dependent failure, unrelated to this change.

Seven unit tests added. The two that pin the ceilings derive them from the dimension set rather than asserting literals, so they fail if the signal set ever changes.

## Checklist

- [x] New public symbols re-exported from `crates/libsy/src/lib.rs`.
- [ ] Python-side checklist items are not applicable — Rust-only change, no customer-facing surface change.

## Notes for reviewers

**Which of the issue's three options this takes.** Option 1 (lower the default) no longer applies: `confidence_threshold` is a required TOML field with no default, and the Python paths the issue cites were removed when routing moved to Rust. Option 3 (add a second negative signal) changes routing behaviour and needs recalibration. This is option 2 — happy to re-scope if you prefer option 3.

**Warn, not reject.** The hard overrides (`should_escalate` / `should_deescalate`) run before the scorer and still fire, and the judge still resolves turns, so the route is not inert end to end — only the scorer is. Rejecting at load would break running deployments, including the `confidence_threshold = 1.0` fixture in `config.rs`.

**Placement.** The guard lives in `libsy` so it covers the TOML server path and the PyO3 bindings from one place, and the ceilings sit next to the `SCORE_GAIN` / `SIGNAL_UNIT` / `HARD_SEVERITY` constants they are derived from. Deployments on `capable_first` will see this alongside the existing "capable_first is experimental" warning in `config.rs`; collapsing the pair is an option if it reads as noise.

**The capable ceiling depends on the severity table.** It assumes `HARD_SEVERITY` is the strongest severity the scorer sees, which is what that constant's doc comment already states (critical is intercepted by the override). If a pattern with a severity between `0.7` and `1.0` is ever added to `ERROR_PATTERNS`, `max_capable_confidence` moves and its test fails — intentionally.

**Docs left alone.** #264 also flags four lines in `docs/routing_algorithms/stage_router_routing.md`. #288 is already rewriting that page, so touching it here would only conflict. A follow-up can point it at these helpers once that lands.

## Summary by CodeRabbit

* **New Features**
  * Added confidence-limit utilities for identifying efficient and capable maximum thresholds.
  * Added detection for thresholds that cannot move a picker beyond its default tier.

* **Bug Fixes**
  * Warnings now identify unreachable thresholds, the affected mode, and recommended values.
  * Routing preserves hard overrides and fall-through behavior for undecided turns.

* **Tests**
  * Added coverage for confidence ceilings, mode-specific thresholds, and picker behavior at supported limits.

### DIFF
--- a/crates/libsy/src/algorithms/stage.rs
+++ b/crates/libsy/src/algorithms/stage.rs
@@ -22,8 +22,8 @@ use super::llm_class::{LlmClassifierConfig, LlmTaskClassifier, TaskClassifierCon
 use super::util::prompts::{SystemPromptProcessor, TargetPrompts};
 use super::util::stage::{
     DecisionSource, HandoffNoteConfig, PickerMode, StageClassifier, StageTargets,
-    max_efficient_confidence, record_decision_source, record_routing_decision,
-    scorer_cannot_leave_default,
+    max_capable_confidence, max_efficient_confidence, record_decision_source,
+    record_routing_decision, scorer_cannot_leave_default,
 };
 use super::util::tool_signals::{DEFAULT_RECENT_WINDOW, ToolSignalProcessor};
 use crate::core::algorithm::{Algorithm, Driver};
@@ -159,22 +159,29 @@ fn build_route(
             ),
         });
     }
-    // A threshold above the efficient ceiling silently disables the scorer for
-    // capable_first: every turn reads as "signals were weak" when in fact that
-    // branch cannot fire at all. Warn rather than reject, since the judge and the
-    // hard de-escalation shortcut still work and the setting stays a policy choice.
+    // A threshold above the scorer's ceiling in the direction that leaves the
+    // default tier silently disables the scorer: those turns read as "signals were
+    // weak" when in fact the branch cannot fire at all. Warn rather than reject —
+    // the hard overrides still fire, and the setting stays a policy choice.
     if scorer_cannot_leave_default(config.mode, config.confidence_threshold) {
+        let (picker, other_tier, ceiling) = match config.mode {
+            PickerMode::CapableFirst => ("capable_first", "efficient", max_efficient_confidence()),
+            PickerMode::EfficientFirst => ("efficient_first", "capable", max_capable_confidence()),
+        };
         tracing::warn!(
             confidence_threshold = config.confidence_threshold,
-            efficient_ceiling = max_efficient_confidence(),
-            "stage_router picker \"capable_first\" with confidence_threshold {} cannot de-escalate: \
-             production_intensity is the only signal on the efficient side, so the scorer tops out \
-             at {:.4}. Every turn will fall through to the judge, or to the capable tier when no \
-             judge is configured. Set confidence_threshold at or below {:.4} for the scorer to \
-             reach the efficient tier.",
+            ceiling,
+            "stage_router picker \"{}\" with confidence_threshold {} can never select the {} tier: \
+             the scorer tops out at {:.4} in that direction, so no signal clears the gate. Turns \
+             the hard overrides do not claim fall through to the judge, or to the default tier when \
+             no judge is configured. Set confidence_threshold at or below {:.4} for the scorer to \
+             reach the {} tier.",
+            picker,
             config.confidence_threshold,
-            max_efficient_confidence(),
-            max_efficient_confidence(),
+            other_tier,
+            ceiling,
+            ceiling,
+            other_tier,
         );
     }
     // The tiers are a fixed pair; their targets are whatever the deployment calls
--- a/crates/libsy/src/algorithms/stage.rs
+++ b/crates/libsy/src/algorithms/stage.rs
@@ -22,7 +22,8 @@ use super::llm_class::{LlmClassifierConfig, LlmTaskClassifier, TaskClassifierCon
 use super::util::prompts::{SystemPromptProcessor, TargetPrompts};
 use super::util::stage::{
     DecisionSource, HandoffNoteConfig, PickerMode, StageClassifier, StageTargets,
-    record_decision_source, record_routing_decision,
+    max_efficient_confidence, record_decision_source, record_routing_decision,
+    scorer_cannot_leave_default,
 };
 use super::util::tool_signals::{DEFAULT_RECENT_WINDOW, ToolSignalProcessor};
 use crate::core::algorithm::{Algorithm, Driver};
@@ -157,6 +158,24 @@ fn build_route(
                 config.confidence_threshold
             ),
         });
+    }
+    // A threshold above the efficient ceiling silently disables the scorer for
+    // capable_first: every turn reads as "signals were weak" when in fact that
+    // branch cannot fire at all. Warn rather than reject, since the judge and the
+    // hard de-escalation shortcut still work and the setting stays a policy choice.
+    if scorer_cannot_leave_default(config.mode, config.confidence_threshold) {
+        tracing::warn!(
+            confidence_threshold = config.confidence_threshold,
+            efficient_ceiling = max_efficient_confidence(),
+            "stage_router picker \"capable_first\" with confidence_threshold {} cannot de-escalate: \
+             production_intensity is the only signal on the efficient side, so the scorer tops out \
+             at {:.4}. Every turn will fall through to the judge, or to the capable tier when no \
+             judge is configured. Set confidence_threshold at or below {:.4} for the scorer to \
+             reach the efficient tier.",
+            config.confidence_threshold,
+            max_efficient_confidence(),
+            max_efficient_confidence(),
+        );
     }
     // The tiers are a fixed pair; their targets are whatever the deployment calls
     // them, and the classifier scores onto those names.
--- a/crates/libsy/src/algorithms/util/stage.rs
+++ b/crates/libsy/src/algorithms/util/stage.rs
@@ -334,26 +334,48 @@ pub fn score_signal(signal: &ToolSignals) -> ScoreResult {
     }
 }
 
+/// Confidence a turn scoring `units` maxed signals lands on, after the squash.
+fn confidence_for_units(units: f64) -> f64 {
+    (SCORE_GAIN * SIGNAL_UNIT * units).tanh()
+}
+
 /// Highest confidence the scorer can reach toward the efficient tier.
 ///
 /// `production_intensity` is the only dimension on the efficient side and
 /// [`ratio`] bounds it to `1.0`, so the efficient direction has exactly one
 /// signal unit to work with — there is nothing for it to corroborate with. The
-/// `tanh` squash then pulls that unit down from `SCORE_GAIN * SIGNAL_UNIT` to
-/// roughly `0.4621`, which lands just under the commonly used `0.5` threshold.
+/// `tanh` squash then pulls that unit down to roughly `0.4621`, which lands just
+/// under the commonly used `0.5` threshold.
 pub fn max_efficient_confidence() -> f64 {
-    (SCORE_GAIN * SIGNAL_UNIT).tanh()
+    confidence_for_units(1.0)
+}
+
+/// Highest confidence the scorer can reach toward the capable tier.
+///
+/// Escalation stacks two units, not three: `severity` contributes one at its
+/// [`HARD_SEVERITY`] cap (critical is intercepted by the override), and
+/// `spinning` / `exploring` partition the not-producing case so only one of them
+/// can fire. `production_intensity` is zero whenever either does, so nothing
+/// subtracts. That caps escalation confidence at roughly `0.7616`.
+pub fn max_capable_confidence() -> f64 {
+    confidence_for_units(2.0)
 }
 
 /// True when `confidence_threshold` puts every scorer outcome out of reach of
 /// `mode`'s non-default tier, so the scorer can never change a turn's tier.
 ///
-/// Only [`PickerMode::CapableFirst`] can hit this: leaving its default means
-/// picking efficient, and that side caps at [`max_efficient_confidence`].
-/// Escalation stacks several dimensions, so `efficient_first` always has some
-/// reachable band and is never reported here.
+/// Both pickers can hit this, because `tanh` bounds confidence well below `1.0`
+/// in either direction — `capable_first` from
+/// [`max_efficient_confidence`], `efficient_first` from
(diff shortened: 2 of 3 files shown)


WUKUNTAI-0211:
@coderabbitai review

WUKUNTAI-0211:
@coderabbitai review

WUKUNTAI-0211:
Re-confirmed against current `main` (`9ad6744b`), since this PR has been open a while and `main` has moved 16 commits — three of which touch `crates/libsy/src/algorithms/stage.rs` (#431 retry relocation, #411 `json_object` classifier, #413 `Decision::reasoning` → log). None of them touch the scorer, so the gap this PR addresses is still present.

## Static check

On `main`, `crates/libsy/src/algorithms/util/stage.rs` is unchanged in every part that matters here:

- `SCORE_GAIN = 5.0`, `HARD_SEVERITY = 0.7`, `SIGNAL_UNIT = 0.10` (L40/43/46)
- `raw = SIGNAL_UNIT * (severity / HARD_SEVERITY + spinning + exploring - production_intensity)`, `score = (SCORE_GAIN * raw).tanh()` (L328-330)
- the gate is still `if scored.confidence >= confidence_threshold` (L387)
- the efficient direction still has exactly one contributing term, `production_intensity`, and it still comes from `ratio(writes + edits, recent_ops)`, so it is bounded by `1.0`

## Empirical check

I checked out `9ad6744b` into a clean worktree and added a throwaway brute-force test over the scorer's input space, enumerating `ToolSignals` with:

- `severity` from `0.0` to `1.0` in steps of `0.05`
- `recent_write_count`, `recent_edit_count`, `recent_read_count`, `recent_todowrite_count` each `0..=4`
- `turn_depth` `0..=8`
- `tests_passed` both values
- `compacted = false` (it is a hard escalate, so it can never reach the scorer)

For each combination I called `pick_tier(&signal, PickerMode::CapableFirst, 0.5)` and counted how often it resolved to the efficient tier via the scorer:

```
REPRO264 total=236250 scorer_efficient=0 hard_deescalate=5400 max_negative_confidence=0.4621171573
```

- `scorer_efficient = 0` — across all 236,250 inputs, `DecisionSource::Dimensions` never selected `Tier::Efficient`. Not rare; unreachable.
- `max_negative_confidence = 0.4621171573` — the measured ceiling on negative-side confidence, exactly `tanh(0.5)`, which is 7.6% short of the `>= 0.5` gate.
- `hard_deescalate = 5400` — the only route to the efficient tier is the `should_deescalate` shortcut (`tests_passed && recent write/edit >= 1 && severity <= 0.0`), which bypasses the scorer entirely. The measurements in #264 show that path not firing once across 439 turns.

So with `picker = "capable_first"` and `confidence_threshold = 0.5`, the scorer's offload branch on current `main` is dead code, and each such turn logs as `fall_open` — indistinguishable from "the signals were weak this turn".

## On the existing mitigation

#363 already marks `capable_first` experimental, warns at startup, and makes `picker` / `confidence_threshold` required in the TOML schema. That removes the "both defaults silently conspire" half of #264 — you can no longer arrive here by omitting the knobs. The arithmetic half is untouched: a user who explicitly writes `capable_first` at `0.5` still gets a scorer that cannot fire, with nothing in the logs distinguishing that from weak signals. That is what this PR reports.

The test above was a throwaway for verification and is not part of this PR; happy to fold a permanent version of it in if reviewers would like the ceiling pinned by a test.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -334,6 +334,60 @@ pub fn score_signal(signal: &ToolSignals) -> ScoreResult {
...
+
+/// True when `confidence_threshold` puts every scorer outcome out of reach of
+/// `mode`'s non-default tier, so the scorer can never change a turn's tier.
+///
+/// Both pickers can hit this, because `tanh` bounds confidence well below `1.0`
+/// in either direction — `capable_first` from
+/// [`max_efficient_confidence`], `efficient_first` from
+/// [`max_capable_confidence`]. Above its ceiling the scorer still runs and still
+/// reports a score, but it can only ever confirm the default tier.
+pub fn scorer_cannot_leave_default(mode: PickerMode, confidence_threshold: f64) -> bool {
```
nachiketb-nvidia, line 382: all of these don't have to be pub, they can be pub(crate).

in fact, since they are only used in logs, they can be in-line or function within functions inside `scorer_cannot_leave_default` block.

or it can be just one function doing it all.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -649,6 +703,172 @@ mod tests {
         );
     }
 
+    #[test]
+    fn production_is_the_only_signal_pushing_toward_efficient() {
```
nachiketb-nvidia, line 707: these are too many tests for this small fix, pls change an existing test for this or have only one

crates/libsy/src/algorithms/stage.rs
```diff
@@ -158,6 +159,31 @@ fn build_route(
             ),
         });
     }
+    // A threshold above the scorer's ceiling in the direction that leaves the
+    // default tier silently disables the scorer: those turns read as "signals were
+    // weak" when in fact the branch cannot fire at all. Warn rather than reject —
+    // the hard overrides still fire, and the setting stays a policy choice.
```
nachiketb-nvidia, line 165: fewer comments pls :)

nachiketb-nvidia (changes requested):

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -334,6 +334,60 @@ pub fn score_signal(signal: &ToolSignals) -> ScoreResult {
...
+/// Both pickers can hit this, because `tanh` bounds confidence well below `1.0`
+/// in either direction — `capable_first` from
+/// [`max_efficient_confidence`], `efficient_first` from
+/// [`max_capable_confidence`]. Above its ceiling the scorer still runs and still
+/// reports a score, but it can only ever confirm the default tier.
+pub fn scorer_cannot_leave_default(mode: PickerMode, confidence_threshold: f64) -> bool {
+    let ceiling = match mode {
+        // Leaving the capable default means picking efficient, and vice versa.
+        PickerMode::CapableFirst => max_efficient_confidence(),
+        PickerMode::EfficientFirst => max_capable_confidence(),
```
sabhatinas, line 386: Please use max_efficient_score or max_capable_score for the theoretical limits instead. If new signal buckets get added to the picker logic, these estimations will break and look fragile. Can you add a warning inside picker function itself by refactoring the blocks there if needed to check against mathematical ceilings?

sabhatinas (changes requested):

grahamking:
@WUKUNTAI-0211 Are you still working on this?

bhuvan002:
Closing due to inactivity. Please re-open with the requested changes if this is still needed.

PR #397, opened 2026-08-13
fix(translation): preserve tool IDs across Anthropic

### LINKED ISSUE #178: [bug] tool_use id sanitization is not reversible — replaying rewritten ids breaks multi-turn tool calling on Kimi K2 (vLLM, id template functions.<name>:<index>)
## Summary

`sanitize_anthropic_tool_use_id` (`crates/switchyard-translation/src/util.rs`) rewrites upstream tool-call ids that contain characters outside `[A-Za-z0-9_-]` by replacing each offending char with `_`. The mapping is **lossy (non-injective)** and there is no inverse anywhere in the codebase: on the next turn, the client replays the sanitized id and Switchyard forwards it to the upstream **verbatim** (`codecs/openai_chat/buffered.rs` — tool_call `id` and tool message `tool_call_id`; same on the Responses encoder's `call_id`).

For models whose native tool-call ids use such characters, the upstream model receives a *corrupted version of its own id format* on every continuation turn. Concretely: Kimi K2 served by vLLM (kimi_k2 tool-call parser) emits ids shaped `functions.<name>:<index>`, e.g. `functions.list_skills:0`. Switchyard turns this into `functions_list_skills_0`, and replaying that id makes Kimi K2 **dead-stop instead of continuing the tool loop**: it returns an empty response (buffered: `finish_reason:"stop"`, explicit `tool_calls:[]`, empty content; streaming: `finish_reason:"tool_calls"` with **zero** tool-call deltas). The Anthropic-side client then sees an assistant turn with no `tool_use` where one was expected, which breaks every multi-turn tool loop (agent frameworks surface this as a tool-call parse failure).

## Reproduction / evidence

Anthropic `/v1/messages` client → Switchyard (`format: openai chat`) → vLLM serving a Kimi K2 model. Turn 1 (tool available) works: Kimi emits a valid `tool_calls` entry, Switchyard returns a clean `tool_use` (id sanitized). Turn 2 (replay assistant `tool_use` + `tool_result`) fails: Switchyard forwards the sanitized id upstream and Kimi degenerates.

Byte-matched A/B directly against the vLLM endpoint (temperature 0, identical request bytes, only the replayed assistant `tool_calls[0].id` + tool message `tool_call_id` varied):

| replayed id | tool call emitted on continuation |
|---|---|
| `functions_list_skills_0` (sanitized) | **1/16** (~94% failure — empty response) |
| `functions.list_skills:0` (native) | 16/16 |
| `call_abc123` / `sw_00000001` (foreign) | 16/16 |
| `toolu_01AbCdEfGhJkMnPqRs` (Anthropic-style) | 16/16 |

Notable: completely *foreign* ids are fine — the failure is specific to the sanitized mutation of the model's own id template (`.`/`:` → `_`), which apparently corrupts the model's tool-call bookkeeping for the next generation. Verified identical behavior at an older pin and current main (the sanitizer is unchanged since the initial commit).

Streaming flavor of the degenerate mode (raw upstream, 5/5): final chunk `{"delta":{},"finish_reason":"tool_calls"}` with zero tool-call deltas — Switchyard faithfully translates that to `stop_reason:"tool_use"` with 0 `tool_use` blocks, so the drop is upstream-model behavior triggered by the corrupted id, not a Switchyard streaming bug.

## Why sanitization can't simply be removed

api.anthropic.com enforces `^[a-zA-Z0-9_-]+$` on replayed `tool_use.id` server-side (undocumented — the docs type `id` as plain string; only tool *names* have a documented pattern), so unsanitized ids like `functions.list_skills:0` would 400 against genuine Anthropic backends and strict Anthropic-schema clients. Sanitization is load-bearing; it just needs to be **reversible**.

## Proposed fix: reversible charset-safe encoding + restore on the upstream-bound direction

- **Encode** (in `sanitize_anthropic_tool_use_id`): if the raw id is non-empty, all chars in `[A-Za-z0-9_-]`, and doesn't start with the escape prefix → return unchanged (zero behavior change for `toolu_*` / `call_*` traffic). Otherwise return e.g. `sy64_` + base64url-nopad(raw) — base64url's alphabet is exactly the Anthropic-safe charset, and the single change covers all response-side emit sites.
- **Decode**: a `desanitize_anthropic_tool_use_id` (strip `sy64_`, base64url-decode, pass through unchanged on any failure) applied at the upstream-bound emit points: openai_chat tool_call `id` + tool message `tool_call_id`, and the Responses encoder's two `call_id` sites.

Properties: injective (no collisions; double-encode disambiguated by the prefix guard), stateless (the original bytes travel inside the id itself — no server-side map needed), and the upstream model always gets back exactly the id it emitted. The "regenerate + id map" alternative doesn't work in a stateless server (the original→generated mapping is destroyed at response time and can't be rebuilt from the replayed conversation).

Rough size: ~50–70 lines across `util.rs` + the four upstream-bound call sites, plus tests (a regression test replaying a `functions.<name>:<index>`-style id through `/v1/messages`, buffered + streaming).

## Related

- PR #173 (Anthropic outbound normalization, "fixing … tool IDs …") normalizes ids on the Anthropic-backend-bound direction — one-way. The restore direction proposed here should coordinate with / live alongside that layer.
- #111 / #46 touched the same subsystem (tool-turn translation fidelity) but not id reversibility.

Happy to submit a PR implementing the above if the approach sounds right.

ting-hong-shieh:
## What

- Replace lossy Anthropic tool-use ID sanitization with `sy64_`-prefixed, unpadded Base64URL encoding.
- Decode encoded `tool_use.id` and `tool_result.tool_use_id` values at Anthropic ingress, for both buffered and streaming paths.
- Keep the provider-neutral IR and OpenAI Chat / Responses encoders free of Anthropic-specific escaping.
- Keep ordinary `[A-Za-z0-9_-]` IDs unchanged, escape the reserved prefix, and pass malformed encoded values through unchanged.

## Why

Some OpenAI-compatible models emit tool-call IDs containing characters that Anthropic does not accept. Kimi K2, for example, can emit IDs such as `functions.list_skills:0` and expects the same ID on the next turn.

The previous sanitizer replaced unsupported characters with underscores. That mapping could not be reversed, so Switchyard replayed a different ID to the upstream model and could break multi-turn tool calling. The new encoding keeps Anthropic-facing IDs within the accepted character set while Anthropic decoders restore the original ID before it enters the neutral IR.

Closes #178

## Before / after

Input upstream tool-call ID: `functions.list_skills:0`

Before:

```json
{
  "anthropic_tool_use_id": "functions_list_skills_0",
  "replayed_openai_tool_call_id": "functions_list_skills_0",
  "replayed_openai_tool_result_id": "functions_list_skills_0"
}
```

After:

```json
{
  "anthropic_tool_use_id": "sy64_ZnVuY3Rpb25zLmxpc3Rfc2tpbGxzOjA",
  "replayed_openai_tool_call_id": "functions.list_skills:0",
  "replayed_openai_tool_result_id": "functions.list_skills:0"
}
```

## How tested

- [x] `cargo test -p switchyard-translation`
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/ -v -m "not integration"` — 147 passed, 2 deselected
- [x] Anthropic → OpenAI Chat streaming regression restores the original ID

## Checklist

- [x] No new classes or downstream public symbols.
- [x] Unit and integration tests cover the bug fix.
- [x] README and `--help` are unchanged because this does not add a configuration or command surface.

## Notes for reviewers

`sy64_` is reserved for encoded IDs. An upstream ID that already starts with this prefix is encoded again before it reaches an Anthropic client, which keeps decoding unambiguous. Invalid prefixed input is left unchanged.

No live provider call was made.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved tool-call and tool-result IDs accurately when translating between Anthropic, OpenAI Chat, and Responses formats.
  * Added reversible handling for IDs containing special characters, Unicode, empty values, or reserved prefixes.
  * Prevented malformed encoded IDs from being altered unexpectedly.
* **Tests**
  * Added coverage for request and streaming translations, including regression cases for unsafe tool IDs.


ayushag-nv:
@ting-hong-shieh Thanks for putting this up. Can you always do the before and after the change snapshot of the results. It helps me to do better and fast reviews.

ting-hong-shieh:
Thanks, @ayushag-nv. I ran the same in-process buffered translation at the PR base (`b256d936`) and current head (`18ef28bd`): OpenAI Chat response → Anthropic response → Anthropic replay → OpenAI Chat request. No provider call was involved.

Input upstream tool-call ID: `functions.list_skills:0`

Before (`b256d936`):

```json
{
  "anthropic_tool_use_id": "functions_list_skills_0",
  "replayed_openai_tool_call_id": "functions_list_skills_0",
  "replayed_openai_tool_result_id": "functions_list_skills_0",
  "upstream_emitted_id": "functions.list_skills:0"
}
```

The unsupported characters are replaced with underscores, and both replayed OpenAI IDs differ from the ID emitted by the upstream model.

After (`18ef28bd`):

```json
{
  "anthropic_tool_use_id": "sy64_ZnVuY3Rpb25zLmxpc3Rfc2tpbGxzOjA",
  "replayed_openai_tool_call_id": "functions.list_skills:0",
  "replayed_openai_tool_result_id": "functions.list_skills:0",
  "upstream_emitted_id": "functions.list_skills:0"
}
```

The Anthropic-facing ID stays within the accepted character set, while the OpenAI tool call and tool result both recover the exact upstream ID on replay.

ayushag-nv:
@nachiketb-nvidia @elyasmnvidian to review

ting-hong-shieh:
Resolved the merge conflict and updated the branch. Thanks!

nachiketb-nvidia (changes requested):
Requesting changes here: the buffered path restores reversible Anthropic-safe tool IDs, but the streaming path does not.

  Concrete issue:
  - Anthropic stream `tool_use.id = sy64_ZnVuY3Rpb25zLmxpc3Rfc2tpbGxzOjA`
  - Expected OpenAI tool-call ID: `functions.list_skills:0`
  - Actual OpenAI stream output keeps the encoded `sy64_...` ID

  Suggested fix:
  - Keep `sanitize_anthropic_tool_use_id(...)` when encoding to Anthropic.
  - Decode IDs when reading from Anthropic, not when writing to OpenAI/Responses.
  - Apply that in:
    - Anthropic buffered `tool_use.id`
    - Anthropic buffered `tool_result.tool_use_id`
    - Anthropic stream `content_block_start` tool IDs
  - Remove `desanitize_anthropic_tool_use_id(...)` calls/imports from OpenAI Chat / Responses buffered encoders.
  - After that cleanup, OpenAI Chat / Responses encoders should just emit the neutral IR IDs directly.
  - Add one streaming regression test for Anthropic -> OpenAI Chat or Responses. one test should be enough, remove the rest
  This keeps the Anthropic-specific escaping localized to the Anthropic codec and makes buffered + streaming behavior consistent.

ting-hong-shieh:
Thanks, @nachiketb-nvidia. Addressed this in e534e88c:

- moved ID decoding to the Anthropic buffered decoder for `tool_use.id` and `tool_result.tool_use_id`
- moved streaming ID decoding to Anthropic `content_block_start`
- removed Anthropic-specific desanitization from OpenAI Chat and Responses encoders
- replaced the forward streaming test with one Anthropic → OpenAI Chat regression and removed the redundant standalone request tests; existing buffered tests now cover tool calls and results

Validation:
- `cargo fmt --all --check`
- `cargo test --workspace`
- Ruff and mypy
- Pytest: 147 passed, 2 deselected

No live provider call was made. Could you please take another look?

nachiketb-nvidia (approved):
thanks!

grahamking:
@ting-hong-shieh Could you rebase this? It's a small import conflict. Then it's ready to go.

Thanks for the great work as usual!

ting-hong-shieh:
Rebased-equivalent, but done as a merge — `main` moved 45 commits and this branch
already carries two earlier `main` merges, so replaying the two original commits
re-opened six conflicts across three files including `codecs/responses/buffered.rs`,
rather than the one import conflict. The merge keeps it at that one conflict and leaves
the reviewed diff untouched: still +122/-29 across 8 files, the only change from what you
approved being the merged import list.

`cargo fmt --all --check` clean, `cargo clippy -p switchyard-translation --all-targets`
clean, `cargo test -p switchyard-translation` 154 passed.

Happy to redo it as a true rebase if you'd rather have the linear history — it just needs
the conflict resolutions re-derived by hand.

ting-hong-shieh:
@grahamking one thing I cannot do from here: the push reset the fork workflow approval, so
`CI`, `Package portability`, `Documentation` and `Getting Started` are sitting at
`action_required` on `8a9ad776`. They were all green on the previous head. Could you
approve the runs when you get a moment? The PR is `MERGEABLE` otherwise.

PR #404, opened 2026-08-13
fix(translation): accept SSE data without a space

### LINKED ISSUE #399: [bug] SSE decoder drops data: fields without a space after the colon
## Symptom

The shared SSE decoder silently drops valid `data:` fields when the colon is not followed by a space. An upstream stream that emits `data:{...}` instead of `data: {...}` therefore produces no normalized response chunks.

The [WHATWG event-stream parser](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream) treats one space after the colon as optional: it collects the value after the colon and removes a single leading space only when one is present.

## Reproduction

Add this focused regression test next to `decode_stream_parses_sse_bytes_into_ir_chunks` in `crates/switchyard-translation/src/helpers.rs`:

```rust
#[test]
fn decode_stream_accepts_sse_data_without_optional_space() -> Result<(), LlmClientError> {
    let sse = b"data:{\"choices\":[{\"delta\":{\"content\":\"Hello\"}}]}\n\n\
         data:[DONE]\n\n"
        .to_vec();
    let bytes = stream::once(async move { Ok::<Vec<u8>, LlmClientError>(sse) });
    let chunks = decode_all(bytes, WireFormat::OpenAiChat)?;
    assert_eq!(text_of(&chunks), "Hello");
    Ok(())
}
```

Run:

```bash
cargo test -p switchyard-translation decode_stream_accepts_sse_data_without_optional_space -- --nocapture
```

The assertion fails with:

```text
assertion `left == right` failed
  left: ""
 right: "Hello"
```

`parse_json_sse_frame` currently filters with `line.strip_prefix("data: ")`, so both the payload and the no-space `data:[DONE]` marker are ignored.

## Expected vs. actual

- **Expected:** `data:{...}` and `data: {...}` decode to the same event payload. A no-space `data:[DONE]` terminates the stream.
- **Actual:** no-space data fields are treated as empty frames and silently discarded.

## Environment

- OS / arch: macOS 26.5.2, arm64
- Install path: source checkout
- Inbound format: OpenAI Chat Completions, streaming
- Backend: local byte-stream fixture; no provider credentials

## Additional context

The existing decoder test covers only `data: {...}`. A fix could parse the field name at the first colon, accept both spellings, and remove at most one optional leading space from the value.

I searched open and closed issues and pull requests for `data: without space`, `optional space SSE`, `parse_json_sse_frame`, and `SSE parser drops data`; I found no matching report or in-flight fix.

I'm happy to send a PR with the parser change and regression coverage if this direction looks right.

ting-hong-shieh:
## What

- accept both `data:{...}` and `data: {...}` SSE fields
- remove at most one optional ASCII space after the field-name colon
- cover a no-space JSON payload and `data:[DONE]` in the shared stream decoder

## Why

The SSE parser matched only the literal `data: ` prefix, so valid upstream events without the optional space were silently ignored.

Closes #399

## Validation

- `cargo test -p switchyard-translation decode_stream_accepts_sse_data_without_optional_space -- --nocapture` — passed
- `cargo test -p switchyard-translation` — passed
- `cargo fmt --all --check` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed
- `cargo test --workspace` — passed with loopback access enabled for the repository's local mock servers
- `git diff --check` — passed

No live provider calls or secrets were used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response parsing to accept data fields with or without a space after the colon.
  * Preserved support for termination markers such as `[DONE]`.
* **Tests**
  * Added regression coverage for both supported data field formats.

### DIFF
--- a/crates/switchyard-translation/src/helpers.rs
+++ b/crates/switchyard-translation/src/helpers.rs
@@ -583,6 +583,17 @@ mod tests {
     }
 
     #[test]
+    fn decode_stream_accepts_sse_data_without_optional_space() -> Result<(), LlmClientError> {
+        let sse = b"data:{\"choices\":[{\"delta\":{\"content\":\"Hello\"}}]}\n\n\
+             data:[DONE]\n\n"
+            .to_vec();
+        let bytes = stream::once(async move { Ok::<Vec<u8>, LlmClientError>(sse) });
+        let chunks = decode_all(bytes, WireFormat::OpenAiChat)?;
+        assert_eq!(text_of(&chunks), "Hello");
+        Ok(())
+    }
+
+    #[test]
     fn stream_helpers_replay_same_format_provider_fields() -> Result<(), BoxError> {
         let provider_event = json!({
             "id": "chatcmpl-test",
--- a/crates/switchyard-translation/src/sse.rs
+++ b/crates/switchyard-translation/src/sse.rs
@@ -36,7 +36,10 @@ pub(crate) fn parse_json_sse_frame(
     let data = frame
         .lines()
         .filter(|line| !line.is_empty() && !line.starts_with(':'))
-        .filter_map(|line| line.strip_prefix("data: ").map(|l| l.to_string()))
+        .filter_map(|line| {
+            line.strip_prefix("data:")
+                .map(|value| value.strip_prefix(' ').unwrap_or(value).to_string())
+        })
         .fold(String::new(), |mut a, b| {
             a.reserve(b.len() + 1);
             a.push_str(&b);


ting-hong-shieh:
## Before / after output snapshot

I ran the same in-process SSE decode at the PR base (`b256d936`) and current head (`a1f747bd`). No HTTP or provider call was involved.

Input:

```text
data:{"choices":[{"delta":{"content":"Hello"}}]}

data:[DONE]
```

Before (`b256d936`):

```json
{"decoded_text":"","event_count":0}
```

Both no-space `data:` fields were ignored, so the text and terminal marker disappeared.

After (`a1f747bd`):

```json
{"decoded_text":"Hello","event_count":1}
```

The JSON field is decoded and the no-space `data:[DONE]` marker terminates the stream without producing an extra event.

nachiketb-nvidia:
closing this in favour of https://github.com/NVIDIA-NeMo/Switchyard/pull/447

PR #405, opened 2026-08-13
fix(server): validate configured base URLs

### LINKED ISSUE #398: [bug] --dry-run accepts an invalid llm client base_url
## Symptom

On current `main`, a deployment with `base_url = "not a url"` passes
`switchyard-server --dry-run`, prints `server OK: test-route`, and exits 0. The
same deployment also starts normally; its first model request then fails with
HTTP 502 because the upstream request cannot be constructed.

## Reproduction

No API key or live upstream is required.

Create `routes.toml`:

```toml
schema_version = 1

[llm_clients.invalid]
format = "openai_chat"
base_url = "not a url"

[targets.invalid]
id = "upstream-model"
llm_client = "invalid"

[routes.invalid]
id = "test-route"
type = "passthrough"
target = "invalid"
```

Run the dry check:

```bash
cargo run --locked -p switchyard-server -- \
  --config routes.toml --dry-run
```

It exits 0 and prints:

```text
server OK: test-route
```

Start the same deployment:

```bash
cargo run --locked -p switchyard-server -- \
  --config routes.toml --host 127.0.0.1 --port 43173
```

Then send a request from another terminal:

```bash
curl -i http://127.0.0.1:43173/v1/chat/completions \
  -d '{"model":"test-route","messages":[{"role":"user","content":"hello"}]}'
```

The response is:

```text
HTTP/1.1 502 Bad Gateway

{"error":{"code":"upstream_configuration_error","message":"failed to build upstream request: builder error","type":"upstream_error"}}
```

## Expected vs. actual

- **Expected:** `--dry-run` rejects `llm_clients.invalid.base_url` as an invalid
  absolute HTTP(S) URL and identifies the affected client. This check should not
  require connecting to an upstream provider.
- **Actual:** The non-empty value passes configuration validation. Its invalid
  syntax is discovered only when the first request is sent.

## Environment

- Python version: N/A (Rust server path)
- Install path: source build with `cargo run --locked`
- Backend: none; the malformed URL prevents an upstream request from being built

## Additional context

[`build_backend`](https://github.com/NVIDIA-NeMo/Switchyard/blob/b256d936f1d77bf13ec9bec399ea0a253e07ca05/crates/switchyard-server/src/config.rs#L758-L768)
currently checks only whether `base_url` is empty. The URL reaches
[`reqwest::Client::post`](https://github.com/NVIDIA-NeMo/Switchyard/blob/b256d936f1d77bf13ec9bec399ea0a253e07ca05/crates/libsy-llm-client/src/client.rs#L272-L283)
when a model call is made.

A local syntax check during deployment construction would catch the error while
preserving `--dry-run`'s no-network behavior. Validating the resolved endpoint
with the same URL-joining rules used at request time would also avoid creating a
second interpretation of `base_url`.

I searched open and closed issues and pull requests for `base_url`, `dry-run`,
`invalid URL`, `builder error`, and `upstream_configuration_error`, and found no
matching report.

I'm happy to send a PR for this if the expected validation behavior sounds right.

ting-hong-shieh:
## What

- validate each configured client's resolved upstream endpoint during deployment construction
- require an absolute `http` or `https` URL with a host
- use the existing backend-specific endpoint joining before validation
- add config-level invalid URL cases and a process-level `--dry-run` regression test

## Why

An invalid `base_url` passed `--dry-run` and failed only when the first request tried to build its upstream HTTP request. Construction now rejects the same resolved endpoint that request-time code would use, without contacting the upstream.

Closes #398

## Validation

- `cargo test -p switchyard-server rejects_non_http_base_urls_during_construction -- --nocapture` — passed
- `cargo test -p switchyard-server --test cli dry_run_rejects_invalid_base_url -- --nocapture` — passed
- `cargo test -p switchyard-server` — passed
- `cargo fmt --all --check` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed
- `cargo test --workspace` — passed with loopback access enabled for the repository's local mock servers
- `git diff --check` — passed

No live provider calls or secrets were used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for backend endpoint URLs, requiring absolute HTTP or HTTPS URLs with a valid host.
  * Improved configuration error reporting for malformed, relative, or unsupported endpoint URLs.
  * Added command-line coverage to ensure invalid URLs cause dry-run startup failures.

### DIFF
--- a/crates/switchyard-server/src/config.rs
+++ b/crates/switchyard-server/src/config.rs
@@ -262,11 +262,42 @@ fn count_tokens_priority(target_name: &str, model_id: &ModelId) -> usize {
         .unwrap_or(3)
 }
 
+/// A client endpoint, parsed when the config loads rather than checked afterwards.
+///
+/// Holding a `HttpBaseUrl` is proof the value is an absolute HTTP(S) URL, so no
+/// later stage has to re-check it or can forget to.
+#[derive(Clone, Debug)]
+pub(crate) struct HttpBaseUrl(reqwest::Url);
+
+impl HttpBaseUrl {
+    pub(crate) fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl<'de> Deserialize<'de> for HttpBaseUrl {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let raw = String::deserialize(deserializer)?;
+        let url = reqwest::Url::parse(raw.trim()).map_err(|error| {
+            serde::de::Error::custom(format!("base_url must be an absolute HTTP(S) URL: {error}"))
+        })?;
+        if !matches!(url.scheme(), "http" | "https") || url.host_str().is_none() {
+            return Err(serde::de::Error::custom(
+                "base_url must be an absolute HTTP(S) URL",
+            ));
+        }
+        Ok(Self(url))
+    }
+}
+
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct LlmClientConfig {
     pub(crate) format: ClientFormat,
-    pub(crate) base_url: String,
+    pub(crate) base_url: HttpBaseUrl,
     api_key_env: Option<String>,
     #[serde(default)]
     forward_auth: bool,
@@ -913,12 +944,6 @@ fn build_backend(
     config: &LlmClientConfig,
     extra_body: &BTreeMap<String, Value>,
 ) -> ServerResult<Backend> {
-    let base_url = config.base_url.trim();
-    if base_url.is_empty() {
-        return Err(ServerError::new(format!(
-            "llm client {client_name} base_url must not be empty"
-        )));
-    }
     if config.max_retries > MAX_CONFIGURED_RETRIES {
         return Err(ServerError::new(format!(
             "llm client {client_name} max_retries must be at most {MAX_CONFIGURED_RETRIES}"
@@ -952,7 +977,7 @@ fn build_backend(
         })
         .transpose()?;
     let http = HttpBackendConfig {
-        base_url: base_url.to_string(),
+        base_url: config.base_url.as_str().to_string(),
         api_key,
         forward_auth: config.forward_auth,
         extra_headers: config.extra_headers.clone(),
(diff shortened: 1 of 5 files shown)


ting-hong-shieh:
## Before / after CLI snapshot

I ran the same invalid deployment through the real `switchyard-server --dry-run` binary at the PR base (`b256d936`) and current head (`be133106`).

Relevant input:

```toml
[llm_clients.invalid]
format = "openai_chat"
base_url = "not a url"
```

Before (`b256d936`):

```text
exit_code: 0
stdout: server OK: test-route
stderr:
```

The malformed URL passed construction and dry-run validation.

After (`be133106`):

```text
exit_code: 1
stdout:
stderr: invalid server config <config>: llm client invalid base_url must resolve to an absolute HTTP(S) URL: relative URL without a base
```

This was a local dry run. No socket was bound and no provider call or credential was involved.

crates/switchyard-server/src/config.rs
```diff
@@ -800,11 +800,30 @@ fn build_backend(
...
     };
-    Ok(match config.format {
+    let backend = match config.format {
         ClientFormat::OpenAiChat => Backend::OpenAiChat(http),
         ClientFormat::OpenAiResponses => Backend::OpenAiResponses(http),
         ClientFormat::AnthropicMessages => Backend::Anthropic(http),
-    })
+    };
+    validate_backend_url(client_name, &backend)?;
+    Ok(backend)
```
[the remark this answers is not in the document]
ting-hong-shieh: Addressed in 0fffe8ce.

Same-input before/after snapshot:

- Input: a syntactically invalid `base_url = "not a url"` on `[llm_clients.unused]`, with no target referencing that client.
- Before (`be133106`): the new regression test failed with `unexpected error: configuration unexpectedly succeeded`.
- After (`0fffe8ce`): `ServerConfig::build_clients` validates every declared client before target construction, and `test config::tests::rejects_invalid_unreferenced_llm_client ... ok` confirms the unused invalid client is rejected with the `llm client unused base_url` validation context.

Checks run:

- `cargo test --workspace`
- `cargo fmt --all --check`
- `git diff --check`

No provider call was made; this is config construction only.

crates/switchyard-server/src/config.rs
```diff
@@ -800,11 +802,30 @@ fn build_backend(
...
+}
+
+// Validate the endpoint after the backend applies the same format-specific URL
+// joining used for requests, so dry-run and request construction cannot diverge.
+fn validate_backend_url(client_name: &str, backend: &Backend) -> ServerResult<()> {
```
nachiketb-nvidia: this PR doesn't follow "parse, don't validate". With rust's type system, we can implement a custom URL type to make sure things are A-okay.

something like:

```
#[derive(Clone, Debug)]
  pub struct HttpBaseUrl(reqwest::Url);

  impl<'de> serde::Deserialize<'de> for HttpBaseUrl {
      fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
      where
          D: serde::Deserializer<'de>,
      {
          let raw = String::deserialize(deserializer)?;
          let url = reqwest::Url::parse(raw.trim()).map_err(serde::de::Error::custom)?;

          if !matches!(url.scheme(), "http" | "https") || url.host_str().is_none() {
              return Err(serde::de::Error::custom(
                  "base_url must be an absolute HTTP(S) URL",
              ));
          }

          Ok(Self(url))
      }
  }

  Then:

  struct LlmClientConfig {
      format: ClientFormat,
      base_url: HttpBaseUrl,
      ...
  }
```

crates/switchyard-server/src/config.rs
```diff
@@ -1097,6 +1118,37 @@ target = "weak"
         Ok(())
     }
 
+    #[test]
+    fn rejects_non_http_base_urls_during_construction() {
```
nachiketb-nvidia: we dont need this test, just keep the one in cli.rs

nachiketb-nvidia (changes requested):

nachiketb-nvidia:
@ting-hong-shieh a request, love the PRs, but can you type out the replies yourself rather than AI write them?

It will be better for everyone involved. Easier to read and understand for you and me. Thanks!

ting-hong-shieh:
@nachiketb-nvidia  I switched it to the type `HttpBaseUrl` and removed the extra test. Kept the CLI regression test.
Let me know if there are any remaining issues, thanks!

nachiketb-nvidia:
thank you!

nachiketb-nvidia (approved):

PR #411, opened 2026-08-13
fix: support json_object classifier responses

### LINKED ISSUE #409: stage_router classifier hardcodes response_format type=json_schema; incompatible with providers that only support json_object
## Problem

The stage_router LLM classifier (judge) **hardcodes** `response_format: { type: "json_schema" }` in every judge request. Many provider gateways (and some model backends) do not support `json_schema` structured output — they only support `response_format: { type: "json_object" }`. On those providers the judge call fails with HTTP 400, the verdict is dropped, and the router silently degrades to always picking the efficient tier (judge becomes dead weight — costs a call, never produces a verdict).

## Reproduction

Config:

```toml
[routes.code]
type = "stage_router"
capable_target = "pro"
efficient_target = "fast"
picker = "efficient_first"
confidence_threshold = 0.5

[routes.code.classifier]
target = "flash"          # model that supports json_object but NOT json_schema
base_threshold = 0.5
```

Server log:

```
WARN ... libsy: judge verdict unavailable; routing without one
  judge_model="deepseek-v4-flash"
  reason="upstream_non_5xx"
  error=client call to target "deepseek-v4-flash" failed: upstream returned HTTP 400:
  {"error":{"message":"This response_format type is unavailable now","type":"invalid_request_error"}}
```

The same model accepts `json_object` without issue:

```bash
# json_schema -> 400
curl ... -d '{"response_format":{"type":"json_schema","json_schema":{...}}}'
# -> {"error":{"message":"This response_format type is unavailable now"}}

# json_object -> 200 (with "json" in prompt)
curl ... -d '{"response_format":{"type":"json_object"}}'
# -> 200 OK
```

## Root cause (source-verified, v0.2.0)

The `json_schema` type is enforced at the contract layer with no config-level override:

1. **`algorithms/util/classifier_contract.rs`** — `from_inner_schema()` wraps every schema in a hardcoded `json_schema` envelope:

```rust
// line 72-83
let validator = compile_schema(&schema)?;
Self::from_response_format(
    prompt_template,
    json!({
        "type": "json_schema",          // <- hardcoded
        "json_schema": {
            "name": "switchyard_classifier_response",
            "strict": true,
            "schema": schema,
        }
    }),
    Some(validator),
)
```

2. **`from_response_format()`** requires the `/json_schema/schema` JSON pointer — a `json_object` response_format would fail validation:

```rust
// line 101-105
response_format
    .pointer("/json_schema/schema")
    .ok_or_else(|| LibsyError::AlgorithmError {
        message: "response schema has no json_schema.schema".to_string(),
    })?;
```

3. **`algorithms/llm_class.rs:33`** — the capability-classifier schema is compiled in via `include_str!`:

```rust
const SCHEMA_TEMPLATE: &str = include_str!("../prompts/capability-classifier/schema.json");
```

The schema file itself starts with `"type": "json_schema"`.

There is **no configuration field** on `ClassifierContractConfig`, `StageRouterConfig`, or the classifier sub-block in TOML to select `json_object` instead.

## Impact

Any target model whose provider only supports `response_format: { type: "json_object" }` **cannot serve as a stage_router classifier**. In practice this excludes models on providers that expose `json_object` but not `json_schema` — a common combination on OpenAI-compatible gateways in front of open-weight models.

When the judge model lacks `json_schema` support, the failure is **silent and costly**: every ambiguous turn still fires a judge call (which 400s), the verdict is discarded, and the router falls open to the default tier. The only signal is a `WARN` log line.

## Proposal

Allow the response_format type to be configured per classifier target, with a sensible default. For example:

```toml
[routes.code.classifier]
target = "flash"
base_threshold = 0.5
response_format_type = "json_object"   # or "json_schema" (default)
```

When `json_object` is selected:
- Send `response_format: { type: "json_object" }` instead of the `json_schema` wrapper.
- Still validate the parsed JSON against the schema client-side (the `Validator` already exists for `from_inner_schema`).
- Inject the schema description into the system prompt (or rely on the existing prompt which already instructs the judge on output shape).

This keeps `json_schema` as the default (strict, zero-prompt-overhead) while unblocking providers that only offer `json_object`.

## Environment

- switchyard-server / switchyard-libsy 0.2.0 (crates.io)
- Provider: OpenAI-compatible gateway (DeepSeek backend)
- Model: deepseek-v4-flash — accepts `json_object`, rejects `json_schema` with HTTP 400

ting-hong-shieh:
## What changed

- Add `response_format_type` to packaged classifier configuration, with `json_schema` as the unchanged default and `json_object` as an opt-in mode.
- Support the option in `stage_router` classifiers and in the built-in `llm_classifier` capability and escalation modes.
- In `json_object` mode, send `{"type":"json_object"}`, include the packaged verdict schema in the classifier prompt, and validate the returned object against that schema locally.
- Expose the setting through the Rust task-classifier config and Python bindings, and document the TOML options.
- Add mock-backed regression coverage for provider requests, prompt schema guidance, local schema rejection, fallback routing, Python configuration, and invalid Python values.

Custom classifier mode remains on its user-supplied JSON Schema contract. #429 tracks the separate contract decisions needed before adding JSON Object mode there.

## Why

Packaged classifiers always sent JSON Schema structured output. Providers that support JSON Object mode but not JSON Schema mode could not run those classifiers, and the server TOML parser rejected attempts to select another response format.

The default path remains unchanged for providers that support JSON Schema.

## Before and after

These results use the local mock upstream; no provider or LLM call is involved.

On `main`, adding the proposed TOML field fails during configuration parsing:

```text
unknown field `response_format_type`
```

With this change, the mock classifier request contains:

```json
{"response_format":{"type":"json_object"}}
```

The system prompt contains the packaged verdict schema. A mock verdict that is valid for the Rust type but violates that schema by adding an unexpected property is rejected locally, so the stage router follows its configured `model/weak` fallback instead of routing to `model/strong` from the supplied `p_solve` value.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace --exclude switchyard-py`
- `cargo test -p switchyard-server` — 32 unit and 27 integration tests passed
- `uv run ruff check .`
- `uv run mypy switchyard`
- `uv run pytest tests/ -q` — 138 passed, 2 skipped
- `make -C docs publish`

Direct `cargo test --workspace` reaches a macOS PyO3 extension-module linker error for `switchyard-py` because Python symbols are unavailable to the standalone test binary. The same binding builds through maturin and passes the Python suite above.

Closes #409.

## Summary by CodeRabbit

- **New Features**
  - Added configurable classifier response formats: JSON Schema (default) and JSON Object.
  - JSON Object mode supports providers without JSON Schema support by including guidance in the prompt and validating results locally.
  - Added configuration support for Python bindings, stage-router classifiers, and built-in capability and escalation routes.
  - Invalid response-format values now produce clear configuration errors.

- **Bug Fixes**
  - Improved handling of malformed classifier verdicts, preventing invalid results from affecting routing.

- **Documentation**
  - Documented response-format options and validation behavior.

### DIFF
--- a/switchyard_rust/libsy.py
+++ b/switchyard_rust/libsy.py
@@ -92,7 +92,7 @@ if TYPE_CHECKING:
             recent_turn_window: int | None = None,
             max_output_tokens: int = 4096,
             prompt: str | None = None,
-            response_format_type: str = "json_schema",
+            response_format_type: Literal["json_schema", "json_object"] = "json_schema",
         ) -> None: ...
 
     @final
--- a/switchyard_rust/libsy.py
+++ b/switchyard_rust/libsy.py
@@ -92,6 +92,7 @@ if TYPE_CHECKING:
             recent_turn_window: int | None = None,
             max_output_tokens: int = 4096,
             prompt: str | None = None,
+            response_format_type: str = "json_schema",
         ) -> None: ...
 
     @final
--- a/tests/test_libsy_minimal_bindings.py
+++ b/tests/test_libsy_minimal_bindings.py
@@ -110,6 +110,48 @@ async def test_classifier_config_accepts_a_prompt_override() -> None:
     assert response["model"] == "weak"
 
 
+async def test_classifier_config_accepts_json_object_output() -> None:
+    """Verify that Python can select JSON Object mode for a classifier judge."""
+
+    class JudgeClient(EchoClient):
+        async def call(self, request: dict[str, Any]) -> dict[str, Any]:
+            self.calls.append(request)
+            return {
+                "model": self.model,
+                "outputs": [
+                    {
+                        "role": "assistant",
+                        "content": [
+                            {
+                                "type": "text",
+                                "text": (
+                                    '{"crux":"bounded task","primary_rule":"SUP-1",'
+                                    '"capability_boundary":"supported","p_solve":0.9}'
+                                ),
+                            }
+                        ],
+                        "stop_reason": "end_turn",
+                    }
+                ],
+            }
+
+    judge = JudgeClient("judge")
+    algorithm = algorithms.llm_task_classifier(
+        LlmTarget("judge", judge),
+        LlmTarget("weak", EchoClient("weak")),
+        LlmTarget("strong", EchoClient("strong")),
+        config=TaskClassifierConfig(0.5, response_format_type="json_object"),
+    )
+
+    _, response = await algorithm.run(request_body())
+
+    assert judge.calls[0]["output"]["response_format"] == {"type": "json_object"}
+    prompt = judge.calls[0]["instructions"][0]["content"][0]["text"]
+    assert "JSON Schema" in prompt
+    assert '"p_solve"' in prompt
+    assert response["model"] == "weak"
+
+
 async def test_random_weights_and_seed_are_reproducible() -> None:
     def algorithm():
         return algorithms.random(
--- a/tests/test_libsy_minimal_bindings.py
+++ b/tests/test_libsy_minimal_bindings.py
@@ -207,14 +207,22 @@ async def test_classifier_config_accepts_json_object_output() -> None:
             }
 
     judge = JudgeClient("judge")
+    weak = EchoClient("weak")
     algorithm = algorithms.llm_task_classifier(
-        LlmTarget("judge", judge),
-        LlmTarget("weak", EchoClient("weak")),
-        LlmTarget("strong", EchoClient("strong")),
+        "judge",
+        "weak",
+        "strong",
         config=TaskClassifierConfig(0.5, response_format_type="json_object"),
     )
 
-    _, response = await algorithm.run(request_body())
+    _, response = await run_algorithm(
+        algorithm,
+        {
+            "judge": judge,
+            "weak": weak,
+            "strong": EchoClient("strong"),
+        },
+    )
 
     assert judge.calls[0]["output"]["response_format"] == {"type": "json_object"}
     prompt = judge.calls[0]["instructions"][0]["content"][0]["text"]
--- a/tests/test_libsy_minimal_bindings.py
+++ b/tests/test_libsy_minimal_bindings.py
@@ -231,6 +231,16 @@ async def test_classifier_config_accepts_json_object_output() -> None:
     assert response["model"] == "weak"
 
 
+def test_classifier_config_rejects_unknown_response_format() -> None:
+    invalid_response_format: Any = "yaml"
+
+    with pytest.raises(
+        ValueError,
+        match="response_format_type must be 'json_schema' or 'json_object'",
+    ):
+        TaskClassifierConfig(0.5, response_format_type=invalid_response_format)
+
+
 async def test_random_weights_and_seed_are_reproducible() -> None:
     def algorithm():
         return algorithms.random(
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -14,7 +14,9 @@ use switchyard_protocol::{ContentBlock, Decision, Message, ModelId, Role};
 use super::fall_through::{DefaultTarget, FallThrough};
 use super::util::DEFAULT_JUDGE_MAX_OUTPUT_TOKENS;
 use super::util::affinity::AffinityRouter;
-use super::util::classifier_contract::{ClassifierContract, ClassifierContractConfig};
+use super::util::classifier_contract::{
+    ClassifierContract, ClassifierContractConfig, ClassifierResponseFormat,
+};
 use super::util::escalation::{self, EscalationJudge, EscalationJudgeConfig, EscalationPolicy};
 use super::util::llm_judge::{
     ClassifierInput, JsonSchemaDecoder, JudgeClassifier, JudgePolicy, JudgeRuntimeConfig,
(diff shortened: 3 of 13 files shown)


hallelujah-shih:
Great work on this — exactly the gap I hit in #409.

One thing I noticed while reviewing the diff: the `response_format_type` field is added to `StageClassifierConfig` and wired through `task_classifier_config()`, but the `LlmClassifier` route variant in `crates/switchyard-server/src/config.rs` (the `LlmClassifierModeConfig::Capability` and `::Escalation` arms, around lines 806–832) still constructs its `TaskClassifierConfig` via the bare `classifier_contract(prompt)` helper, which does not forward `response_format_type`.

As a result, an `llm_classifier` route has no way to select `json_object` mode — the field would be silently ignored (or rejected as unknown if added at that TOML level). The libsy layer (`TaskClassifierConfigWire`) already supports the field; it is just not plumbed through the server-side `LlmClassifier` config path.

Would it be possible to also add `response_format_type` to the `LlmClassifier` route config and pass `.with_response_format_type(...)` at both construction sites (capability + escalation)? Happy to file a follow-up issue if you prefer to track it separately.

ting-hong-shieh:
Thanks for catching this, @hallelujah-shih. I agree it belongs in this PR rather than a follow-up.

Addressed in c5d6d3f0:

- Added response_format_type to standalone llm_classifier routes, defaulting to json_schema.
- Forwarded it through both capability and escalation classifier contracts.
- Kept custom mode on its configured JSON Schema instead of silently ignoring json_object.
- Added integration coverage for the exact json_object request and prompt-embedded verdict schemas, while retaining default json_schema coverage.
- Updated the TOML and routing documentation.

Validated with the full switchyard-server test suite (32 unit + 27 integration), workspace clippy with warnings denied, cargo fmt, and the strict documentation build.

ayushag-nv:
> The stage classifier always sent JSON Schema structured output. Providers that support JSON Object mode but not JSON Schema mode could not run the classifier, and the TOML parser rejected any attempt to select another response format.
> 
> The default path remains unchanged for providers that support JSON Schema.
@ting-hong-shieh why specific to stage router ? is this generalizable to normal classifier ?

ting-hong-shieh:
Yes. The current PR head generalizes this to the built-in `llm_classifier` route as well as `stage_router`.

`response_format_type` is supported by both the capability and escalation modes and is forwarded into their classifier contracts. The regression tests cover the exact `json_object` request and prompt-embedded verdict schema for both paths.

Custom mode is unchanged and continues to use its user-supplied JSON Schema contract. If you would also like `json_object` support for custom classifiers, I can include that in this PR.

ayushag-nv:
> Yes. The current PR head generalizes this to the built-in `llm_classifier` route as well as `stage_router`.
> 
> `response_format_type` is supported by both the capability and escalation modes and is forwarded into their classifier contracts. The regression tests cover the exact `json_object` request and prompt-embedded verdict schema for both paths.
> 
> Custom mode is unchanged and continues to use its user-supplied JSON Schema contract. If you would also like `json_object` support for custom classifiers, I can include that in this PR.

Also, seems like you are considering prompt schema to have json_schema thing in there. What if no response schema is there or just user mention json_object in the response schema thing and user hardcodes it in the prompt? Can you think around those cases as well.

switchyard_rust/libsy.py
```diff
@@ -92,6 +92,7 @@ def __init__(
             recent_turn_window: int | None = None,
             max_output_tokens: int = 4096,
             prompt: str | None = None,
+            response_format_type: str = "json_schema",
```
ayushag-nv: nit: Since only two values are accepted at runtime, could this use Literal["json_schema", "json_object"] instead of str?

tests/test_libsy_minimal_bindings.py
```diff
@@ -181,6 +181,56 @@ async def call(self, request: dict[str, Any]) -> dict[str, Any]:
     assert response["model"] == "weak"
 
 
+async def test_classifier_config_accepts_json_object_output() -> None:
```
ayushag-nv, line 184: nit: Could we add a small regression test confirming that an invalid response_format_type raises ValueError?

ayushag-nv:
@ting-hong-shieh I would suggest leave the PR in the current form and create a follow up issue. Don't want to overload this PR

ayushag-nv (approved):
rest looks good. Once you resolve comments, I will merge it. Thanks

tests/test_libsy_minimal_bindings.py
ting-hong-shieh: Added test_classifier_config_rejects_unknown_response_format in 9e3dcbee. It verifies that an unsupported value raises ValueError and checks the accepted-value message.

switchyard_rust/libsy.py
ting-hong-shieh: Updated in 9e3dcbee: response_format_type is now annotated as Literal["json_schema", "json_object"].

ting-hong-shieh:
@ayushag-nv Addressed both inline comments in 9e3dcbee and resolved the threads. The Python annotation now uses Literal for the two accepted values, and the new regression test verifies that an unsupported value raises ValueError. Full pytest passes with 138 tests and 2 skips; Ruff and mypy also pass. I opened #429 for the custom-classifier and prompt/schema contract questions so this PR stays scoped.

PR #412, opened 2026-08-13
feat(examples): update LiteLLM for run stream

nachiketb-nvidia:
## What

Update the experimental LiteLLM integration for the Python libsy `run_stream` API merged in #392.

## Why

The example still used the removed `LlmTarget` and managed `Algorithm.run()` APIs. It also created one LiteLLM client per model even though each `Step.CallModel` request already contains the selected model ID.

## How

- drive the Stage router directly with `async for` and `match Step...` inside `main()`
- pass model IDs directly to `algorithms.stage_router`
- use one model-agnostic `LiteLLMSyClient` for every model call
- read the selected LiteLLM alias from `call.request["model"]`
- migrate the focused unit, E2E, and README examples
- refresh the example lockfile against current `main`

## What to review

- the direct Python `run_stream` control flow in `example.py`
- one-client dispatch through the model already selected in each normalized request
- preservation of efficient and capable Stage-router behavior

## Validation

- `uv run ruff check examples/experimental/litellm`
- `uv run --project examples/experimental/litellm --locked --python 3.12 pytest examples/experimental/litellm/tests -m "not e2e" -v` (39 passed)
- live one-client `run_stream` call against the configured OpenAI-compatible endpoint (passed with a non-empty response)

## Summary by CodeRabbit

- **New Features**
  - LiteLLM routing examples now select models dynamically for each request.
  - Added streaming router workflows that handle decisions, model calls, failures, and completion.
  - A single LiteLLM client can now serve requests across multiple models.

- **Documentation**
  - Updated the LiteLLM README and examples to reflect model selection and streaming routing.

- **Tests**
  - Expanded coverage for dynamic model routing, shared client usage, request handling, and end-to-end responses.

### DIFF
--- a/examples/experimental/litellm/src/switchyard_litellm/client.py
+++ b/examples/experimental/litellm/src/switchyard_litellm/client.py
@@ -208,7 +208,10 @@ def _optional_mapping(
     return _mapping(value, field)
 
 
-def _payload(request: Mapping[str, object], model: str) -> dict[str, Any]:
+def _payload(request: Mapping[str, object]) -> dict[str, Any]:
+    model = request.get("model")
+    if not isinstance(model, str) or not model:
+        raise ValueError("model must be a non-empty string")
     if request.get("stream") is True:
         raise ValueError("stream=True is not supported")
     _reject_sequence_payload(request, "instructions")
@@ -226,7 +229,7 @@ def _payload(request: Mapping[str, object], model: str) -> dict[str, Any]:
         raise ValueError("reasoning.raw is not supported")
 
     payload: dict[str, Any] = {
-        "model": model,
+        "model": f"openai/{model}",
         "messages": _messages(request),
         "stream": False,
     }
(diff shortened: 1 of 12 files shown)


examples/experimental/litellm/src/switchyard_litellm/client.py
```diff
@@ -226,7 +229,7 @@ def _payload(request: Mapping[str, object], model: str) -> dict[str, Any]:
         raise ValueError("reasoning.raw is not supported")
 
     payload: dict[str, Any] = {
-        "model": model,
+        "model": f"openai/{model}",
```
grahamking, line 239: Hard to tell from the diff, but is this test code? Otherwise why prepend `openai` like this?

grahamking (approved):

examples/experimental/litellm/src/switchyard_litellm/client.py
nachiketb-nvidia: added a comment to explain the litellm quirk

PR #415, opened 2026-08-13
fix(translation): preserve chat reasoning details

pst2154:
## What changed

- preserve bare assistant reasoning when translating Responses history to Chat Completions
- retain and replay structured `reasoning_details` in buffered Chat requests and responses
- carry structured reasoning details through the provider-neutral streaming representation
- add regression coverage for tool-call history and buffered/streamed round trips

## Why

The Chat request encoder removed every normalized reasoning block before sending conversation history upstream. The Chat decoder also ignored `reasoning_details`, so structured or encrypted reasoning could not survive a subsequent turn. This particularly hurt reasoning models served through an `openai_chat` backend while the same workloads behaved correctly through `openai_responses`.

## Impact

Reasoning-capable Chat backends now receive their previous-turn reasoning context. Plaintext reasoning uses the compatible `message.reasoning` field, while structured details retain their original order and JSON shape.

## Validation

- `cargo clippy -p switchyard-protocol -p switchyard-translation --all-targets -- -D warnings`
- `cargo test -p switchyard-protocol -p switchyard-translation`
- `uv run ruff check .`
- `uv run pytest tests/ -q` (139 passed)
- live two-turn GLM-5.2 check through a Responses endpoint backed by OpenRouter Chat Completions

## Summary by CodeRabbit

* **New Features**
  * Preserved structured reasoning details across responses, streamed content, and format translations.
  * Added support for importing and exporting OpenAI reasoning details, including encrypted reasoning and signatures.
  * Added compatibility with Anthropic and Responses streaming formats, with readable reasoning text when available.
  * Retained legacy reasoning text as a fallback when structured details are unavailable.

* **Tests**
  * Added coverage for reasoning detail preservation, streaming, translation, and fallback behavior.

### DIFF
--- a/crates/protocol/src/llm.rs
+++ b/crates/protocol/src/llm.rs
@@ -87,7 +87,8 @@ pub enum ContentBlock {
         text: String,
         /// Provider signature used to validate or continue the reasoning block.
         signature: Option<String>,
-        /// Structured reasoning details that must be replayed without modification.
+        /// Structured reasoning details, such as an encrypted `{ "type":
+        /// "reasoning.encrypted", "data": "..." }` object, replayed without modification.
         #[serde(default, skip_serializing_if = "Vec::is_empty")]
         details: Vec<Value>,
         /// Plaintext fallback when structured details contain no displayable text.
--- a/crates/protocol/src/llm.rs
+++ b/crates/protocol/src/llm.rs
@@ -91,9 +91,6 @@ pub enum ContentBlock {
         /// "reasoning.encrypted", "data": "..." }` object, replayed without modification.
         #[serde(default, skip_serializing_if = "Vec::is_empty")]
         details: Vec<Value>,
-        /// Plaintext fallback when structured details contain no displayable text.
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        fallback_text: Option<String>,
     },
     /// Image content.
     Image {
--- a/crates/protocol/src/llm.rs
+++ b/crates/protocol/src/llm.rs
@@ -90,6 +90,9 @@ pub enum ContentBlock {
         /// Structured reasoning details that must be replayed without modification.
         #[serde(default, skip_serializing_if = "Vec::is_empty")]
         details: Vec<Value>,
+        /// Plaintext fallback when structured details contain no displayable text.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        fallback_text: Option<String>,
     },
     /// Image content.
     Image {
--- a/crates/protocol/src/llm.rs
+++ b/crates/protocol/src/llm.rs
@@ -87,6 +87,9 @@ pub enum ContentBlock {
         text: String,
         /// Provider signature used to validate or continue the reasoning block.
         signature: Option<String>,
+        /// Structured reasoning details that must be replayed without modification.
+        #[serde(default, skip_serializing_if = "Vec::is_empty")]
+        details: Vec<Value>,
     },
     /// Image content.
     Image {
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -193,11 +193,18 @@ impl AggLlmResponse {
                             text,
                         });
                     }
-                    ContentBlock::Reasoning { text, .. } => {
-                        chunks.push(LlmResponseChunk::ReasoningDelta {
-                            index: output_index,
-                            text,
-                        });
+                    ContentBlock::Reasoning { text, details, .. } => {
+                        if !details.is_empty() {
+                            chunks.push(LlmResponseChunk::ReasoningDetailsDelta {
+                                index: output_index,
+                                details,
+                            });
+                        } else {
+                            chunks.push(LlmResponseChunk::ReasoningDelta {
+                                index: output_index,
+                                text,
+                            });
+                        }
                     }
                     ContentBlock::ToolCall(tool) => {
                         let args = serde_json::to_string(&tool.arguments).unwrap_or_default();
@@ -272,6 +279,13 @@ pub enum LlmResponseChunk {
         /// Reasoning fragment.
         text: String,
     },
+    /// Adds structured reasoning details to one output index.
+    ReasoningDetailsDelta {
+        /// Provider output index.
+        index: usize,
+        /// Reasoning detail objects in provider order.
+        details: Vec<Value>,
+    },
     /// Adds or updates a tool call at one index.
     ToolCallDelta {
         /// Tool-call index within the response.
@@ -322,6 +336,7 @@ pub struct ResponseAccumulator {
     model: Option<String>,
     text: String,
     reasoning: Option<String>,
+    reasoning_details: Vec<Value>,
     tool_calls: BTreeMap<usize, PartialToolCall>,
     usage: Usage,
     stop_reason: Option<StopReason>,
@@ -359,6 +374,9 @@ impl ResponseAccumulator {
                     .get_or_insert_with(String::new)
                     .push_str(&text);
             }
+            LlmResponseChunk::ReasoningDetailsDelta { details, .. } => {
+                self.reasoning_details.extend(details);
+            }
             LlmResponseChunk::ToolCallDelta {
                 index,
                 id,
@@ -388,10 +406,11 @@ impl ResponseAccumulator {
     /// tool calls (by ascending delta index) — a single assistant output.
     pub fn finish(self) -> AggLlmResponse {
         let mut content = Vec::new();
-        if let Some(reasoning) = self.reasoning {
+        if self.reasoning.is_some() || !self.reasoning_details.is_empty() {
             content.push(ContentBlock::Reasoning {
-                text: reasoning,
+                text: self.reasoning.unwrap_or_default(),
                 signature: None,
+                details: self.reasoning_details,
             });
         }
         if !self.text.is_empty() {
@@ -611,6 +630,7 @@ mod tests {
                 ContentBlock::Reasoning {
                     text: "think".to_string(),
                     signature: None,
+                    details: Vec::new(),
                 },
                 ContentBlock::Text {
                     text: "answer".to_string(),
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -193,11 +193,17 @@ impl AggLlmResponse {
                             text,
                         });
                     }
-                    ContentBlock::Reasoning { text, details, .. } => {
+                    ContentBlock::Reasoning {
+                        text,
+                        details,
+                        fallback_text,
+                        ..
+                    } => {
                         if !details.is_empty() {
                             chunks.push(LlmResponseChunk::ReasoningDetailsDelta {
                                 index: output_index,
                                 details,
+                                fallback_text,
                             });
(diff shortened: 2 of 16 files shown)


crates/protocol/src/stream.rs
```diff
@@ -193,11 +193,18 @@ impl AggLlmResponse {
...
+                            chunks.push(LlmResponseChunk::ReasoningDetailsDelta {
+                                index: output_index,
+                                details,
+                            });
+                        } else {
+                            chunks.push(LlmResponseChunk::ReasoningDelta {
+                                index: output_index,
+                                text,
+                            });
+                        }
```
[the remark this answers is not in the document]
pst2154: Addressed in 350f837b. The neutral IR now retains fallback_text alongside structured details, aggregation preserves both values, OpenAI Chat round-trips the fallback for encrypted-only details, and Anthropic/Responses streaming emits it when the details have no displayable text. Added buffered, streaming, and aggregate regression coverage. Verified the protocol/translation tests, strict clippy, all non-PyO3 workspace tests, and 139 Python tests.

crates/switchyard-translation/src/codecs/common.rs
```diff
@@ -41,6 +41,22 @@ pub(crate) fn reasoning_text_from_blocks(content: &[ContentBlock], separator: &s
...
+        .filter_map(|detail| {
+            detail
+                .get("text")
+                .or_else(|| detail.get("summary"))
+                .and_then(Value::as_str)
+                .filter(|text| !text.is_empty())
+        })
+        .collect::<Vec<_>>();
+    (!parts.is_empty()).then(|| parts.join("\n"))
+}
```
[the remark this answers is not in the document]
pst2154: Addressed in 350f837b. reasoning_text_from_details now validates text as a non-empty string before falling back to a non-empty summary. Added a streaming regression test with an empty text field and a usable summary. Verified the protocol/translation tests, strict clippy, all non-PyO3 workspace tests, and 139 Python tests.

crates/switchyard-translation/src/codecs/openai_chat/stream.rs
```diff
@@ -105,14 +106,35 @@ fn decode_openai_chat_stream(
...
+                .get("reasoning_details")
+                .and_then(Value::as_array)
+                .filter(|details| !details.is_empty())
+            {
+                let fallback_text = if reasoning_text_from_details(details).is_none() {
+                    ["reasoning_content", "reasoning"]
+                        .into_iter()
+                        .find_map(|key| delta.get(key).and_then(Value::as_str))
+                        .filter(|text| !text.is_empty())
+                        .map(ToOwned::to_owned)
```
grahamking: This finds the first of those two fields that is populate? Like this:
```
for key in ["reasoning_content", "reasoning"] {
        if let Some(text) = delta.get(key).and_then(Value::as_str) {
                if !text.is_empty() {
                        result = Some(text.to_owned());
                        break;
                }
        }
}
```

Could you make a function like `let fallback_text = get_reasoning(&reasoning_details, ["reasoning_content", "reasoning"])` ?

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
```diff
@@ -804,6 +863,39 @@ fn encode_message_without_tool_results_to_openai(
...
+            .iter()
+            .filter_map(|block| match block {
+                ContentBlock::Reasoning { fallback_text, .. } => fallback_text.as_deref(),
+                _ => None,
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+        if !fallback.is_empty() {
+            message_json["reasoning"] = Value::String(fallback);
+        }
```
grahamking: Could you extract the `if` and `else` branch into a function? The name of the function will make it readable:
```
if reasoning_details.is_empty() {
  do_case_1()
} else {
  do_case_2()
}
```

It's a bit hard to wade through right now.

crates/switchyard-translation/src/codecs/common.rs
```diff
@@ -41,6 +41,27 @@ pub(crate) fn reasoning_text_from_blocks(content: &[ContentBlock], separator: &s
         .join(separator)
 }
 
+/// Extracts displayable text from structured reasoning details.
+pub(crate) fn reasoning_text_from_details(details: &[Value]) -> Option<String> {
```
grahamking, line 45: Like this! Very nice.

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
```diff
@@ -389,13 +396,52 @@ impl FormatCodec for OpenAiChatCodec {
...
 fn prepend_openai_reasoning_blocks(content: &mut Vec<ContentBlock>, object: &Map<String, Value>) {
+    if let Some(details) = object
+        .get("reasoning_details")
+        .and_then(Value::as_array)
+        .filter(|details| !details.is_empty())
+    {
+        let detail_text = reasoning_text_from_details(details);
+        let fallback_text = ["reasoning_content", "reasoning"]
+            .into_iter()
+            .find_map(|key| object.get(key).and_then(Value::as_str))
```
grahamking: This PR does a lot of extracting fields deep into the JSON. Would it make more sense to deserialize the whole thing once into an OpenAI compatible object?

grahamking: I recognize that's probably a larger change to switchyard-translation. Still, you can see it happening a lot right?

crates/protocol/src/llm.rs
```diff
@@ -87,6 +87,12 @@ pub enum ContentBlock {
         text: String,
         /// Provider signature used to validate or continue the reasoning block.
         signature: Option<String>,
+        /// Structured reasoning details that must be replayed without modification.
+        #[serde(default, skip_serializing_if = "Vec::is_empty")]
+        details: Vec<Value>,
```
grahamking, line 93: Can you give an example in the comment where reasoning is not a String?

grahamking:
Codex review suggested some simplifications - could you look into them?

 - Remove fallback_text from `ContentBlock::Reasoning` (https://github.com/NVIDIA-NeMo/Switchyard/blob/350f837b923bd9fbb1b67c2df4cc876d22519c87/crates/protocol/src/llm.rs#L84-L96). text already contains either extracted display text or the legacy fallback. Carry that text with `ReasoningDetailsDelta`, and let each translation boundary decide whether details already represent it. This eliminates duplicated state in the content block and accumulator.

 - Consolidate the request/response/stream test fixtures. Keep separate behavioral tests, but avoid repeating the full model, policy, detail objects, and tool-call payload in every test.

 - Remove the comments that merely restate test names. I know our AGENTS asks for them, we need to fix that.

crates/switchyard-translation/src/codecs/openai_chat/stream.rs
pst2154: Addressed in ad2e175e. I extracted `first_nonempty_string` and use it in both buffered and streaming Chat decoders. It selects the first populated alias and correctly continues to `reasoning` when `reasoning_content` exists but is empty; the encrypted-details stream regression now covers that case.

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
pst2154: Addressed in ad2e175e. The call site now delegates to `encode_openai_message_reasoning`, whose two branches are named `encode_openai_message_plaintext_reasoning` and `encode_openai_message_structured_reasoning`. This keeps the message assembly path short and makes the two wire representations explicit.

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
pst2154: Yes, the repeated field extraction is visible. I kept this bug-fix PR on the existing tolerant `Value`-based codec architecture because a full Chat DTO would be a broader change: these codecs intentionally accept provider extensions and preserve unknown fields. In ad2e175e I consolidated the repeated alias lookup into one checked helper. I would treat a complete typed wire model as a separate design change so its strictness and extension behavior can be reviewed independently.

crates/protocol/src/llm.rs
pst2154: Addressed in ad2e175e. The field comment now gives an encrypted structured detail as the concrete non-string example: `{ "type": "reasoning.encrypted", "data": "..." }`.

pst2154:
(Codex) I reviewed these suggestions and pushed 98da83e8.

- Removed `fallback_text` from `ContentBlock::Reasoning` and removed the duplicate fallback state from the accumulator. `ReasoningDetailsDelta` now carries normalized `text`, and each encoder decides whether structured details already represent it.
- Consolidated only fixtures that were genuinely identical across request/response/stream tests (model, normalized policy, structured details, and tool call) in `tests/common`; behavior-specific payloads and assertions remain separate.
- Removed the new test comments that merely repeated their test names, keeping only the comment that explains why preservation is disabled.

Validation: `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --exclude switchyard-py` (502 passed), `uv run ruff check .`, and `uv run pytest tests/ -v` (139 passed). The unexcluded workspace test still hits the existing local macOS PyO3 link failure for missing Python symbols.

ting-hong-shieh:
Cross-posting a measurement from #449 that bears on this PR's encode path. No claim on scope here — just data you would want before this lands.

@DuanZGit ran a matrix against a production DeepSeek thinking endpoint (`deepseek-v4-flash`), varying only the extra field on the assistant history message in a `user → assistant(tool_calls) → tool` sequence:

| extra field on assistant history message | result |
|---|---|
| (none) | 400 `reasoning_content in thinking mode must be passed back` |
| `reasoning` | 400, same message |
| `reasoning_content` | 200 |
| `reasoning_content: ""` | 200 |
| `reasoning_details: [...]` | 400, same message |
| unknown field | 400, same message |

Their full comment, including the redacted fixture: https://github.com/NVIDIA-NeMo/Switchyard/issues/449#issuecomment-5308413313

Reading this PR against that table: `encode_openai_message_plaintext_reasoning` writes `message["reasoning"]`. For an assistant message decoded from `reasoning_content` with no `reasoning_details`, `details` is empty, so that branch runs and the replayed history carries `reasoning` — row 2. Meanwhile `encode_response` in the same file still emits `reasoning_content`, so the request and response directions currently disagree on the field name.

If that reading is right, this PR fixes the round trip for targets that accept `reasoning` while DeepSeek-style thinking targets stay at 400. Worth saying precisely what is mine and what is not: the table is @DuanZGit's measurement against their endpoint, and my contribution is only the code reading above. I have not run this against a live endpoint.

Rows 5 and 6 also suggest DeepSeek does a presence check on `reasoning_content` rather than strict unknown-field validation, which may matter for how defensively the field needs to be gated.

For context, I have a branch for #449 that encodes `reasoning_content` with a capability gate and a `push_lossy` diagnostic on drop. I have asked in #449 whether to rebase it onto this PR once it lands rather than open anything competing, and my preference is to do exactly that. If you would rather carry the gate and the tests here, I am happy to hand them over instead.

grahamking (approved):

PR #420, opened 2026-08-14
fix(client): strip api-key and OpenAI org/project headers before forwarding

### LINKED ISSUE #419: Client api-key and OpenAI org/project headers are forwarded upstream
## Summary

`RESERVED_HEADERS` in `crates/libsy-llm-client/src/client.rs` strips `authorization` and `x-api-key` from forwarded client metadata, but does not strip three headers of the same class:

- `api-key` (Azure OpenAI's credential header, the `x-`-less sibling of `x-api-key`)
- `openai-organization` (OpenAI billing/tenant selector)
- `openai-project` (OpenAI project selector)

`switchyard-server`'s `metadata_from_headers` stores the entire inbound `HeaderMap`, and `send_once` runs `forward_metadata_headers` (copies every non-reserved client header) before `apply_auth` sets the backend credential. Because those three names are not reserved, a client-supplied copy rides upstream next to the backend's own key.

This is a trust-boundary gap: the point of the reserved list is that a caller cannot influence the credential or tenant identity used upstream. It works for `authorization` and `x-api-key`; it misses the three above.

## Impact

- **OpenAI backends**: a client `OpenAI-Organization` (or `OpenAI-Project`) is forwarded. An invalid value makes the upstream return `401 mismatched_organization`. A valid value that the backend key can access bills and attributes the request to that org or project instead of the intended one. In a multi-tenant deployment behind one shared key this lets one caller redirect billing or attribution within the key's scope.
- **Azure OpenAI backends** that authenticate with `api-key`: a client `api-key` is forwarded as the credential header. If the operator did not set `api-key` via `extra_headers`, the client-supplied value is what reaches upstream.
- On plain OpenAI and Gemini, `api-key` is ignored (the bearer wins), so there the observable effect is a client-controlled header leaving the proxy rather than an auth swap.

No operator secret is leaked to the client, and the org/project vector is bounded to tenants the backend key can already reach. Reported here with a fix; happy to route through `psirt@nvidia.com` instead if you prefer to treat it as a security report.

## Affected

`switchyard-server` 0.2.0 and current `main`. The `RESERVED_HEADERS` list is identical in both.

## Reproduction

Point a Switchyard passthrough route at a mock upstream that records inbound headers, configure the backend credential, and send the client headers.

Backend config (`sy.toml`):

```toml
schema_version = 1
[llm_clients.local]
format = "openai_chat"
base_url = "http://127.0.0.1:19999/v1"
api_key_env = "MOCK_KEY"
[targets.cap]
id = "captured-model"
llm_client = "local"
[routes.primary]
id = "captured-model"
type = "passthrough"
target = "cap"
```

Mock upstream that appends each request's headers to a file:

```python
import json, sys
from http.server import BaseHTTPRequestHandler, HTTPServer
CANNED = {"id":"x","object":"chat.completion","created":0,"model":"captured-model",
          "choices":[{"index":0,"message":{"role":"assistant","content":"pong"},"finish_reason":"stop"}],
          "usage":{"prompt_tokens":1,"completion_tokens":1,"total_tokens":2}}
class H(BaseHTTPRequestHandler):
    def do_POST(self):
        self.rfile.read(int(self.headers.get("content-length", 0)))
        with open("captured.jsonl", "a") as f:
            f.write(json.dumps({k.lower(): v for k, v in self.headers.items()}) + "\n")
        b = json.dumps(CANNED).encode()
        self.send_response(200); self.send_header("content-type","application/json")
        self.send_header("content-length", str(len(b))); self.end_headers(); self.wfile.write(b)
    def log_message(self, *a): pass
HTTPServer(("127.0.0.1", 19999), H).serve_forever()
```

Run:

```bash
python3 mock.py &
MOCK_KEY="OFFICE_KEY" cargo run -p switchyard-server -- --config sy.toml --port 19000 &

BODY='{"model":"captured-model","messages":[{"role":"user","content":"ping"}]}'
curl -s -o /dev/null localhost:19000/v1/chat/completions -H 'content-type: application/json' -H 'OpenAI-Organization: CANARY_ORG' -d "$BODY"
curl -s -o /dev/null localhost:19000/v1/chat/completions -H 'content-type: application/json' -H 'OpenAI-Project: CANARY_PROJECT' -d "$BODY"
curl -s -o /dev/null localhost:19000/v1/chat/completions -H 'content-type: application/json' -H 'api-key: CANARY_AZURE_KEY' -d "$BODY"
curl -s -o /dev/null localhost:19000/v1/chat/completions -H 'content-type: application/json' -H 'x-api-key: CANARY_X_API_KEY' -d "$BODY"
```

Observed in `captured.jsonl` (repeated 5/5 per header):

| Client header | Reaches upstream? |
|---|---|
| `OpenAI-Organization: CANARY_ORG` | yes |
| `OpenAI-Project: CANARY_PROJECT` | yes |
| `api-key: CANARY_AZURE_KEY` | yes |
| `x-api-key: CANARY_X_API_KEY` | no (stripped, as intended) |

The upstream `authorization` stays `Bearer OFFICE_KEY` in every case, confirming the leak is the extra header riding alongside the backend credential.

## Fix

Add `api-key`, `openai-organization`, and `openai-project` to `RESERVED_HEADERS` so `forward_metadata_headers` drops the client copy. An operator that legitimately needs one of these upstream still sets it through the backend's `extra_headers`, which are applied after forwarding and therefore unaffected. PR with the change and a regression test follows.

Atharva-Kanherkar:
Fixes #419.

## What

`RESERVED_HEADERS` in `crates/libsy-llm-client/src/client.rs` strips `authorization` and `x-api-key` from forwarded client metadata, but not three headers of the same class:

- `api-key`, Azure OpenAI's credential header (the `x-`-less sibling of `x-api-key`)
- `openai-organization` and `openai-project`, OpenAI's billing and project selectors

`forward_metadata_headers` copies every non-reserved client header onto the upstream request, so a caller can set those three next to the backend's own credential. An invalid `openai-organization` makes OpenAI return `401 mismatched_organization`; a valid one bills a different org the key can reach; on an Azure deployment a client `api-key` is forwarded as the credential header.

This PR adds the three names to `RESERVED_HEADERS`. An operator that legitimately needs one of them upstream still sets it through the backend's `extra_headers`, which are applied after forwarding and are unaffected.

## Change

- Add `api-key`, `openai-organization`, `openai-project` to `RESERVED_HEADERS`.
- Refresh the comment above the list to describe the intent (the old comment pointed at a Python `_SENSITIVE_HEADERS` list that is not present in this tree).
- Add `strips_client_credential_and_tenant_headers`, a regression test that sends all three as client headers and asserts none reach the upstream while the backend bearer is preserved.

## Test

```
cargo test -p switchyard-llm-client
```

All existing tests pass; the new test and the existing `forwards_metadata_headers_except_reserved` both cover the reserved set. Reproduction steps for the original behavior are in #419.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented caller-provided metadata from overriding protected Azure and OpenAI authentication headers.
  * Ensured backend authorization remains intact while tenant and credential selector headers are filtered.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -29,6 +29,7 @@ use crate::raw::RawResponse;
 
 // Headers this client owns or that are hop-by-hop. Backends apply an explicitly
 // enabled caller credential after generic metadata forwarding skips these.
+// Azure/OpenAI credentials and tenant selectors are also backend-owned.
 const RESERVED_HEADERS: &[&str] = &[
     "host",
     "content-length",
@@ -41,6 +42,9 @@ const RESERVED_HEADERS: &[&str] = &[
     "x-api-key",
     "chatgpt-account-id",
     "x-openai-fedramp",
+    "api-key",
+    "openai-organization",
+    "openai-project",
     "anthropic-beta",
     "anthropic-version",
     "content-type",
@@ -1807,6 +1811,18 @@ mod tests {
             "accept-encoding",
             http::HeaderValue::from_static("gzip, br"),
         );
+        headers.insert(
+            "api-key",
+            http::HeaderValue::from_static("client-azure-key"),
+        );
+        headers.insert(
+            "openai-organization",
+            http::HeaderValue::from_static("org-client"),
+        );
+        headers.insert(
+            "openai-project",
+            http::HeaderValue::from_static("proj-client"),
+        );
         let request = Request {
             llm_request: LlmRequest {
                 model: Some("gpt".to_string()),
@@ -1836,6 +1852,9 @@ mod tests {
             .ok_or("request recording should be enabled")?;
         let received = received.first().ok_or("expected one upstream request")?;
         assert!(!received.headers.contains_key("accept-encoding"));
+        assert!(!received.headers.contains_key("api-key"));
+        assert!(!received.headers.contains_key("openai-organization"));
+        assert!(!received.headers.contains_key("openai-project"));
         Ok(())
     }
 


crates/libsy-llm-client/src/client.rs
```diff
@@ -1813,6 +1819,60 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn strips_client_credential_and_tenant_headers()
```
nachiketb-nvidia: we can probably remove this test, and put these checks in an existing test which tests for the other headers, net new tests need more maintainence

crates/libsy-llm-client/src/client.rs
```diff
@@ -42,6 +41,13 @@ const RESERVED_HEADERS: &[&str] = &[
     "cookie",
     "set-cookie",
     "x-api-key",
+    // `api-key` is Azure OpenAI's credential header (the `x-`-less sibling of
```
nachiketb-nvidia: nit: overcommenting makes the code harder to read even for agents. pls put these comments as a one-liner in the comments above RESERVED_HEADERS

nachiketb-nvidia (changes requested):
I have a few nits and needs rebase, but otherwise LGTM!
thank you very much for the PR!

crates/libsy-llm-client/src/client.rs
Atharva-Kanherkar: thats right, lol. thanks for the review!

crates/libsy-llm-client/src/client.rs
Atharva-Kanherkar: thats informative too.

Atharva-Kanherkar:
@nachiketb-nvidia thank you for your review. switchyard is awesome, i use it with opencode + gemini + some custom setup. it has some bugs, but its obvious for a standard new dev project. nice work, please keep this project going:)
ill try to file bugs here and there more, as I feel that there might be some. thanks! <3

nachiketb-nvidia (approved):

PR #428, opened 2026-08-14
fix(server): redact transport error details

### LINKED ISSUE #423: [bug]: transport 502 echoes base_url including ?key=
## symptom
switchyard-server 0.2.0 copies reqwest's transport error into the client 502. that string includes the full upstream url. if the admin put the provider key in `base_url` as `?key=` (google's documented gemini shape), the caller gets the key.

header bearer keys and http userinfo (`http://key@host`) are not echoed. query params are.

## reproduction
no server listening on 19998.

```toml
schema_version = 1
[llm_clients.local]
format = "openai_chat"
base_url = "http://127.0.0.1:19998/v1?key=CANARY_ADMIN_QUERY_KEY"
authorization = "Bearer CANARY_ADMIN_HEADER_KEY"
[targets.cap]
id = "captured-model"
llm_client = "local"
[routes.primary]
id = "captured-model"
type = "passthrough"
target = "cap"
```

```bash
switchyard-server --config sy.toml --port 9000
  -d '{"model":"captured-model","messages":[{"role":"user","content":"x"}],"max_tokens":8}'
```

actual: http 502. `error.message` is:

```
error sending request for url (http://127.0.0.1:19998/v1?key=CANARY_ADMIN_QUERY_KEY/chat/completions)
```

5/5. a 502 whose `base_url` has no query string does not contain the canary. a successful chat does not either. `/health`, `/v1/models`, `/v1/stats` stay clean.

## expected vs. actual
- expected: client-visible errors do not contain credentials from `base_url`.
- actual: `?key=` stays in the url string. the extra_headers bearer is not echoed.

## environment
- os / arch: macos arm64
- inbound format (chat completions / anthropic messages / responses): chat completions, also stream and `/v1/messages`
- backend (openai / anthropic / nvidia inference hub / other): closed local port (transport failure). live header-auth chats to openai/gemini/anthropic/openrouter did not echo the bearer.

## additional context
`client_error` maps `LlmClientError::Transport` / `Timeout` to `source.to_string()`. reqwest includes the url. there is no redaction step. `HttpBackendConfig`'s `Debug` impl redacts `api_key`, but that is not the path the client sees.

this is a canary-only report. no real provider keys in the public captures.

wire captures and replay tests: https://github.com/Atharva-Kanherkar/kairo/pull/6

ting-hong-shieh:
## What

- Return fixed client-visible messages for upstream transport and timeout failures instead of exposing their source errors.
- Preserve the existing HTTP status, error type, and error code for both classifications.
- Add regression coverage for a credential-bearing upstream URL, timeout sanitization, and unchanged non-secret response-decoding details.

## Why

`client_error` previously copied `LlmClientError::Transport` and `LlmClientError::Timeout` source strings into HTTP error responses. Reqwest transport errors may include the complete request URL, so credentials stored in `base_url` query parameters could be returned to callers.

The server now treats transport and timeout sources as internal details at the client-facing error boundary. This is generic and does not depend on a provider or query-parameter name.

Closes #423

## How tested

- `cargo test -p switchyard-server client_ -- --nocapture`
- `cargo test -p switchyard-server --test server transport_error_hides_credential_bearing_upstream_url -- --exact --nocapture`
- `cargo test -p switchyard-server --test server request_and_upstream_errors_use_the_inbound_wire_format -- --exact --nocapture`
- `git diff --check`

All tests use local or synthetic errors with `CANARY_ADMIN_QUERY_KEY`. No provider endpoint or real credential was used.

## Notes for reviewers

- Transport failures remain HTTP 502 with `upstream_error` type and code.
- Timeout failures remain HTTP 504 with `upstream_error` type and `upstream_timeout` code.
- `InvalidResponse` continues to expose its source detail unchanged.
- No new logging was added.

## Summary by CodeRabbit

- **Bug Fixes**
  - Upstream connection failures now display a generic `502 Bad Gateway` message without exposing request URLs or credentials.
  - Timeout failures now return a standardized gateway-timeout message.
  - Invalid-response errors continue to provide relevant source details.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -1033,22 +1033,32 @@ fn client_error(error: &LlmClientError) -> Response {
             "upstream_error",
             "upstream_error",
         ),
-        LlmClientError::Transport { source } | LlmClientError::InvalidResponse { source } => {
-            error_response(
-                StatusCode::BAD_GATEWAY,
-                source.to_string(),
-                "upstream_error",
-                "upstream_error",
-            )
-        }
+        // A transport source is reqwest's, and it renders the full request URL, including any
+        // credentials configured in its query string.
+        LlmClientError::Transport { source } => redacted_error_response(
+            StatusCode::BAD_GATEWAY,
+            "upstream transport error",
+            source.to_string(),
+            "upstream_error",
+            "upstream_error",
+        ),
+        // Decoding failures describe the response body, not the request target.
+        LlmClientError::InvalidResponse { source } => error_response(
+            StatusCode::BAD_GATEWAY,
+            source.to_string(),
+            "upstream_error",
+            "upstream_error",
+        ),
         LlmClientError::ResponseTranslation(message) => error_response(
             StatusCode::BAD_GATEWAY,
             message,
             "upstream_error",
             "upstream_error",
         ),
-        LlmClientError::Timeout { source } => error_response(
+        // A timeout carries the same request context as a transport error.
+        LlmClientError::Timeout { source } => redacted_error_response(
             StatusCode::GATEWAY_TIMEOUT,
+            "upstream request timed out",
             source.to_string(),
             "upstream_error",
             "upstream_timeout",
@@ -1078,6 +1088,9 @@ struct ApiError {
     message: String,
     error_type: &'static str,
     code: &'static str,
+    // Detail for the request log only. Errors whose source can carry the configured upstream
+    // URL keep the full text here and send a fixed message on the wire.
+    log_detail: Option<String>,
 }
 
 impl ApiError {
@@ -1092,7 +1105,13 @@ impl ApiError {
             message: message.into(),
             error_type,
             code,
+            log_detail: None,
         }
+    }
+
+    fn with_log_detail(mut self, detail: impl Into<String>) -> Self {
+        self.log_detail = Some(detail.into());
+        self
     }
 
     fn into_response(self, wire_format: WireFormat) -> Response {
@@ -1113,9 +1132,11 @@ impl ApiError {
             }),
         };
         let mut response = (self.status, Json(body)).into_response();
-        response
-            .extensions_mut()
-            .insert(RequestLogError(self.message.clone()));
+        response.extensions_mut().insert(RequestLogError(
+            self.log_detail
+                .clone()
+                .unwrap_or_else(|| self.message.clone()),
+        ));
         response.extensions_mut().insert(self);
         response
     }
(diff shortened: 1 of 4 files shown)


ting-hong-shieh:
@coderabbitai review

crates/switchyard-server/src/lib.rs
```diff
@@ -867,23 +867,30 @@ fn client_error(error: &LlmClientError) -> Response {
...
-                source.to_string(),
-                "upstream_error",
-                "upstream_error",
-            )
-        }
+        // Transport sources may include the full request URL, including configured
+        // query credentials. Keep transport details out of client-visible errors.
+        LlmClientError::Transport { .. } => error_response(
+            StatusCode::BAD_GATEWAY,
+            "upstream transport error",
```
nachiketb-nvidia, line 1040: we shouldn't let go of the `source`, we can format it directly into the error message through string formatting

nachiketb-nvidia (changes requested):

grahamking:
We're doing this at the `switchyard-runner` level now: https://github.com/NVIDIA-NeMo/Switchyard/pull/537

Would that cover your use case @ting-hong-shieh ? In general redaction is the responsibility of the integration not of Switchyard, and at this early stage of the project clear errors are very important, so I think this could be closed.

ting-hong-shieh:
#537 covers a different sink. It is a runner-level summary for plugins; this PR is
about the HTTP error body `switchyard-server` returns to the caller, which an
integration cannot rewrite.

That body still carries the key on today's `main` (27fc1ce9), four days after #537
merged. Running the regression test against unmodified `main`:

```
credential leaked in "error sending request for url
(http://127.0.0.1:45673/v1?key=CANARY_ADMIN_QUERY_KEY/chat/completions)"
```

@nachiketb-nvidia was right that the first diff dropped the source. `error_response`
fed one string to both the wire body and `RequestLogError`, so redacting the message
redacted the operator log with it. Those are now split: the client gets a fixed
message, the request log keeps the full source. Clear errors where they are needed,
nothing for the caller.

Checking the rest of the class rather than just the reported path, the same leak
exists on the streaming side. `bytes_stream()` failures are wrapped as `Transport`
in `libsy-llm-client`, and `sse.rs` rendered that straight into the SSE event, past
the `client_error` boundary. Fixed there too, with the full text still going to the
`tracing::warn!` the path already emitted. The redaction downcasts the boxed stream
error back to `LlmClientError`, so I pinned that boxing behaviour with a test in
`switchyard-translation` — otherwise a later change there would silently undo this.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
and `cargo test --workspace` are clean.

Happy to close if you would still rather redaction live in the integration, but the
HTTP body and the SSE frames are not reachable from there.

ting-hong-shieh:
Closing this, per @grahamking's point that redaction belongs to the integration and that clear errors matter more at this stage.

For the record, the behaviour this addressed is still on `main` as of 27fc1ce9: when the upstream URL carries a credential and the transport fails, `switchyard-server` returns that URL, key included, in the HTTP error body. #537 covers the runner-side route summary, not that body. If it becomes worth handling later, the regression test in this branch (`crates/switchyard-server/tests/server.rs`) reproduces it directly.

Thanks for the review, @nachiketb-nvidia.

PR #430, opened 2026-08-14
fix(server): use normalized session IDs in routing stats

afourniernv:
## What

- build routing-log context from protocol-normalized `Metadata.session_id`, so the canonical `x-switchyard-session-id` and recognized client session headers are recorded consistently with routing
- prefer the normalized session ID when canonical and legacy headers disagree
- preserve `proxy_x_session_id` as a routing-log-only fallback for existing benchmark and legacy clients
- keep the public API, configuration, and routing-log JSON schema unchanged

## Why

Fixes [SWITCH-1191](https://linear.app/nvidia/issue/SWITCH-1191/oss-020observability-x-switchyard-session-id-is-not-recorded-in-native).

I independently reproduced the reported behavior on untouched `origin/main` at `a17efa94` before applying the fix:

```text
canonical_status=404 Not Found
legacy_status=200 OK
recorded_session_id="legacy-session"
```

This was not a later regression from previously working canonical session stats. The native session-stats feature initially shipped with a separate legacy-header parser, while request routing already consumed normalized protocol metadata. This change makes routing stats use that existing normalized value instead of parsing session identity a second time.

## How tested

- [x] pre-fix reproduction on untouched `origin/main`: canonical lookup `404`, legacy lookup `200`, JSONL recorded the legacy ID
- [x] `cargo test -p switchyard-server --test server routing_log` — 2 passed, covering canonical precedence, legacy-only compatibility, and canonical session lifetime through streaming completion
- [x] `cargo test -p switchyard-server` — 32 unit and 27 integration tests passed
- [x] `cargo test --workspace --exclude switchyard-py`
- [x] `git diff --check`

No live provider call was needed; the failure is entirely in inbound metadata normalization and terminal routing-log recording, and the reproduction uses the local mock upstream.

## Checklist

- [x] Regression tests cover the bug fix and legacy compatibility.
- [x] README updated for the customer-facing session-ID contract.
- [x] Commit carries the required DCO sign-off.

## Notes for reviewers

The normalized session `String` is cloned into the independently owned routing-log context because request metadata moves into the algorithm while streaming logging may outlive request execution. The legacy header remains log-only so it cannot influence affinity, escalation, or overflow state.

## Summary by CodeRabbit

- **New Features**
  - Session statistics now support canonical session IDs through the `x-switchyard-session-id` header.
  - Statistics remain compatible with the legacy `proxy_x_session_id` header when no canonical ID is provided.
  - Streaming responses record final session usage totals after the response completes.

- **Bug Fixes**
  - Routing logs and session statistics now consistently use the canonical session ID when both headers are present.
  - Updated documentation explains model-grouped totals and behavior when records or routing logs are unavailable.

### DIFF
--- a/crates/switchyard-server/README.md
+++ b/crates/switchyard-server/README.md
@@ -88,8 +88,10 @@ for equal weighting. The optional `seed` reproduces the selection sequence for t
 Pass `--routing-log-file PATH` to append one JSON record after each completed routed response.
 Streaming responses are recorded after the stream drains. When enabled,
 `GET /v1/routing/session-stats?session_id=ID` rescans the durable log and returns call and token
-totals for that exact `proxy_x_session_id`, grouped by served model. The endpoint returns `404` when
-the session has no records and is not registered when routing logging is disabled.
+totals for that normalized session ID, normally supplied as `x-switchyard-session-id`, grouped by
+served model. The legacy `proxy_x_session_id` remains a fallback when no normalized session ID is
+present. The endpoint returns `404` when the session has no records and is not registered when
+routing logging is disabled.
 
 An `llm_classifier` route sends each task to `classifier_target` for a capability verdict, then
 routes to `weak_target` or `strong_target`. Beyond the three targets it accepts these keys; only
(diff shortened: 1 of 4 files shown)


afourniernv:
@grahamking when you have a chance, could you take a first pass on this? I reproduced SWITCH-1191 on untouched main before applying the fix; the exact pre-fix behavior and the permanent regression coverage are in the PR description. Thanks!

grahamking (approved):

grahamking:
Thanks for the fix!

PR #431, opened 2026-08-14
feat(libsy-llm-client): Move retry logic from libsy to libsy-llm-client

grahamking:
This PR moves LLM calling retry code from `crates/libsy/src/algorithms/fall_through.rs` to `crates/libsy-llm-client/src/run.rs`, because the HTTP work belong in `libsy-llm-client`. 

`FallThrough` holds almost no state now.

Previously retry was handled by the `FallThrough` algorithm wrapper. It would issue a `CallModel`, inspect the HTTP response, and then ask the algorithm for a different model on failure. Now that retry is outside `libsy`, `FallThrough` returns to it's design goal of being a set of classifiers with "fall through" if an earlier one does not select a model.

For the client to retry we need a list of ModelId to try. `Driver::call_model` now takes `Vec<ModelId>`, which are the models the algorithm wants us to try in order, the selected one first. Algorithms behave like this already, but the multiple-models is somewhat hidden by the interaction with FallThrough. This makes it explicit.

Note this is not compatible with `AffinityRouter`, which also needs to migrate out of `libsy`. Later.

AI usage: GPT 5.6 Sol high and Opus 5 high generated a plan each. Grok 4.6 xhigh compared them. GPT 5.6 Sol medium implemented. Sonnet 5 high reviewed. With lots of manual stuff in between and during.

## Summary by Sonnet 5

This PR relocates model-call retry/fallback from the `FallThrough` algorithm (`crates/libsy`) into the HTTP client layer (`crates/libsy-llm-client/src/run.rs`). Previously `FallThrough` tracked excluded targets, remembered context-window overflows for a session, and republished a new `Decision` on every fallback hop. Now `Driver::call_model` takes an ordered `Vec<ModelId>`, and `libsy-llm-client`'s `call_first_available` tries each candidate in turn, exhausting each one's HTTP retry budget before advancing.

`FallThrough` is reduced to producing that ordered candidate list. Rust/Python bindings, docs, and tests are all updated to match — it's a large, mostly-mechanical refactor (21 files, +583/-824) with good new test coverage for the moved logic (run.rs gets 4 solid new tests covering fallback policy, retry-budget exhaustion, and stream boundaries).


grahamking:
@coderabbit review

crates/libsy/src/core/algorithm.rs
```diff
@@ -85,25 +66,25 @@ impl CallModel {
     /// Take the call's contents without answering it, dropping the promise — the routing
     /// outcome plus the request as the algorithm would have sent it, after any rewriting.
     ///
-    /// Should only be called if `decision.is_answer_call` is true as that is the final call.
+    /// Should only be called if `is_answer_call` is true as that is the final call.
```
nachiketb-nvidia, line 69: if it should be never called, should we make it so this state is unrepresentable?

grahamking: Yes! In the upcoming `RoutingOutcome` change we remove `into_parts` and `is_answer_call`. A `CallModel` becomes always a call the algo needs to decide, never a final decision. We drop `Driver::decide` and `Step::Decision`.

We'll end up with just `Step::CallModel` (algo still thinking) and `Step::Done(RoutingOutcome)` (algo decided).

The `RoutingOutcome` has the ordered list of model IDs to call and the `Request` to send (already stamped with the model ID). Optionally it has the `Response` too for escalation router if it's already generated it as part of the decision.

But to do all of that I needed to split retries out of `FallThrough` first, which is this PR.

nachiketb-nvidia (approved):

PR #439, opened 2026-08-15
fix: preserve JSON object key order in proxied payloads

### LINKED ISSUE #438: Proxied JSON payloads are re-serialized with alphabetized object keys, breaking order-enforcing structured-output backends
## Summary

When a request passes through the router, every JSON object in the payload is re-serialized with its keys sorted alphabetically (serde_json's default `Map` is BTreeMap-backed; the workspace does not enable the `preserve_order` feature). Array order is preserved; only object key order changes.

For most fields this is cosmetic. For `response_format.json_schema.schema` it is semantic: order-enforcing structured-output backends (vLLM with its default xgrammar backend) use the `properties` declaration order as the *generation order*. Reordering the schema therefore changes what the model is forced to generate.

I'm opening a companion PR alongside this issue — the fix is a one-line `Cargo.toml` change — so the details below double as its rationale.

## Reproduction (model-independent)

1. Configure a `passthrough` route to any vLLM target started with `--enable-log-requests`.
2. Send a chat completion with a `json_schema` response format whose properties are deliberately non-alphabetical:

   ```json
   "properties": {
     "zebra_field": {"type": "string"},
     "apple_field": {"type": "string"},
     "mango_field": {"type": "string"}
   },
   "required": ["zebra_field", "apple_field", "mango_field"]
   ```

3. vLLM's request log shows what the router actually forwarded:

   ```
   StructuredOutputsParams(json={'additionalProperties': False, 'properties':
   {'apple_field': ..., 'mango_field': ..., 'zebra_field': ...},
   'required': ['zebra_field', 'apple_field', 'mango_field'], ...})
   ```

   The `properties` map is alphabetized while the `required` array keeps its order, and the model's output follows the imposed alphabetical order.

Observed on the v0.2.0 release image; the serialization path is unchanged on current `main`.

## Impact we measured

We run a LoRA fine-tuned capability judge (Nemotron 3.5 Lightning) as an `llm_classifier` target on a local vLLM. The model was fine-tuned to emit its verdict keys in a fixed order ending with `p_solve`. Behind the router, the alphabetized schema forces a different key order, and in 36/200 judgments (18%) the model stalled after the third key and emitted whitespace until `max_tokens` (the classic grammar/EOS deadlock), producing an invalid verdict and a fail-open on every occurrence. Interposing a 30-line proxy that only restores the original key order dropped this to 0/100 with fully stable decisions, which isolates key order as the cause. Even without a stall, forcing `capability_boundary` to be generated before `crux` measurably changed verdicts.

A generalist judge model is robust to the reordering, which is presumably why this has gone unnoticed.

## Fix (companion PR)

The PR enables serde_json's `preserve_order` feature at the workspace level, so the router forwards object keys as the client sent them. It is one line in `Cargo.toml` plus the lockfile (`indexmap` is already in the dependency tree), and `cargo test --workspace` passes. If you'd rather avoid the workspace-wide feature — e.g., carry passthrough JSON as `RawValue` so the router never re-serializes user payloads — I'm happy to rework in that direction.

## Environment

- switchyard-server 0.2.0 (release image), config: `llm_classifier` + `passthrough` routes
- Backend: vLLM v0.27.1 (`--served-model-name judge`), default structured outputs backend (xgrammar), DGX Spark (aarch64)

himorishige:
Fixes #438.

## What

Enables serde_json's `preserve_order` feature for the workspace, so JSON objects in proxied payloads are forwarded with their keys in the order the client sent them instead of alphabetized (the BTreeMap default). One line in `Cargo.toml`, plus the lockfile update (`indexmap` was already in the dependency tree).

## Why

Object key order is semantic for `response_format.json_schema`: order-enforcing structured-output backends (vLLM with its default xgrammar backend) use the `properties` declaration order as the generation order. The router currently re-serializes every object with alphabetized keys, silently changing what downstream models are forced to generate. A model-independent repro is in #438.

With an order-sensitive fine-tuned judge behind an `llm_classifier` route we measured an 18% judgment-runaway rate (grammar/EOS whitespace deadlock at `max_tokens`, each occurrence an invalid verdict and a fail-open); restoring key order drops it to zero.

## Verification

- `cargo test --workspace` passes with the feature enabled
- Passthrough probe: a schema sent as `zebra/apple/mango` reaches the backend in that order (previously `apple/mango/zebra`)
- E2E: an identical 100-judgment probe through the router against a local vLLM target went from 36/200 runaways — with all 10 request shapes flipping between tiers across reruns — to 0 runaways and 0 flips once key order is preserved

## Notes

- `preserve_order` swaps serde_json's Map backing to IndexMap workspace-wide. Anything relying on alphabetical serialization would change behavior; the test suite passes, and for a proxy, forwarding payloads unmodified seems the safer default.
- Array order was already preserved; only object keys were affected.
- If you'd rather avoid the workspace-wide feature (e.g., carry passthrough JSON as `RawValue` instead), happy to rework.

## Summary by CodeRabbit

* **Improvements**
  * JSON object fields now retain their original order when processed and displayed, providing more consistent and predictable output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2122,6 +2122,7 @@ version = "1.0.150"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e8014e44b4736ed0538adeecded0fce2a272f22dc9578a7eb6b2d9993c74cfb9"
 dependencies = [
+ "indexmap",
  "itoa",
  "memchr",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ parking_lot = "0.12"
 rand = "0.10"
 reqwest = { version = "0.13.4", default-features = false, features = ["json", "rustls", "stream"] }
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+serde_json = { version = "1", features = ["preserve_order"] }
 switchyard-libsy = { path = "crates/libsy", version = "0.2.0" }
 switchyard-llm-client = { path = "crates/libsy-llm-client", version = "0.2.0" }
 switchyard-protocol = { path = "crates/protocol", version = "0.2.0" }
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -119,6 +119,7 @@ impl From<LlmResponseChunk> for LlmResponseStreamEvent {
 /// Not `Clone` — the `Stream` variant owns a single-consumption stream. A buffered
 /// backend returns `Agg` directly; a streaming one returns `Stream` and the consumer
 /// drives it, folding to an [`AggLlmResponse`] when it needs the whole response.
+#[allow(clippy::large_enum_variant)]
 pub enum LlmResponse {
     /// Live, single-consumption response stream.
     Stream(LlmResponseStream),


grahamking (approved):

grahamking:
@himorishige nice catch thank you!

nachiketb-nvidia (approved):
Thank you!

grahamking:
@himorishige It looks like adding the preserve order feature made `Serialize` and/or `Deserialize` impls bigger, tripping a new clippy lint.

264 bytes on the heap is fine. Can you add `#[allow(clippy::large_enum_variant)]` to that enum? We need `cargo clippy` to pass to be able to merge.

Thanks!

himorishige:
Thanks @grahamking! Added `#[allow(clippy::large_enum_variant)]` to `LlmResponse` in f4d30e3. Kept the variant unboxed as you suggested, so the public API is unchanged.

PR #443, opened 2026-08-15
feat(libsy): add request-fit routing gate for two-tier routes

### LINKED ISSUE #442: Proposal: request-fit routing gate for two-tier routes
# Proposal: length-based difficulty proxy for chat routing
Status: draft (v2, replaces the "request-fit routing gate" framing).

# Motivation
Switchyard's scorecard/difficulty classifier today only covers **coding** tasks. For **chat** routes we have no notion of "easy" vs "hard" at all — everything either goes to one target, or routing depends on a judge we haven't built for chat yet.

We want a cheap, no-training-required proxy for difficulty in chat routes, so we can send obviously-simple requests to the efficient model and obviously- demanding ones to the capable model, without waiting for a real chat scorecard.

Length is the cheapest signal we have:

- Long inputs are more likely to need a large context window.
- Small models tend to degrade faster than large ones on long-context retrieval/reasoning, even when the input technically fits.
It is not a good proxy for difficulty in general — a one-line prompt can be a hard theorem, a 50-page doc can be a trivial extraction — but it is cheap, available today, and directionally correct for the "long input" half of the problem.

# What this is not
- Not a difficulty classifier. It never says "this prompt is short, so send it to the cheap model." It only escalates on length. Absence of length escalation means "no opinion", not "confirmed easy".
- Not a replacement for a future chat scorecard. It's a stopgap for the one signal we can act on without training anything.

# Options considered
1. Length threshold only (chosen for v1) — estimate input size (chars/4), escalate to the capable model above a threshold, otherwise leave the decision to whatever comes next (default target, or a judge if one is configured). Zero training, zero extra latency, immediately deployable.
1. Keyword/regex heuristics — detect phrases suggesting hard reasoning ("prove", "derive", "why does X fail") or simple tasks ("summarize", "list", "extract"). Cheap, but brittle and high maintenance; rejected for v1, worth revisiting as a cheap additional signal later.
1. Self-triage by the efficient model — have the small model judge its own confidence/difficulty before answering. No new infra, but costs an extra call (or requires prompting tricks), and ties routing quality to the small model's self-assessment, which is exactly what we don't trust for hard cases. Rejected for v1, interesting as a second-tier signal.
1. Trained chat difficulty classifier (the real fix) — a chat-side scorecard analogous to the coding one. Most accurate, highest cost (data, training, maintenance). This is the actual long-term answer; this proposal is the stopgap until it exists.
1. Do nothing — fine only until we deploy an efficient tier with a small context window; then every long chat session pays for a wasted call to the small model before falling back.

v1 ships option 1. Options 2–3 are candidate follow-ups layered on top of the same classifier slot. Option 4 is the real destination; this issue is not meant to replace or preempt that work.

# Design
One classifier, matching the existing LlmTaskClassifier pattern: classifier core, FallThrough wrapper, one RouteConfig variant.
```
est = estimate_input_tokens(request)   # chars/4 over instructions + messages + tools
est >= escalate_over_input_tokens  → route to capable tier
otherwise                          → abstain
```
"Abstain" matters for composability: if a route has no judge configured, abstain is equivalent to your plain if/else — it just falls through to the default (efficient) target. If a route does have a judge configured later (e.g. a future chat scorecard), the length check runs first as a cheap pre-filter, and the judge only runs for requests that don't already need escalation. Same classifier, no special-casing needed either way.

Config:
```
[routes.auto]
type = "length_gate"
efficient_target = "small-model"
capable_target = "big-model"
escalate_over_input_tokens = 24000
```
Threshold is in tokens even though the estimator counts characters — no tokenizer dependency for v1. If chars/4 proves too inaccurate for some model family, add a per-route chars_per_token knob later.

# Tests
Estimator: text, tool calls/results, multimodal blocks, instructions; behavior exactly at the threshold.
Abstain path: no judge configured → falls through to default target (equivalent to plain if/else).
Abstain path: judge configured → judge decides for non-escalated requests, is never called for escalated ones.
Server: TOML round-trip, unknown fields rejected, missing threshold rejected.

# Open questions
Naming: `length_gate` vs something else?
v1 as a standalone route type only, or also expose the classifier for stage_router-style composition from day one?
Is chars/4 good enough across the tokenizers we actually deploy, or does v1 need the chars_per_token knob now rather than later?
Should we track this issue as a placeholder for option 4 (trained chat scorecard), or open that as a separate proposal once this ships?

## Related work
See #445 / #453 for a complementary chat-routing approach: instead of a length proxy, it uses an LLM-judged capability card and in-session "regret" signals (user corrections/re-asks) to escalate. The two are meant to be composable —
length_gate is a domain-agnostic pre-filter for context-window risk, while #445's approach targets the "no verifier" difficulty-judgment problem itself.
A route could plausibly use both: length_gate as an early abstain-capable pre-check, falling through to the conversation classifier for everything else.

panpan0000:
Closes #442.

Implements the request-fit routing gate: a one-way, length-based escalation
for two-tier routes, so requests the weak tier would reject (or handle
poorly) skip straight to the strong tier instead of paying one wasted call
plus its latency.

## What it does

- `RequestFitClassifier` estimates the request's input size in tokens
  (`chars/4` over instructions, messages, tool calls/results, media, and tool
  definitions) and escalates to the strong tier at or above a configured
  waterline. Below it, the classifier abstains and the next classifier in the
  cascade decides — length says nothing about difficulty.
- `RequestFit` is the assembled standalone route (gate first, weak tier as the
  default target), mirroring `LlmTaskClassifier`.
- New `type = "request_fit"` route config:

  ```toml
  [routes.auto]
  type = "request_fit"
  weak_target = "small-model"
  strong_target = "big-model"
  escalate_over_input_tokens = 24000
  ```

- The reactive context-window eviction stays: the gate makes overflows rare,
  and eviction catches the ones the estimate misses (regression-tested).

## Known limitations

- The estimator is deliberately coarse and counts characters, not tokens.
  Media sizing is asymmetric: URL media is under-counted (its URL string),
  while inline base64 is over-counted. The gate only ever escalates, so an
  under-estimate degrades to today's behavior (one wasted call), never to a
  wrong-tier answer.
- `escalate_over_input_tokens` is inclusive (`>=`).
- `max_output_tokens` is not yet folded into the estimate; output-budget
  gating is a follow-up.

## Tests

- Estimator: text, instructions, tool calls/results, tool definitions, inline
  media, and exact-threshold behavior.
- Gate: abstains below the waterline, escalates at/above it.
- Route: short → weak, long → strong, weak overflow → eviction retries strong.
- Server: TOML round-trip, and missing/zero threshold, identical targets, and
  unknown fields rejected.

## Summary by CodeRabbit

* **New Features**
  * Added request-fit routing that estimates request size, including messages, tools, files, media, and other inputs.
  * Automatically routes shorter requests to a weaker model and larger requests to a stronger model based on a configurable token threshold.
  * Added server configuration for selecting models and setting escalation thresholds.
* **Bug Fixes**
  * Preserved retry and fallback behavior when requests exceed the weaker model’s context limit.
  * Added validation for invalid thresholds and identical routing targets.

### DIFF
--- a/crates/libsy/src/algorithms.rs
+++ b/crates/libsy/src/algorithms.rs
@@ -11,6 +11,7 @@ pub mod llm_class;
 pub mod noop;
 pub mod passthrough;
 pub mod rand;
+pub mod request_fit;
 pub mod stage;
 
 pub mod util;
--- a/crates/libsy/src/algorithms/request_fit.rs
+++ b/crates/libsy/src/algorithms/request_fit.rs
@@ -1,0 +1,470 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! One-way request-size gate for two-tier routes.
+//!
+//! [`RequestFitClassifier`] estimates the request's input size and escalates to the
+//! strong tier when it crosses a configured waterline: the weak model would reject the
+//! request (its context window is smaller) or handle it poorly (long-context quality
+//! degrades faster for small models). Below the waterline the classifier abstains —
+//! length says nothing about difficulty, so the cascade's next classifier decides.
+//!
+//! [`RequestFit`] is the assembled standalone route: the gate, then the weak tier as
+//! the default target.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use serde_json::Value;
+use switchyard_protocol::{
+    ContentBlock, FileSource, ImageSource, MediaSource, ModelId, Request, Response,
+};
+
+use super::fall_through::{DefaultTarget, FallThrough};
+use crate::core::algorithm::{Algorithm, Driver};
+use crate::core::classifier::{Classification, Classifier, Score};
+use crate::{LibsyError, Result};
+
+/// Telemetry label for this algorithm's spans, metrics, and logs.
+const ALGORITHM_NAME: &str = "request_fit";
+
+/// Rough characters per token across common BPE tokenizers. Deliberately coarse —
+/// the gate is a pre-flight check, not billing.
+const CHARS_PER_TOKEN: u64 = 4;
+
+/// Fixed cost for JSON numbers, booleans, and nulls in tool schemas and arguments.
+const SCALAR_CHARS: u64 = 8;
+
+/// Estimates the request's input size in tokens as `chars / 4`, rounded up.
+///
+/// Counts instructions, message content (including tool calls and results), media
+/// payloads, and tool definitions. Overestimates slightly rather than missing an
+/// escalation the weak tier would have rejected.
+pub(crate) fn estimate_input_tokens(request: &Request) -> u64 {
+    let llm = &request.llm_request;
+    let mut chars = 0u64;
+    for instruction in &llm.instructions {
+        chars = chars.saturating_add(blocks_chars(&instruction.content));
+    }
+    for message in &llm.messages {
+        chars = chars.saturating_add(blocks_chars(&message.content));
+    }
+    for tool in &llm.tools {
+        chars = chars
+            .saturating_add(tool.name.chars().count() as u64)
+            .saturating_add(
+                tool.description
+                    .as_deref()
+                    .map_or(0, |d| d.chars().count() as u64),
+            )
+            .saturating_add(value_chars(&tool.parameters));
+    }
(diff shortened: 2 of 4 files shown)


nachiketb-nvidia (changes requested):
@panpan0000 I have some questions on this, left a comment on the issue related.

panpan0000:
hold until discussion on https://github.com/NVIDIA-NeMo/Switchyard/issues/442 addressed

nachiketb-nvidia:
Thank you!

we are currently focussed on task-oriented routing at the moment, and simple chats are not in scope. Closing this MR for now

Issue #445, opened 2026-08-15
Proposal: conversation routing without a verifier


panpan0000:
# Proposal: Conversation Routing Without a Verifier

Status: draft.

## TL;DR

Switchyard's difficulty classifier (capability classifier + escalation judge) was built for **coding tasks**. Every rule in it assumes a machine-verifiable outcome — a test passes, a build succeeds, a checker approves.

**Conversation** tasks — chat, Q&A, writing, summarization, tutoring, roleplay — don't have that. There's no verifier. The only judge of quality is the user.

Right now, those requests all fall into a dead zone: `primary_rule=none, capability_boundary=unmatched`. The classifier spits out a `p_solve` with nothing grounding it, and the escalation judge has no tests or builds to read. Routing a conversation request to the efficient or capable model is basically a coin flip.

## Why not just bolt on a "conversation" rule card?

Because the existing capability card is really a *verifiable-outcome* card. Its definition of success ("whole task correct, as judged by the final verifier") and its `p_solve` calibration both assume a verifier exists.

Conversation needs a different definition of success — "rated acceptable by a reference evaluator" — and its own calibration set. Mixing that into the same `p_solve` scale would corrupt the calibration that's already tuned for coding.

## Proposal — three layers, cheapest first

1. **Decide the value prop.** What does efficient vs. capable even mean for conversation — draft vs. polished? Correct vs. incorrect? Everything downstream depends on this.

2. **Cheap default + regret signal.** Don't try to forecast difficulty from the first message. Route by a deployment-level default, and escalate based on in-session user feedback — corrections, "that's wrong," "rephrase that," "not what I asked." The user is a free, reliable verifier, so use it.

3. **Only if a real classifier is justified:** build a separate `mode = "conversation"` (alongside capability/escalation/custom), with its own prompt, schema, and rules grounded in failure modes you can actually see from the instructions — factual faithfulness, multi-step reasoning, strict format adherence, high-stakes domains, long-document fidelity, multilingual handling. Calibrate against LLM-as-judge labels, since there's no verifier to calibrate against.

Prototype through the existing `custom` classifier mode before promoting anything to a first-class mode.

## Non-goals

- No changes to the coding capability card or escalation judge — those stay as-is for coding.
- No difficulty guessing here. The request-fit length gate (#442) is a separate, domain-agnostic mechanism and isn't part of this proposal.

## Relationship to #442 / #453

[...]

PR #453, opened 2026-08-17
feat(server): add config-only conversation routing example

panpan0000:
Addresses #445 (first two implementation paths from the proposal comment — both config-only, no new code).

## What's in this PR

A checked-in example deployment, `examples/conversation-routing/conversation-routing.toml`, with two two-tier conversation routes built entirely from the existing config surface:

| Route | Mechanism | Behavior |
|---|---|---|
| `switchyard/conversation` | `llm_classifier` custom mode | Pre-hoc scoring card: a judge reads the turn against a conversation capability card (CONV-1..9) and names `efficient` or `capable`. Conservative default: capable. |
| `switchyard/conversation-regret` | `llm_classifier` escalation mode | Regret-driven escalation: efficient answers by default; when the judge sees user regret (correction, re-ask, dissatisfaction), the session latches to capable. |

Both prompts and the response schema live inline in the example, so `switchyard-server --config examples/conversation-routing/conversation-routing.toml` runs as-is (replace the model ids with your own tiers).

## Tests

Two new integration tests in `crates/switchyard-server/tests/server.rs` drive the **checked-in example file itself** (via `include_str!`) against the mock upstream:

- `conversation_example_config_builds_both_routes` — the example parses and registers both routes.
- `conversation_example_routes_by_card_and_escalates_on_regret` — end-to-end: the scoring card routes to capable through the target selector (not the fallback), and the regret route runs efficient → escalates on the same turn the user says "wrong" → stays capable without another judge call.

The mock upstream gained a conversation-schema verdict branch and a regret-detection branch (user-role messages only, so the prompt's own "wrong" phrasing doesn't self-trigger).

## Notes / limitations

- The scoring card is deliberately uncalibrated; tune CONV rules against your own traffic (user-regret logs are the free calibration label).
- Escalation latches one-way for the session and does not decay.
- No Rust library changes: this validates the two zero-code paths from the proposal before any first-class `mode = "conversation"` work.

## Summary by CodeRabbit

- **New Features**
  - Added a ready-to-use, configuration-only example for routing conversations between efficient and capable model tiers.
  - Supports capability scoring, regret-based escalation, and one-way session latching after confirmed failures.
  - Added OpenAI-compatible usage guidance, model customization, and calibration instructions.

- **Tests**
  - Added integration coverage for both routing paths, escalation behavior, and prevention of repeated evaluation calls.

### DIFF
--- a/crates/switchyard-server/tests/server.rs
+++ b/crates/switchyard-server/tests/server.rs
@@ -166,11 +166,30 @@ async fn upstream_chat(
     let custom_target_schema = body
         .pointer("/response_format/json_schema/schema/properties/decision/properties/target")
         .is_some();
+    // The conversation example's scoring-card schema names its tiers
+    // "efficient"/"capable"; return a capable verdict so the checked-in example
+    // config can be exercised end-to-end.
+    let conversation_target_schema = body
+        .pointer("/response_format/json_schema/schema/properties/decision/properties/target/enum")
+        .and_then(Value::as_array)
+        .is_some_and(|values| values == &vec![json!("efficient"), json!("capable")]);
     let requests_invalid_verdict = body["messages"].as_array().is_some_and(|messages| {
         messages.iter().any(|message| {
             message["content"]
                 .as_str()
                 .is_some_and(|content| content.contains("invalid verdict"))
+        })
+    });
+    // The conversation example's regret route escalates when the judge sees the
+    // user push back; "wrong" stands in for that correction in the condensed
+    // trajectory the judge reads. System messages are skipped — the regret
+    // prompt itself names correction phrases like "wrong".
+    let requests_regret = body["messages"].as_array().is_some_and(|messages| {
+        messages.iter().any(|message| {
+            message["role"] == "user"
+                && message["content"]
+                    .as_str()
+                    .is_some_and(|content| content.contains("wrong"))
         })
     });
     let requests_schema_invalid_verdict = body["messages"].as_array().is_some_and(|messages| {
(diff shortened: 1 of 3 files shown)


examples/conversation-routing/conversation-routing.toml
```diff
@@ -0,0 +1,185 @@
...
+    "primary_rule": {
+      "type": "string",
+      "enum": ["CONV-1", "CONV-2", "CONV-3", "CONV-4", "CONV-5", "CONV-6", "CONV-7", "CONV-8", "CONV-9", "none"]
+    },
+    "p_solve": {"type": "number", "minimum": 0.0, "maximum": 1.0}
+  },
+  "required": ["decision", "crux", "primary_rule", "p_solve"],
+  "additionalProperties": false
+}
+'''
```
ayushag-nv, line 122: @panpan0000 This looks interesting. Can you explain how are you handling final target selection policy with this response schema in the example ??

panpan0000: Good question — the target selection only reads one field: policy.selector = "/decision/target" pulls decision.target (efficient/capable) straight out of the judge's JSON and uses it as-is. 
The other fields — crux, primary_rule, p_solve — aren't consumed by the routing policy at all; 

they're carried in the response purely for offline calibration/debugging (e.g. comparing p_solve against actual outcomes later). 

If the judge ever returns something the selector can't parse or that doesn't match the target enum, it falls back to default_target = "capable", which is the conservative default.

crates/switchyard-server/tests/server.rs
```diff
@@ -166,26 +166,54 @@ async fn upstream_chat(
     let custom_target_schema = body
         .pointer("/response_format/json_schema/schema/properties/decision/properties/target")
         .is_some();
+    // The conversation example's scoring-card schema names its tiers
```
nachiketb-nvidia, line 169: can we remove these tests? these include live tests which we don't want, and an example doesn't need tests, thanks

panpan0000:
Given the maintainer feedback on #443 that simple chats are currently out of scope, 
should #453 be deferred/closed as well? 
This PR is only a configuration example for #445 and does not add production routing behavior.

PR #454, opened 2026-08-17
feat(server): expose deployment config

nachiketb-nvidia:
## What

Add `GET /v1/config`, returning the validated TOML deployment as normalized JSON.

## Why

Clients need the configured route, target, and LLM client topology to interpret forthcoming decision-only routing responses.

## How

- serialize the existing typed config with Serde
- retain the serialized config in TOML-backed `ServerState` values
- omit `api_key_env` from serialization
- return `404` for programmatically constructed states that have no TOML deployment
- document the endpoint and cover its response shape with one focused integration test

## Example

`GET /v1/config` returns the complete normalized deployment, including defaulted values:

```json
{
  "schema_version": 1,
  "llm_clients": {
    "provider": {
      "format": "openai_chat",
      "base_url": "https://example.test/v1",
      "forward_auth": false,
      "extra_headers": {},
      "max_retries": 2
    }
  },
  "targets": {
    "fast": {
      "id": "model/fast",
      "llm_client": "provider",
      "extra_body": {}
    }
  },
  "routes": {
    "default": {
      "type": "passthrough",
      "id": "switchyard/default",
      "context_window": null,
      "tool_calling": null,
      "reasoning": null,
      "target": "fast"
    }
  }
}
```

The source TOML contains `api_key_env = "PATH"`; neither that environment-variable name nor its resolved value appears in the response.

## What to review

- the `/v1/config` JSON contract
- omission of `api_key_env`
- behavior for non-TOML `ServerState` construction

## Validation

- `cargo test -p switchyard-server config_endpoint_returns_loaded_toml_as_json`
- commit hooks: `cargo fmt`, `cargo clippy`

### DIFF
--- a/crates/libsy/src/algorithms/util/classifier_contract.rs
+++ b/crates/libsy/src/algorithms/util/classifier_contract.rs
@@ -4,13 +4,13 @@
 //! Prompt and structured-output contracts shared by LLM classifiers.
 
 use jsonschema::Validator;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 
 use crate::{LibsyError, Result};
 
 /// Provider-side structured-output mode used by a classifier judge.
-#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq)]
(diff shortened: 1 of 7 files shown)


nachiketb-nvidia:
closed until further need arises, could lead to security risks

PR #456, opened 2026-08-17
feat(server): add decision endpoint

nachiketb-nvidia:
## What

- Add `POST /v1/decision` for resolving a route without serving its answer target.
- Accept an explicit inbound wire format and provider request body.
- Return the selected target and ordered fallback targets from libsy's `RoutingOutcome`.
- Include each target's name, model ID, client format, base URL, and `extra_body`.
- Include an optional buffered `response` when routing itself produced the answer.

## Why

Decision-only integrations need Switchyard to run its routing logic while leaving the selected model call to the caller. The response provides the connection settings needed to reproduce that call without exposing API keys, credential environment names, configured headers, retry policy, or runtime client internals.

Some algorithms, including escalation and advisor routing, may produce a reusable answer while deciding. Returning that answer prevents an external integration from paying for the same answer twice.

## How

The endpoint decodes the nested provider request, resolves its Switchyard route, and uses `libsy::drive` to obtain the terminal `RoutingOutcome`. Every `CallModel` emitted before that outcome is a routing dependency, such as a classifier or judge call, and is served normally. The endpoint does not pass the outcome to `switchyard-llm-client` for another answer call.

The outcome's `selected_model_id` and ordered `fallback_models` are resolved through the route's canonical TOML target configuration. No separate target or client descriptors are retained. When `RoutingOutcome.response` is present, the existing `LlmResponse::into_agg` helper buffers it and the translation crate encodes it in the requested `input_format`. The `response` field is omitted when the outcome contains no answer.

## Request and response

The nested `request` uses the format named by `input_format` and selects a Switchyard route through its `model` field:

```http
POST /v1/decision
Content-Type: application/json

{
  "input_format": "openai_chat",
  "request": {
    "model": "switchyard/general",
    "messages": [
      {
        "role": "user",
        "content": "Explain speculative decoding briefly."
      }
    ]
  }
}
```

An ordinary routing decision preserves libsy's selected/fallback order and omits `response`:

```json
{
  "selected": {
    "target": "economy",
    "model": "model/weak",
    "llm_client": {
      "format": "openai_chat",
      "base_url": "https://example.com/v1"
    },
    "extra_body": {
      "service_tier": "priority"
    }
  },
  "fallbacks": [
    {
      "target": "quality",
      "model": "model/strong",
      "llm_client": {
        "format": "openai_chat",
        "base_url": "https://example.com/v1"
      },
      "extra_body": {}
    }
  ]
}
```

When routing produced the answer, `response` contains its buffered provider payload:

```json
{
  "selected": {
    "target": "economy",
    "model": "model/weak",
    "llm_client": {
      "format": "openai_chat",
      "base_url": "https://example.com/v1"
    },
    "extra_body": {}
  },
  "fallbacks": [],
  "response": {
    "id": "chatcmpl-example",
    "object": "chat.completion",
    "model": "model/weak",
    "choices": [
      {
        "index": 0,
        "message": {
          "role": "assistant",
          "content": "Speculative decoding verifies draft tokens with a larger model."
        },
        "finish_reason": "stop"
      }
    ]
  }
}
```

Credentials, credential environment names, configured headers, and retry settings are intentionally omitted. `base_url` and `extra_body` are included because the caller needs them to reproduce the target call.

## What to review

- `RoutingOutcome` is the endpoint's source of truth; obsolete `Step::Decision` and answer-call interception are gone.
- Routing dependencies execute, but the endpoint makes no post-routing answer call.
- An answer produced while routing is returned once; ordinary decisions retain the existing response shape.
- Selected and fallback target order matches libsy exactly.
- Authentication configuration and headers are never serialized.

## Validation

- `cargo test -p switchyard-server decision_returns_callable_target_and_routing_answer`

### DIFF
--- a/crates/switchyard-server/src/config.rs
+++ b/crates/switchyard-server/src/config.rs
@@ -23,8 +23,7 @@ use switchyard_llm_client::{
 use switchyard_protocol::{ModelId, RoutedLlmClient, WireFormat};
 
 use crate::{
-    CallerAuthKind, CountTokensTarget, DecisionLlmClient, DecisionTarget, ModelCapabilities,
-    ServerError, ServerResult, ServerState,
+    CallerAuthKind, CountTokensTarget, ModelCapabilities, ServerError, ServerResult, ServerState,
 };
 
 const SUPPORTED_SCHEMA_VERSION: u32 = 1;
@@ -45,22 +44,31 @@ pub fn load_server_state(path: impl AsRef<Path>) -> ServerResult<ServerState> {
 }
 
 fn server_state_from_toml(toml: &str) -> ServerResult<ServerState> {
-    let config: ServerConfig = toml::from_str(toml)
-        .map_err(|error| ServerError::new(format!("failed to parse TOML: {error}")))?;
-    config.build()
+    let config: Arc<ServerConfig> = Arc::new(
+        toml::from_str(toml)
+            .map_err(|error| ServerError::new(format!("failed to parse TOML: {error}")))?,
+    );
+    let state = config.build()?;
+    Ok(state.with_config(config))
 }
 
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
-struct ServerConfig {
+pub(crate) struct ServerConfig {
     schema_version: u32,
     #[serde(default)]
-    llm_clients: BTreeMap<String, LlmClientConfig>,
-    targets: BTreeMap<String, TargetConfig>,
+    pub(crate) llm_clients: BTreeMap<String, LlmClientConfig>,
+    pub(crate) targets: BTreeMap<String, TargetConfig>,
     routes: BTreeMap<String, RouteConfig>,
 }
 
 impl ServerConfig {
+    pub(crate) fn routing_target_names(&self, route_name: &str) -> Option<Vec<&str>> {
+        self.routes
+            .get(route_name)
+            .map(RouteConfig::routing_target_names)
+    }
+
     fn build(&self) -> ServerResult<ServerState> {
         if self.schema_version != SUPPORTED_SCHEMA_VERSION {
             return Err(ServerError::new(format!(
@@ -104,7 +112,6 @@ impl ServerConfig {
             let algorithm = build_algorithm(route_name, config, &targets)?;
             let (client, caller_auth) = self.build_route_clients(route_name, config, &clients)?;
             let count_tokens_target = self.build_count_tokens_target(config, &clients);
-            let decision_targets = self.build_decision_targets(config)?;
             routes.push((
                 config.id().clone(),
                 algorithm,
@@ -112,7 +119,7 @@ impl ServerConfig {
                 caller_auth,
                 capabilities,
                 count_tokens_target,
-                decision_targets,
+                Some(route_name.clone()),
             ));
         }
         ServerState::new_with_capabilities(routes)
@@ -235,40 +242,6 @@ impl ServerConfig {
                 client: client.clone(),
             })
     }
-
-    /// Retains only the non-secret settings an external caller needs after routing.
-    fn build_decision_targets(
-        &self,
-        route: &RouteConfig,
-    ) -> ServerResult<BTreeMap<ModelId, DecisionTarget>> {
-        route
-            .routing_target_names()
-            .into_iter()
-            .map(|target_name| {
-                let target = self.targets.get(target_name).ok_or_else(|| {
-                    ServerError::new(format!("route references unknown target {target_name}"))
-                })?;
-                let client = self.llm_clients.get(&target.llm_client).ok_or_else(|| {
-                    ServerError::new(format!(
-                        "target {target_name} references unknown llm client {}",
-                        target.llm_client
-                    ))
-                })?;
-                Ok((
-                    target.id.clone(),
-                    DecisionTarget {
-                        target: target_name.to_string(),
-                        model: target.id.clone(),
-                        llm_client: DecisionLlmClient {
-                            format: client.format.wire_format(),
-                            base_url: client.base_url.clone(),
-                        },
-                        extra_body: target.extra_body.clone(),
-                    },
-                ))
-            })
-            .collect()
-    }
 }
 
 // Prefer known Claude families, then preserve the route's target order.
@@ -283,9 +256,9 @@ fn count_tokens_priority(target_name: &str, model_id: &ModelId) -> usize {
 
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
-struct LlmClientConfig {
-    format: ClientFormat,
-    base_url: String,
+pub(crate) struct LlmClientConfig {
+    pub(crate) format: ClientFormat,
+    pub(crate) base_url: String,
     api_key_env: Option<String>,
     #[serde(default)]
     forward_auth: bool,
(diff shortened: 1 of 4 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -562,6 +625,88 @@ async fn openai_responses(
...
+async fn run_decision_only(route: &RouteEntry, request: Request) -> libsy::Result<ModelId> {
+    let stream = Arc::clone(&route.algorithm).run_stream(request);
+    tokio::pin!(stream);
+
+    while let Some(step) = stream.next().await {
+        match step? {
+            // Decision-only behavior: take the selected answer call and stop before model I/O.
+            Step::CallModel(call) if call.is_answer_call => {
+                let (_, models) = call.into_parts();
+                return models.into_iter().next().ok_or(LibsyError::NoTargets);
```
bhuvan002: Instead of returning the first should we instead return all in case the caller needs to use the fallback model(s)?

crates/switchyard-server/src/lib.rs
```diff
@@ -285,6 +322,31 @@ impl ServerState {
...
+        route: &'a RouteEntry,
+        selected_model: &ModelId,
+    ) -> Option<DecisionResponse<'a>> {
+        let config = self.config.as_ref()?;
+        let (target_name, target) = config
+            .routing_target_names(route.config_name.as_ref()?)?
+            .into_iter()
+            .rev()
+            .filter_map(|name| config.targets.get_key_value(name))
+            .find(|(_, target)| target.id == *selected_model)?;
```
bhuvan002: If we are routing between two targets with the same model, say Opus 4.8 low effort and Opus 4.8 max effort, is it possible to to route by target name instead of model ID if that information isn't lossy?

crates/switchyard-server/src/lib.rs
```diff
@@ -562,6 +625,88 @@ async fn openai_responses(
...
+
+/// Drives an algorithm until its answer call is known, without executing that final call.
+async fn run_decision_only(route: &RouteEntry, request: Request) -> libsy::Result<ModelId> {
+    let stream = Arc::clone(&route.algorithm).run_stream(request);
+    tokio::pin!(stream);
+
+    while let Some(step) = stream.next().await {
+        match step? {
+            // Decision-only behavior: take the selected answer call and stop before model I/O.
+            Step::CallModel(call) if call.is_answer_call => {
```
grahamking: This case is now in `Done(outcome)`. The selected model is `outcome.selected_model_id`. You don't need `into_parts`.

crates/switchyard-server/src/lib.rs
```diff
@@ -562,6 +625,88 @@ async fn openai_responses(
...
+                let (_, models) = call.into_parts();
+                return models.into_iter().next().ok_or(LibsyError::NoTargets);
+            }
+            // Normal algorithm execution: classifiers and judges must answer before a final
+            // target can be selected.
+            Step::CallModel(call) => serve_decision_dependency(route, *call).await?,
+            // Published decisions are observability events. The answer CallModel above is the
+            // executable selection and therefore the endpoint's source of truth.
+            Step::Decision(_) => {}
```
grahamking: `Step::Decision` is gone.

crates/switchyard-server/src/config.rs
```diff
@@ -232,6 +235,40 @@ impl ServerConfig {
...
+                    DecisionTarget {
+                        target: target_name.to_string(),
+                        model: target.id.clone(),
+                        llm_client: DecisionLlmClient {
+                            format: client.format.wire_format(),
+                            base_url: client.base_url.clone(),
+                        },
+                        extra_body: target.extra_body.clone(),
+                    },
+                ))
```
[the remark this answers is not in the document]
nachiketb-nvidia: check now?

crates/switchyard-server/src/config.rs
```diff
@@ -232,6 +235,40 @@ impl ServerConfig {
...
+                Ok((
+                    target.id.clone(),
+                    DecisionTarget {
+                        target: target_name.to_string(),
+                        model: target.id.clone(),
+                        llm_client: DecisionLlmClient {
+                            format: client.format.wire_format(),
+                            base_url: client.base_url.clone(),
+                        },
+                        extra_body: target.extra_body.clone(),
```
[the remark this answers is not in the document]
nachiketb-nvidia: check now?

crates/switchyard-server/src/config.rs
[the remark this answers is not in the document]
nachiketb-nvidia: what about now?

crates/switchyard-server/src/config.rs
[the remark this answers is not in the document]
nachiketb-nvidia: what about now?

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: removed

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: updated it all!

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: in the API we're returning everything one would need to make their own model call. So we return two targets and their info here

nachiketb-nvidia: or am i misunderstanding?

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: done, changed API to have fallbacks as well

crates/switchyard-server/src/lib.rs
bhuvan002: I meant the `find` here will select one particular `target` corresponding to the selected model ID. But if there are two separate targets with the same model ID (like with different extra_body fields) then this would drop one of the targets. I guess this is an algorithm limitation and not something you can fix with the new API route, so resolving this.

crates/switchyard-server/src/lib.rs
```diff
@@ -564,6 +646,95 @@ async fn openai_responses(
...
+                Ok(response) => Some(response),
+                Err(error) => return server_error(error.to_string()),
+            }
+        }
+        None => None,
+    };
+    match state.decision_response(route, &outcome, response) {
+        Some(response) => Json(response).into_response(),
+        None => error_response(
+            StatusCode::UNPROCESSABLE_ENTITY,
```
bhuvan002: Would this be better as a 5xx server error instead of a client bad request? Since it's not the client's fault if the server is set up with a bad config or the algorithm returns a bad target

grahamking (approved):

crates/switchyard-server/src/lib.rs
grahamking: Claude said something similar:

>   The final commit, fix(server): allow shared target model ids, deletes the uniqueness check that an earlier commit in this same PR (fix(server): align decision endpoint with routing outcomes) added specifically to keep decision_response unambiguous.
> 
>  That earlier check rejected configs where two targets in one route shared a model id, with the message "model ids must be unique within a route." Removing it reopens exactly the ambiguity it existed to prevent:
>
 > - decision_response's resolve closure (lib.rs) picks the first target in routing_target_names() order whose id matches the outcome's model (weak_target before strong_target in capability mode).
>  - build_route_clients (config.rs), which builds the ClientRouter actually used to serve calls, inserts into a HashMap keyed by model id in callable_target_names() order (weak_target then strong_target then judge) — so the last insert wins, i.e. strong_target's client.
>
>  If two targets in a route share a model id, /v1/decision will report the base_url/extra_body of one target while a real call for that model would actually go through a different target's client/config 

I'm not sure how much of a real problem that is. Two identical model ids will very likely have identical `base_url`'s. 

I've seen Claude complain about this a few times in my work. Possibly we should decide what the rules are on re-using models, and enforce them when parsing the config TOML. Likely fine for now. Feel free to resolve.

crates/switchyard-server/src/config.rs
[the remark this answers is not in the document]
nachiketb-nvidia: why would this be a leak? this is supposed to be the request body stuff, this is stuff needed to run the decided model

crates/switchyard-server/src/lib.rs
nachiketb-nvidia: yeah "target" here is the differentiator. I removed a useless check where we drop two targets with the same model so it should just return the target name and everything you'd need to call that target

PR #460, opened 2026-08-17
fix(translation): replay assistant reasoning as reasoning_content too

### LINKED ISSUE #449: [bug] reasoning_content stripped from requests: breaks reasoning-required upstreams (DeepSeek thinking) on multi-turn
## Symptom

Multi-turn chat with a reasoning-required OpenAI-compatible upstream (e.g. DeepSeek's thinking models) fails on the second turn with:

```
400 invalid_request_error: The `reasoning_content` in the thinking mode must be passed back to the API.
```

and, in tool-calling flows:

```
400 invalid_request_error: Messages with role 'tool' must be a response to a preceding message with 'tool_calls'
```

The first turn works fine; any follow-up turn that carries back the assistant message from turn 1 fails.

## Reproduction

1. Configure a route whose backend is a reasoning/thinking OpenAI-compatible model that requires `reasoning_content` to be passed back verbatim on subsequent turns (DeepSeek v4 flash/pro "deepseek-reasoner"-style models do this).
2. Send a first request; the response includes `reasoning_content` on the assistant message (switchyard's `encode_response` correctly emits it).
3. Send a second request whose `messages` array includes that assistant message **with** its `reasoning_content` (i.e. a client that faithfully passes history back).

Expected: second turn succeeds.

Actual: switchyard strips the reasoning field from the request before forwarding, so the upstream rejects the turn with the 400 above.

```bash
# turn 1 (ok) — response contains reasoning_content
curl -s http://127.0.0.1:4000/v1/chat/completions \
  -d '{"model":"smart1","messages":[{"role":"user","content":"hi"}],"stream":false}'

# turn 2 (400) — history includes assistant reasoning_content from turn 1
curl -s http://127.0.0.1:4000/v1/chat/completions \
  -d '{"model":"smart1","messages":[{"role":"user","content":"hi"},
       {"role":"assistant","content":"...","reasoning_content":"thinking..."}],"stream":false}'
```

## Expected vs. actual

- **Expected:** the assistant message's `reasoning_content` is forwarded to the upstream so reasoning-required providers can validate the turn.
- **Actual:** `reasoning_content` is silently dropped on the request path; providers that enforce "reasoning_content must be passed back" reject with 400.

## Root cause (code pointer)

`crates/switchyard-translation/src/codecs/openai_chat/buffered.rs`, `encode_message_without_tool_results_to_openai`:

```rust
let content_blocks = message
    .content
    .iter()
    .filter(|block| {
        !matches!(
            block,
            ContentBlock::ToolCall(_) | ContentBlock::Reasoning { .. }
        )
    })
```

`ContentBlock::Reasoning` is explicitly filtered out of the encoded request, even though the response encoder (`encode_response`, same file) re-emits it as `reasoning_content`. This asymmetry breaks round-tripping for providers that require reasoning to be echoed back (DeepSeek thinking, some vLLM reasoning templates).

Note: `prepend_openai_reasoning_blocks` merges incoming reasoning into content blocks on the decode side, but there is no corresponding encode-side path that restores a standalone `reasoning_content` field on assistant messages in the request.

## Suggested direction

When encoding a request for an OpenAI-compatible upstream, preserve `ContentBlock::Reasoning` from assistant messages as a top-level `reasoning_content` (or `reasoning`) field instead of dropping it — ideally gated by a policy/extension so providers that reject unknown fields are unaffected. This mirrors the response-side behavior and makes switchyard safe for reasoning-required providers.

## Environment

- Switchyard: main @ 2026-08-14 (recent commits: `feat(libsy-llm-client)` etc.)
- Inbound format: Chat Completions
- Backend: OpenAI-compatible reasoning model (DeepSeek thinking)
- Route type: deterministic routing with a classifier LLM

## Additional context

Workaround used in production: a thin gateway in front of switchyard caches the last assistant `reasoning_content` per session and re-injects it into outbound requests when the client omits it. This confirms the upstream requirement and that the fix belongs in the translation layer. Happy to prepare a patch + tests if the direction is approved.

ting-hong-shieh:
## What

Replays assistant reasoning under both OpenAI-compatible spellings instead of `reasoning` alone.

`encode_openai_message_plaintext_reasoning` and the structured fallback in `encode_openai_message_structured_reasoning` now go through one helper that writes `reasoning` and `reasoning_content` with the same text. No public API change; `TargetCapabilities` is untouched.

## Why

#415 stopped request history from dropping reasoning, which was the original report in #449. It replays it as `message["reasoning"]`, while the response encoder in the same file writes `reasoning_content`, so the two directions disagree on the name.

Reasoning-required upstreams look for `reasoning_content` specifically and treat it as present-or-absent rather than reading the alias. @DuanZGit measured this against a production DeepSeek thinking endpoint in #449, varying only the extra field on the assistant history message:

| extra field on assistant history message | result |
|---|---|
| (none) | 400 `reasoning_content in thinking mode must be passed back` |
| `reasoning` | 400, same message |
| `reasoning_content` | 200 |
| `reasoning_content: ""` | 200 |
| `reasoning_details: [...]` | 400, same message |
| unknown field | 400, same message |

So the second turn still fails on those upstreams today even though the reasoning is replayed.

Sending both spellings rather than swapping one for the other keeps providers that read `reasoning` working exactly as they do now. The rows above also indicate the rejection is a presence check on `reasoning_content`, not strict unknown-field validation, since an unknown field alone produced the same error.

I considered gating the field on a new `TargetCapabilities` flag and decided against it: `main` already sends `reasoning` unconditionally, so a provider strict enough to reject `reasoning_content` is already rejecting `reasoning`. A second spelling of the same text does not add a new class of field, and a capability flag would add public surface for no additional protection.

Closes #449

## How tested

The checklist below is Python-oriented and this change is Rust-only, so those items are **not applicable**. Commands actually run, on Linux with cargo 1.96.1, at the head of this branch:

```
cargo fmt --all --check
cargo test --workspace
```

All three clean.

Added to `crates/switchyard-translation/tests/request_translation.rs`:

- the shape reported in #449 — an assistant turn carrying reasoning and `tool_calls`, followed by the tool result — replays both spellings with the same text, keeps the tool calls, and keeps the tool result following the call it answers;
- a turn with no reasoning gains neither key.

Also verified while writing it, beyond the new tests:

- the decode side already accepts either spelling via `first_nonempty_string(object, &["reasoning_content", "reasoning"])`, so a message carrying both decodes to one reasoning block rather than two;
- re-encoding a translated request is byte-stable, so the two keys neither accumulate nor drift across passes.

**Live verification of this branch's exact shape.** I have no live endpoint, so the tests above are deterministic regression coverage only. @DuanZGit ran the dual-write against their production DeepSeek endpoint ([#449](https://github.com/NVIDIA-NeMo/Switchyard/issues/449#issuecomment-5323601266)):

| assistant history extra | result |
|---|---|
| both `reasoning` and `reasoning_content`, same text — what this PR sends | **200** |
| `reasoning` only — what `main` sends today | 400, unchanged |
| `reasoning` and `reasoning_content: ""` | 200 |

Row two is `main`'s current behavior, and row one is this PR's. That measurement is theirs, not mine; my own evidence remains the code path and the tests.

- [ ] `uv run ruff check .` clean — n/a, no Python changed
- [ ] `uv run mypy switchyard` clean — n/a, no Python changed
- [ ] `uv run pytest tests/` green — n/a, no Python changed
- [ ] Manual smoke — n/a, no live endpoint available; see above

## Checklist

- [ ] One class per file; filename = `snake_case` of the primary class. — n/a, no Python changed
- [ ] New public symbols exported from `switchyard/__init__.py.__all__`. — n/a, no new public symbols
- [ ] README / `--help` updated if customer-facing surface changed. — no CLI or README surface changed
- [x] Commits signed off (`Signed-off-by:`) per the DCO.

## Notes for reviewers

**One existing test changed.** `responses_reasoning_item_merges_into_next_assistant_message_for_openai_chat`, added by #415, asserts an exact message object and fails when a second key appears. I added `reasoning_content` to its expected value rather than relaxing the assertion, so it still pins the exact wire shape. Calling it out because it is not my test.

**Scope limit worth stating.** `encode_request` consults `exact_preserved_request` first, so when source and target are both `OpenAiChat` and `PreservationPolicy` is `InMemory` or `Embed`, the original body is replayed and this code never runs. Reaching it requires cross-format translation, `PreservationPolicy::Disabled`, or a request built directly from the IR. A gateway that terminates the client and re-emits is in scope; a straight same-format proxy is not.

**Signed reasoning is still not replayed.** The plaintext path matches only `signature: None`, so Anthropic `thinking` blocks that carry a signature continue to be excluded, unchanged by this PR. `anthropic_thinking_blocks_do_not_leak_into_openai_chat_messages` pins that behavior and still passes.

**If you would rather not send both**, the alternative is to make `reasoning_content` replace `reasoning` for targets that declare support, which reintroduces the capability flag. Say the word and I will switch it; I preferred this shape because it needs no new public surface and cannot regress a provider that reads `reasoning` today.


ting-hong-shieh:
Checked this against the branch, and the named test does not fail — `openai_chat_reasoning_details_round_trip_in_assistant_history` passes as written, before and after this change.

The reason is what the decoder does with the fixture. That fixture's `reasoning_details` carry readable text, so `prepend_openai_reasoning_blocks` sets the block's `text` from `reasoning_text_from_details`, not from the `"fallback text"` string in the body. `encode_openai_message_structured_reasoning` only collects a fallback when the block text *differs* from the details text, so for that fixture the fallback is empty, `set_openai_reasoning_text` returns early, and neither key is emitted. The `reasoning` absence assertion still holds.

The case you are describing is real, though — it is the *other* test. `openai_chat_encrypted_reasoning_details_retain_fallback` uses encrypted details with no readable text, so the fallback is emitted, and that is the structured-path case where both keys now appear. Nothing was asserting `reasoning_content` there, so my new tests covered the plaintext path only.

Added that assertion in 737adc1c. Thanks — the finding pointed at a genuine coverage gap even though the failure prediction did not hold.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean at that commit.

nachiketb-nvidia (commented):
can you reply to coderabbit if you're done?
Also, I don't understand the bug here, can you dumb it down a bit for me?

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
```diff
@@ -907,9 +905,22 @@ fn encode_openai_message_structured_reasoning(
...
+// upstreams look for `reasoning_content` specifically and reject a follow-up turn
+// whose assistant history lacks it, treating the field as present-or-absent rather
+// than reading the alias, so both are written. The decode side already accepts
+// either spelling.
+fn set_openai_reasoning_text(message: &mut Value, reasoning: String) {
+    if reasoning.is_empty() {
+        return;
     }
+    message["reasoning"] = Value::String(reasoning.clone());
+    message["reasoning_content"] = Value::String(reasoning);
```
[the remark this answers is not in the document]
ting-hong-shieh: Replying here so this thread is not left looking open — I answered in a top-level comment earlier, which was the wrong place.

Summary: the predicted failure does not happen, but the finding pointed at a real coverage gap, which is fixed in 737adc1c.

`openai_chat_reasoning_details_round_trip_in_assistant_history` passes as written. Its `reasoning_details` carry readable text, so the decoder takes the block's text from `reasoning_text_from_details` rather than from the `"fallback text"` string in the body. The structured encoder only emits a fallback when those two differ, so for that fixture nothing is emitted and the `reasoning` absence assertion still holds.

The case being described is the other test, `openai_chat_encrypted_reasoning_details_retain_fallback`, where the details are encrypted and carry no readable text. That one does emit the fallback, and it is the structured-path case where both keys now appear. Nothing was asserting `reasoning_content` there, so that assertion is now added.

ting-hong-shieh:
Sure, and sorry — I opened with the field names, which is the middle of the story.

**The one-line version:** on the second turn of a conversation, we hand the model's own previous thinking back to it under a name it does not recognise, and it refuses the request.

Walking through it with a reasoning model behind an `openai_chat` backend:

**Turn 1.** The user asks something. The model answers and also returns its private reasoning. Switchyard passes both back to the client. This part works.

**Turn 2.** The client sends the conversation history back so the model can continue. Switchyard re-encodes that history, including the assistant's previous turn. Upstreams in thinking mode require the previous reasoning to come back with it — if it is missing, they refuse to continue, because they will not silently re-think a turn they have already paid for.

The refusal looks like this:

```
400  The `reasoning_content` in the thinking mode must be passed back to the API
```

**The catch.** Two spellings exist in the OpenAI-compatible world. Most providers read `reasoning`. DeepSeek-style thinking models read `reasoning_content`, and they check for that exact key — they do not fall back to the alias.

Before #415 we sent neither, so the reasoning was simply lost and every second turn failed. #415 fixed the loss and sends `reasoning`. For providers that read that key, it now works. For the ones in this issue, the reasoning is present on the wire but under the wrong name, so they still refuse.

So the assistant history message we send today looks like this:

```json
{ "role": "assistant", "content": "...", "reasoning": "the model's earlier thinking" }
```

and this PR makes it look like this:

```json
{ "role": "assistant", "content": "...",
  "reasoning": "the model's earlier thinking",
  "reasoning_content": "the model's earlier thinking" }
```

Same text under both names, so whichever key a provider reads, it sees the same reasoning.

**Why both rather than switching.** Swapping `reasoning` for `reasoning_content` would fix these upstreams and break the ones #415 was written for. Sending both costs one duplicated string in the request body and cannot regress a provider that reads either.

The reason I am confident about the name and not guessing: @DuanZGit ran it against a live DeepSeek endpoint in #449, changing only that field on the history message:

| field sent | result |
|---|---|
| nothing | 400 |
| `reasoning` | 400, same error |
| `reasoning_content` | 200 |
| `reasoning_content: ""` | 200 |

Row two is this PR's reason for existing — it is the state of `main` today.

One honest caveat: I have not run that matrix myself, and I have no live endpoint. My own evidence is the code path and the tests. @DuanZGit offered in #449 to re-run it against this branch, which would be worth taking them up on before merge, since sending *both* keys is a combination their table did not cover.

DuanZGit:
Ran the full matrix against our live DeepSeek endpoint on the exact shape this PR sends — the dual-key combination, which the earlier table didn't cover.

**Result: this PR is what fixes it.** The precise shape it serializes (row 3) is the first one that passes:

| # | reasoning on assistant history message | result |
|---|----------------------------------------|--------|
| 1 | neither field | 400 ❌ |
| 2 | `reasoning` only (what `main` sends today) | 400 ❌ |
| 3 | **`reasoning` + `reasoning_content` (PR #460)** | **200 ✅** |
| 4 | `reasoning_content` only | 200 ✅ |
| 5 | `reasoning` + `reasoning_content:""` | 200 ✅ |
| 6 | `reasoning:""` + `reasoning_content` | 200 ✅ |

In plain terms for the questions raised: the upstream refuses only when the `reasoning_content` key is absent — row 3 proves that carrying both spellings satisfies it, and row 2 is literally the current state of `main` (the reasoning is present but under the wrong name), which is why #449 stays open after #415. The dual-write can't regress the providers #415 fixed, because it keeps `reasoning` exactly where it was and merely adds the spelling DeepSeek reads.

Two notes on trustworthiness: the decode-side claim in the PR and the `no reasoning → neither key` case are already pinned by the tests, and our live runs above independently confirm the wire-level behavior this PR relies on (presence check on `reasoning_content`). We ran all six against `deepseek-v4-flash`; keys stayed server-side.

This was the re-run I offered in #449 — happy this landed before merge. From our side this is ready to go.

nachiketb-nvidia (approved):

crates/switchyard-translation/tests/request_translation.rs
```diff
@@ -1010,6 +1010,85 @@ fn responses_reasoning_items_attach_to_tool_call_turn_for_openai_chat() -> TestR
     Ok(())
 }
 
+// Reasoning-required upstreams look for `reasoning_content` and treat it as
+// present-or-absent, so replaying only `reasoning` fails the next turn. Both
+// spellings carry the same text, and tool calls stay alongside them.
+#[test]
+fn openai_chat_replays_reasoning_under_both_spellings() -> TestResult {
```
nachiketb-nvidia, line 1217: lets combine 2 tests into one, and make the comments one line

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
```diff
@@ -907,9 +905,22 @@ fn encode_openai_message_structured_reasoning(
         })
         .collect::<Vec<_>>()
         .join("\n");
-    if !fallback.is_empty() {
-        message["reasoning"] = Value::String(fallback);
+    set_openai_reasoning_text(message, fallback);
+}
+
+// Writes replayed reasoning under both OpenAI-compatible spellings.
```
nachiketb-nvidia: lets reduce the comments here to one line as well

nachiketb-nvidia (changes requested):

grahamking:
@ting-hong-shieh Thanks for this. It makes sense.

But it will increase costs right? There is now double the thinking tokens going in. And depending on how the inference service handles the request, we might also have double those going back out.

Is there a way to make it configurable, so there's only `reasoning` or `reasoning_contents`, never both? We already have the `format` field, so maybe extending that by having two kinds of `openai_chat` format? I'm not sure.

crates/switchyard-translation/tests/request_translation.rs
ting-hong-shieh: Done — the two cases are one test now (`openai_chat_replays_reasoning_under_both_spellings`),
with the no-reasoning turn appended to the same conversation so both assertions run off one
translation, and the comment cut to one line.

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
ting-hong-shieh: Done — cut to one line.

ting-hong-shieh:
@grahamking Measured it. The honest answer is that the absolute cost is fixed and small,
and the percentage depends entirely on what else is in the context — so a single number
would have been misleading.

**Absolute:** the duplicate key adds exactly the reasoning text again, once per assistant
turn that carries plaintext reasoning. Five turns of 2000-character reasoning is 10,075
extra bytes in the request body, and that number does not move no matter how large the
rest of the conversation is.

**As a share of the body**, on synthetic OpenAI-Chat conversations run through the engine
(serialized request bytes, this PR versus the same body with the second key stripped):

| shape | overhead |
|---|---|
| chat, short context (80-char user turns, no tools) | +73.4% |
| chat, long user prompts (4k chars) | +27.1% |
| agentic, 8k-char tool outputs | +18.5% |
| agentic, 40k-char tool outputs | +4.7% |
| short reasoning, 20k-char tool outputs | +1.0% |

So for agentic traffic it is single digits to ~20%; the alarming numbers only appear when
reasoning dominates a small context.

Two corrections to things I nearly told you:

**It is request-side only, in this codebase.** The change lives in the request encoder;
`encode_response` is untouched and already emits `reasoning_content` alone. I cannot speak
for what a given inference service does with the extra field, but nothing here doubles the
response path.

**It is not confined to the plaintext path.** I first thought histories carrying
`reasoning_details` were exempt. They are only exempt when the details carry recoverable
text — `openai_chat_encrypted_reasoning_details_retain_fallback` shows encrypted details
still emit the fallback under both keys.

On making it configurable: I looked for a way to infer the right spelling instead, and
there is not one. The obvious idea is to remember which key arrived and replay that — but
`encode_response` normalizes everything to `reasoning_content` on the way back to the
client, so a client replaying a Switchyard-produced history always sends
`reasoning_content` regardless of which upstream is behind us. The arrived spelling
carries no information about the upstream. (It would also mean threading a field through
27 `ContentBlock::Reasoning` construction sites across `protocol`, `libsy` and the three
codecs.)

Which leaves your instinct as the structurally correct one: the required spelling is a
property of the upstream, it cannot be derived from the request, so it belongs on the
route or format config. Two shapes:

- **default to both, let a route opt into one** — nothing breaks on upgrade, operators
  only touch it to save tokens
- **default to one, let a route opt into both** — cheaper by default, but someone finds
  out via a 400

I would take the first. What I do not have is grounds to pick a single spelling as the
default: @DuanZGit's row 4 shows the live DeepSeek endpoint accepts `reasoning_content`
alone, but I have no evidence about what OpenRouter and the other documented backends read
on input, and I would rather not guess that in public.

Happy to implement the config here, or land this as the narrow correctness fix and do it
separately — your call on scope.

Separately: this branch is merged up to `main` (one import conflict) and
@nachiketb-nvidia's two review points are addressed.

grahamking:
> no evidence about what OpenRouter and the other documented backends read on input

OpenRouter normalizes everything to "reasoning". It will also include "reasoning_details" (e.g. summary, encrypted reasoning). I never sends you "reasoning_content".

grahamking:
@ting-hong-shieh Thanks for those numbers, that's really helpful!

A big part of Switchyard's goal is to save users money. Costing up to 20% more is too much.

Would you be OK closing this PR, and starting a new one that makes reasoning vs reasoning_content configurable in the TOML config? This is the option we want:

> default to one, let a route opt into both 

meaning it stays as it is today, but a flag in the `targets` entry allows using the older DeepSeek style. Most API's have transitioned to "reasoning", so that should be the default.

Something like this?
```
[targets.medium]
id = "deepseek/deepseek-v4-flash-0731"
llm_client = "deepseek"
reasoning_format = "deepseek" # or "openai", the default
```

```
enum ReasoningFormat {
  /// "reasoning" field
  #[default]
  OpenAI,
  /// "reasoning_content" field
  DeepSeek,
}
```

Most of the code should live in `switchyard-translation`.

PR #461, opened 2026-08-17
feat(libsy): count escalation judge verdicts

yanyanz89:
## What

`EscalationPolicy` maps the judge's verdict to a target and then discards it. `escalation.rs` contains no metric, span, or tracing call. This adds a counter, plus its entry in the metrics reference.

```
switchyard.escalation_judge_verdicts{judge_model, verdict}
```

`verdict` is `escalate` or `hold`. One counter, two labels, one call site.

## Why

`switchyard.classifier_fail_open` (#205) counts the judge's **failures**. Nothing counts its **decisions**, so an operator can see that the judge broke but not that it escalated on every turn, or on none.

That hides a documented failure mode. `confirmations` defaults to 2 and the streak is retained per session, and `fall_through.rs` notes that "requests without a session ID use unretained per-run state". A client omitting `x-switchyard-session-id` therefore resets the streak every turn and never latches, while still paying for a judge call on every turn.

Three arms, six turns each, against a deterministic local upstream:

| Arm | session header | judge | `decisions` | verdicts |
|---|---|---|---|---|
| A | present | escalates | capable=5, efficient=1 | escalate=2 |
| B | **absent** | escalates | efficient=6 | escalate=6 |
| C | present | declines | efficient=6 | hold=6 |

**B and C are identical on `decisions` and opposite on verdicts.** B burns a judge call every turn for nothing; C is healthy. The remedies differ: B needs a session header or `confirmations = 1`.

Arm A shows why this is not a duplicate of `switchyard.decisions`: six decisions, two verdicts. The judge stops being consulted once the streak reaches `confirmations`, so the two diverge with session length.

## Correctness and cardinality

The verdict is recorded in `EscalationPolicy::to_classification`, off the judge's own boolean, rather than downstream off the tier the verdict resolves to. The two are not interchangeable: a deployment may point the capable and efficient targets at the same model, which `Classifier::routing_tier` already handles explicitly, and a decline would then be indistinguishable from an escalation by target alone. There is a regression case for exactly this.

The fail-open branch is deliberately not recorded. `to_classification` returns `Ambiguous` when the verdict is missing, and that consultation is already counted by `classifier_fail_open`. The two metrics partition every consultation exactly once.

`verdict` is typed `&'static str`, so the label set is bounded by the type rather than by convention. `judge_model` is already used by `classifier_fail_open`. No request or response content enters label space.

## How tested

Not run: Python gates; this PR touches Rust observability and Markdown only.

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test -p switchyard-llm-client --test observability` green (11 passed)

Four cases in `crates/libsy-llm-client/tests/observability.rs`, mirroring `classifier_fail_open_records_each_failure_stage`:

| Case | Asserts |
|---|---|
| escalate verdict | `verdict="escalate"` is 1, **and the request still routed to the efficient tier** |
| decline | `verdict="hold"` is 1 |
| failed judge call | verdict counter **absent**, `classifier_fail_open` is 1 |
| both tiers on one model, decline | `verdict="hold"` is 1 |

The first pins non-duplication: the default `confirmations = 2` means a single verdict is counted while traffic stays put. The third pins the fail-open boundary. The fourth is the same-target regression.

Stashing only the recording calls while keeping the tests fails as expected:

```
assertion `left == right` failed: case escalate did not count the verdict
  left: None
 right: Some(1)
```

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. — N/A
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. — N/A
- [x] README / `--help` updated if customer-facing surface changed. — N/A; `docs/internal/metrics_reference.md` updated with the metric, its cardinality row, and a triage entry.

## Notes for reviewers

- `reason` is excluded from labels: it is free-form text from the judge and would need a bounded taxonomy designed first. Reasonable follow-up.
- This counts **consultations**, not traffic. The metrics reference says so explicitly, since the distinction is easy to miss.
- `linj/escalation-recovery-delatch` appears to touch the same area. This is meant to complement it: a de-latch path adds a third transition, which makes verdict counts more useful rather than less. Happy to rebase or defer.

## Summary by CodeRabbit

* **New Features**
  * Added monitoring for escalation decisions, including escalation and hold verdicts by judging model.
  * Improved visibility into routing behavior, confirmation thresholds, and fail-open events.

* **Bug Fixes**
  * Corrected escalation reporting so unresolved verdicts remain ambiguous and failed consultations are counted accurately.

* **Documentation**
  * Documented the new escalation verdict metrics, label details, counting behavior, and troubleshooting guidance.

### DIFF
--- a/crates/libsy-llm-client/tests/observability.rs
+++ b/crates/libsy-llm-client/tests/observability.rs
@@ -33,8 +33,9 @@ use tracing_subscriber::layer::{Context as LayerContext, SubscriberExt};
 use tracing_subscriber::registry::LookupSpan;
 
 use switchyard_libsy::{
-    AffinityRouter, Algorithm, Classifier, Driver, LibsyError, LlmClassifierConfig,
-    LlmTaskClassifier, PickerMode, StageRouter, StageRouterConfig, Step, TaskClassifierConfig,
+    AffinityRouter, Algorithm, Classifier, ClassifierContractConfig, Driver, EscalationJudgeConfig,
+    LibsyError, LlmClassifierConfig, LlmTaskClassifier, PickerMode, StageRouter, StageRouterConfig,
+    Step, TaskClassifierConfig,
 };
 use switchyard_llm_client::{ClientRouter, RunObservation, RunObserver};
 use switchyard_protocol::ModelId;
@@ -495,6 +496,26 @@ fn classifier_router(
                 base_threshold: 0.5,
                 ..TaskClassifierConfig::default()
             },
+        },
+    )?))
+}
+
+/// Escalation-mode counterpart to [`classifier_router`], on the packaged prompt and
+/// contract. `EscalationJudgeConfig::default()` requires two consecutive escalate
+/// verdicts to latch, so a single verdict is recorded without moving any traffic.
+fn escalation_router(
+    judge_model: &str,
+    efficient_model: &str,
+    capable_model: &str,
+) -> switchyard_libsy::Result<Arc<dyn Algorithm>> {
+    Ok(Arc::new(LlmTaskClassifier::new(
+        LlmClassifierConfig::Escalation {
+            judge_target: ModelId::from(judge_model),
+            efficient_target: ModelId::from(efficient_model),
+            capable_target: ModelId::from(capable_model),
+            contract: ClassifierContractConfig::default(),
+            config: EscalationJudgeConfig::default(),
+            max_output_tokens: 256,
         },
     )?))
 }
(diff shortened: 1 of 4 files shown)


nachiketb-nvidia:
This is duplicating already customizable behaviour. You can do this with llm_classifier already off main.

closing PR, pls comment if im wrong. Thanks!

yanyanz89:
Thank you @nachiketb-nvidia, I may be missing the intended replacement.

My understanding is that custom `llm_classifier` can expose a selected target/decision, but it is not equivalent to this metric for packaged `mode = "escalation"` because this PR counts the escalation judge's resolved verdicts per consultation, before they are collapsed into a target.

The concrete case I was trying to make observable is:

- `confirmations = 2`
- no `x-switchyard-session-id`
- judge returns `escalate` every turn

That produces the same final decisions as a healthy judge returning `hold` every turn: traffic stays on the efficient tier. But the operator action is different: the first case needs a session header or `confirmations = 1`, while the second is healthy.

It also differs after latch: traffic decisions continue, but judge consultations stop.

Is the recommendation to model this through custom `llm_classifier` instead of packaged escalation mode, or is there already a main-branch metric that exposes resolved escalation judge verdicts/consultations? If there is an equivalent path, I’m happy to drop this; otherwise I think this metric covers a gap in escalation observability.

PR #462, opened 2026-08-17
fix(translation): decode Anthropic structured output into requests

### LINKED ISSUE #452: [bug]: /v1/messages drops output_format / json_schema
## symptom

`/v1/messages` drops anthropic `output_format` / `json_schema`. the forwarded openai body is only `model`, `messages`, `max_completion_tokens`. http 200, no warning. the same process keeps `response_format` on `/v1/chat/completions`.

## reproduction

```toml
schema_version = 1
[llm_clients.local]
format = "openai_chat"
base_url = "http://127.0.0.1:9996/v1"
authorization = "Bearer SY-CAPTURE"
[targets.cap]
id = "captured-model"
llm_client = "local"
[routes.primary]
id = "captured-model"
type = "passthrough"
target = "cap"
```

```bash
# any capture that records the outbound body
switchyard-server --config sy.toml --port 9004
curl -s localhost:9004/v1/messages \
  -d '{"model":"captured-model","max_tokens":64,"output_format":{"type":"json_schema","schema":{"type":"object","properties":{"city":{"type":"string"},"ok":{"type":"boolean"}},"required":["city","ok"],"additionalProperties":false}},"messages":[{"role":"user","content":"ping"}]}'
```

forwarded keys, 5/5: `max_completion_tokens`, `messages`, `model`. no `response_format`, no `json_schema`.

same process, openai ingress (control), 5/5:

```bash
curl -s localhost:9004/v1/chat/completions \
  -d '{"model":"captured-model","max_tokens":16,"response_format":{"type":"json_schema","json_schema":{"name":"city","strict":true,"schema":{"type":"object","properties":{"city":{"type":"string"},"ok":{"type":"boolean"}},"required":["city","ok"],"additionalProperties":false}}},"messages":[{"role":"user","content":"hi"}]}'
```

forwarded body includes `response_format.type=json_schema`.

live gemini 2.5 flash, same prompt "return the city paris and ok true.", 3/3: caller gets a markdown fence (` ```json ... ``` `), not parseable json. one run truncated mid-fence.

## expected vs. actual

- expected: `output_format` maps onto `response_format` (the field the openai route already forwards), or the request is rejected.
- actual: the schema is stripped. the model is not asked for structured output.

## environment

- switchyard version (or commit sha): 0.2.0 (`switchyard-server`)
- python version (`python --version`): n/a (rust)
- os / arch: macos arm64
- inbound format (chat completions / anthropic messages / responses): anthropic messages
- backend (openai / anthropic / nvidia inference hub / other): openai-compatible mock for the wire, live gemini 2.5 flash via `https://generativelanguage.googleapis.com/v1beta/openai`

## additional context

claude code volume path is `ANTHROPIC_BASE_URL` -> `/v1/messages`. agents that send `output_format` get a fence they cannot `json.loads`.

wire captures: https://github.com/Atharva-Kanherkar/kairo/pull/11

ting-hong-shieh:
## What

Decodes Anthropic structured output into the neutral request contract, so a schema arriving on `/v1/messages` reaches the upstream instead of being dropped.

- `decode_anthropic_output_format` inverts the existing `encode_anthropic_output_format`. `output_config.format` is Anthropic's current field and wins; the top-level `output_format` is the earlier beta spelling that Anthropic still accepts, read as a fallback and added to the known-field list so it is not also copied into provider extensions.
- A format that cannot be mapped now reports through `push_lossy` instead of vanishing.

## Scope: enforcement is deliberately not addressed here

The mapping produces a schema without `strict: true`, so the upstream is asked to follow the schema rather than required to. Anthropic's `output_config.format` is the enforced kind, so this is a real downgrade — but setting `strict` unconditionally would reject Anthropic schemas that are legal today (any optional property), turning an unenforced response into a 400.

At a maintainer's request this PR stays in limited scope and restores the schema only. Enforcement is tracked in #467, which records the IR gap behind it and the accepted-subset difference that makes it more than a one-line change.

One smaller decision that is in scope: the neutral contract requires `json_schema.name` and Anthropic identifies the schema only by position, so requests decoded this way carry `"response"`. That value is visible on the wire — happy to derive it or use something else.

## Why

`crates/switchyard-translation/src/codecs/anthropic/buffered.rs` hard-coded the field when decoding a request:

```rust
output: OutputParams {
    max_output_tokens,
    response_format: None,
},
```

Three of the four legs already existed, so the asymmetry was the whole defect:

| path | before |
|---|---|
| OpenAI Chat decode `response_format` → IR | present |
| IR → OpenAI Chat encode `response_format` | present |
| IR → Anthropic encode `output_config.format` | present, added by #296 |
| Anthropic decode → IR | missing |

That also explains the control case in the report: the OpenAI ingress works because its decoder reads the field.

Worth noting that the reproduction in #452 uses `output_format`, the deprecated beta spelling. The current `output_config.format` was dropped in exactly the same way, so the defect was wider than the report showed.

Closes #452

## How tested

The checklist below is Python-oriented and this change is Rust-only, so those items are **not applicable**. Commands actually run, on Linux with cargo 1.96.1, at the head of this branch:

```
cargo fmt --all --check
cargo test --workspace
```

All three clean.

Seven tests in `crates/switchyard-translation/tests/request_translation.rs`: the current field, the legacy field, both present at once (current wins), no structured output at all, the three unmappable shapes each producing a diagnostic, `LossyConversionPolicy::Reject` turning that into an error, and an `effort`-only `output_config` still decoding reasoning effort untouched.

Also verified while writing it:

- same-format replay under default preservation is byte-identical to the source;
- the neutral shape is built field by field, so only `schema` is copied and a client cannot inject `strict` or other sibling keys through it;
- deeply nested schemas cannot drive the recursive constraint-stripper without bound — `serde_json` refuses to parse past roughly 128 levels, which I measured rather than assumed.

**Not verified.** I have not reproduced the original behavior or any fix against a live endpoint, Gemini or otherwise. The evidence is the translation path plus deterministic regression coverage, and I have not confirmed whether Anthropic accepts a `format.type` other than `json_schema` — that case currently produces a diagnostic and drops the format.

- [ ] `uv run ruff check .` clean — n/a, no Python changed
- [ ] `uv run mypy switchyard` clean — n/a, no Python changed
- [ ] `uv run pytest tests/` green — n/a, no Python changed
- [ ] Manual smoke — n/a, no live endpoint available; see above

## Checklist

- [ ] One class per file; filename = `snake_case` of the primary class. — n/a, no Python changed
- [ ] New public symbols exported from `switchyard/__init__.py.__all__`. — n/a, no new public symbols
- [ ] README / `--help` updated if customer-facing surface changed. — no CLI or README surface changed
- [x] Commits signed off (`Signed-off-by:`) per the DCO.

## Notes for reviewers

**Scope limit.** `encode_request` consults `exact_preserved_request` first, so a same-format Anthropic request under the default `PreservationPolicy::InMemory` is replayed verbatim and none of this runs — I verified the replay is byte-identical. Reaching the decode path requires cross-format translation, `PreservationPolicy::Disabled`, or a request built directly from the IR.

**A rename that follows from that.** Under `Disabled`, an Anthropic-to-Anthropic request arriving with the legacy `output_format` is re-emitted as `output_config.format`. Normalizing to the current field is defensible, but it is a rename on the wire and an upstream that only understands the beta spelling would break. I can preserve the incoming spelling instead if you prefer.

**Behavior change for strict callers.** Callers on `LossyConversionPolicy::Reject` now get a translation error where an unmappable format was previously accepted and forwarded unconstrained. Intended, but visible.

### DIFF
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -406,11 +406,14 @@ fn decode_anthropic_output_format(
         )?;
         return Ok(None);
     }
-    let Some(schema) = format.get("schema") else {
+    // A non-object schema would be forwarded verbatim into the neutral contract and
+    // reach the upstream as a malformed `json_schema.schema`, so it is refused here
+    // rather than handed on.
+    let Some(schema) = format.get("schema").filter(|schema| schema.is_object()) else {
         push_lossy(
             diagnostics,
             policy,
-            "Anthropic structured output requires format.schema; the requested format was dropped",
+            "Anthropic structured output requires format.schema to be an object; the requested format was dropped",
         )?;
         return Ok(None);
     };
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -28,6 +28,11 @@ use crate::util::{
     json_string, push_lossy, stable_id, string_value, validate_request_capabilities,
 };
 
+// Schema name applied when converting Anthropic structured output to the neutral
+// contract. Anthropic identifies the schema only by position, while the neutral
+// OpenAI shape requires a name, so requests that arrive this way share one.
+const ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME: &str = "response";
+
 /// Format codec for Anthropic Messages payloads.
 pub struct AnthropicMessagesCodec;
 
@@ -58,7 +63,7 @@ impl FormatCodec for AnthropicMessagesCodec {
                 .map(ToOwned::to_owned),
             output: OutputParams {
                 max_output_tokens,
-                response_format: None,
+                response_format: decode_anthropic_output_format(body),
             },
             sampling: SamplingParams {
                 temperature: body.get("temperature").and_then(Value::as_f64),
@@ -146,6 +151,7 @@ impl FormatCodec for AnthropicMessagesCodec {
                 "top_k",
                 "thinking",
                 "output_config",
+                "output_format",
                 "stream",
             ],
         );
@@ -358,6 +364,32 @@ impl FormatCodec for AnthropicMessagesCodec {
             diagnostics: Vec::new(),
         })
     }
+}
+
+// Reads Anthropic's structured-output schema into the neutral response format.
+//
+// `output_config.format` is the current field; the top-level `output_format` is
+// the earlier beta spelling that Anthropic still accepts, so both are read and
+// the current one wins. The neutral contract is OpenAI-shaped and requires a
+// schema name that Anthropic never sends, so one is supplied here.
+fn decode_anthropic_output_format(body: &Map<String, Value>) -> Option<Value> {
+    let format = body
+        .get("output_config")
+        .and_then(Value::as_object)
+        .and_then(|config| config.get("format"))
+        .or_else(|| body.get("output_format"))
+        .and_then(Value::as_object)?;
+    if format.get("type").and_then(Value::as_str) != Some("json_schema") {
+        return None;
+    }
+    let schema = format.get("schema")?;
+    Some(json!({
+        "type": "json_schema",
+        "json_schema": {
+            "name": ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME,
+            "schema": schema.clone(),
+        },
+    }))
 }
 
 /// Maps the neutral OpenAI-shaped JSON schema to Anthropic's output format.
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -28,11 +28,6 @@ use crate::util::{
     json_string, push_lossy, stable_id, string_value, validate_request_capabilities,
 };
 
-// Schema name applied when converting Anthropic structured output to the neutral
-// contract. Anthropic identifies the schema only by position, while the neutral
-// OpenAI shape requires a name, so requests that arrive this way share one.
-const ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME: &str = "response";
-
 /// Format codec for Anthropic Messages payloads.
 pub struct AnthropicMessagesCodec;
 
@@ -367,16 +362,8 @@ impl FormatCodec for AnthropicMessagesCodec {
     }
 }
 
-// Reads Anthropic's structured-output schema into the neutral response format.
-//
-// `output_config.format` is the current field; the top-level `output_format` is
-// the earlier beta spelling that Anthropic still accepts, so both are read and
-// the current one wins. The neutral contract is OpenAI-shaped and requires a
-// schema name that Anthropic never sends, so one is supplied here.
-//
-// A format that cannot be mapped is reported rather than dropped in silence: the
-// caller asked for constrained output and would otherwise receive prose with no
-// indication that the constraint never reached the upstream.
+// Reads the current `output_config.format`, or the beta `output_format` it replaced,
+// into the neutral OpenAI-shaped response format.
 fn decode_anthropic_output_format(
     body: &Map<String, Value>,
     diagnostics: &mut Vec<TranslationDiagnostic>,
@@ -406,9 +393,7 @@ fn decode_anthropic_output_format(
         )?;
         return Ok(None);
     }
-    // A non-object schema would be forwarded verbatim into the neutral contract and
-    // reach the upstream as a malformed `json_schema.schema`, so it is refused here
-    // rather than handed on.
+    // A non-object schema would reach the upstream as a malformed `json_schema.schema`.
     let Some(schema) = format.get("schema").filter(|schema| schema.is_object()) else {
         push_lossy(
             diagnostics,
@@ -420,7 +405,8 @@ fn decode_anthropic_output_format(
     Ok(Some(json!({
         "type": "json_schema",
         "json_schema": {
-            "name": ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME,
+            // Anthropic identifies the schema by position; the neutral shape needs a name.
+            "name": "response",
             "schema": schema.clone(),
         },
     })))
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -55,6 +55,7 @@ impl FormatCodec for AnthropicMessagesCodec {
                     })
             })
             .transpose()?;
+        let response_format = decode_anthropic_output_format(body, &mut diagnostics, policy)?;
         let mut request = LlmRequest {
             model: body
                 .get("model")
@@ -63,7 +64,7 @@ impl FormatCodec for AnthropicMessagesCodec {
                 .map(ToOwned::to_owned),
             output: OutputParams {
                 max_output_tokens,
-                response_format: decode_anthropic_output_format(body),
+                response_format,
             },
             sampling: SamplingParams {
                 temperature: body.get("temperature").and_then(Value::as_f64),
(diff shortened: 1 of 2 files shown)


ting-hong-shieh:
The schema-type point is correct and now fixed in 9268d2c1.

I had validated the format object and its `type` but read `schema` by presence alone, which is inconsistent on its face. Confirmed the consequence before fixing it — a string, number, array, or `null` was copied verbatim and forwarded:

```
schema="nonsense" -> {"json_schema":{"name":"response","schema":"nonsense"},"type":"json_schema"}  diagnostics=0
schema=42         -> {"json_schema":{"name":"response","schema":42},"type":"json_schema"}          diagnostics=0
schema=[1,2]      -> {"json_schema":{"name":"response","schema":[1,2]},"type":"json_schema"}       diagnostics=0
schema=null       -> {"json_schema":{"name":"response","schema":null},"type":"json_schema"}        diagnostics=0
```

`format.schema` must now be an object, refused through the same `push_lossy` path as the other unmappable shapes, with the four cases added to the regression list.

On `strict`: agreed, and it is the open question at the top of the PR description — I would rather a maintainer decide it than pick silently, because it determines whether this PR actually closes #452.

I did consider your suggestion of emitting a lossiness diagnostic when the conversion is relaxed, and decided against it for now. The diagnostic would fire on every successfully mapped schema, which under `LossyConversionPolicy::Reject` would turn each structured-output request into a translation error — a much larger regression than the problem it documents. It would only make sense paired with a decision to keep `strict` off deliberately, so I would rather resolve the field first. Happy to add it if the maintainers land on relaxed-and-documented.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean at 9268d2c1.

crates/switchyard-translation/tests/request_translation.rs
```diff
@@ -1403,6 +1403,215 @@ fn openai_request_translates_system_developer_and_reasoning_to_anthropic() -> Te
     Ok(())
 }
 
+// Builds an Anthropic request whose structured output uses the given field shape.
+fn anthropic_structured_output_request(output: Value) -> Value {
```
nachiketb-nvidia, line 1407: is it possible to make this all one cleaner test?

nachiketb-nvidia:
> Should the neutral format carry strict: true?

What does `strict: true` intended to do?

crates/switchyard-translation/tests/request_translation.rs
ting-hong-shieh: Done in ff3c2e35 — seven tests down to three, and the block is 128 lines shorter.

The mapping cases were near-identical: each rebuilt a request, translated it, and asserted on `response_format`. They are now one labelled table covering the current field, the legacy spelling, the current field winning over the legacy one, no structured output, an effort-only config, and the six shapes that cannot be mapped. Each row carries the schema it expects and whether a diagnostic is required.

Two cases stayed separate because they are not mapping assertions and folding them in would have meant special-casing rows: strict policy turning a drop into an error, and reasoning effort surviving alongside a schema in the same `output_config`. Happy to pull those in too if you would rather have literally one test.

Two things I did to keep the table from becoming the weaker kind of test:

- every assertion carries its row label, so a failure still says which case broke rather than just a line number;
- I checked it still has teeth by reverting the schema-object validation and confirming the table fails, then restoring it.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean at ff3c2e35.

ting-hong-shieh:
It marks the schema as enforced rather than advisory.

OpenAI splits this in two. Plain JSON mode gets you syntactically valid JSON with no guarantee it matches your schema. `strict: true` is what makes the API constrain generation to the schema, and their guide is explicit that "only Structured Outputs ensure schema adherence."

Anthropic's `output_config.format` is already the enforced kind. So the mapping as it stands converts a guarantee into a best-effort request, which matters here because the symptom in #452 is exactly an unenforced schema — the caller got a markdown fence back rather than parseable JSON. Forwarding the schema without `strict` may leave that unchanged.

**But setting it unconditionally would be worse**, and this is the part I had not pinned down when I opened the PR. The two providers accept different schema subsets, and OpenAI rejects an out-of-subset schema at request time when `strict` is on:

| | Anthropic `output_config.format` | OpenAI `strict: true` |
|---|---|---|
| every property in `required` | not required — optional properties are allowed | **required** |
| `additionalProperties: false` | required | required |
| `allOf` | supported, with limits | **not allowed** |
| string `format`, `minItems` | supported | restricted for fine-tuned models |

The first row is the problem. An Anthropic schema with any optional property is perfectly legal today, and switching on `strict` would turn that request into a 400 instead of a fence. That is a worse failure than the one we are fixing, and it would hit schemas that work today.

So I do not think either constant is right. What I would suggest instead: set `strict: true` only when the schema provably satisfies the strict subset — every property present in `required`, `additionalProperties: false`, and none of the forbidden composition keywords — and omit it otherwise. Enforcement where it is safe, and a request that still works where it is not.

That also mirrors what this codec already does in the other direction: `strip_anthropic_unsupported_constraints` narrows an OpenAI schema to what Anthropic accepts rather than passing it through and hoping.

Worth noting the reporter's schema in #452 satisfies the strict subset, so that path would go from "fence" to enforced — the reported case actually gets fixed rather than merely forwarded.

Happy to implement that predicate here, or to leave `strict` off entirely and document the limitation, whichever you prefer. I did not want to pick between them silently since it changes what the upstream does with a caller's schema.

crates/switchyard-translation/tests/request_translation.rs
nachiketb-nvidia: do we need this much testing? every new test leads to more test burden and maintainence

nachiketb-nvidia:
@ting-hong-shieh lets create a new issue to address IR defficiencies such as strictness, keep this MR as is, in limited scope

crates/switchyard-translation/tests/request_translation.rs
ting-hong-shieh: Cut further in 639b92d6 — three tests to two, and the block is now 125 lines against the 215 I opened with.

What went: four rows differed only in which non-object value stood in for the schema, which tests one predicate four times, so that is one row now. The effort case rides along with the first row instead of owning a test. The per-row diagnostic flag is derived from the case rather than spelled out.

What is left is one row per behavior:

| row | what breaks if it goes |
|---|---|
| current field | the fix itself |
| legacy beta field | the alias I added silently stops working |
| current wins over legacy | a client mid-migration sends both and gets the stale schema |
| no structured output | we invent a `response_format` that was never asked for |
| unsupported format type | the drop goes silent again |
| non-object schema | malformed schema forwarded upstream — the bug CodeRabbit caught here |

Plus one test for strict policy returning an error, which is a different code path from a diagnostic.

I would rather not cut past this, and here is my reasoning: two of those rows exist because the review process on this PR found real defects — the non-object schema was a genuine bug, and the silent drop was the thing #449 complained about. Removing them would delete the regression cover for defects we found this week. If you still want it smaller, tell me which behaviors you are willing to stop guarding and I will drop those rows; I just do not want to pick them myself.

On the burden point more generally, I think you are right that the first version was too much, and the fault was shape rather than count — seven near-identical functions each rebuilding a request. One table costs far less to maintain than seven bodies, since a new case is a row rather than thirty lines.

I also checked the trimmed table still has teeth rather than just being shorter, by breaking three things and confirming it fails each time: removing the schema-object validation, removing the legacy alias, and changing the diagnostic message. All three are caught; restored, it passes.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean at 639b92d6.

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -360,6 +367,65 @@ impl FormatCodec for AnthropicMessagesCodec {
     }
 }
 
+// Reads Anthropic's structured-output schema into the neutral response format.
```
nachiketb-nvidia: nit: can we reduce the comments to a one-liner, fewer comments, better for humans (and even agents)

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -28,6 +28,11 @@ use crate::util::{
     json_string, push_lossy, stable_id, string_value, validate_request_capabilities,
 };
 
+// Schema name applied when converting Anthropic structured output to the neutral
+// contract. Anthropic identifies the schema only by position, while the neutral
+// OpenAI shape requires a name, so requests that arrive this way share one.
+const ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME: &str = "response";
```
nachiketb-nvidia: do we need this if its used just once? don't need as many comments too

nachiketb-nvidia (commented):
looks good, just a few nits, pls address the coderabbit

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -360,6 +367,62 @@ impl FormatCodec for AnthropicMessagesCodec {
...
+        )?;
+        return Ok(None);
+    };
+    Ok(Some(json!({
+        "type": "json_schema",
+        "json_schema": {
+            "name": ANTHROPIC_STRUCTURED_OUTPUT_SCHEMA_NAME,
+            "schema": schema.clone(),
+        },
+    })))
```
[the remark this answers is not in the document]
ting-hong-shieh: Replying in the thread so it is not left looking open — I answered in a top-level comment earlier, which was the wrong place.

The schema-type point was correct and is fixed in 9268d2c1. `format.schema` had been read by presence alone, so a string, number, array, or `null` was copied verbatim and forwarded:

```
schema="nonsense" -> {"json_schema":{"name":"response","schema":"nonsense"},"type":"json_schema"}  diagnostics=0
schema=42         -> {"json_schema":{"name":"response","schema":42},"type":"json_schema"}          diagnostics=0
schema=[1,2]      -> {"json_schema":{"name":"response","schema":[1,2]},"type":"json_schema"}       diagnostics=0
schema=null       -> {"json_schema":{"name":"response","schema":null},"type":"json_schema"}        diagnostics=0
```

It must now be an object, refused through the same `push_lossy` path as the other unmappable shapes, with a regression row covering it.

On the `strict` half: a maintainer has asked to keep this PR in limited scope and track enforcement separately, so it is deliberately unchanged here and moves to a follow-up issue. I had also considered your suggestion of emitting a lossiness diagnostic for the relaxed conversion and think it needs its own severity rather than `push_lossy` — as written it would fire on every successfully mapped schema, which under `LossyConversionPolicy::Reject` turns each structured-output request into a translation error.

ting-hong-shieh:
Done — #467 tracks enforcement, and this PR stays as it is.

The issue records why it is not a one-liner: the IR has nowhere to say whether a schema must be enforced, so the field is carried as one provider's spelling and reinterpreted at the other boundary. Anthropic loses enforcement on the way out to OpenAI Chat, and gains it on the way in, neither with a diagnostic. It also captures the accepted-subset difference — an Anthropic schema with any optional property is legal today and would start returning 400 if `strict` were switched on blindly — plus the three directions I can see, without proposing one.

Also addressed here:

- both nits in 62aa79a5 — the single-use schema-name constant is inlined and the block comments are cut to the line each one needed;
- the CodeRabbit thread now has a reply on the thread itself. That was my mistake: I answered it in a top-level comment earlier, which left the thread looking untouched. The substance was that its schema-type finding was right and is fixed in 9268d2c1; its `strict` half is what #467 now carries.

I updated the PR description too, since it still opened with `strict` as a blocking question.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean at 62aa79a5.

nachiketb-nvidia (approved):

PR #470, opened 2026-08-18
fix(translation): encode data URI images as Anthropic base64 sources

### LINKED ISSUE #468: Anthropic translation forwards data: URI image_url as a URL-type source, breaking base64 vision requests
## Summary

When routing an OpenAI-format chat completion request containing a `data:` URI in an `image_url.url` content part to an `anthropic_messages`-format target, Switchyard forwards the string verbatim into Anthropic's `image` content block as a URL-type source (`source: {type: "url", url: "data:image/png;base64,..."}`). Anthropic's API correctly rejects this, since a URL-type source must be a real `http(s)://` link, not a base64 data URI. The resulting error is opaque to callers: it's returned as a generic upstream 400 with message `"Only HTTPS URLs are supported."`, giving no indication that the actual problem is the missing data-URI translation.

## Expected behavior

Per OpenAI's chat completions API, `image_url.url` may be either a real `http(s)://` URL or a `data:<mime>;base64,<data>` URI (this is the standard way OpenAI-compatible clients send inline images without external hosting). When translating to Anthropic's Messages API format, Switchyard should detect the `data:` prefix, parse out the media type and base64 payload, and emit Anthropic's base64-type image source instead:

```json
{"type": "image", "source": {"type": "base64", "media_type": "image/png", "data": "<payload>"}}
```

URLs that are already `http(s)://` should presumably continue to pass through as Anthropic's URL-type source unchanged.

## Steps to reproduce

Configuration: a target with `format = "anthropic_messages"` pointing at the real Anthropic API (`base_url = "https://api.anthropic.com"`).

Request (OpenAI-format, sent to Switchyard's `/v1/chat/completions`):

```json
{
  "model": "<anthropic-backed route>",
  "messages": [{
    "role": "user",
    "content": [
      {"type": "text", "text": "What color is this?"},
      {"type": "image_url", "image_url": {"url": "data:image/png;base64,<...>"}}
    ]
  }]
}
```

Observed response: HTTP 400, `{"error": {"code": "upstream_error", "message": "Only HTTPS URLs are supported.", "type": "upstream_error"}}`.

## Confirmation this is translation-side, not client/config error

- The same request against a `format = "openai_chat"` target (Google's OpenAI-compatible Gemini endpoint, passthrough with no format translation) succeeds and correctly describes the image — so the base64 payload itself is well-formed and the issue is specific to the Anthropic translation path.
- The literal error string `"Only HTTPS URLs are supported."` does not appear anywhere in the compiled Switchyard binary (checked via binary grep), confirming it's Anthropic's own API error being relayed through, not a Switchyard-side validation message — i.e. Switchyard is constructing a request Anthropic considers malformed, rather than rejecting it before the request is sent.

## Environment

- Switchyard built from `main` via `cargo build --release` (Docker image built same-day as this report).
- Reached via a target with `llm_client` pointing at `format = "anthropic_messages"`, `base_url = "https://api.anthropic.com"`.

bharadwaj-pendyala:
## What

Anthropic image blocks now carry a `base64` source when the incoming OpenAI `image_url.url` is a `data:` URI, rather than a `url` source holding the whole data URI.

One match arm in `encode_one_anthropic_block` plus a `split_base64_data_uri` helper, both in `crates/switchyard-translation/src/codecs/anthropic/buffered.rs`. Three regression tests.

## Why

`decode_image_source` maps every `image_url.url` to `ImageSource::Url` whatever the scheme (`crates/switchyard-translation/src/codecs/openai_chat/buffered.rs:547`), and the Anthropic encoder rendered that variant as a URL source. An inline image therefore reached Anthropic as:

```json
{"type": "image", "source": {"type": "url", "url": "data:image/png;base64,aW1hZ2U="}}
```

Anthropic rejects that with `"Only HTTPS URLs are supported."`, and the error says nothing about the missing translation. Data URIs are how OpenAI-compatible clients send images without hosting them first, so this hits every base64 vision request routed to an `anthropic_messages` target.

`encode_one_anthropic_tool_result_block` delegates image blocks to the same function, so tool-result images are covered by construction. I could not write a translation-level test for that path: both `openai_chat` (`codecs/openai_chat/buffered.rs:129`) and `responses` (`codecs/responses/buffered.rs:419`) collapse a tool output to a single `ContentBlock::Text`, so no `ToolResult` carrying an image reaches the encoder from either source format today.

Closes #468

## How tested

Rust gates, run on this commit against base `1700f62`:

- [x] `cargo test --workspace --exclude switchyard-py` green, 559 passed, 0 failed

`switchyard-py` is excluded because it does not link on my macOS box (`linking with cc failed`, aws-lc-sys objects built for macOS 26.5 linked at 11.0). I confirmed that failure is pre-existing by stashing this change and rebuilding, so it is my toolchain, not the diff. Everything else in the workspace runs.

`openai_data_uri_image_translates_to_anthropic_base64_source` fails on unmodified `main`:

```
< "type": String("url"),
< "url": String("data:image/png;base64,aW1hZ2U="),
> "data": String("aW1hZ2U="),
> "media_type": String("image/png"),
> "type": String("base64"),
```

Python gates, all clean, though this change touches no Python:

- [x] `uv run ruff check .` clean, "All checks passed!"
- [x] `uv run mypy switchyard` clean, "no issues found in 21 source files"
- [x] `uv run pytest tests/` green, 145 passed, 2 skipped
- [x] Manual smoke: none against a live endpoint. Confirming the wire shape against Anthropic needs a key, so I pinned it with the translation crate's own tests instead.

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. Not applicable, Rust only, no new file.
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. Not applicable, the helper is private to the module.
- [x] README / `--help` updated if customer-facing surface changed. Not applicable, no surface change.

## Notes for reviewers

**Why the encoder and not the decoder.** Normalizing in `decode_image_source` so the neutral IR carries `ImageSource::Base64` is the other option, and it is arguably the better boundary. It costs something today: `detail` lives only on the `Url` variant (`crates/protocol/src/llm.rs:136`), so an `openai_chat` to `openai_chat` translation would start silently dropping `"detail": "high"` on inline images. Fixing that properly means adding `detail` to `ImageSource::Base64`, a public protocol change I did not want to make unasked in a bug fix. Encoding it here keeps the constraint where it applies, since Anthropic is the only target that refuses a data URI. Happy to move it if you would rather change the IR, and I raised the same choice on #468.

**What is deliberately left alone.** A data URI with no media type, or with a percent-encoded payload, still goes out as a URL source, unchanged from today. Both are wrong for Anthropic, but guessing a MIME type or re-labelling percent-encoded bytes as base64 would corrupt the payload rather than relay it, so the test `percent_encoded_data_uri_image_stays_an_anthropic_url_source` pins the pass-through. RFC 2397 also makes the `;base64` marker case-insensitive; I match it lowercase only, since that is what OpenAI-compatible clients emit, and an unmatched marker just falls back to today's behaviour. Say the word if you want either handled.

This was written with AI assistance, which CONTRIBUTING allows. I can explain and defend every line.

### DIFF
--- a/crates/switchyard-translation/tests/request_translation.rs
+++ b/crates/switchyard-translation/tests/request_translation.rs
@@ -1581,6 +1581,60 @@ fn anthropic_unmappable_output_format_is_rejected_under_strict_policy() -> TestR
     Ok(())
 }
 
+// Verifies inline `data:` images become Anthropic base64 sources, since Anthropic's URL
+// source rejects anything that is not an http(s) link. A percent-encoded payload is not raw
+// base64, so it keeps the URL source it had before rather than reaching Anthropic mislabelled.
+#[test]
+fn openai_data_uri_images_translate_to_anthropic_sources() -> TestResult {
+    let engine = TranslationEngine::default();
+    let cases = [
+        (
+            "base64 data URI",
+            json!({"url": "data:image/png;base64,aW1hZ2U=", "detail": "high"}),
+            json!({"type": "base64", "media_type": "image/png", "data": "aW1hZ2U="}),
+        ),
+        (
+            "data URI carrying extra parameters",
+            json!({"url": "data:image/jpeg;charset=binary;base64,aW1hZ2U="}),
+            json!({"type": "base64", "media_type": "image/jpeg", "data": "aW1hZ2U="}),
+        ),
+        (
+            "percent-encoded data URI",
+            json!({"url": "data:image/png;base64,YQ%3D%3D"}),
+            json!({"type": "url", "url": "data:image/png;base64,YQ%3D%3D"}),
+        ),
+    ];
+
+    for (case, image_url, expected_source) in cases {
+        let body = json!({
+            "model": "claude-sonnet-4-20250514",
+            "messages": [{
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "What color is this?"},
+                    {"type": "image_url", "image_url": image_url}
+                ]
+            }]
+        });
+
+        let output = engine
+            .translate_request(
+                WireFormat::OpenAiChat,
+                WireFormat::AnthropicMessages,
+                &body,
+                &TranslationPolicy::default(),
+            )?
+            .body;
+
+        assert_eq!(
+            output["messages"][0]["content"][1],
+            json!({"type": "image", "source": expected_source}),
+            "{case}"
+        );
+    }
+    Ok(())
+}
+
 // Verifies Anthropic receives its supported schema subset without mutating the neutral contract.
 #[test]
 fn openai_schema_constraints_are_removed_from_anthropic_output_format() -> TestResult {
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -908,9 +908,13 @@ fn encode_one_anthropic_block(block: &ContentBlock) -> Vec<Value> {
             })]
         }
         ContentBlock::Image { source } => vec![match source {
-            ImageSource::Url { url, .. } => {
-                json!({"type": "image", "source": {"type": "url", "url": url}})
-            }
+            ImageSource::Url { url, .. } => match split_base64_data_uri(url) {
+                Some((media_type, data)) => json!({
+                    "type": "image",
+                    "source": {"type": "base64", "media_type": media_type, "data": data},
+                }),
+                None => json!({"type": "image", "source": {"type": "url", "url": url}}),
+            },
             ImageSource::Base64 { media_type, data } => json!({
                 "type": "image",
                 "source": {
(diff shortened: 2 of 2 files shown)


bharadwaj-pendyala:
On the tool-result test suggestion: I checked, and that path is not reachable through `translate_request` today. `openai_chat` builds every tool message as `ToolResult { content: vec![ContentBlock::Text { .. }] }` (`codecs/openai_chat/buffered.rs:129`), and `responses` does the same for `function_call_output` (`codecs/responses/buffered.rs:419`), so no `ToolResult` carrying an image ever reaches the encoder from either source format. An Anthropic `image` block inside a tool result decodes to `ImageSource::Raw` and is relayed verbatim, so it never hits the new arm either.

`encode_one_anthropic_tool_result_block` still delegates images to `encode_one_anthropic_block`, so the fix applies by construction if a source format ever starts carrying them. I have updated the PR body to say that rather than claim test coverage I cannot write.

crates/switchyard-translation/tests/request_translation.rs
```diff
@@ -1403,6 +1403,108 @@ fn openai_request_translates_system_developer_and_reasoning_to_anthropic() -> Te
     Ok(())
 }
 
+// Verifies inline `data:` images become Anthropic base64 sources, since Anthropic's
+// URL source rejects anything that is not an http(s) link.
+#[test]
+fn openai_data_uri_image_translates_to_anthropic_base64_source() -> TestResult {
```
nachiketb-nvidia: nit: lets reduce to one test, which tests all three, maybe using loops

nachiketb-nvidia (approved):
nit, but otherwise LGTM!

Thank you!

bharadwaj-pendyala:
Done, the three tests are now one table-driven test over the three data URI shapes, matching the `for (case, ...) in cases` pattern already used further down the file.

Also rebased onto `main` at 6babb3b. The branch had gone 11 commits behind, and #462 added tests at the same spot in `request_translation.rs`; the resolution keeps both sets.

Re-verified the fix still does something on current `main`. Reverting only `buffered.rs` to main's copy and keeping the new test:

```
---- openai_data_uri_images_translate_to_anthropic_sources stdout ----
assertion failed: `(left == right)`: base64 data URI
<        "type": String("url"),
<        "url": String("data:image/png;base64,aW1hZ2U="),
>        "data": String("aW1hZ2U="),
>        "media_type": String("image/png"),
>        "type": String("base64"),
```

With the fix restored, `cargo test -p switchyard-translation` is 141 passed, 0 failed. `cargo fmt --all -- --check` and `cargo clippy -p switchyard-translation --all-targets -- -D warnings` are both clean.

One thing I made explicit in the comment while I was in there. `split_base64_data_uri` declines a percent-encoded payload and a data URI with no media type, so both keep relaying as a URL source exactly as they did before this PR. Handling `data:image/png;base64,YQ%3D%3D` properly means percent-decoding the body before treating it as base64, which felt like more than this fix should take on. Happy to add it here if you would rather it were covered.

PR #483, opened 2026-08-19
fix: sanitize Responses reasoning across backend handoffs

### LINKED ISSUE #481: Responses backend rejects plaintext reasoning after cross-provider routing
## Problem

When a stage router switches between a local OpenAI Responses-compatible model and a strict OpenAI Responses backend, provider-specific reasoning items can make the next target reject otherwise valid conversation history.

### Local to strict OpenAI

Some local servers, including llama.cpp, return reasoning items with plaintext `content` and an empty `encrypted_content`. Replaying that history to the strict backend fails with:

```text
Invalid 'input[3].content': array too long. Expected an array with maximum length 0, but got an array with length 1 instead.
```

### Strict OpenAI to local

The reverse handoff has a separate incompatibility. A signed, encrypted reasoning item normalized for strict OpenAI cannot be consumed by llama.cpp:

```text
item['content'] is not an array
```

Adding an empty array is also rejected:

```text
item['content'] is empty
```

## Expected behavior

Switchyard should normalize Responses reasoning items for the selected target while preserving messages, function calls, and function-call outputs:

- With the default `responses_reasoning = "preserve_encrypted"` policy, remove unsigned/plaintext reasoning items and retain signed encrypted reasoning with an empty plaintext `content` array.
- With an explicit `responses_reasoning = "drop"` policy, remove all reasoning items for a local backend that cannot consume another provider's encrypted representation.
- Preserve the remaining conversation and tool-call history in order.

The policy is explicit rather than inferred from model names, URLs, or authentication, so authenticated local servers and unauthenticated hosted-compatible servers can be configured correctly.

## Implementation and validation

A focused implementation is available on the fork branch:

https://github.com/srchandrupatla/Switchyard/tree/fix/responses-reasoning-handoff

Relevant commits:

- `30c648c2` — sanitize unsigned Responses reasoning handoffs
- `ab9b9e8f` — add the explicit, model-agnostic Responses reasoning replay policy
- `0971862f` — keep the existing public backend configuration API source-compatible

Validation completed so far:

- Formatting and Clippy pass.
- 55 `switchyard-llm-client` unit tests pass.
- 11 client observability tests pass.
- The full non-PyO3 Rust workspace test suite passes.
- Dedicated HTTP-level regression tests cover both handoff directions.
- A release build on macOS passed a real llama.cpp cloud-to-local request with HTTP 200.
- The reverse local-to-`gpt-5.6-sol` request also completed successfully.
- Two isolated 20-request bidirectional stress runs completed with 20/20 successes.

I am continuing production stress testing before opening a PR.

srchandrupatla:
## What

Add an explicit per-client policy for replaying OpenAI Responses reasoning items across backend handoffs.

- Introduce `ResponsesReasoningPolicy` with `preserve_encrypted` (default) and `drop` modes.
- Sanitize reasoning at the outbound backend boundary while preserving messages, tool calls, and tool results.
- Expose the policy as `responses_reasoning` for `openai_responses` clients in TOML.
- Keep the existing public `HttpBackendConfig` construction API source-compatible.
- Document the new configuration and its intended use for local Responses-compatible backends.

## Why

When a routed Codex conversation moved between a local llama.cpp backend and the hosted Codex Responses backend, provider-specific reasoning items were replayed unchanged. Plaintext local reasoning sent to the hosted backend could be rejected with:

```text
Invalid 'input[3].content': array too long. Expected an array with maximum length 0
```

Encrypted hosted reasoning is similarly not meaningful to a local backend. This made otherwise valid multi-turn sessions fail only after the router changed providers.

The policy is explicit rather than inferred from model names, URLs, or authentication. Existing clients retain the conservative `preserve_encrypted` behavior; local clients can opt into `responses_reasoning = "drop"`.

Closes #481

## How tested

- [x] `uv run ruff check .`
- [x] `uv run pytest tests/ -v -m "not integration"` — 143 passed, 2 deselected
- [x] Relevant Rust workspace tests, including 55 client tests, 11 observability tests, 45 server tests, and 37 server integration tests
- [x] Full `cargo test --workspace` on macOS with only the upstream PyO3 test-link workaround applied temporarily — all tests passed; the workaround was reverted and is not part of this PR
- [x] Live llama.cpp + Codex routing smoke and stress tests — 59/59 requests passed, including 20 cloud-to-local and 20 local-to-cloud reasoning-history handoffs and concurrent four-slot traffic

### macOS PyO3 note

The repository currently enables PyO3's deprecated `extension-module` feature unconditionally. On macOS, that prevents a plain `cargo test --workspace` from linking `switchyard-py` because test binaries need `libpython`, while extension modules deliberately do not link it. PyO3 recommends letting maturin >= 1.9.4 enable extension-module mode only for packaging. Switchyard already uses maturin 1.13.1.

This same macOS failure was previously documented, but not fixed, in #411. I verified that removing only `extension-module` from the PyO3 dependency makes the complete workspace compile and pass on macOS. That unrelated packaging change was reverted to keep this PR focused and is proposed separately in #484; Linux CI on the current upstream base is green.

PyO3 guidance: https://pyo3.rs/main/building-and-distribution.html#the-pyo3_build_extension_module-environment-variable

## Checklist

- [x] Change is focused on Responses reasoning handoffs.
- [x] Unit and configuration tests cover both policies and invalid configurations.
- [x] README and TOML schema documentation are updated.
- [x] All commits include DCO `Signed-off-by` trailers.
- [x] No model-name, URL, or authentication heuristics are used.

## Notes for reviewers

The sanitizer is deliberately owned by `libsy-llm-client`, where the final outbound backend capability is known. `preserve_encrypted` removes plaintext reasoning while retaining encrypted reasoning with an empty content array; `drop` removes all reasoning items for backends that cannot safely consume provider-specific reasoning state.

## Summary by CodeRabbit

* **New Features**
  * Added configurable reasoning-history handling for OpenAI Responses clients.
  * Added `preserve_encrypted` and `drop` policies, preserving messages and tool-call history as appropriate.
  * Added validation to prevent this setting from being used with unsupported client formats.

* **Documentation**
  * Documented the new configuration option, defaults, supported values, and local-server usage.

* **Tests**
  * Added coverage for hosted and local Responses behavior, including encrypted and unsigned reasoning items.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2301,6 +2301,7 @@ dependencies = [
  "opentelemetry_sdk",
  "parking_lot",
  "reqwest",
+ "serde",
  "serde_json",
  "switchyard-libsy",
  "switchyard-protocol",
--- a/crates/libsy-llm-client/Cargo.toml
+++ b/crates/libsy-llm-client/Cargo.toml
@@ -29,6 +29,7 @@ parking_lot.workspace = true
 http.workspace = true
 httpdate.workspace = true
 serde_json.workspace = true
+serde.workspace = true
 tokio.workspace = true
 tracing.workspace = true
 tracing-opentelemetry.workspace = true
--- a/crates/libsy-llm-client/README.md
+++ b/crates/libsy-llm-client/README.md
@@ -59,7 +59,8 @@ switchyard-translation = { path = "../switchyard-translation" }   # for WireForm
 ```rust
 use std::collections::BTreeMap;
 use switchyard_llm_client::{
-    Backend, HttpBackendConfig, ModelConfig, TranslatingLlmClient,
+    Backend, HttpBackendConfig, ModelConfig, ResponsesReasoningPolicy,
+    TranslatingLlmClient,
 };
 
 fn build_client() -> switchyard_llm_client::Result<TranslatingLlmClient> {
@@ -70,6 +71,7 @@ fn build_client() -> switchyard_llm_client::Result<TranslatingLlmClient> {
         extra_headers: BTreeMap::new(),
         extra_body: BTreeMap::new(),
         max_retries: 2,
+        responses_reasoning: ResponsesReasoningPolicy::default(),
     };
 
     let models = [ModelConfig::new(
@@ -246,6 +248,9 @@ fn build_multi_format_client(
   transport failures, timeouts, HTTP 408/429, and 5xx responses. Buffered body
   transport failures are retried; streaming body failures are not replayed after
   the response has been returned.
+- `HttpBackendConfig::responses_reasoning` controls reasoning-item replay for
+  Responses backends. `PreserveEncrypted` keeps signed provider state without
+  plaintext; `Drop` removes reasoning while retaining messages and tool history.
 
 Retries replay the same upstream request to the same model. Each candidate's
 `max_retries` budget is exhausted before candidate fallback advances to the next
--- a/crates/libsy-llm-client/README.md
+++ b/crates/libsy-llm-client/README.md
@@ -59,8 +59,7 @@ switchyard-translation = { path = "../switchyard-translation" }   # for WireForm
 ```rust
 use std::collections::BTreeMap;
 use switchyard_llm_client::{
-    Backend, HttpBackendConfig, ModelConfig, ResponsesReasoningPolicy,
-    TranslatingLlmClient,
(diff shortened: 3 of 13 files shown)


nachiketb-nvidia (commented):
could you address the coderabbit comments first and reply to them, thanks!

bhuvan002:

PR #489, opened 2026-08-19
fix(llm-client): do not add a fifth cache_control block

fedecia:
## What

`enable_anthropic_prompt_caching` counts the `cache_control` markers already in the
request body and abstains once four are present, instead of always inserting one.

One guard plus a small recursive counter. No behaviour change below the limit.

## Why

The function marks the final message content block for every `Backend::Anthropic`.
Anthropic and Bedrock both cap a request at **four** blocks carrying `cache_control`,
counting tools, system blocks and message blocks **together**. A caller that has
already spent the budget therefore gets a fifth marker and the request fails upstream:

```
A maximum of 4 blocks with cache_control may be provided. Found 5.
```

The insert is idempotent per block, so a client that marks its own final block costs
nothing. The failing shape is a client that spends four breakpoints elsewhere — on
tools and system blocks — and leaves the last message block unmarked.

Claude Code does exactly that, on some request shapes and not others, which is why it
presents as flakiness rather than a hard failure. Observed with `switchyard/*` routes
to Bedrock via a PortKey gateway: **0 of 4** non-streaming calls succeeded before the
change, **31 of 31** after, then 133 calls in one live session with no failure of any
kind. Streaming turns were unaffected throughout, so a session keeps working while
some calls die.

Worth naming, because it is the symptom a user actually reports: Claude Code's
**auto mode fails first**. Its permission classifier is one of the affected request
shapes, so the tool-approval path breaks while ordinary conversation keeps working,
and the surfaced message is that Claude Code cannot reach a model. The real error is
only in the router log.

Introduced by #233, which enables Anthropic prompt caching by default. Its review
checklist covers the case where the final block is already marked — that case is
handled — but not the total across all blocks.

No issue filed; this PR is the report. Happy to open one if you would rather track it
that way.

## How tested

The Python gates do not apply — this is a Rust-only change to `crates/libsy-llm-client`.

- [x] `cargo clippy -p switchyard-llm-client --all-targets -- -D warnings` clean
- [x] `cargo test -p switchyard-llm-client` green — 53 tests, including three new
- [x] Manual smoke: a 4-marker body returns HTTP 200 against the live gateway where it
      previously returned 400; a 5-marker body still returns 400, which is correct
      because the caller, not the client, is over the limit

New tests:

- `anthropic_prompt_caching_abstains_at_the_cache_control_limit` — a body at four
  markers is left untouched
- `anthropic_prompt_caching_still_marks_one_below_the_limit` — three markers still get
  the fourth, so caching is not silently disabled
- `anthropic_prompt_caching_marks_final_message` (existing) — unchanged

## Checklist

- [x] Unit tests added for the bug fix
- [x] Commits signed off (`Signed-off-by`) per the DCO
- [ ] One class per file — n/a, Rust
- [ ] New public symbols exported — n/a, both additions are private
- [ ] README / `--help` updated — n/a, no customer-facing surface change

## Notes for reviewers

**The limit is a constant here (`MAX_CACHE_CONTROL_BLOCKS = 4`).** It matches what
Anthropic and Bedrock enforce today. If another Anthropic-compatible backend allows a
different number, this wants to come from the backend rather than the crate.

**Abstaining is the deliberate choice over evicting.** When the budget is spent, a
caller that placed four breakpoints itself needs them more than we need a fifth, and
dropping one of theirs to make room would silently change a caching decision they
made. The cost is that the final message block is not cached on those requests.

**The counter walks the whole body.** It is O(nodes) on a serialized request and runs
once per call. If that is not acceptable on the hot path, it could count only `tools`,
`system` and `messages` — but the recursive form cannot miss a marker in a shape we
have not anticipated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prompt caching behavior for Anthropic requests by respecting the four-marker limit.
  * Prevented unnecessary cache markers when the limit has already been reached.
  * Added coverage for requests at the limit and those with one available slot.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -754,8 +754,33 @@ fn merge_extra_body(body: &mut Value, extra_body: &BTreeMap<String, Value>) {
     }
 }
 
+// Anthropic and Bedrock both cap a request at four blocks carrying
+// `cache_control`, counting tools, system blocks and message blocks together.
+const MAX_CACHE_CONTROL_BLOCKS: usize = 4;
+
+// Counts the blocks upstream will see carrying a `cache_control` marker. The
+// marker's own value holds no such key, so it is never counted twice.
+fn count_cache_control_blocks(value: &Value) -> usize {
+    match value {
+        Value::Object(map) => {
+            usize::from(map.contains_key("cache_control"))
+                + map.values().map(count_cache_control_blocks).sum::<usize>()
+        }
+        Value::Array(items) => items.iter().map(count_cache_control_blocks).sum(),
+        _ => 0,
+    }
+}
+
 // Marks the final message content block as the Anthropic prompt-cache breakpoint.
 fn enable_anthropic_prompt_caching(body: &mut Value) {
+    // Abstain once the caller has spent the budget itself. Adding a fifth marker
+    // turns a request that was valid on arrival into an upstream HTTP 400, and a
+    // caller that placed four breakpoints deliberately needs them more than we
+    // need a fifth. When the final block is already marked this returns early
+    // and changes nothing, which is what the insert below would have done.
+    if count_cache_control_blocks(body) >= MAX_CACHE_CONTROL_BLOCKS {
+        return;
+    }
     let Some(content) = body
         .get_mut("messages")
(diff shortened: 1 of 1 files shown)


nachiketb-nvidia (approved):
thanks for the catch!

PR #487, opened 2026-08-19
feat(libsy): pin the classifier target across a user turn

ryan-lempka:
Adds a new algorithm utility (`turn_pin.rs`) that enables detection of whether the last message in the conversation was the user, and then re-runs the LLM classifier on that user message. If the classifier is not re-run you use the prior routed model selected. This feature will be useful when using the LLM Classifier algorithm in interactive agent settings.

Setting can be enabled via `classify_trigger = user_turn | new_session | every_request` within the routing config TOML. And `session_affinity` is removed from the config options which can now be achieved via `classify_trigger = new_session`.

This PR addresses the core of https://github.com/NVIDIA-NeMo/Switchyard/issues/347.

## Summary by CodeRabbit

- **New Features**
  - Added optional `reclassify = "on_user_turn"` routing for capability, custom LLM, and stage classifiers.
  - Classifier decisions persist across tool-call continuations and refresh on new user turns.
  - Abstentions clear the retained routing decision.

- **Bug Fixes**
  - Added validation preventing reclassification with session affinity or escalation mode.

- **Documentation**
  - Documented configuration, behavior, limitations, and follow-up turn handling.

PR #492, opened 2026-08-20
feat: classifier decision gate for sub-agent routing via passthrough

ayushag-nv:
## What

- Enables LLM Classifier as decision gate for sub-agent boundary for the PassThrough Algorithm
- Reimplemented PassThrough using FallThrough 
- You can now configure custom prompts and targets for subagents and use llm classifier for routing. 
- Fixes a issue in sub-agent usage metric recording during sub agent stream sending response.completed event (caught by codex while doing demo) 

There can be different scenarios this can be useful 
1. No Routing at Parent Level but routing at sub-agent level only ==> PassThrough + SubAgent Aware Composition
2. Routing at Parent Level (Classifier, Stage , escalation..etc) ==> {Classifier, Stage..etc} + SubAgent Aware Composition

Scoped this PR to passthrough only, but the design is good enough to extend to other routers as well. 

Code by: GPT 5.6 Sol High , Assisted and Designed by: Me 

## Why

We should be able to delegate different kind of sub-agent work to differnet models based on their domain , difficulty etc.

Linear: [SWITCH-1270](https://linear.app/nvidia/issue/SWITCH-1270/add-a-multi-target-decision-gate-for-sub-agent-routing)

## How tested

Tested both Claude Code and Codex against a standalone Switchyard server using the local NVIDIA configurations. Export `NVIDIA_API_KEY` before starting either server.

Claude Code:

```bash
RUST_LOG='switchyard_server::request=info,libsy=debug' \
  cargo run -p switchyard-server -- \
  --config subagent-classifier-claude.toml --port 4000

env -u ANTHROPIC_API_KEY \
  ANTHROPIC_BASE_URL=http://127.0.0.1:4000 \
  ANTHROPIC_AUTH_TOKEN=unused \
  claude --model claude-sonnet-4-6
```

Codex:

```bash
RUST_LOG='switchyard_server::request=info,libsy=debug' \
  cargo run -p switchyard-server -- \
  --config subagent-classifier-gpt-5.6.toml --port 4000

OPENAI_BASE_URL=http://127.0.0.1:4000/v1 \
  OPENAI_API_KEY=unused \
  codex --model gpt-5.6-terra
```

In both runs, parent-agent requests stayed on the configured parent target. Delegated requests were classified between two sub-agent targets, and later requests from the same sub-agent reused the affinity decision.

## Notes for reviewers

Suggested review order:

1. `crates/protocol/src/metadata.rs` - detects delegated requests from Claude Code and Codex metadata.
2. `crates/libsy/src/algorithms/util/subagent.rs` - extracts the delegated prompt and gates classification to sub-agent traffic.
3. `crates/libsy/src/algorithms/util/affinity.rs` - scopes affinity to the child agent identity.
4. `crates/libsy/src/algorithms/passthrough.rs` - composes the gate, classifier, affinity, and parent fallback.
5. `crates/switchyard-server/src/config.rs` - parses the TOML and builds the configured passthrough route.

Current scope is passthrough only. Clients must request the configured route ID; an explicit unregistered model name is rejected before the sub-agent classifier runs.

## Summary by CodeRabbit

* **New Features**
  * Added configurable passthrough routing to parent targets and optional delegated sub-agent targets.
  * Added classifier-based sub-agent selection with session affinity and configurable fallback behavior.
  * Added support for custom sub-agent classifier settings, including policies, schemas, and output limits.

* **Bug Fixes**
  * Prevented delegated-work affinity from affecting root or maintenance requests.
  * Improved handling of missing or unusable delegated prompts by abstaining safely.

### DIFF
--- a/crates/libsy/src/algorithms/passthrough.rs
+++ b/crates/libsy/src/algorithms/passthrough.rs
@@ -1,26 +1,103 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Single-target routing for direct model calls and integration diagnostics.
+//! Direct parent routing with an optional delegated-work classifier cascade.
 
 use std::sync::Arc;
 
 use switchyard_protocol::{ModelId, Request};
 
-use crate::core::algorithm::{Algorithm, Driver};
-use crate::{Result, RoutingOutcome};
+use super::fall_through::{DefaultTarget, FallThrough};
+use super::util::affinity::AffinityRouter;
+use super::util::subagent::{SubagentGate, SubagentOverride};
+use super::util::turn_pin::ClassifyTrigger;
+use crate::core::algorithm::{self, Algorithm, Driver};
+use crate::core::classifier::Classifier;
+use crate::core::state::State;
+use crate::{LibsyError, Result, RoutingOutcome};
 
-/// Routing algorithm that always selects one configured target.
+/// Routes parent traffic directly and optionally classifies delegated sub-agent work.
 pub struct Passthrough {
-    target: ModelId,
+    parent_target: ModelId,
+    route: FallThrough<State>,
+}
+
+/// Runtime components for classifying and retaining delegated sub-agent work.
+pub struct PassthroughSubagentConfig {
+    /// Targets the delegated-work classifier may select.
+    pub targets: Vec<ModelId>,
+    /// Classifier invoked for the first request from each identified child.
+    pub classifier: Arc<dyn Classifier<State>>,
+    /// Child target used when `classifier` abstains.
+    pub default_target: ModelId,
+    /// Controls whether each child is classified once or on every request.
+    pub classify_trigger: ClassifyTrigger,
+    /// Unsupported for child routing because child identity must come from harness metadata.
+    pub message_hash_fallback: bool,
+}
+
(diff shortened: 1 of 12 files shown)


ayushag-nv:
@coderabbit full review

ryan-lempka:
If we go with `classify_trigger` approach per https://github.com/NVIDIA-NeMo/Switchyard/pull/487 would it make sense for `subagent` to be another trigger option?

ayushag-nv:
> If we go with `classify_trigger` approach per #487 would it make sense for `subagent` to be another trigger option?

@ryan-lempka I think that will make sense when we want to trigger Nd re trigger within a session either main agent or sub agent. Sub-Agent can be more than a trigger, with the approach in this PR , it kind of isolates sub agent routing. Now you can add more triggers inside it with your approach. Wdyt ?

crates/libsy/src/algorithms/passthrough.rs
```diff
@@ -1,26 +1,103 @@
...
+    pub subagent: Option<PassthroughSubagentConfig>,
 }
 
 impl Passthrough {
-    /// Creates an algorithm that always selects `target`.
-    pub fn new(target: impl Into<ModelId>) -> Self {
-        Passthrough {
-            target: target.into(),
-        }
+    /// Creates direct parent routing, optionally with a decision gate for sub-agents.
```
messiaen: What is the advantage of putting this is `PassThrough`? Could we have a `SubAgent` algorithm?

grahamking (approved):

PR #498, opened 2026-08-20
feat(libsy-llm-client): report rejected extra_body keys as a configuration error

### LINKED ISSUE #260: feat(libsy-llm-client): report rejected extra_body keys as a configuration error

gburachas:
Closes #260.

## What

When an upstream rejects a request parameter that this client injected from a target's `extra_body`, the call now fails with a configuration error naming the offending keys, instead of passing the upstream 400 through.

```
upstream rejected a request parameter for model qwen/qwen3-35b-a3b. This target sets
extra_body keys [enable_thinking], which the model does not accept. Remove or correct
them in the target's extra_body.
```

## Why

`extra_body` is a per-target default from the deployment TOML, not something the caller asked for:

```toml
[targets.weak]
id = "some-model"
llm_client = "weak"
extra_body = { enable_thinking = false }
```

If the model does not accept one of those keys, every request to that target fails, and the caller sees an error about a parameter they never set. The error currently gives no indication that the cause is a config file.

## Change from the first version of this PR

The original implementation retried once with the injected keys removed. @ayushag-nv pointed out two problems with that, both correct:

1. It dropped the whole `extra_body` when only one key was at fault.
2. The TOML is static, so a bad key is bad on every request. The retry would silently double the call volume for the lifetime of the deployment rather than fixing anything.

The retry is gone. What remains is the detection plus the error message, which is what was asked for. The diff is 84 lines of production code; the rest is tests.

## Detection

A parameter reject is a 400 or 422 whose body names a parameter (`PARAM_REJECT_PHRASES`, `PARAM_REJECT_STATUSES`). Deliberately narrow, because a false positive would report a genuine upstream failure as a configuration error and send the operator to edit a file that is correct:

- 429 and 5xx never qualify; they are transient.
- Context-window rejections have their own error variant and never match.
- If the target injected nothing, the upstream error passes through unchanged, since the rejected parameter is then the caller's.

The upstream response body is not copied into the message. It can quote the request back, and the key names are enough to act on.

## How tested

Three tests in `crates/libsy-llm-client/src/client.rs`, using the existing wiremock setup:

- `an_extra_body_reject_names_the_keys_and_does_not_retry` — asserts the error names the key and, via `.expect(1)` on the mock, that the request is sent exactly once
- `a_param_reject_without_injected_keys_surfaces_the_upstream_error` — the passthrough case
- `param_reject_classification_is_narrow` — the 429/5xx and context-overflow boundaries

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `uv run ruff check .` all pass. Commit signed off per the DCO.

One file changed.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -51,6 +51,29 @@ const RESERVED_HEADERS: &[&str] = &[
     "accept-encoding",
 ];
 
+// Body phrases marking a 4xx as a rejection of a request *parameter* rather than of
+// the request's content. A server that refuses an unknown top-level field names it
+// one of these ways; the reasoning-control knobs are listed explicitly because they
+// are the fields operators most often configure per target. Matched against the raw
+// body, which covers both JSON error envelopes and the plain-text bodies some
+// proxies return.
+const PARAM_REJECT_PHRASES: &[&str] = &[
+    "chat_template_kwargs",
+    "enable_thinking",
+    "unexpected keyword",
+    "unknown parameter",
+    "unknown field",
+    "unrecognized",
+    "extra_forbidden",
+    "additional properties",
+    "extra fields not permitted",
+    "invalid parameter",
+];
+
+// Statuses that can mean "request malformed / unknown parameter". A 429 or 5xx is a
+// transient fault and must never be treated as a parameter reject.
+const PARAM_REJECT_STATUSES: &[u16] = &[400, 422];
+
 const INITIAL_RETRY_DELAY: Duration = Duration::from_millis(250);
 const MAX_RETRY_BACKOFF: Duration = Duration::from_secs(2);
 const MAX_RETRY_AFTER: Duration = Duration::from_secs(60);
@@ -221,7 +244,7 @@ impl TranslatingLlmClient {
             strip_anthropic_incompatible_fields(&mut body);
             strip_unsigned_thinking_blocks(&mut body);
         }
-        merge_extra_body(&mut body, backend.extra_body());
+        let injected_extra_body = merge_extra_body(&mut body, backend.extra_body());
         if matches!(backend, Backend::Anthropic(_)) {
             enable_anthropic_prompt_caching(&mut body);
         }
@@ -267,6 +290,19 @@ impl TranslatingLlmClient {
                     return Ok(response);
                 }
                 Err(failure) => {
+                    // The upstream refused a parameter and this target injects some. The
+                    // caller's request was serviceable; the deployment is misconfigured, and
+                    // it will be misconfigured on every later request too, so this is
+                    // reported as a configuration error naming the keys rather than passed
+                    // through as an upstream failure.
+                    if !injected_extra_body.is_empty() && failure.is_param_reject() {
+                        span.record("outcome", "error");
+                        if let Some(status) = failure.status {
+                            span.record("status_code", status.as_u16());
+                        }
+                        span.record("will_retry", false);
+                        return Err(extra_body_rejected(model, &injected_extra_body));
+                    }
                     let will_retry = attempt < max_retries && failure.is_retryable();
                     span.record("outcome", "error");
                     if let Some(status) = failure.status {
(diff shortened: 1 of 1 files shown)


crates/libsy-llm-client/src/client.rs
```diff
@@ -1648,6 +1729,174 @@ mod tests {
         Ok(())
     }
 
+    #[test]
```
ayushag-nv, line 1816: @gburachas So many AI written tests here, try to merge the similar ones. Keep them minimal. As littel as possible and as much as needed,

ayushag-nv (changes requested):
@gburachas I like your idea of prevention by retrying on fail and reject the extra body. There are two concerns here
1. FOr eg if one key is rejected in extra body, we are removing the entire extra body which is not right
2. This is trouble some for the entire serving, becaue the toml file is fixed and it will always do 2 calls, one fail and one retry. Better fail with a reasonable Error message, so that user can fix their toml. 

So, if you want to add something, I will suggest to add a minimal viabl error message. That should be all. No need for this PR.

crates/libsy-llm-client/src/client.rs
gburachas: Sounds good! Changing to erroring mode.

gburachas:
@ayushag-nv agreed on both points; the PR is reworked.

The retry is removed. On a parameter reject where this client injected `extra_body` keys, the call now fails with a configuration error naming those keys, so the operator can find them in the TOML:

```
upstream rejected a request parameter for model qwen/qwen3-35b-a3b. This target sets
extra_body keys [enable_thinking], which the model does not accept. Remove or correct
them in the target's extra_body.
```

If the target injected no `extra_body`, the upstream error passes through unchanged, since the rejected parameter is then the caller's own.

The detection is unchanged and stays narrow: only a 400 or 422 whose body names a parameter, never a 429 or 5xx.

gburachas:
Tests cut from four to two, and the retry is gone. Both points taken.

The PR now has a single commit: I amended and force-pushed rather than adding a follow-up commit, so the earlier retry-based version is no longer in the history. `git log` on the branch shows one commit, not two.

What is left:

- `an_extra_body_reject_is_reported_as_a_configuration_error` — merges the two previous integration tests. First half: an injected key produces a configuration error naming it, with `.expect(1)` on the mock pinning that the request is sent once. Second half: with no `extra_body`, the upstream 400 passes through unchanged.
- `param_reject_classification_is_narrow` — now a table rather than separate assertions, covering the 400/422 cases that qualify and the unrelated-400, 429 and 503 cases that must not.

I dropped the standalone `merge_extra_body_reports_only_the_keys_it_injected` test: the pre-existing `extra_body_adds_defaults_without_overriding_the_request` already covers the merge, and the integration test above fails if the injected key is not reported, since the message would not name it.

85 lines of production code, 129 of tests, one file.

ayushag-nv (changes requested):
@gburachas I think we don't need a 200+ lines worth of PR for just throwing a simple error message. Can you please simplify it and do it under 50 lines. Tests are unncessary mostly here, codex/claude tends to generate a lot of tests

bhuvan002:

Issue #493, opened 2026-08-20
[feature] Support sub-agent-aware model routing


ayushag-nv:
## Problem

Coding-agent harnesses can delegate work to child agents. Each child receives a specific task from its parent, but the model best suited for that work may differ from the model serving the parent or other children.

Switchyard needs to route delegated sub-agent work independently while preserving the parent's routing behavior. This must be a server-side routing capability so it works across supported and custom clients.

The core composition is:

```text
Parent routing policy + Sub-agent routing policy
```

The parent and sub-agent policies must be configurable independently.

## Requirements

Sub-agent-aware routing should:

- Distinguish parent-agent requests from delegated sub-agent requests.
- Identify each child using a stable `session_id + agent_id`.
- Support a fixed target for all delegated work without making a classifier call.
- Support classification across any number of sub-agent targets.
- Give the classifier only its configured system prompt and the delegated task prompt produced by the parent.
- Retain a child's selected target when configured to classify once per child.
- Keep sibling sub-agent assignments independent.
- Leave parent-agent routing unchanged.
- Exclude harness-maintenance requests from sub-agent routing.
- Use a configured default child target when classification fails or abstains.
- Log the child identity, whether classification ran, and the selected target.

## Proposed configuration

A route may define an optional sub-agent routing policy independently of its parent routing policy.

### Fixed sub-agent target

All delegated work goes to one target:

```toml
[routes.agent]
id = "agent"
type = "passthrough"
target = "parent"

[routes.agent.subagent]
type = "passthrough"
target = "worker"
```

The same sub-agent policy should compose with other parent algorithms:

```toml
[routes.agent]
id = "agent"
type = "stage_router"
# Parent Stage Router configuration

[routes.agent.subagent]
type = "passthrough"
target = "worker"
```

### Classified sub-agent target

Delegated work is classified across multiple targets:

```toml
[routes.agent]
id = "agent"
type = "passthrough"
target = "parent"

[routes.agent.subagent]
type = "llm_classifier"
classifier_target = "classifier"
targets = ["worker", "reviewer"]
default_target = "worker"
classify_trigger = "new_session"
prompt = "Select the appropriate target for the delegated task."
response_schema = '''
{
  "type": "object",
  "properties": {
    "target": {
      "type": "string",
      "enum": ["worker", "reviewer"]
    }
  },
  "required": ["target"],
  "additionalProperties": false
}
'''
policy = { type = "target_selector", selector = "/target" }
```

[...]

PR #505, opened 2026-08-20
feat(libsy): generalize subagent awareness to all algos + subagent UX improvement

ayushag-nv:
## What

#492 added llm classifier as subagent gate in passthrough (single target interaction)

This makes the subagent work more generalizable while keeping base algorithms intact. Earlier I was disturbing passthrough from being passthrough. Now, passthrough can remain as it is. But if a user wants to configure sub_agents additionally , then they do something like this now 
```
let algorithm = Passthrough::new(parent_target);
let parent: Arc<dyn Algorithm> = Arc::new(algorithm);
attach_subagent_router(route_name, parent, subagents.as_ref(), targets) // this will construct subagent config if there and construct SubagentRouter with the parent, otherwise just the parent
```

This way parent algo remains same. We just add stuff in config file only.

Right now, added subagent awareness to PassThrough and StageRouter. Add more later if needed.

- earlier the UX of config was specific to llm classifier for subagent. Now it becomes more intuitive and follow similar pattern as main agent. 
   eg : 
   ```
          route.main_agent 
          type: passthrough
          
          route.main_agent.sub_agents
          type:passthrough | llm_classifier 
    ```

## Why

every algorithm should be able to get sub-agent routing configs

Relates #493

Linear: [SWITCH-1277](https://linear.app/nvidia/issue/SWITCH-1277/support-independent-sub-agent-routing-on-passthrough-and-stage-routes) 

## Notes for reviewers

start with subagent.rs, then config.rs 

## Summary by CodeRabbit

* **New Features**
  * Added fixed-target routing for delegated sub-agent requests.
  * Added support for configurable delegated-work routing using either a fixed target or an LLM classifier.
  * Improved routing behavior for parent requests and delegated sub-agent work.

* **Documentation**
  * Updated configuration and routing guides with the new `subagents` settings.
  * Documented fixed-target delegated routing and removed references to the legacy configuration option.

* **Bug Fixes**
  * Improved validation for incompatible passthrough routing configurations.

### DIFF
--- a/crates/libsy/src/algorithms/subagent.rs
+++ b/crates/libsy/src/algorithms/subagent.rs
@@ -64,14 +64,16 @@ impl SubagentRouter {
             });
         }
 
-        let mut subagent = FallThrough::new_with_state(config.targets).with_name("subagent");
-        match config.classify_trigger {
-            ClassifyTrigger::EveryRequest => {}
+        let mut subagent = match config.classify_trigger {
+            ClassifyTrigger::EveryRequest => {
+                FallThrough::new_with_state(config.targets).with_name("subagent")
+            }
             ClassifyTrigger::NewSession => {
                 let affinity = Arc::new(AffinityRouter::for_subagents());
-                subagent = subagent
+                FallThrough::new_with_state(config.targets)
+                    .with_name("subagent")
                     .with_processor(affinity.clone())
-                    .with_classifier(affinity);
+                    .with_classifier(affinity)
             }
             ClassifyTrigger::UserTurn => {
                 return Err(LibsyError::AlgorithmError {
@@ -79,7 +81,7 @@ impl SubagentRouter {
                         .to_string(),
                 });
             }
-        }
+        };
         subagent = subagent
             .with_classifier(Arc::new(SubagentGate::new(config.classifier)))
             .with_classifier(Arc::new(DefaultTarget::new(config.default_target)));
--- a/crates/libsy/src/algorithms/subagent.rs
+++ b/crates/libsy/src/algorithms/subagent.rs
@@ -1,0 +1,305 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Delegated sub-agent routing around an arbitrary parent algorithm.
+
+use std::sync::Arc;
+
+use switchyard_protocol::{Metadata, ModelId, Request};
+
+use super::fall_through::{DefaultTarget, FallThrough};
+use super::util::affinity::{AffinityRouter, ClassifyTrigger};
+use super::util::subagent::SubagentGate;
+use crate::core::algorithm::{self, Algorithm, Driver};
+use crate::core::classifier::Classifier;
+use crate::core::state::State;
+use crate::{LibsyError, Result, RoutingOutcome};
+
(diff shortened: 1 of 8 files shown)


crates/libsy/src/algorithms/subagent.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        let mut subagent = FallThrough::new_with_state(config.targets).with_name("subagent");
+        match config.classify_trigger {
+            ClassifyTrigger::EveryRequest => {}
+            ClassifyTrigger::NewSession => {
+                let affinity = Arc::new(AffinityRouter::for_subagents());
+                subagent = subagent
+                    .with_processor(affinity.clone())
+                    .with_classifier(affinity);
+            }
```
grahamking, line 77: Could you do 
```
let mut subagent = match config.classify_trigger {
```
and move line 67 into `EveryRequest` variant?

crates/libsy/src/algorithms/subagent.rs
```diff
@@ -0,0 +1,302 @@
...
+        if request
+            .metadata
+            .as_ref()
+            .is_some_and(Metadata::is_subagent_work)
+        {
+            self.subagent.execute(driver, request).await
+        } else {
+            self.parent.clone().route(driver, request).await
+        }
+    }
```
grahamking, line 109: Now that is a nice method!

crates/switchyard-server/src/config.rs
```diff
@@ -396,23 +396,84 @@ struct EscalationClassifierRouteConfig {
...
     policy: ClassifierPolicyConfig,
+    classify_trigger: ClassifyTrigger,
+    message_hash_fallback: bool,
+    recent_turn_window: Option<usize>,
+    max_output_tokens: u64,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct LlmClassifierRouteConfig {
```
grahamking, line 415: Putting `#[serde(default)]` here on the struct would allow you to remove it from all the fields.

grahamking: You'd add `Default` in `derive`. The customized default on `max_output_tokens` can stay as field level overrides struct level.

crates/switchyard-server/src/config.rs
```diff
@@ -996,6 +1040,104 @@ const fn default_max_retries() -> u32 {
...
 
+// Resolve nested policy once so every supported parent builds the same child route.
+fn build_subagent_router_config(
+    route_name: &str,
+    config: Option<&SubagentRouteConfig>,
+    targets: &BTreeMap<String, ModelId>,
+) -> ServerResult<Option<SubagentRouterConfig>> {
+    let Some(config) = config else {
+        return Ok(None);
+    };
```
grahamking: Can you remove the `Option` from both `config` and the return value? If there's no config don't call it.

grahamking (approved):

grahamking:
Claude says:

> The refactor is a faithful, well-tested decomposition of existing behavior with genuinely simpler fallback-filtering logic than before.

crates/libsy/src/algorithms/subagent.rs
ayushag-nv: Nice suggestion. Fixed it

crates/switchyard-server/src/config.rs
ayushag-nv: Yes make sense. Done

PR #506, opened 2026-08-20
feat(prefill-router): add Transformers feature extraction

nachiketb-nvidia:
## What

Add a private `prefill-router` crate that exposes a backend-neutral Rust contract for extracting pooled prefill hidden states.

The first backend embeds Hugging Face Transformers through PyO3 and supports:

- batched prompt extraction
- chat-template keyword arguments
- all, upper-half, or selected hidden-state layers
- last-token and mean pooling over attention-masked tokens
- explicit batch size and maximum sequence length
- lazy model loading and unloading

## Why

Prefill routing needs the encoder forward path before checkpoint transforms and routing policy can be implemented. Keeping that path behind `PrefillForward` lets the initial Transformers implementation be replaced by a native backend such as Candle without changing its Rust callers.

This MR intentionally does not add PCA, MLP scoring, checkpoint loading, caching, or routing.

## How

- Keep the Transformers implementation in a private embedded Python module.
- Disable Hugging Face remote-code execution; supported models must load through installed Transformers code.
- Decode Python output into private raw parts and only construct `ForwardOutput` when every requested layer and pooling mode is present with the expected shape.
- Use attention masks for pooling so both left- and right-padded tokenizers are correct.
- Use a crate-local uv environment for CPU Torch and Transformers test dependencies.
- Import the reference `PrefillExtractor` directly from llm-router commit [`8a9d3509`](https://github.com/NVIDIA-AI-Blueprints/llm-router/commit/8a9d3509fbde879d9795258081bab9553b458e04) as a dev-only dependency. The immutable GitHub archive avoids the upstream repository's unavailable Git LFS checkpoint during installation.
- Use PyO3's rpath helper and the activated `VIRTUAL_ENV` so `cargo test` can locate CPython and its site-packages without `LD_LIBRARY_PATH` or `PYTHONPATH`.

## Diff size

The raw `+2,324` lines break down as:

- **1,550 generated lockfile lines:** 1,540 in `uv.lock` and 10 in `Cargo.lock`.
- **774 source and configuration lines:** the Rust contract and PyO3 adapter, embedded Transformers extraction, manifests, venv setup, and parity harness.

The copied 183-line Python oracle has been removed. The parity test now runs the upstream package directly from its pinned commit.

## Exactness

The extraction path is derived from NVIDIA AI Blueprints llm-router commit [`8a9d3509`](https://github.com/NVIDIA-AI-Blueprints/llm-router/blob/8a9d3509fbde879d9795258081bab9553b458e04/src/model_router_toolkit/prefill/extract.py).

The parity test downloads and caches `Qwen/Qwen3.5-0.8B`, then runs the Rust-backed implementation and the pinned upstream `PrefillExtractor` sequentially with the same:

- two prompts
- chat-template arguments
- complete layer set
- last and mean pooling
- batching and truncation limits

It asserts equality of the complete tensors for every prompt, layer, pooling mode, and hidden dimension. This is not a shape or tolerance-only comparison.

## Validation

```bash
cargo test -p prefill-router
cargo clippy -p prefill-router --all-targets -- -D warnings

cd crates/prefill-router
uv sync --locked
source .venv/bin/activate
cargo test matches_the_reference_transformers_tensors_exactly -- --ignored
```

All passed locally. The ignored exactness test downloads `Qwen/Qwen3.5-0.8B` on its first run.

## What to review

- Fidelity of `python/transformers_forward.py` to the linked llm-router extraction path.
- Whether `PrefillForward`, `ForwardRequest`, and the parsed `ForwardOutput` form the right replacement boundary for a future native implementation.
- The direct, pinned upstream oracle dependency used by the exact tensor test.
- The intentionally narrow boundary between extraction and later scoring/routing work.

Closes [SWITCH-1278](https://linear.app/nvidia/issue/SWITCH-1278/create-prefill-router-crate-and-implement-transformers-forward-path).


crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,181 @@
...
+        self._torch = None
+        self.n_layers = 0
+        self.hidden_dim = 0
+
+    def _ensure_loaded(self) -> None:
+        if self._model is not None:
+            return
+
+        import torch
+        from transformers import AutoModelForCausalLM, AutoTokenizer
```
ayushag-nv, line 63: imports at the top

nachiketb-nvidia: these are intentionally lazy so as to avoid bringing in torch and transformers until forward pass

crates/prefill-router/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[project]
+name = "prefill-router-tests"
+version = "0.0.0"
+requires-python = ">=3.10"
+dependencies = [
+    "torch>=2.0",
+    "transformers>=4.45",
```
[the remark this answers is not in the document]
nachiketb-nvidia: the crate-local pyproject.toml is explicitly a CPU test environment. GPU runtime packaging is separate scope.

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,181 @@
...
+            trust_remote_code=True,
+        )
+        if self._tokenizer.pad_token is None:
+            self._tokenizer.pad_token = self._tokenizer.eos_token
+
+        load_kwargs: dict[str, Any] = {
+            "dtype": dtype,
+            "cache_dir": cache_dir,
+            "trust_remote_code": True,
+        }
```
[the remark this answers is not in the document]
nachiketb-nvidia: outdated for the locked Transformers 5.15.1 environment. The live Qwen parity test already passed with dtype.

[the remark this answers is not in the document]
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,181 @@
...
+            hidden_states = outputs.hidden_states
+            sequence_lengths = attention_mask.sum(dim=1)
+            for batch_index in range(input_ids.shape[0]):
+                sequence_length = int(sequence_lengths[batch_index].item())
+                for layer in layers:
+                    hidden = hidden_states[layer][batch_index, :sequence_length, :].float()
+                    if "last" in pools:
+                        all_last[layer].append(hidden[-1].cpu())
+                    if "mean" in pools:
+                        all_mean[layer].append(hidden.mean(dim=0).cpu())
```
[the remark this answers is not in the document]
nachiketb-nvidia: check now

crates/prefill-router/src/error.rs
```diff
@@ -0,0 +1,28 @@
...
+    },
+
+    /// The embedded implementation returned an invalid result.
+    #[error("invalid prefill result: {0}")]
+    InvalidResult(String),
+}
+
+pub(crate) fn python_error(operation: &'static str, source: PyErr) -> PrefillRouterError {
+    PrefillRouterError::Python { operation, source }
+}
```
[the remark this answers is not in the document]
nachiketb-nvidia: check now

crates/prefill-router/python/transformers_forward.py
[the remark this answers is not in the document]
nachiketb-nvidia: should be fixed now

crates/prefill-router/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+                }
+                if rows.iter().any(|row| row.len() != self.hidden_dim) {
+                    return Err(PrefillRouterError::InvalidResult(format!(
+                        "{pooling} layer {layer} contains a vector with the wrong hidden width"
+                    )));
+                }
+            }
+        }
+
+        Ok(self)
```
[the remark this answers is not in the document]
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Hugging Face Transformers prefill forward used by the Rust crate."""
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
```
bhuvan002: Why is parallelism set to false?

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def _detect_device(torch: Any) -> str:
+    override = os.environ.get("ROUTER_DEVICE", "").lower()
```
bhuvan002: If this is a switchyard specific env var can we prefix it with `SWITCHYARD_*` or something else if there is already some convention? This should also be documented

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
+
+
+def _detect_device(torch: Any) -> str:
```
bhuvan002, line 11: Would be nice to add some kind of logging here showing the device selection

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+        if self._model is not None:
+            return
+
+        import torch
+        from transformers import AutoModelForCausalLM, AutoTokenizer
+
+        self._torch = torch
+        device = self._device_override or _detect_device(torch)
+        dtype = torch.float32 if device == "cpu" else torch.bfloat16
+        cache_dir = self._cache_dir or os.environ.get("HF_HUB_CACHE")
```
bhuvan002: nit: Referencing `from huggingface_hub.constants import HF_HUB_CACHE` instead of hardcoding env variable here would be cleaner

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+        pooling_modes: list[str] | None = None,
+        batch_size: int = 4,
+        max_length: int = 2048,
+    ) -> dict[str, Any]:
+        """Extract pooled hidden states using the blueprint's direct indexing."""
+        self._ensure_loaded()
+
+        if extract_layers == "all":
+            layers = list(range(self.n_layers))
+        elif extract_layers is None:
```
bhuvan002: Is this a default setting to extract second half layers, or do the semantics actually mean that `None` means second half? Look a little confusing

crates/prefill-router/python/transformers_forward.py
```diff
@@ -0,0 +1,178 @@
...
+            with self._torch.no_grad():
+                outputs = self._model(
+                    input_ids=input_ids,
+                    attention_mask=attention_mask,
+                    output_hidden_states=True,
+                    use_cache=False,
+                )
+
+            hidden_states = outputs.hidden_states
+            for batch_index in range(input_ids.shape[0]):
```
bhuvan002: We probably don't need to loop over both batch index and layer here. Can you see if this can be vectorized by stacking the selected hidden states and broadcasting the attention mask over the layer/hidden dimensions?
Something like
```
hidden = torch.stack(
    [hidden_states[layer] for layer in layers],
    dim=0,
).float()  # [L, B, S, D]

mask = attention_mask.bool()  # [B, S]

mean = (
    (hidden * mask[None, :, :, None]).sum(dim=2)
    / mask.sum(dim=1)[None, :, None]
)  # [L, B, D]

last_idx = mask.sum(dim=1) - 1
batch_idx = torch.arange(mask.shape[0], device=mask.device)

last = hidden[:, batch_idx, last_idx, :]  # [L, B, D]
```

bhuvan002 (changes requested):

sabhatinas:
Is there a set of ground truth prompts you compared the reference python implementation with?

crates/prefill-router/python/transformers_forward.py
sabhatinas: Unless the in-memory hidden states collection with transformers only supports tp/pp=1, we should be fine to set any default. disabling sounds reasonable to me by default

crates/prefill-router/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Extra keyword arguments passed to the tokenizer's chat template.
+    pub chat_template_kwargs: serde_json::Map<String, serde_json::Value>,
+    /// Hidden-state layers to extract.
+    pub layers: LayerSelection,
+    /// Pooling modes to calculate for every requested layer.
+    pub pooling: Vec<Pooling>,
+    /// Maximum number of prompts in each model forward pass.
+    pub batch_size: usize,
+    /// Maximum tokenized length of each prompt.
+    pub max_length: usize,
```
sabhatinas, line 44: why do we expose this?

crates/prefill-router/python/transformers_forward.py
bhuvan002: Tokenizer parallelism here is unrelated to tp/pp, it controls cpu parallelism during tokenization. So unless there’s a specific reason to disable it I don’t think we should explicitly turn it off and potentially limit tokenization performance

crates/prefill-router/src/lib.rs
nachiketb-nvidia: its exposed by the python code

nachiketb-nvidia: we could just have defaults

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: removed

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: removed, a lot of this just straight remnants of python code from prefill router

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: logged in rust side so tracing subscriber is aware

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: removed, more remnants from the python impl

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: default setting is now "upper_half"

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: parallelized to the best this can do, and a bunch of other improvements

bhuvan002 (approved):

nachiketb-nvidia:
>Is there a set of ground truth prompts you compared the reference python implementation with?

@sabhatinas I compared with straight up inputs and the pre-fill router python code implementation directly via uv and dev extras

PR #512, opened 2026-08-21
feat(server): decouple upstream model name from target id via upstream_model

al157:
# feat(server): decouple upstream model name from target id via `upstream_model`

## Problem

A target's `id` currently serves two roles at once:

1. **Routing label** — how routes, fallback logic, stats, and logs refer to the target.
2. **Upstream model name** — the exact model id written into requests sent to the provider.

This makes it impossible to configure **the same model served by multiple providers**
(e.g. `deepseek-v4-flash` from DeepSeek official and from an aggregator). Because
`Random::new` deduplicates targets by model id (`rand.rs:52-55`) and the client keys
its models by id, identical ids collide:

```
target X reuses model id deepseek-v4-flash on llm client Y; only one target per id
is kept and the other is dropped.
random route pool: random targets must be unique
```

Users who want multi-supplier redundancy for one model — a core reliability use case —
currently cannot express it in the config surface.

## Solution

Introduce an explicit, optional `upstream_model` field that separates the two concerns:

- **Lookup key** stays the target `id` (e.g. `deepseek-v4-flash@sensenova`) — routing,
  fallback exclusion, error reporting, stats, and response restamping are untouched.
- **Wire name** becomes `upstream_model` when set (e.g. `deepseek-v4-flash`), falling
  back to `id` so every existing single-supplier setup is unaffected.

### Changes (4 files, +82/−11)

| Crate | File | Change |
|---|---|---|
| switchyard-server | `src/config.rs` | `TargetConfig` gains `#[serde(default)] upstream_model: Option<ModelId>`; registration passes key = `id`, wire = `upstream_model.unwrap_or(id)` |
| libsy-llm-client | `src/client.rs` | `ModelConfig` gains a `wire_model` field (4th ctor arg). Lookup remains keyed by `model_name`; `call_rewrite_model` and `count_tokens` now write `wire_model` into the upstream request |
| libsy-llm-client | `src/client.rs` tests | new `wire_model_remaps_the_upstream_name` test (wiremock asserts the upstream receives the bare model name); existing ctor call sites updated |
| libsy-llm-client / switchyard-server | `run.rs`, `tests/server.rs` | mechanical ctor-arg updates |

`libsy` is unchanged: algorithms only ever see the routing `ModelId`.

## Example

```toml
[targets."deepseek-v4-flash@deepseek"]
id = "deepseek-v4-flash@deepseek"          # unique routing label
upstream_model = "deepseek-v4-flash"       # actual model id sent upstream
llm_client = "deepseek"

[targets."deepseek-v4-flash@sensenova"]
id = "deepseek-v4-flash@sensenova"
upstream_model = "deepseek-v4-flash"
llm_client = "sensenova"

[routes.pool]
type = "random"
targets = ["deepseek-v4-flash@deepseek", "deepseek-v4-flash@sensenova"]
weights = [1, 1]
```

Requests to `pool` now spread across providers, each receiving the correct bare model
id on the wire.

## Verification

- `cargo check -p switchyard-server -p switchyard-llm-client`: 0 errors
- `cargo test -p switchyard-server -p switchyard-llm-client`: 153 passed / 0 failed
  (includes the new remap test)
- `cargo fmt --check`: clean; `cargo clippy`: no warnings
- End-to-end (3-provider pool, live keys): 6 consecutive calls → HTTP 200 ×6,
  distributed across ≥3 distinct supplier instances, each upstream receiving the bare
  model name; zeroing one instance's weight redirects all traffic with no errors

## Compatibility

Configs without `upstream_model` behave identically (wire name falls back to `id`);
all pre-existing tests pass unchanged apart from the added constructor argument.

## Summary by CodeRabbit

* **New Features**
  * Added support for separate routing and upstream model names.
  * Targets can optionally specify the model identifier sent to upstream services.
  * When unspecified, the routing identifier is used automatically.
  * Model remapping now applies consistently to token counting and completion requests.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -59,21 +59,29 @@ const MAX_RETRY_AFTER: Duration = Duration::from_secs(60);
 /// pin a wire format, plus any `other_backends` reachable over additional formats.
 #[derive(Clone, Debug)]
 pub struct ModelConfig {
+    /// Lookup key — the name callers address this model by (e.g. a target id that
+    /// must be unique across a deployment).
     model_name: ModelId,
+    /// The model id written into upstream requests. May differ from `model_name`
+    /// when a target remaps the model en route to its backend.
+    wire_model: ModelId,
     default_backend: Backend,
     other_backends: Option<Vec<Backend>>,
 }
 
 impl ModelConfig {
     /// A model named `model_name` served by `default_backend`, optionally reachable
-    /// over additional wire formats via `other_backends`.
+    /// over additional wire formats via `other_backends`. Upstream requests carry
+    /// `wire_model`; pass the same name when no remap is wanted.
     pub fn new(
         model_name: impl Into<ModelId>,
         default_backend: Backend,
         other_backends: Option<Vec<Backend>>,
+        wire_model: impl Into<ModelId>,
     ) -> Self {
         Self {
             model_name: model_name.into(),
+            wire_model: wire_model.into(),
             default_backend,
             other_backends,
         }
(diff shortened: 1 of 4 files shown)


grahamking:
Thanks @al157 ! Nice and concise PR.

We do need to do something here. We have too many IDs. Some are unique and some are not, and we sometimes use non-unique ones as if they were unique.

This PR adds two more IDs so I think it's not what we want to do. I'm going to close it, but thanks for prompting us to rethink this.

PR #513, opened 2026-08-21
feat(libsy): expose routing metadata from state.extra to Python

yanaSelin:
## Summary

Exposes internal routing decision signals from Rust algorithm state to Python callers via `PyDecision`. Previously, the only way to read routing signals from Python was to parse the reasoning string. Now every `Decision` step carries a typed metadata snapshot that Python can query by key.

## Python contract

`Decision` (exposed as `switchyard.libsy.Decision`) gains two new methods:

```python
decision.get(key: str) -> float | None
decision.get_str(key: str) -> str | None
```

`get` returns a scalar (`f32`) value or `None` if the key is absent or holds a non-scalar value.  
`get_str` returns a string value or `None` if the key is absent or holds a non-string value.

### Keys written by `StageClassifier` (tool-signal path)

Present when the algorithm ran `StageClassifier` (i.e. there were tool signals to score). Absent when the session has no tool calls yet.

| Key | Type | Description |
|-----|------|-------------|
| `signal_severity` | `float` | Severity dimension of the tool signal |
| `signal_spinning` | `float` | Spinning/stuck dimension |
| `signal_exploring` | `float` | Exploration dimension |
| `signal_production_intensity` | `float` | Production-intensity dimension |
| `signal_score` | `float` | Composite score (`score_signal().score`) |
| `signal_confidence` | `float` | `abs(score)` — confidence in the direction |

All six are written **before** the `Resolved` / `ConsultClassifier` branch, so they are present regardless of which path was taken (including the below-threshold fallback that sets confidence to 0).

### Keys written by `TaskClassifierPolicy` (LLM judge path)

Present only when the judge ran and returned a verdict. Absent if the judge timed out, errored, or was not configured.

| Key | Type | Description |
|-----|------|-------------|
| `judge_p_solve` | `float` | Model's estimated probability of solving the task |
| `judge_crux` | `str` | One-line description of the task's core difficulty |
| `judge_primary_rule` | `str` | Primary routing rule that applied |
| `judge_capability_boundary` | `str` | Capability boundary the task sits at |

### Example usage

```python
async for step in stream:
    if isinstance(step, Decision):
        severity   = step.get("signal_severity")    # float | None
        confidence = step.get("signal_confidence")  # float | None
        p_solve    = step.get("judge_p_solve")       # float | None
        crux       = step.get_str("judge_crux")      # str   | None
```

## Implementation

- New `WithExtra` trait on `state.rs` — lets `FallThrough<S>` snapshot `extra` without losing genericity; `()` returns an empty map for stateless compositions
- `Step::Decision` changed from a tuple variant to a struct variant carrying `extra: HashMap<String, StateValue>`
- `Driver::decide()` takes the snapshot as a second argument
- `StageClassifier::score()` writes signal fields before the `match outcome` so both `Resolved` and `ConsultClassifier` branches capture them
- `JudgePolicy` gains a default no-op `record_to_state`; `TaskClassifierPolicy` overrides it to write the judge fields
- `PyDecision` in `switchyard-py` carries the snapshot and exposes `get` / `get_str` via `#[pymethods]`

## Summary by CodeRabbit

* **New Features**
  * Added routing decision events to execution streams, including the selected model and whether the call is an answer request.
  * Exposed routing metadata through the Python API with typed accessors.
  * Added support for recording classifier and judge signals in execution state.
  * Routing paths now consistently report selected models and associated metadata.

* **Bug Fixes**
  * Improved handling of decision events during failed and completed executions.
  * Prevented decision events from interrupting normal model-call processing.


nachiketb-nvidia:
Thank you!

The extra metadata being emitted here does not need to go through Decision. Direct emission via OTEL for any subscriber to get back.
There is prior implementation of this, see https://github.com/NVIDIA-NeMo/Switchyard/blob/a8c4d2ed201e068b0c65495af2a6850bde6aa4a2/crates/libsy/src/algorithms/util/stage.rs#L275

Closing, pls re-open if you feel its necessary

PR #516, opened 2026-08-21
fix: give the upstream Codex tools instead of dropping them

### LINKED ISSUE #515: [bug] Codex tools never reach a translated upstream: `additional_tools` items and `custom` tools are dropped
### What happens

A Codex CLI session through a route that translates the wire format (`openai_responses` in, `openai_chat` upstream) runs to completion and **executes nothing**. The model prints its tool call as prose and the turn ends. Codex reports success, so the failure is silent — exit status is 0 and the rollout looks like a normal short turn.

Same session, same prompt, on a passthrough (`openai_responses` → `openai_responses`) route: the tool runs.

### Reproduction

Codex **0.146.1**, provider `wire_api = "responses"` pointed at Switchyard, a two-tier route whose efficient tier is an `openai_chat` client.

```
codex exec --dangerously-bypass-approvals-and-sandbox \
  'Run exactly this shell command and show me its output: echo CONTROL-PROBE'
```

The reply is the tool call, as text:

```json
{
  "tool": "exec",
  "arguments": {
    "cmd": "echo CONTROL-PROBE && echo '---' && date -u",
    "max_output_tokens": 5000,
    "timeout": 15000
  }
}
```

`~/.codex/sessions/**/rollout-*.jsonl` for that session contains **no** `function_call`, `custom_tool_call`, or `*_output` item at all.

### Root cause

Two request-decoding gaps compose. Both are in `crates/switchyard-translation/src/codecs/responses/buffered.rs`.

**1. `additional_tools` input items are not decoded.**

Codex 0.146 declares its whole toolset in an *input item*, not in the request's `tools` array. The captured inbound request has **no `tools` key at all** — only `tool_choice: "auto"` — and `input[0]` is:

```json
{"type": "additional_tools", "role": "developer", "tools": [ ... ~24 KB ... ]}
```

`decode_responses_input` has no arm for that type, so it reaches the catch-all and becomes a `ContentBlock::Unknown` inside a **user** message. The upstream is therefore offered zero tools, while Codex's own prompt still tells the model the tools exist. Hence the prose call.

**2. `custom` tools are dropped.**

The item's `tools` array holds ordinary Responses tool specs, and one of them is:

```json
{"type": "custom", "name": "exec", "description": "...", "format": {"type": "text"}}
```

`decode_responses_tools` has arms for `namespace`, `function`, and a bare `name`; everything else falls through to `push_responses_id_tool`, which requires an `id`, returns `false`, and **has its return value discarded**. So a `custom` tool disappears with no diagnostic. The string `custom` appears nowhere under `crates/switchyard-translation/src/codecs/`.

Fixing only (2) is not sufficient: with (1) unfixed the tool specs are never read in the first place.

### Why a passthrough route is unaffected

Responses → Responses keeps the item as sent, so the upstream sees the declaration. That is why this looks like a model or a prompt problem until the inbound and outbound bodies are compared.

### Evidence

Same config, same prompt, two servers differing only by the fix:

| server | rollout tool items | command ran |
|---|---|---|
| current `main` | none | no |
| patched | `function_call` + `function_call_output` | yes |

The efficient tier answered both turns (`DeepSeek-V4-Pro`), so this is not an escalation artefact.

### Suggested fix

Decode the `additional_tools` item into `LlmRequest::tools` and consume it from the input; add a `custom` arm that advertises the tool to a chat upstream as a function taking one string, and turn the call back into a `custom_tool_call` on the response path (buffered and streaming). Both mappings can ride in the request's `ProviderExtensions` under prefixed keys, the way `codex_namespaces` already does for #384, so no provider-neutral type grows a Codex-specific field and encoding back to Responses stays lossless.

PR follows.

fedecia:
Fixes #515.

A Codex CLI session through a translating route runs to completion and **executes nothing**. The client reports success, so the failure is silent. Two request-decoding gaps compose to produce it; neither shows on a Responses-to-Responses passthrough.

**1. `additional_tools` is not decoded.** Codex 0.146 declares its whole toolset in an input item — `{"type": "additional_tools", "role": "developer", "tools": [...]}` — and sends no top-level `tools` key at all. `decode_responses_input` has no arm for that type, so it hits the catch-all and becomes an opaque `Unknown` block inside a user message. The upstream is offered zero tools.

**2. `custom` tools are dropped.** The item's `exec` entry is `{"type": "custom", ...}`. `decode_responses_tools` had arms for `namespace`, `function`, and a bare name; everything else fell to `push_responses_id_tool`, which needs an `id`, returns `false`, and has its return value discarded — so the tool vanished without a diagnostic.

The model is still told about the tools by Codex's own prompt, so it prints the call as prose JSON and the turn ends.

## What this does

- Decodes the `additional_tools` item into `LlmRequest::tools`, and consumes the item from the input so the same 24 KB of schemas is not also sent as conversation.
- Advertises a `custom` tool to a chat upstream as a function taking one required string, and turns such a call back into a `custom_tool_call` — buffered and streaming — so the freeform contract holds. The streaming path renames `response.function_call_arguments.delta`/`.done` for the announced item ids only, since a delta event identifies its item by id alone.
- Recovers the freeform input from any argument shape a model may answer with (`{"input": …}`, a bare string, a single differently-named argument), because a dropped call costs the whole turn.
- Re-emits the declaration as an `additional_tools` item when encoding back to Responses, so a same-format hop hands the upstream the shape the client sent.

Both mappings ride in the request's `ProviderExtensions` under prefixed keys, the same approach `codex_namespaces` takes for #384: no provider-neutral type grows a Codex-specific field, and no codec forwards the keys upstream.

A `grammar`-formatted custom tool degrades to an unconstrained string, since a JSON Schema cannot express a grammar. That is strictly better than dropping it — the model can still call the tool, and Codex validates the input on receipt.

## Testing

`cargo test --workspace` passes (`switchyard-py` excluded locally: a pre-existing `libpython3.12.dylib` rpath problem, unrelated to this change). `cargo clippy --all-targets` and `cargo fmt --check` are clean.

New tests:

- `crates/switchyard-translation/src/codex_tools.rs` — 6 unit tests over the extension mapping, the argument-shape recovery, the buffered rewrite, the delta rename, and the two no-op cases.
- `crates/switchyard-translation/tests/request_translation.rs` — an `additional_tools` item becomes the upstream toolset and does not leak into the conversation; it survives a same-format hop unchanged; a `custom` tool reaches a chat upstream as a callable function; a `custom_tool_call` in the replayed history becomes a chat tool call.

Each new test was confirmed to fail before the change.

## Live verification

One routed Codex 0.146.1 session, same config and same prompt, against two servers differing only by this commit:

| server | rollout tool items | command ran |
|---|---|---|
| `main` @ 053a61e2 | none | no |
| this branch | `function_call` + `function_call_output` | yes |

The efficient tier answered both turns, so this is not an escalation artefact.

## Summary by CodeRabbit

* **New Features**
  * Added support for Codex custom tools and additional tools.
  * Custom tool calls and results now translate correctly between Responses and Chat formats.
  * Custom tools preserve their metadata, input values, and original placement.
  * Streaming responses now restore custom tool names and arguments accurately.

* **Bug Fixes**
  * Prevented tool declarations from appearing as conversation content.
  * Preserved custom tools during same-format Responses translation.

* **Tests**
  * Added coverage for custom tools, additional tools, tool-call history, and streaming behavior.

### DIFF
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -88,7 +88,32 @@ impl FormatCodec for OpenAiResponsesCodec {
             policy,
         )?;
         let mut tool_namespaces = Map::new();
-        request.tools = decode_responses_tools(body.get("tools"), &mut tool_namespaces);
+        let mut custom_tools = Map::new();
+        let mut additional_tools = Map::new();
+        request.tools =
+            decode_responses_tools(body.get("tools"), &mut tool_namespaces, &mut custom_tools);
+        // Codex 0.146 declares its tools in an `additional_tools` input item rather
+        // than in `tools`, so a request that offers a full toolset arrives with no
+        // `tools` key at all. Decoding the item here is what lets those tools reach
+        // an upstream; left in the input they become opaque user content and the
+        // model is offered nothing to call.
+        let specs = crate::codex_tools::additional_tool_specs(body.get("input"));
+        if !specs.is_empty() {
+            let decoded = decode_responses_tools(
+                Some(&Value::Array(specs)),
+                &mut tool_namespaces,
+                &mut custom_tools,
+            );
+            for tool in decoded {
+                // A name already declared in `tools` wins: it is the more specific
+                // declaration, and two entries would be an invalid toolset.
+                if request.tools.iter().any(|known| known.name == tool.name) {
+                    continue;
+                }
+                crate::codex_tools::record_additional_tool(&mut additional_tools, &tool.name);
+                request.tools.push(tool);
+            }
+        }
         request.tool_choice = body
             .get("tool_choice")
             .and_then(decode_responses_tool_choice);
@@ -109,6 +134,8 @@ impl FormatCodec for OpenAiResponsesCodec {
             ],
         );
         crate::codex_namespaces::attach_tool_namespaces(&mut request.extensions, tool_namespaces);
+        crate::codex_tools::attach_custom_tools(&mut request.extensions, custom_tools);
+        crate::codex_tools::attach_additional_tools(&mut request.extensions, additional_tools);
         Ok(DecodedRequest {
             request,
             diagnostics,
@@ -147,21 +174,53 @@ impl FormatCodec for OpenAiResponsesCodec {
         if !instructions.is_empty() {
             body.insert("instructions".to_string(), Value::String(instructions));
         }
-        body.insert(
-            "input".to_string(),
+        // Tools the client declared through an `additional_tools` item go back
+        // there, so a Responses-to-Responses hop hands the upstream the same
+        // request shape it was given.
+        let additional_names = crate::codex_tools::additional_tool_names(&request.extensions);
+        let (additional, plain): (Vec<ToolDefinition>, Vec<ToolDefinition>) = request
+            .tools
+            .iter()
+            .cloned()
+            .partition(|tool| additional_names.contains(&tool.name));
+        let input = if additional.is_empty() {
             encode_responses_input(
                 &request.messages,
                 &mut diagnostics,
                 _policy,
                 crate::codex_namespaces::tool_namespaces(&request.extensions),
-            )?,
-        );
-        if !request.tools.is_empty() {
+            )?
+        } else {
+            // The single-user-message shorthand is a bare string, which has no room
+            // for a declaration item, so the list form is required here.
+            let mut items = encode_responses_input_items(
+                &request.messages,
+                &mut diagnostics,
+                _policy,
+                crate::codex_namespaces::tool_namespaces(&request.extensions),
+            )?;
+            items.insert(
+                0,
+                json!({
+                    "type": "additional_tools",
+                    "role": "developer",
+                    "tools": encode_responses_tools(
+                        &additional,
+                        crate::codex_namespaces::tool_namespaces(&request.extensions),
+                        &crate::codex_tools::custom_tool_names(&request.extensions),
+                    ),
+                }),
+            );
+            Value::Array(items)
+        };
+        body.insert("input".to_string(), input);
+        if !plain.is_empty() {
             body.insert(
                 "tools".to_string(),
                 encode_responses_tools(
-                    &request.tools,
+                    &plain,
                     crate::codex_namespaces::tool_namespaces(&request.extensions),
+                    &crate::codex_tools::custom_tool_names(&request.extensions),
                 ),
             );
         }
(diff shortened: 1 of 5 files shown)


fedecia:
Pushed a second commit after live testing found a defect in the first.

A strict Responses backend validates the item-id prefix **per item type**. When the restored `custom_tool_call` kept the `fc_` id the function call had, Azure rejected the item on the *next* request, because the client persists it and replays it:

```
Invalid 'input[8].id': 'fc_2'. Expected an ID that begins with 'ctc'.
```

So a routed session ran its first tool call and then died on turn two, which reads as an unrelated bug. The rewrite now re-prefixes the id and keeps the suffix, and the stream state maps the old id to the new one so the argument deltas — and any other event carrying `item_id` — name the id the client was given.

The suite missed this because every earlier probe happened to have the model pick a `function` tool, which never reaches the restore path. Verified live afterwards: a two-turn routed Codex session produced 5 `custom_tool_call` items, all `ctc`-prefixed (both the re-prefixed short ids and the upstream's own long ones), both turns ran their commands, and the session crossed tiers to an Azure Responses backend and back with no upstream error.

nachiketb-nvidia (commented):
can you adress the coderabbit comments and reply to them? thanks!

also needs a signed commit push

grahamking:
@fedecia I think you might have something here, but there's a bit more work.

1. Plz address the comments above.
2. Could you please re-write the whole description in English in your own words? That will help us understand what you're talking about.
3. It shouldn't be almost a thousand lines. Ask your agent to simplify it dramatically, and then to merge overlapping tests and keep only the most important ones.

fedecia:
@grahamking sure, Codex CLI declares tools through Responses API extensions that Switchyard currently drops during Responses→Chat Completions translation. This leaves the model with no callable tools. We want Switchyard to translate those declarations and round-trip custom tool calls/results, while keeping Responses→Responses requests lossless.

>More specifically, Codex CLI sends its tool declarations inside an additional_tools input item and represents free-form tools with the Responses API’s custom tool type. Switchyard’s normalized translation path currently does not recognize those shapes, so a Responses→Chat Completions route silently drops the tools. The upstream model still sees Codex’s textual instructions, but receives no actual tool definitions and emits attempted calls as plain text.
The proposed change lifts these declarations into the normalized request, maps each custom tool to a function with a single string argument when crossing a Chat Completions backend, and converts the resulting calls and outputs back to the original custom_tool_call format. For Responses→Responses routes, the original additional_tools shape is preserved. This is confined to the translation codec and does not introduce a new public API or routing configuration.

bhuvan002:
Closing due to inactivity. If this is still needed could you please re-open and address the comments from coderabbit? Thanks!

PR #518, opened 2026-08-21
feat(libsy): let stage algorithms fall_open tier be set at runtime

ryan-lempka:
**Background:** Today the stage algorithm has a fallback tier that comes from a config setting called `picker`. When stage algorithm is uncertain of the right model to route to it falls back to the picker. This is currently set inside of the config and fixed for the life of the process.

**What:** This PR adds in the ability to change the fallback tier at runtime. Introduces two new functions `set_fall_open` and `clear_fall_open` so the fallback tier is now read from session state, falling back to picker when nothing is set. Nothing calls these functions yet, so stage behavior is unchanged.

**Why:** A first step towards hierarchical algorithms. This will allow another algorithm (such as LLM Classifier) to set which tier the stage algorithm falls back to at some point during the session. Given a large number of stage algorithm turns will default to the picker we can use LLM Classifier to adjust it based on difficulty. Follow-on PRs will look enable doing this at the start of the session, and also at the start of a user-turn.

## Summary by CodeRabbit

* **New Features**
  * Added persistent fall-open routing overrides for selecting the strong or weak tier.
  * Added support for clearing an override and returning to the configured default.
  * Overrides remain active without suppressing normal signal routing.

### DIFF
--- a/crates/libsy/src/algorithms/stage.rs
+++ b/crates/libsy/src/algorithms/stage.rs
@@ -10,23 +10,23 @@
 //!
 //! Signals do not decide every turn. An under-threshold turn abstains and falls
 //! through to the optional [`LlmTaskClassifier`] — the capability route's judge,
-//! joined in unchanged — and then to the picker's default tier. The judge is
-//! asked per turn and its verdict is never pinned to the session.
+//! joined in unchanged — and then to the picker's default tier, or to an
+//! override a decider ahead of stage set in its place.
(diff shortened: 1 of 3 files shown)


ryan-lempka:
@coberabbitai review

ayushag-nv (approved):
looks good. In the next PR or in an example form , would love to see how tier information gets set in session state.

Issue #495, opened 2026-08-20
docs/feat(libsy): classify_trigger = "user_turn" needs a session id, and cannot fall back to a message hash


gburachas:
## Summary

`classify_trigger = "user_turn"` (added in #487) only takes effect when the caller supplies a session id. That requirement is stated in the source but not in the TOML reference that operators read, and the setting that would otherwise work around it is rejected.

## Detail

`TurnPin` in `crates/libsy/src/algorithms/util/turn_pin.rs` stores the chosen target in the composition's `State`, and its doc comment is explicit:

> Without a session id there is no retained state, and every turn is classified.

That is correct and matches `crates/libsy/src/algorithms/fall_through.rs`, which keeps one state value per session id and uses "unretained per-run state" otherwise. So with no session id, `user_turn` behaves the same as `every_request`.

Two things follow.

**1. The TOML reference does not say so.** `docs/reference/toml_schema.md` describes the setting as:

> `user_turn` judges each new user message and holds that target across the tool calls between.

An operator reading only that would reasonably expect the target to be held, and would see one classifier call per request instead, with nothing in the logs indicating why.

**2. The fallback that exists for this case is rejected here.** `message_hash_fallback` derives an identity from the first user message for exactly this situation, but `llm_class.rs:346` and `llm_class.rs:418` reject it unless `classify_trigger = "new_session"`:

```rust
if self.message_hash_fallback && self.classify_trigger != ClassifyTrigger::NewSession {
    return Err(LibsyError::AlgorithmError {
        message: "message_hash_fallback requires classify_trigger = new_session".to_string(),
    });
}
```

This affects callers that do not send a session id, which includes benchmark harnesses and direct API clients.

## Suggested changes

1. Document the requirement in `docs/reference/toml_schema.md`: `user_turn` requires a session id, and without one it behaves as `every_request`.
2. Allow `message_hash_fallback` with `classify_trigger = "user_turn"`, keying the retained target on the hash of the first user message as it already does for `new_session`.

The first is a documentation fix and stands on its own. The second is a small change to the two validation sites above plus the identity lookup, and I am happy to open a PR for it if that is wanted. If the restriction in (2) is deliberate, then documenting the reason would be enough.

gburachas:
Split across two pull requests:

- documentation: #510 by @ketpatil77
- runtime: #519, which allows `message_hash_fallback` with `classify_trigger = "user_turn"`

[...]

PR #519, opened 2026-08-21
fix(libsy): allow message_hash_fallback with classify_trigger = user_turn

gburachas:
Addresses the runtime half of #495. The documentation half is #510 by @ketpatil77.

## What

`message_hash_fallback` is currently rejected unless `classify_trigger = "new_session"`. This allows it for `"user_turn"` as well, and continues to reject it for `"every_request"`.

## Why

`classify_trigger = "user_turn"` keeps the chosen target across the tool calls following a user message. The target is held per session, so a request with no session id has nowhere to keep it and the classifier runs again on every request. `user_turn` then behaves as `every_request`, with nothing to indicate it.

`message_hash_fallback` exists for that case: it identifies the conversation by a hash of its first user message. It was restricted to `new_session`, which is the one configuration where the problem does not arise.

Callers without a session id include benchmark harnesses and direct API clients. Two of the profiles in `benchmark/routing-profiles/` set `classify_trigger = "user_turn"`, so they are affected.

## The change

The condition in the two config validators and in `LlmTaskClassifier::from_classifier` becomes:

```rust
if message_hash_fallback && classify_trigger == ClassifyTrigger::EveryRequest
```

`every_request` retains nothing between requests, so a fallback identity has nothing to key and the setting stays rejected there.

No other code was needed. `affinity_router` in `crates/libsy/src/algorithms/llm_class.rs` already builds `AffinityRouter::new().with_release_on_user_turn()` for `user_turn`, and already applies `with_message_hash_fallback()` for any trigger. The validation was the only thing preventing the combination.

## Evidence

`user_turn_holds_its_target_without_a_session_id` drives two requests with no session id: an opening user message, then a continuation whose last message is not the user's. It records which targets were called.

| `message_hash_fallback` | Recorded calls |
|---|---|
| `true` (this PR) | `judge`, `efficient`, `efficient` |
| `false` | `judge`, `efficient`, `judge`, `efficient` |

The second row is the current behaviour for anyone without a session id: the judge runs on the continuation as well, which is what `user_turn` is meant to avoid.

## Tests

Two, both in `crates/libsy/src/algorithms/llm_class.rs`:

- `user_turn_holds_its_target_without_a_session_id` — the behaviour above.
- `message_hash_fallback_is_allowed_on_every_retaining_trigger` — the configuration is accepted for both retaining triggers. The `every_request` rejection is already covered by the existing `invalid_classifier_config_is_rejected`, so it is not repeated.

## Note on #510

That PR edits the same table in `docs/reference/toml_schema.md`, on the `classify_trigger` row; this one edits the `message_hash_fallback` row. If #510 merges first I will rebase.

## How tested

- `cargo fmt --all --check`
- `cargo test --workspace`
- `uv run pytest tests/ -m "not integration"` — 113 passed
- `uv run ruff check .`
- Commit signed off per the DCO

## Summary by CodeRabbit

* **New Features**
  * Enabled message-hash fallback for routes using either new-session or user-turn classification.
  * Improved user-turn affinity reuse when session metadata is unavailable.

* **Bug Fixes**
  * Updated validation to consistently reject message-hash fallback for every-request classification.

* **Documentation**
  * Updated configuration guidance to reflect the expanded classification support.

### DIFF
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -342,9 +342,12 @@ impl TaskClassifierConfig {
                 message: "max_output_tokens must be at least 1".to_string(),
             });
         }
-        if self.message_hash_fallback && self.classify_trigger != ClassifyTrigger::NewSession {
+        // Only `every_request` is rejected: it retains nothing, so a fallback identity has
+        // nothing to key. Both retaining triggers can use it.
+        if self.message_hash_fallback && self.classify_trigger == ClassifyTrigger::EveryRequest {
             return Err(LibsyError::AlgorithmError {
-                message: "message_hash_fallback requires classify_trigger = new_session"
+                message: "message_hash_fallback requires classify_trigger = new_session or \
+                          user_turn"
                     .to_string(),
             });
         }
@@ -414,9 +417,12 @@ impl CustomClassifierConfig {
                 message: "max_output_tokens must be at least 1".to_string(),
             });
         }
-        if self.message_hash_fallback && self.classify_trigger != ClassifyTrigger::NewSession {
+        // Only `every_request` is rejected: it retains nothing, so a fallback identity has
+        // nothing to key. Both retaining triggers can use it.
+        if self.message_hash_fallback && self.classify_trigger == ClassifyTrigger::EveryRequest {
             return Err(LibsyError::AlgorithmError {
-                message: "message_hash_fallback requires classify_trigger = new_session"
+                message: "message_hash_fallback requires classify_trigger = new_session or \
+                          user_turn"
                     .to_string(),
             });
         }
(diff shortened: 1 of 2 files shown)


gburachas:
Context on where this sits, since two earlier pull requests of mine were closed and this is not a revival of either.

**#351** (`fix(libsy): harden the classifier judge path`) was closed for inactivity. @ayushag-nv had asked for it to be split, because it addressed three unrelated issues at once, and I did not respond in time. It is not being reopened. Its three parts are being submitted separately: the first was #497 (merged), and the remaining two are issues #277 and #279.

**#352** (`feat(libsy-llm-client): harden upstream request handling`) I closed myself, for the same reason: its description said "Closes #277, closes #260, closes #345". Its parts became #498 (open) and issue #345 (parked, pending a design response).

**This pull request is new work**, not part of either. It addresses one issue, #495, which was filed after both were closed. #495 has a documentation half and a runtime half; @ketpatil77 took the documentation half in #510, and this is the runtime half. A separate pull request rather than an extension of #510 because it changes validation and behaviour, and #510 states it deliberately excludes that.

One issue, three lines of production change, two tests.

ayushag-nv (changes requested):
Overall PR looks good. Just a docs comment: 

Can you update the docs at the following locations with the updated message_hash_fallback behavior. 
a) docs/routing_algorithms/llm_classifier_routing.md
b) crates/switchyard-server/README.md

bhuvan002:

PR #524, opened 2026-08-22
[feature] Cross-platform, declarative way to point any AI agent at switchyard-server

hyeonggyu:
> Baseline: `origin/main` @ `053a61e` (post-`#501`). The proxy is now the standalone Rust binary `switchyard-server`; agents reach it over the wire by pointing their base-URL / model env at it.

## Problem

Every agent is currently wired to the server by hand. The operator must know each agent's base-URL and model environment variables or manually write per-agent configuration files (e.g., OpenCode, OpenClaw). This manual setup introduces several issues:

* Configuration errors and drift (missing `/v1` path segments, invalid model IDs, missing options like `ANTHROPIC_CUSTOM_MODEL_OPTION`).
* Cross-platform platform variances (POSIX `exec` vs. Windows process spawning, `PATHEXT` script/binary resolution).
* Resource leaks from leftover temporary configuration files when config-file-based agents terminate or encounter signal interrupts.

## Proposed solution

A cross-platform native Rust subcommand (`switchyard agent run`) that parses declarative agent specifications, renders environment variables and temporary configurations, and transparently manages agent execution.

1. **Agent spec (JSON / YAML / TOML)**
* `binary`: Executable name or path (resolved across platforms via `$PATH`).
* `env`: Environment variable map supporting `{base_url}`, `{base_url_v1}`, and `{model}` placeholders.
* `config_template` / `config_filename`: Optional temporary configuration rendering for config-file agents.
* Built-in specs shipped for standard agents (Claude Code, Codex, OpenClaw, OpenCode, Hermes, generic OpenAI/Anthropic SDKs).

2. **Cross-platform execution & lifecycle management**
* **Unix (Linux / macOS):** Uses `std::os::unix::process::CommandExt::exec` to replace the process image without overhead.
* **Windows:** Spawns a child process using `Command::spawn`, inherits standard I/O (`Stdio::inherit()`), and forwards child exit codes upon completion.
* **Binary resolution:** Uses OS-agnostic resolution handling executable extensions on Windows (`PATHEXT` for `.exe`, `.cmd`, `.bat`).
* **RAII cleanup:** Implements RAII wrappers combined with signal handlers (`SIGINT`, `SIGTERM`, `Ctrl+C`) to ensure temporary configuration files are purged on exit.

### User-facing example

```bash
# 1) Start the server
switchyard-server --config routes.toml --port 4000 &

# 2) Run any agent via a single command across Linux, macOS, or Windows
[...]

PR #525, opened 2026-08-22
fix(server): add --detach and --pidfile for background/supervised mode (duplicated PR)

hyeonggyu:
## Summary

`switchyard-server` only runs in the foreground today, so it dies with the
launching terminal and offers no first-class way to run as a managed background
service. This adds a detached execution mode:

- `--detach`: re-executes the binary under the system `setsid` (a new session,
  stdio disconnected) **before** the Tokio runtime does significant work, then
  the parent exits and the child keeps serving. The detached child is immune to
  the launching terminal's job-control signals (Ctrl-Z / terminal close).
- `--pidfile <PATH>`: records the detached child's pid (defaults to
  `<temp>/switchyard-server.pid`) so operators can signal it later.

Implementation:
- `crates/switchyard-server/src/daemon.rs` (new): `detach_into_background` and
  `write_pidfile`. Uses the stable `setsid` binary rather than unstable std
  `setsid` features. `--detach` is stripped from the re-exec args to avoid a
  detach recursion.
- `crates/switchyard-server/src/cli.rs`: `ServerArgs` gains `detach` and
  `pidfile` (both `pub(crate)`).
- `crates/switchyard-server/src/main.rs`: detach runs before `cli::run`.

## Test plan

- [x] `cargo build -p switchyard-server` — clean, no warnings.
- [x] `switchyard-server --config routes.toml --port 4123 --detach --pidfile /tmp/sy.pid` → parent exits (rc=0), child runs in its own session (SID == PID, PPID == 1), `GET /health` returns `200`.
- [x] `kill -TERM <pid>` (graceful drain, as a future `stop` would) → server exits, `GET /health` stops answering.

## Out of scope (not in this PR)

The draft issue also proposed `stop` / `status` subcommands and systemd /
launchd service templates. This PR ships only `--detach` + `--pidfile` (the
minimal background-mode primitive); follow-ups can add process-control
subcommands and service templates on top of the pidfile.

## Scope notes

- Unix-only detach (`cfg(unix)`); non-Unix keeps the existing foreground path.
- No change to request/response handling or routing; server behavior is
  identical once running, only the launch lifecycle changes.

## Summary by CodeRabbit

* **New Features**
  * Added support for running the server as a detached background process.
  * Added a `--pidfile` option for configuring the process ID file location.
  * The server now records its background process ID and reports startup failures.
  * A temporary-directory PID file is used by default.

### DIFF
--- a/crates/switchyard-server/src/cli.rs
+++ b/crates/switchyard-server/src/cli.rs
@@ -16,6 +16,13 @@ use switchyard_server::{
 const DEFAULT_HOST: IpAddr = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
 const DEFAULT_PORT: u16 = 4000;
 
+/// Default pidfile path used by `--detach` when `--pidfile` is omitted.
+pub(crate) fn default_pidfile() -> PathBuf {
+    let mut dir = std::env::temp_dir();
+    dir.push("switchyard-server.pid");
+    dir
+}
+
 /// Command-line arguments accepted by the Rust server binary.
 #[derive(Debug, Parser)]
 #[command(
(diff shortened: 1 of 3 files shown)


hyeonggyu:
Closing: duplicate of #524 (reusing the original PR with updated code).

PR #526, opened 2026-08-22
fix(server): add --detach and --pidfile for background/supervised mode

hyeonggyu:
## Summary

`switchyard-server` only runs in the foreground today, so it dies with the
launching terminal and offers no first-class way to run as a managed background
service. This adds a detached execution mode:

- `--detach`: re-executes the binary under the system `setsid` (a new session,
  stdio disconnected) **before** the Tokio runtime does significant work, then
  the parent exits and the child keeps serving. The detached child is immune to
  the launching terminal's job-control signals (Ctrl-Z / terminal close).
- `--pidfile <PATH>`: records the detached child's pid (defaults to
  `<temp>/switchyard-server.pid`) so operators can signal it later.

Implementation:
- `crates/switchyard-server/src/daemon.rs` (new): `detach_into_background` and
  `write_pidfile`. Uses the stable `setsid` binary rather than unstable std
  `setsid` features. `--detach` is stripped from the re-exec args to avoid a
  detach recursion.
- `crates/switchyard-server/src/cli.rs`: `ServerArgs` gains `detach` and
  `pidfile` (both `pub(crate)`).
- `crates/switchyard-server/src/main.rs`: detach runs before `cli::run`.

## Test plan

- [x] `cargo build -p switchyard-server` — clean, no warnings.
- [x] `switchyard-server --config routes.toml --port 4123 --detach --pidfile /tmp/sy.pid` → parent exits (rc=0), child runs in its own session (SID == PID, PPID == 1), `GET /health` returns `200`.
- [x] `kill -TERM <pid>` (graceful drain, as a future `stop` would) → server exits, `GET /health` stops answering.

## Out of scope (not in this PR)

The draft issue also proposed `stop` / `status` subcommands and systemd /
launchd service templates. This PR ships only `--detach` + `--pidfile` (the
minimal background-mode primitive); follow-ups can add process-control
subcommands and service templates on top of the pidfile.

## Scope notes

- Unix-only detach (`cfg(unix)`); non-Unix keeps the existing foreground path.
- No change to request/response handling or routing; server behavior is
  identical once running, only the launch lifecycle changes.

## Summary by CodeRabbit

* **New Features**
  * Added an option to run the server in the background.
  * Added configurable PID file support, with a default location in the system temporary directory.
  * The PID file is created automatically along with any missing parent directories.
* **Bug Fixes**
  * Startup failures during background execution are now reported and handled cleanly.

### DIFF
--- a/crates/switchyard-server/src/cli.rs
+++ b/crates/switchyard-server/src/cli.rs
@@ -17,10 +17,24 @@ const DEFAULT_HOST: IpAddr = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
 const DEFAULT_PORT: u16 = 4000;
 
 /// Default pidfile path used by `--detach` when `--pidfile` is omitted.
+///
+/// Uses a private runtime directory rather than the shared temporary
+/// directory, so the path is not a predictable world-writable location another
+/// user could pre-create (symlink / toctou) to hijack the recorded pid.
 pub(crate) fn default_pidfile() -> PathBuf {
-    let mut dir = std::env::temp_dir();
-    dir.push("switchyard-server.pid");
-    dir
+    if let Ok(dir) = std::env::var("XDG_RUNTIME_DIR") {
+        let mut p = PathBuf::from(dir);
+        p.push("switchyard-server.pid");
+        return p;
+    }
+    // Fall back to a per-user private state dir (~/.local/state/switchyard),
+    // mirroring the debug-log location used elsewhere in the project.
+    let mut p = PathBuf::from(std::env::var("HOME").unwrap_or_else(|_| ".".to_string()));
+    p.push(".local");
+    p.push("state");
+    p.push("switchyard");
+    p.push("switchyard-server.pid");
+    p
 }
 
 /// Command-line arguments accepted by the Rust server binary.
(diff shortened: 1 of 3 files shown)


grahamking:
@hyeonggyu You should use `systemd`. That's the modern way to do this. See `dev-server/switchyard.service` for an example.

hyeonggyu:
@grahamking,

Agreed — systemd is the right primary path, and dev-server/switchyard.service already shows the shape we want:
 * Foreground ExecStart
 * Restart=on-failure
 * journald log capture
 * DynamicUser / LoadCredential
 * CAP_NET_BIND_SERVICE
> [!NOTE]
> The intent was never to reimplement supervision. --detach/--pidfile exists for the cases where no supervisor is available to own the lifecycle — containers, CI runners, one-off dev boxes where systemd isn't PID 1. There the backgrounding can't be delegated to a process manager, so the server has to outlive the launching shell itself.
> 
So instead of pushing this as the supervision story, I'll:
 * Document systemd as the recommended way (a new "Running as a Service" page that points at dev-server/switchyard.service), and
 * Keep --detach/--pidfile explicitly scoped as the supervisor-less fallback, with its trade-offs called out (logs disconnected, no auto-restart).
That matches the direction you're pointing at while keeping the container/CI case covered.

PR #528, opened 2026-08-23
refactor(relay): migrate dynamic plugin to `switchyard-runner`

bbednarski9:
## What

Stacks the current NeMo Relay dynamic-plugin implementation from #488 on top of #517's `switchyard-runner` extraction.

This baseline copy intentionally preserves #488 behavior. Subsequent commits will replace the plugin's duplicated deployment configuration, target-client construction, and route execution with the runner APIs, then remove only behavior that is demonstrably redundant.

## Stack

- Base: #517 (`gk-switchyard-runner`)
- Source baseline: #488 (`feature/nemo-relay-plugin-owned-http-client`)
- #488 remains open and unchanged for comparison.

## Initial validation

- `git diff --check`
- `cargo metadata --no-deps --format-version 1`

`cargo fmt` is not available in the current local Cargo toolchain; CI will run the repository Rust checks.

## Related

Relates to #488
Relates to #517

## Summary by CodeRabbit

* **New Features**
  * Added native NeMo Relay integration for supported buffered and streaming LLM requests.
  * Added configurable deployment routing, request translation, response handling, and failure reporting.
  * Added routing and usage telemetry, including managed-call version requirements and continuation support.
  * Added plugin configuration, installation guidance, and distributable bundle packaging.

* **Bug Fixes**
  * Prevented duplicate attribution of serving calls in telemetry.

* **Documentation**
  * Added NeMo Relay integration details to the project documentation and changelog.


cjagwani:
Integration-contract mismatch at current `0a7f2684`: the README example declares `[[plugins.dynamic]]` with `plugin_id = "nvidia.switchyard"`, but Hermes #77915 activates dynamic plugins through Relay's `load_dynamic_plugin_activation_specs(...)`. In Relay 0.8.0-rc.3 that helper accepts only `manifest` and `config`, requires `plugins.dynamic[i].manifest`, derives the plugin ID/kind from `relay-plugin.toml`, and rejects `plugin_id` as an unknown field. Relay's 0.8 plugin-configuration guide documents the same manifest-backed shape.

Could this PR align the example with the embedded-Hermes activation path and add a focused contract test that parses the documented `plugins.toml` through `load_dynamic_plugin_activation_specs` before `initialize_with_dynamic_plugins`? Passing those specs is the explicit activation consent even though the manifest correctly keeps `defaults.enabled = false`. This is the configuration NemoClaw will eventually need to generate and seal.

tests/relay_plugin/test_package_bundle.py
```diff
@@ -0,0 +1,105 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for the native Relay plugin bundle packager."""
```
afourniernv, line 4: Are these tests run in CI? The Python job only runs pytest tests/, so this directory appears to be skipped.

not sure if it's worth including or not given that there is no regression protection here and this needs to be manually trigged.

crates/switchyard-nemo-relay-plugin/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        "switchyard.runner.streaming",
+        priority,
+        move |name, request, next| {
+            let runtime = Arc::clone(&runtime);
+            let plugin_runtime = plugin_runtime.clone();
+            async move {
+                let Some(inbound) = protocol_from_call(&name) else {
+                    return next.call(request).await;
+                };
+                let decoded = runtime.decode_request(inbound, &request, true)?;
```
afourniernv: Could we check model before decoding? It looks like we translate and clone the body even when we’re just going to pass the request through.

Maybe I'm wrong?

crates/switchyard-nemo-relay-plugin/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+                    .execute_stream(
+                        inbound,
+                        decoded,
+                        Arc::new(move |event| emit_event(&stream_plugin_runtime, event)),
+                    )
+                    .await;
+                emit_events(&plugin_runtime, execution.events);
+                execution
+                    .result
+                    .map(|stream| Box::pin(stream) as LlmJsonAsyncStream)
```
afourniernv: Do we need to box this again? ReturnedEventStream is already the same type as LlmJsonAsyncStream, so I think this can just return execution.result.

crates/switchyard-nemo-relay-plugin/src/runtime.rs
```diff
@@ -0,0 +1,927 @@
...
+                == Some("gateway")
+            && metadata.session_id.as_deref() == Some("gateway-gateway");
+        if relay_gateway_placeholder {
+            metadata.session_id = None;
+        }
+        metadata.http_headers = Some(headers);
+        metadata.wire_format = Some(inbound);
+        Ok(Request {
+            llm_request,
+            raw_request: Some(request.content.clone()),
```
afourniernv: Once we check that Switchyard manages this request, can we pass it by value and move content here? This currently clones the full JSON body, while translation already retains another copy for replay.

crates/switchyard-nemo-relay-plugin/src/runtime.rs
```diff
@@ -0,0 +1,927 @@
...
+        let relay_gateway_placeholder = !headers.contains_key("x-switchyard-session-id")
+            && headers
+                .get("x-nemo-relay-source")
+                .and_then(|value| value.to_str().ok())
+                == Some("gateway")
+            && metadata.session_id.as_deref() == Some("gateway-gateway");
+        if relay_gateway_placeholder {
+            metadata.session_id = None;
+        }
+        metadata.http_headers = Some(headers);
```
ericevans-nv, line 98: Could you avoid forwarding arbitrary inbound Relay headers to the upstream provider? `forward_metadata_headers` sends every non-reserved header from the request metadata, so caller-controlled headers such as `x-ee-review-probe` cross the Relay-to-provider boundary. Please restrict this to an explicit allowlist or headers configured for that provider.

crates/switchyard-nemo-relay-plugin/src/runtime.rs
```diff
@@ -0,0 +1,927 @@
...
+    inbound: WireFormat,
+    metadata: Json,
+    emit_event: RoutingEventEmitter,
+) -> Result<ReturnedEventStream, String> {
+    let served_model = response.served_model().cloned();
+    let chunks = match response.llm_response {
+        LlmResponse::Agg(response) => response.into_stream(),
+        LlmResponse::Stream(chunks) => chunks,
+    };
+    let chunks = Box::pin(chunks.map(move |item| {
```
ericevans-nv, line 382: Could you emit token-usage metrics after the streaming response reports its final usage? The provider returns `prompt_tokens`, `completion_tokens`, and `total_tokens` in the final stream event, but Switchyard currently emits only the request and overhead metrics. This causes `switchyard.routing.llm_tokens` to omit streaming requests.

crates/switchyard-nemo-relay-plugin/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        "switchyard.runner.buffered",
+        priority,
+        move |name, request, next| {
+            let runtime = Arc::clone(&runtime);
+            let plugin_runtime = plugin_runtime.clone();
+            async move {
+                let Some(inbound) = protocol_from_call(&name) else {
+                    return next.call(request).await;
+                };
+                let decoded = runtime.decode_request(inbound, &request, false)?;
```
ericevans-nv: Could you determine route ownership from the raw `model` field before decoding the full payload? `decode_request(...)?` can return a Switchyard translation error for a request whose model is not configured as a Switchyard route, so execution never reaches `runtime.manages` or `next.call`. The streaming interceptor has the same ordering.

tests/relay_plugin/test_package_bundle.py
bbednarski9: Im just going to move this to /tests/relay-plugin/test_package_bundle.py so it runs with CI

crates/switchyard-nemo-relay-plugin/src/lib.rs
bbednarski9: Nice little perf improvement, thanks

crates/switchyard-nemo-relay-plugin/src/lib.rs
bbednarski9: [d7df841](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/d7df841ae219775eca9e4bbbf331699ad8489f46)

crates/switchyard-nemo-relay-plugin/src/runtime.rs
bbednarski9: [9493eb1](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/9493eb1191afd7f174c967b2c778ae28fcf4382d)

crates/switchyard-nemo-relay-plugin/src/runtime.rs
bbednarski9: I don't think I want to maintain a 3rd place with header filtering. We already filter on both ends: Relay dispatch and Switchyard `libsy-llm-client` provider egress. If there are specific fields that should be disallowed by one of these two existing filters, I think you should upstream to those. See crates/libsy-llm-client/src/client.rs:35

crates/switchyard-nemo-relay-plugin/src/runtime.rs
```diff
@@ -0,0 +1,912 @@
...
+            && headers
+                .get("x-nemo-relay-source")
+                .and_then(|value| value.to_str().ok())
+                == Some("gateway")
+            && metadata.session_id.as_deref() == Some("gateway-gateway");
+        if relay_gateway_placeholder {
+            metadata.session_id = None;
+        }
+        metadata.http_headers = Some(headers);
+        metadata.wire_format = Some(inbound);
```
afourniernv: Was the same-format restriction intentional? I tested a Responses request through Relay with both a Chat target and an Anthropic target. Both return 400 before the upstream is called. The old Relay plugin and switchyard-server allow caller and target formats to differ. Removing this assignment and passing `inbound` directly to `check_caller_format` made both cases work. Could we keep those two formats separate and add a cross-format test?

crates/switchyard-nemo-relay-plugin/src/runtime.rs
```diff
@@ -0,0 +1,912 @@
...
+
+fn finalize_buffered_response(
+    translation_engine: &TranslationEngine,
+    inbound: WireFormat,
+    response: Response,
+) -> Result<Json, String> {
+    let LlmResponse::Agg(response) = response.llm_response else {
+        return Err("Switchyard returned a stream for a buffered request".into());
+    };
+    translation::encode_response(translation_engine, inbound, &response)
```
afourniernv: I think we also need to keep the request extensions through response encoding. I tested a Codex MCP tool named `mcp__docs__search`: the plugin returns that flattened name with no namespace in both buffered and streaming responses, while switchyard-server returns `name: search` and `namespace: mcp__docs`. Passing the decoded request extensions into both encoders fixed it. Could we preserve them here while keeping the current translation checks?

crates/switchyard-nemo-relay-plugin/src/runtime.rs
bbednarski9: [f276e85](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/f276e8575b9100a14dab2b40f612e50f6d0622ee)

crates/switchyard-nemo-relay-plugin/src/lib.rs
bbednarski9: [3b3c97b](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/3b3c97b76dd9343dfaccfa3c0915240782d5c7ec)

crates/switchyard-nemo-relay-plugin/src/runtime.rs
bbednarski9: [dd38b96](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/dd38b96f78943f064778bf5654b32d995d7494dc)

crates/switchyard-nemo-relay-plugin/src/runtime.rs
bbednarski9: [4c7e64f](https://github.com/NVIDIA-NeMo/Switchyard/pull/528/commits/4c7e64ffad99052d880f52b6b3daa9fe8d418f97)

grahamking:
Codex said this:

In-band stream failures lose telemetry and structured error handling. Provider error frames are represented as successful stream items containing StreamError or DecodeError. 

The code detects them at [runtime.rs:381](https://github.com/NVIDIA-NeMo/Switchyard/blob/10a0868d873b798563d8c3e511abbc325635e2cc/crates/switchyard-nemo-relay-plugin/src/runtime.rs#L381), but emits failure events only for outer Err items at [runtime.rs:409](https://github.com/NVIDIA-NeMo/Switchyard/blob/10a0868d873b798563d8c3e511abbc325635e2cc/crates/switchyard-nemo-relay-plugin/src/runtime.rs#L409). The encoder’s structured upstream-error value is then flattened into a string at [runtime.rs:421](https://github.com/NVIDIA-NeMo/Switchyard/blob/10a0868d873b798563d8c3e511abbc325635e2cc/crates/switchyard-nemo-relay-plugin/src/runtime.rs#L421).

I added a temporary diagnostic regression test: the client stream failed, but zero failure events were emitted instead of the expected mark and counter. Add coverage for normalized StreamError and DecodeError, preserve the structured Relay error contract, and emit safe failure telemetry exactly once.

grahamking (approved):

PR #530, opened 2026-08-24
fix: encode Responses image and file inputs as wire shapes, not IR enums

shoemoney:
## What

Fix the OpenAI Responses request encoder so image and file content uses Responses wire shapes instead of serialized protocol enums.

- Image URLs stay strings, base64 images become data URLs, and detail is preserved.
- Files use the flat Responses fields: file_id, or file_data with an optional filename.
- Shared OpenAI media mapping lives in a private codec module. Chat and Responses still build their own outer wire shapes.
- Sources that cannot be mapped keep the existing lossy text fallback and diagnostic.

## Why

A Chat image routed to an OpenAI Responses backend was encoded with an object under image_url. The Responses API requires a string there and rejected the request with a 400. Files had the same underlying problem: the encoder serialized Switchyard protocol enums instead of the flat Responses wire format.

Same-format round trips usually replay the preserved request, so they did not expose this. Fresh Chat-to-Responses and Anthropic-to-Responses translations did.

## How tested

- Focused translation tests cover Chat image and file input, plus Anthropic base64 image and document input.
- cargo fmt --all --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test -p switchyard-translation
- cargo test --workspace
- Captured the full outbound request through switchyard-server with a local upstream and checked the flat image and file shapes.
- Sent buffered and streaming Chat image requests through switchyard-server to the live OpenAI Responses API. Both returned valid Chat responses.

## API impact

No public Rust or Python API changes. The observable change is limited to correcting invalid Responses request bodies.

## Checklist

- [x] Commits signed off for DCO.

### DIFF
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -11,7 +11,10 @@ use crate::codecs::common::{
     collect_responses_reasoning_text, encrypted_reasoning_data, encrypted_reasoning_item_id,
     is_known_role_name, provider_extensions, reasoning_text_from_blocks, text_from_blocks,
 };
-use crate::codecs::openai_chat::{decode_file_source, decode_image_source};
+use crate::codecs::openai_chat::{
+    decode_file_source, decode_image_source, file_source_text, image_source_text, openai_file_part,
+    openai_image_part,
+};
 use crate::codecs::{
     DecodedRequest, DecodedResponse, EncodedRequest, EncodedResponse, FormatCodec,
 };
@@ -19,9 +22,9 @@ use crate::diagnostic::TranslationDiagnostic;
 use crate::error::{Result, TranslationError};
 use crate::format::{FormatId, WireFormat};
 use crate::llm::{
-    AggLlmResponse, ContentBlock, InstructionBlock, LlmRequest, MediaSource, Message, OutputParams,
-    ProviderExtensions, ReasoningParams, ResponseOutput, Role, SamplingParams, StopReason,
-    ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
+    AggLlmResponse, ContentBlock, FileSource, ImageSource, InstructionBlock, LlmRequest,
+    MediaSource, Message, OutputParams, ProviderExtensions, ReasoningParams, ResponseOutput,
+    Role, SamplingParams, StopReason, ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
 };
 use crate::policy::{DeterministicIdPolicy, TranslationPolicy};
 use crate::util::{
@@ -1407,9 +1410,17 @@ fn encode_responses_content(
             ContentBlock::Refusal { text } => {
                 blocks.push(json!({"type": "refusal", "refusal": text}));
             }
-            ContentBlock::Image { source } => {
-                blocks.push(json!({"type": "input_image", "image_url": source}));
-            }
+            ContentBlock::Image { source } => match responses_image_part(source) {
+                Some(part) => blocks.push(part),
+                None => {
+                    push_lossy(
+                        diagnostics,
+                        policy,
+                        "Responses codec could not map image content",
+                    )?;
+                    blocks.push(json!({"type": "input_text", "text": image_source_text(source)}));
+                }
+            },
             ContentBlock::Audio { source } => blocks.push(match source {
                 MediaSource::Raw(raw) => json!({"type": "input_text", "text": json_string(raw)}),
                 MediaSource::Url { url, media_type } => json!({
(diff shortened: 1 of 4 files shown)


ayushag-nv:
@shoemoney Thanks for putting up the PR. I would highly appreciate a humanly written PR description which I can consume better and do a proper review.

crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -1187,6 +1206,29 @@ fn encode_responses_content(
...
+    if let Some(detail) = image_url.get("detail") {
+        part["detail"] = detail.clone();
+    }
+    Some(part)
+}
+
+// Maps IR file sources to Responses input_file parts when possible.
+fn responses_file_part(source: &FileSource) -> Option<Value> {
+    let mut part = openai_file_part(source)?;
+    part["type"] = Value::String("input_file".to_string());
```
afourniernv: I think this still leaves us with the Chat shape. `openai_file_part` gives us `{"type":"file","file":{...}}`, so changing only `type` produces `{"type":"input_file","file":{...}}`. I tried that against `/v1/responses` and it returned a 400—the file fields need to be at the top level. Could we lift the inner `file` object here instead? [OpenAI’s type definition](https://github.com/openai/openai-python/blob/0f18e73a808e3910ce13ec764eae8bd1ea6fa6fe/src/openai/types/responses/response_input_file_content_param.py#L21-L46)

crates/switchyard-translation/tests/request_translation.rs
```diff
@@ -2176,3 +2176,102 @@ fn anthropic_thinking_is_dropped_from_responses_input() -> TestResult {
...
+            {
+                "type": "input_image",
+                "image_url": "https://example.test/image.png",
+                "detail": "high"
+            },
+            {"type": "input_image", "image_url": "data:image/png;base64,aW1hZ2U="},
+            {"type": "input_file", "file": {"file_id": "file_123"}},
+            {
+                "type": "input_file",
+                "file": {"file_data": "ZG9jdW1lbnQ=", "filename": "report.pdf"}
```
afourniernv: I think these expected values need to be flattened too: `file_id`, or `file_data` and `filename`, should sit next to `type`. As written, the test is locking in the Chat wrapper that Responses rejects.

shoemoney:
Fixed in 302bf87: input_file now flattens file_id/file_data/filename to the top level instead of nesting under "file", and the request_translation.rs assertion is updated to match.

shoemoney:
Added doc comments to openai_image_part, image_source_text, openai_file_part, and file_source_text in a2804eb.

shoemoney:
@ayushag-nv Circling back on the description you asked about — I found it had gone stale against the code and I've rewritten it.

Specifically: the `FileSource` bullet and the "Notes for reviewers" section both still described `input_file` as the nested `{"file": {...}}` shape from the original commit. 302bf87 flattened those fields to the block top level (`file_id` / `file_data` / `filename`), which is the shape `/v1/responses` actually accepts and the shape the tests assert. So the old prose was arguing for a design the diff no longer implemented — sorry, that's on me, and it would have made the review harder rather than easier. The description now matches the diff.

No code changed with this; it's the description only. The three checks on the head commit (DCO, Validate PR title, CodeRabbit) are green. If you see a red "Commitlint" run, that workflow is failing at startup on every branch in the repo right now, including your own internal ones — not something from this PR.

afourniernv:
@grahamking I tested the before and after here against OpenAI. The old image shape gets a 400 because image_url is an object, while the corrected shape succeeds. The old nested file shape also gets a 400, while the flat file fields in this PR succeed. The bug and the proposed wire format both check out.

Right now the Responses encoder takes the ImageSource/FileSource IR, builds Chat JSON with the existing helpers, then reshapes it into Responses JSON. That reuses the base64 and raw-source handling, but couples Responses to the Chat shape and copies the full inline payload.

We could encode directly from the IR instead. That matches the other codec paths and avoids the extra copy, but repeats some source-mapping logic. I am leaning that way, but wanted your take before asking the contributor to change it.

bhuvan002:
Thank for for the contribution @shoemoney. The team is discussing the design question raised by @afourniernv  and we hope to get back with an answer next week!

shoemoney:
Force-push is a rebase onto `a70a1fb`, no behavior change. The conflict was in `request_translation.rs`, where main added nine tests and this branch adds two, with no overlap; resolved by taking main's file and re-applying both additions. `cargo test -p switchyard-translation` passes (66 in `request_translation`, 188 across the crate), and `cargo fmt --check` and `cargo clippy --all-targets` are clean.

afourniernv:
@shoemoney I dug through the codec paths again and I think we should change this before landing. We share decode helpers today, but this would be the only encoder that builds another codec's wire shape and then reads it back.

Can we pull out just the common image/file payload, then let Chat and Responses build their own content blocks? That keeps the raw-source handling shared and avoids copying large inline payloads.

afourniernv:
@shoemoney I think the overall fix is correct. What makes me nervous is that no other encoder builds another codec’s wire shape and then reads it back, which is why I asked Graham to take a look.

afourniernv:
@shoemoney I pushed the small refactor we discussed. The shared image/file mapping is now private common code, and Chat and Responses each build their own output shape. I also added the Anthropic document case to the existing test. Workspace tests and clippy pass locally.

afourniernv (approved):
Tested this through switchyard-server against OpenAI Responses in both buffered and streaming modes. I also checked the file shape through the codec and a local HTTP capture. Looks good.

afourniernv:
@coderabbitai resolve

afourniernv:
@shoemoney applied the boundary fix. Ty for your contribution!

PR #533, opened 2026-08-24
feat(libsy): add hierarchical routing

ryan-lempka:
**What**
This PR introduces hierarchical routing as a concrete strategy with one first use case: allow LLM classifier to set the stage algorithm's `picker` to capable or efficient (effectively `picker` is the default stage tier that is used when stage signals cannot decide between `capable` and `efficient`).

**Why**
A first use case of brings together #487, #494, and #518. The LLM Classifier works at the user turn boundary and sets the `picker` for stage. For interactive coding agent use-cases this allows the user to influence the behavior of stage, LLM Classifier to classify the user signal, and the tool loop of the agent only needs the signals provided by stage.

**How**
- New file `hierarchical.rs` which includes `TierPicker`, `TierClassifier`, `HierarchicalRouterConfig`, and `HierarchicalRouter`
- Introduces a trait called `Preroute` which is the contract for when you want to call a model but not decide what model serves the final request
- `TierPicker` implements `Preroute` and chooses capable or efficient based on LLM classification result and uses `set_fall_open` to set the picker for stage
- Changed `build_route` from `stage.rs` to `build_stage_route` so HierarchicalRouter::new can inherit new stage features / settings
- Support for `type = "hierarchical"` in the TOML and added `StageTierConfig` struct to make picking up new stage fields by hierarchical in future easier
- `classify_trigger` is honored for LLM classifier in the TOML so the classification boundary can be adjusted

**Additional note:** `tier_fixtures.rs` moved a bunch of test fixtures from stage so hierarchical can use them.

## Summary by CodeRabbit

* **New Features**
  * Added hierarchical routing, combining task classification with stage-based routing.
  * Added configurable pre-processing hooks that can update requests and shared state before routing.
  * Added support for hierarchical routes, tier selection, fallback behavior, and delegated sub-agent routing.

* **Documentation**
  * Added configuration guidance, examples, and routing overview details for hierarchical routing.
  * Clarified classifier fallback behavior for stage routing.


crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -127,16 +127,21 @@ impl StageTargets {
...
-            Some(Tier::Efficient.label())
+            Some(Tier::Efficient)
         } else {
             None
         }
     }
+
+    /// The tier label for a routed target, or `None` for one outside the pair.
+    pub fn label_for(&self, target: &ModelId) -> Option<&'static str> {
+        self.tier_for(target).map(Tier::label)
```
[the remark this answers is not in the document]
ryan-lempka: Seems like test overkill to me

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,307 @@
...
+            return Ok(());
+        }
+        let (classification, _) = self.judge.score(state, request, Some(driver)).await?;
+        if let Some(winner) = classification.argmax(false)?
+            && let Some(tier) = self.targets.tier_for(&winner.target)
+        {
+            set_fall_open(state, tier);
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
ryan-lempka: Moved `retention_key` out of `AffinityRouter` and reused in `TierSetter` so we now fall back to message hash if `message_hash_fallback = true`.

docs/reference/toml_schema.md
```diff
@@ -211,6 +211,28 @@ optional `handoff_notes` and `classifier` tables and for tuning.
...
+### `hierarchical`
+
+Stacks other algorithms in a hierarchy, where one can set another's
+configuration. Today a classifier sets a stage router's `picker`, the tier it defaults to when its own signals are not confident. See
+[Hierarchical Routing](../routing_algorithms/hierarchical_routing.md).
+
+| Key | Required | Default | Meaning |
+|---|:---:|---|---|
+| `classifier.target` | Yes | — | Target the tier judge is called through. Not a routing destination. |
+| `classifier.base_threshold` | Yes | — | `p_solve` floor that still routes to the efficient tier. In `[0, 1]`. |
```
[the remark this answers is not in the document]
ryan-lempka: `every_request` is instead rejected at construction now, docs updated

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    fn hash_keyed_router() -> Result<Arc<HierarchicalRouter>> {
+        Ok(Arc::new(HierarchicalRouter::new(
+            ModelId::from("strong"),
+            ModelId::from("weak"),
+            HierarchicalRouterConfig {
+                judge_target: ModelId::from(JUDGE),
+                judge: TaskClassifierConfig {
+                    base_threshold: 0.5,
+                    classify_trigger: ClassifyTrigger::UserTurn,
```
grahamking, line 212: Is this `UserTurn` correct here given...

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,322 @@
...
+            router.clone(),
+            unkeyed(turn_request(false)),
+            recorder.serve(),
+        )
+        .await?;
+
+        assert_eq!(
+            recorder.judge_calls(),
+            1,
+            "new_session judges once without a session id"
```
grahamking: ... that it says `new_session` here?

grahamking:
I got this from Codex. Can you check it?

Two requests can choose different tiers

This is a blocking issue in `crates/libsy/src/algorithms/hierarchical.rs` at lines 43 through 75.

Two requests with the same message hash can start at the same time. Both can see that no tier was saved. Both then call the judge. They may use different answers. The last write wins and becomes the saved tier.

This breaks the `new_session` promise for clients that do not send a session ID. Requests with a real session ID are kept in order elsewhere. Hash-based requests are not.

Make requests with the same identity share one judge call, or make later requests wait for the first result. Add a test that starts both requests together and proves that the judge runs once.

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,322 @@
...
+/// Retains the tier per routing identity, so it survives requests that carry no
+/// session ID when `message_hash_fallback` is on. The retained tier is replayed
+/// into state on every request so the cascade below reads it.
+struct TierSetter {
+    judge: Arc<dyn Classifier<State>>,
+    targets: StageTargets,
+    trigger: ClassifyTrigger,
+    message_hash_fallback: bool,
+    tiers: Mutex<HashMap<RoutingIdentity, Tier>>,
+}
```
ayushag-nv, line 40: Shall we move it to stage.rs file as it is specific to stage router only ??

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,322 @@
...
+            return Err(LibsyError::AlgorithmError {
+                message: "hierarchical: classify_trigger must be user_turn or new_session"
+                    .to_string(),
+            });
+        }
+        if config.stage.llm_fallback.is_some() {
+            return Err(LibsyError::AlgorithmError {
+                message: "hierarchical: the stage router cannot also carry a judge".to_string(),
+            });
+        }
```
ayushag-nv: users coming from stage router UX might have configs setup for stage router already, the hierarchical thing should be essentially using that as it is. In this case if llm fallback is set already, may be just emit a warning and unset the llm_fallback. Its just UX , your call @ryan-lempka

crates/libsy/src/algorithms/hierarchical.rs
```diff
@@ -0,0 +1,322 @@
...
+            classify_trigger: ClassifyTrigger::EveryRequest,
+            message_hash_fallback: false,
+            ..config.judge
+        };
+        let judge = LlmTaskClassifier::new(LlmClassifierConfig::Capability {
+            judge_target: config.judge_target,
+            efficient_target: efficient.clone(),
+            capable_target: capable.clone(),
+            config: judge_config,
+        })?;
```
ayushag-nv, line 146: Any particular reason for just using Capability based classifier here ? Can we also configure CustomClassifier with custom prompt and schema ??
Does not need to be extended to custom in this PR, can be a follow up item

crates/libsy/src/algorithms/hierarchical.rs
ryan-lempka: @ayushag-nv I removed this entirely so they can be used together

crates/libsy/src/algorithms/hierarchical.rs
ryan-lempka: I originally had this in `stage.rs` - my thought was we could use a separate file to evolve the ideas of hierarchical / unified routing. And it feels like stage shouldn't carry knowledge of this layer.

crates/libsy/src/algorithms/hierarchical.rs
ryan-lempka: I think `CustomClassifier` feels like a follow up item and could be nice for increased flexibility. For now we only use two tiers and also I wanted to use the tuned default prompt. I can still swap out the prompt with Capability as well so that buys me another level of flexibility before deciding if we need entirely new schema as well.

ryan-lempka:
> I got this from Codex. Can you check it?
> 
> Two requests can choose different tiers
> 
> This is a blocking issue in `crates/libsy/src/algorithms/hierarchical.rs` at lines 43 through 75.
> 
> Two requests with the same message hash can start at the same time. Both can see that no tier was saved. Both then call the judge. They may use different answers. The last write wins and becomes the saved tier.
> 
> This breaks the `new_session` promise for clients that do not send a session ID. Requests with a real session ID are kept in order elsewhere. Hash-based requests are not.
> 
> Make requests with the same identity share one judge call, or make later requests wait for the first result. Add a test that starts both requests together and proves that the judge runs once.

@grahamking this behavior already exists in `AffinityRouter`. I changed the logic to use first write wins to match `AffinityRouter`. I'd vote we file as a separate bug if we feel a longer term fix across affinity/hierarchical is needed. But I don't really agree with Codex this is blocking because all it costs is an extra model call if it occurs. Also if a session id is provided by the client it's a non-issue. So this is really just on the fallback path.

crates/libsy/src/algorithms/hierarchical.rs
ryan-lempka: Fixed

crates/libsy/src/algorithms/hierarchical.rs
ryan-lempka: Fixed

sabhatinas (approved):

ayushag-nv (approved):

PR #537, opened 2026-08-24
feat(runner): expose safe route failure summaries

### LINKED ISSUE #536: [feature] `switchyard-runner` expose safe structured terminal execution failure summaries
This is a design proposal for a public failure-summary API to `switchyard-runner`. This enables in-process switchyard consumers to not need to depend on `RunnerError`, `LibsyError` and `LlmClientError` to produce telemetry. It defines categories, HTTP status codes, target ID, and stream phase, while excluding content and sensitive header information.

Scope: contributed to `switchyard-runner`
Uses: `swithyard-libsy`, `switchyard-protocol`
Used by: in-process consumers like NeMo-Relay

Full plan details below:

## Problem

Embedders that execute Switchyard routes in-process need safe, structured terminal-failure telemetry. Today they must couple themselves to the internal error chain:

```text
RunnerError -> LibsyError::ClientCall -> LlmClientError
```

That coupling is brittle, and using `Display` risks exposing upstream response bodies or free-form provider messages in logs. The NeMo Relay dynamic plugin is an immediate consumer: it needs safe routing-error marks for Relay's OpenTelemetry `/v1/logs` export, including terminal and post-header stream failures.

## Proposed solution

Expose a small, public, forward-compatible summary API from `switchyard-runner`.

Illustrative shape:

```rust
#[non_exhaustive]
pub enum RouteFailureCategory {
    UpstreamHttp,
    ContextWindowExceeded,
    Timeout,
    Transport,
    InvalidResponse,
    RequestTranslation,
    RequestEncoding,
    ResponseTranslation,
    InvalidRequest,
    Configuration,
    Algorithm,
    Other,
}

pub enum RouteFailurePhase {
    BeforeResponse,
    DuringStream,
}

pub struct RouteFailureSummary {
    pub category: RouteFailureCategory,
    pub phase: RouteFailurePhase,
    pub upstream_status: Option<u16>,
    pub target: Option<ModelId>,
}
```

Expose conversion points for both execution modes:

```rust
impl RunnerError {
    pub fn execution_failure_summary(&self) -> RouteFailureSummary;
}

pub fn stream_failure_summary(
    error: &LlmClientError,
    served_model: Option<&ModelId>,
) -> RouteFailureSummary;
```

Exact names and ownership are open to review. The public contract should be the stable summary, not the internal error representation.

The summary is suitable for a consumer to emit a mark such as:

```json
{
  "failure_kind": "context_window_exceeded",
  "terminal": true,
  "phase": "during_stream",
  "target": "strong"
}
```

## Alternatives considered

- **Match `RunnerError`, `LibsyError`, and `LlmClientError` in each embedder.** This works now, but tightly couples consumers to internal error layout and makes future refactors risky.
- **Return only a redacted display string.** `libsy` already has private safe-summary logic, but strings are not a stable telemetry contract and force consumers to parse text.
- **Add a NeMo Relay-specific error type.** The runner should remain host-agnostic; Relay can map the public summary to its own mark schema.
- **Extend Relay's native-plugin ABI with typed HTTP errors.** That may be valuable for client-facing response semantics, but is unnecessary for the immediate observability use case.

## Scope notes

- **Owning surface:** `switchyard-runner`; it adds route-execution context that `libsy` alone cannot provide.
- **Public interface:** yes, a new public Rust API. It should be additive and forward-compatible (`#[non_exhaustive]` category enum or equivalent).
- **Streaming:** `Route::execute()` can successfully return a stream, then later receive `LlmClientError` items while that stream is consumed. The API must cover both pre-response execution errors and post-header stream-item errors.
- **Target/model:** retain the terminal target when known. Streamed `Response` metadata already records the served model.
- **Compatibility:** no deployment-TOML, CLI, Python, or routing-algorithm behavior changes are required.

## Additional context

`libsy` already has private redaction helpers that intentionally omit upstream bodies, free-form provider messages, and boxed source messages. The public summary must preserve that safety posture.

Required safe fields:

- stable failure category;
- optional numeric upstream HTTP status, only when directly available;
- selected/terminal target or model when known;
- whether failure happened before response delivery or during stream consumption.

Must exclude:

- upstream HTTP response bodies;
- free-form provider messages;
- request or response content;
- boxed transport, decode, or FFI source messages.

Acceptance criteria:

- [ ] Public runner API summarizes pre-response route-execution failures.
- [ ] Public API summarizes stream-item failures.
- [ ] Context-window, HTTP, timeout, transport, invalid-response, translation, configuration, algorithm, and unknown errors have safe classifications.
- [ ] Tests verify upstream bodies and arbitrary source/provider messages never appear in the summary.
- [ ] Tests verify terminal-target and safe HTTP-status propagation where available.
- [ ] Documentation states that this is for logs/telemetry, not client-facing HTTP error rendering.

bbednarski9:
## What

Adds a public, redaction-safe terminal-failure summary API to `switchyard-runner`. It covers route-execution failures before response delivery and typed failures yielded by active response streams.

## Why

This is deliberately more than a wrapper over existing error types:

- It creates a stable, host-agnostic telemetry contract at the runner boundary, so integration plugins do not couple to the internal `RunnerError` → `LibsyError` → `LlmClientError` layout.
- It combines protocol error class with runner-only context: whether the failure happened before a response or during stream consumption, the terminal target when known, and a directly available HTTP status.
- It makes redaction structural: provider response bodies, free-form messages, request/response content, and boxed sources have no representation in the public summary.
- It publishes stable string values and uses forward-compatible enums and struct construction, so consumers can record telemetry without parsing display strings or blocking future summary fields.

The NeMo Relay dynamic plugin is the immediate consumer: it will map these summaries into structured routing-error marks for Relay OpenTelemetry logs. Plugin adoption is intentionally deferred to a separate change.

Closes #536

## How tested

- [x] `uv run ruff check .`
- [x] `uv run pytest tests/ -v` (115 passed)
- [x] `uv run --only-group docs mkdocs build --strict`
- [x] Focused regression coverage for redaction, HTTP status, explicit and inferred targets, pre-response errors, and stream errors.

## Checklist

- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use.
- [x] README / `--help` updated if customer-facing surface changed.

## Notes for reviewers

The API intentionally does not change routing, retry, fallback, or client-facing HTTP behavior. It is an additive observability surface that keeps provider diagnostics out of telemetry by default.

### DIFF
--- a/crates/switchyard-runner/src/failure.rs
+++ b/crates/switchyard-runner/src/failure.rs
@@ -1,0 +1,335 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Safe, structured summaries of route-execution failures for telemetry.
+
+use libsy::LibsyError;
+use switchyard_protocol::{LlmClientError, ModelId};
+
+use crate::RunnerError;
+
+/// Stable class of a terminal route-execution failure.
+///
+/// This deliberately carries no provider message, response body, or source
+/// error. It is suitable for logs and telemetry, not client-facing rendering.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum RouteFailureCategory {
+    /// The upstream returned a non-success HTTP response.
+    UpstreamHttp,
+    /// The selected target rejected the request because its context window was exceeded.
+    ContextWindowExceeded,
+    /// The upstream request timed out.
+    Timeout,
+    /// The upstream could not be reached or the request could not be sent.
+    Transport,
+    /// The upstream response could not be decoded or validated.
+    InvalidResponse,
+    /// Decoding the inbound request failed in translation.
+    RequestTranslation,
+    /// Encoding the request for the upstream failed in translation.
+    RequestEncoding,
+    /// Decoding or encoding the response failed in translation.
+    ResponseTranslation,
+    /// The request cannot be served as supplied.
+    InvalidRequest,
+    /// The configured route or client cannot serve the request.
+    Configuration,
+    /// The routing algorithm or driver could not produce an outcome.
+    Algorithm,
+    /// A failure without a safe, more specific category.
+    Other,
+}
+
+impl RouteFailureCategory {
+    /// Returns the stable telemetry value for this category.
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::UpstreamHttp => "upstream_http",
+            Self::ContextWindowExceeded => "context_window_exceeded",
+            Self::Timeout => "timeout",
+            Self::Transport => "transport",
+            Self::InvalidResponse => "invalid_response",
+            Self::RequestTranslation => "request_translation",
+            Self::RequestEncoding => "request_encoding",
+            Self::ResponseTranslation => "response_translation",
+            Self::InvalidRequest => "invalid_request",
+            Self::Configuration => "configuration",
+            Self::Algorithm => "algorithm",
+            Self::Other => "other",
+        }
+    }
+}
+
+/// When a terminal failure occurred relative to response delivery.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum RouteFailurePhase {
+    /// The route failed before returning a response to its caller.
+    BeforeResponse,
+    /// A previously returned streaming response failed while it was consumed.
+    DuringStream,
+}
+
+impl RouteFailurePhase {
+    /// Returns the stable telemetry value for this phase.
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::BeforeResponse => "before_response",
+            Self::DuringStream => "during_stream",
+        }
+    }
+}
+
+/// Safe, structured terminal-failure data for routing telemetry.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct RouteFailureSummary {
+    /// Stable failure classification.
+    pub category: RouteFailureCategory,
+    /// Whether the error preceded response delivery or occurred while streaming.
+    pub phase: RouteFailurePhase,
+    /// Upstream HTTP status, when directly available.
+    pub upstream_status: Option<u16>,
+    /// Selected target that failed, when the runner knows it.
+    pub target: Option<ModelId>,
+}
+
+impl RunnerError {
+    /// Returns a safe telemetry summary for a failure before response delivery.
+    pub fn execution_failure_summary(&self) -> RouteFailureSummary {
+        match self {
+            Self::Algorithm(LibsyError::ClientCall { target, source }) => {
+                client_failure_summary(source, RouteFailurePhase::BeforeResponse, Some(target))
+            }
+            Self::Client(source) => {
+                client_failure_summary(source, RouteFailurePhase::BeforeResponse, None)
+            }
+            Self::Configuration { .. } => summary(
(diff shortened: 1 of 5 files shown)


crates/switchyard-runner/src/failure.rs
```diff
@@ -0,0 +1,335 @@
...
+pub struct RouteFailureSummary {
+    /// Stable failure classification.
+    pub category: RouteFailureCategory,
+    /// Whether the error preceded response delivery or occurred while streaming.
+    pub phase: RouteFailurePhase,
+    /// Upstream HTTP status, when directly available.
+    pub upstream_status: Option<u16>,
+    /// Selected target that failed, when the runner knows it.
+    pub target: Option<ModelId>,
+}
```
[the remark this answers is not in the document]
bbednarski9: addressed in eb25d8e6

crates/switchyard-runner/src/failure.rs
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+}
+
+/// Safe, structured terminal-failure data for routing telemetry.
+#[non_exhaustive]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct RouteFailureSummary {
+    /// Stable failure classification.
+    pub category: RouteFailureCategory,
```
grahamking: Nit, but Rust often calls this `ErrorKind`. https://doc.rust-lang.org/std/?search=ErrorKind

crates/switchyard-runner/src/failure.rs
```diff
@@ -0,0 +1,360 @@
...
+            Self::Timeout => "timeout",
+            Self::Transport => "transport",
+            Self::InvalidResponse => "invalid_response",
+            Self::RequestTranslation => "request_translation",
+            Self::RequestEncoding => "request_encoding",
+            Self::ResponseTranslation => "response_translation",
+            Self::InvalidRequest => "invalid_request",
+            Self::Configuration => "configuration",
+            Self::Algorithm => "algorithm",
+            Self::Other => "other",
```
grahamking: If you don't want to maintain these there are a couple options:

1. Replace the whole function with `format!("{self:?}")`. It allocates a `String` and uses the (relatively expensive) format machinery, but hopefully we're not failing a lot in a tight loop :-)

2. Use `strum_macros` crate, which we already pull-in via `jsonschema` in `libsy`. I wouldn't recommend adding it just for this, but given that it's already there:

```
use strum_macros::IntoStaticStr

#[derive(IntoStaticStr)]
pub enum RouteFailureCategory { ...
}
```

https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html

crates/switchyard-runner/src/failure.rs
```diff
@@ -0,0 +1,360 @@
...
+        match self {
+            Self::BeforeResponse => "before_response",
+            Self::DuringStream => "during_stream",
+        }
+    }
+}
+
+/// Safe, structured terminal-failure data for routing telemetry.
+#[non_exhaustive]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
grahamking: Can you drop the `Eq` and `PartialEq` on this and the new types within it? We typically don't need that, and it adds code in the final binary.

grahamking (approved):

crates/switchyard-runner/src/failure.rs
bbednarski9: Changed to RouteErrorKind. ty for this

crates/switchyard-runner/src/failure.rs
bbednarski9: using strum_macros in switchyard-runner now per 43a9397

PR #539, opened 2026-08-25
feat(prefill-router): run batched prefill inference

nachiketb-nvidia:
## What

- redefine `PrefillForward` as one complete batched prompt-to-probabilities operation
- combine Transformers encoding and checkpoint inference in one embedded Python implementation
- batch tokenizer and encoder work in throughput-oriented chunks of 32
- use Accelerate `device_map="auto"` for automatic multi-GPU placement
- map ordered confidence outputs to caller-provided `ModelId`s in `PrefillRouter`
- remove the public intermediate layer, pooling, and feature-matrix API

## Why

The previous boundary sent hidden-state tensors from Python to Rust only for Rust to send them
back into a second Python checkpoint implementation. Those tensors were an implementation detail,
not a useful crate contract.

The batched boundary also gives a future request collector one operation to call after accumulating
prompts for a short wait window. That scheduler is intentionally outside this MR.

```rust
pub trait PrefillForward: Send {
    fn output_count(&self) -> usize;
    fn forward(&mut self, prompts: &[String]) -> Result<Vec<Vec<f32>>>;
    fn unload(&mut self) -> Result<()>;
}
```

## How

- `TransformersForward` owns both the encoder and checkpoint behind `PrefillForward`.
- The checkpoint supplies the encoder name, selected layers, preprocessing tensors, ensemble
  states, and ordered output count.
- One Python call formats all prompts, tokenizes each chunk together, and runs one encoder forward
  per chunk.
- CUDA defaults to `device_map="auto"`; an explicit CUDA device remains supported.
- The causal LM is loaded compatibly with the reference extractor, then only its base encoder is
  retained so the unused vocabulary head does not compute logits or consume inference memory.
- The forward path extracts the last non-padding token from each selected layer, including with
  left-padded tokenizers, then applies layer normalization, scaling, PCA, and the three-member MLP
  ensemble.
- `PrefillRouter::predict_batch` validates and maps each probability row to semantic `ModelId`s;
  `predict` delegates to the same batched path.

Cost-aware selection, `Algorithm` integration, server configuration, Python bindings, and the
wait-window batch scheduler remain separate integration work.

## What to review

- the batched `PrefillForward` boundary
- throughput default and automatic device placement
- last-non-padding-token extraction and checkpoint operation order
- positional checkpoint-head to `ModelId` binding
- validation at the Python/Rust boundary

## Validation

- `cargo test -p prefill-router`
- `cargo clippy -p prefill-router --all-targets -- -D warnings`
- `ruff check crates/prefill-router/python/transformers_forward.py`
- full handoff test with 71 raw prompts through chat templating, tokenization, the 35B encoder,
  checkpoint preprocessing, and ensemble inference on two GPUs
- all 284 probabilities checked; observed maximum BF16 absolute delta `0.0340490`
- all 71 final routing decisions exactly match the handoff

Final diff against `main`: `+550/-1171`.

Linear: https://linear.app/nvidia/issue/SWITCH-1279


crates/prefill-router/python/export_checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        "chat_template_kwargs": transform.get("chat_template_kwargs", {}),
+        "raw_dim": raw_dim,
+        "pca_dim": int(pca.n_components_),
+        "pca_whiten": bool(pca.whiten),
+    }
+
+
+def export_checkpoint(source: Path, destination: Path) -> None:
+    """Convert one trusted pickle checkpoint to JSON plus safetensors."""
+    checkpoint = torch.load(source, map_location="cpu", weights_only=False)
```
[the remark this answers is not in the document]
nachiketb-nvidia: this is outdated, check now

crates/prefill-router/python/transformers_forward.py
```diff
@@ -37,125 +38,148 @@ def _resolve_device(torch: Any, override: str | None) -> str:
...
+        architecture = checkpoint["architecture"]
+        pipeline = checkpoint["feature_pipeline"]
+        if encoder["view"] != "task_prompt_only":
+            raise ValueError("checkpoint encoder view must be task_prompt_only")
+        if pipeline["pooling"] != "mean of independently standardized selected layers":
+            raise ValueError("unsupported checkpoint feature pooling")
+        if architecture["activation"] != "ReLU":
+            raise ValueError("checkpoint activation must be ReLU")
+        if architecture["ensemble_reduction"] != "mean(sigmoid(logits))":
+            raise ValueError("unsupported checkpoint ensemble reduction")
```
bhuvan002: For these could we log `expected X but got Y` for debugging?

crates/prefill-router/python/transformers_forward.py
```diff
@@ -37,125 +38,148 @@ def _resolve_device(torch: Any, override: str | None) -> str:
...
+        import torch
+
+        checkpoint = torch.load(checkpoint_path, map_location="cpu", weights_only=True)
+        if checkpoint["format_version"] != 1:
+            raise ValueError("unsupported checkpoint format_version")
+
+        encoder = checkpoint["encoder"]
+        architecture = checkpoint["architecture"]
+        pipeline = checkpoint["feature_pipeline"]
+        if encoder["view"] != "task_prompt_only":
```
sabhatinas: can you move all of this to a validation method? seems like there are checks for checkpoint in a lot of places

crates/prefill-router/python/transformers_forward.py
```diff
@@ -37,125 +38,148 @@ def _resolve_device(torch: Any, override: str | None) -> str:
...
+            raise ValueError("checkpoint encoder view must be task_prompt_only")
+        if pipeline["pooling"] != "mean of independently standardized selected layers":
+            raise ValueError("unsupported checkpoint feature pooling")
+        if architecture["activation"] != "ReLU":
+            raise ValueError("checkpoint activation must be ReLU")
+        if architecture["ensemble_reduction"] != "mean(sigmoid(logits))":
+            raise ValueError("unsupported checkpoint ensemble reduction")
+
+        self._numpy = np
+        self._torch = torch
```
bhuvan002: Can we remove these? self._numpy only ever has this value. And somewhere else in the file we're doing np = self._numpy haha

crates/prefill-router/python/transformers_forward.py
```diff
@@ -164,45 +188,74 @@ def extract_batch(
...
+                            state["trunk.0.bias"],
+                        )
+                    )
+                    logits.append(
+                        torch.nn.functional.linear(
+                            trunk,
+                            state[f"heads.{index}.weight"],
+                            state[f"heads.{index}.bias"],
+                        )
+                    )
```
bhuvan002: Why do we need to write the forward pass here in this function, can't we just do `torch.load` and call forward on that?

Or if we really need to, could you convert this to an `nn.Module` with weights loaded by `load_state_dict` and `strict=True` instead?

crates/prefill-router/python/transformers_forward.py
```diff
@@ -164,45 +188,74 @@ def extract_batch(
...
+                    self._torch.arange(hidden.shape[0], device=hidden.device), last_token
+                ]
+                layers.append(pooled.cpu().numpy())
+            predictions.append(self._predict(layers))
+
+        return self._numpy.ascontiguousarray(
+            self._numpy.concatenate(predictions, axis=0), dtype=self._numpy.float32
+        ).tobytes()
+
+    def _predict(self, layers: list[Any]) -> Any:
```
bhuvan002: could do `list[np.ndarray] -> np.ndarray`

crates/prefill-router/python/transformers_forward.py
```diff
@@ -37,125 +38,148 @@ def _resolve_device(torch: Any, override: str | None) -> str:
...
-        elif device == "mps":
-            dtype = torch.float16
-        else:
-            dtype = (
-                torch.bfloat16
-                if torch.cuda.get_device_capability(device)[0] >= 8
-                else torch.float16
-            )
-
+        dtype = (
```
bhuvan002: minor nit but the previous version felt more readable than the nested if else/s here

bhuvan002 (changes requested):

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: can you check now?

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: done

crates/prefill-router/python/transformers_forward.py
nachiketb-nvidia: done

bhuvan002 (approved):

PR #542, opened 2026-08-25
fix(llm-client): fall back when the first stream event is a context overflow

### LINKED ISSUE #531: [bug] Streaming requests never fall back on context overflow: in-stream error events are treated as a successful call
## Symptom

When a request carries `"stream": true`, an upstream that rejects it for context overflow answers `200 OK` + `text/event-stream` with the rejection as the first SSE event; Switchyard declares a streaming call successful as soon as those headers arrive, so ordered candidate fallback (#426 / #431) never fires and the client receives the overflow error. The identical request with `"stream": false` falls back correctly, and since agent clients stream by default, context-window fall-through never engages for exactly the traffic it exists for.

## Reproduction

Route: one model served at two admission ceilings, ordered fallback from the small tier to the large one. Backends are two SGLang deployments of the same model behind an OpenAI-compatible gateway.

```toml
[routes.model_name]
id = "model_name"
type = "random"
targets = ["short_ctx", "long_ctx"]   # short_ctx admits 536,826 tokens; long_ctx admits 1,048,576
weights = [1, 0]
```

```bash
switchyard-server --config routes.toml --port 4000

# input larger than short_ctx's admission ceiling (~600k tokens in our capture)
curl -s http://localhost:4000/v1/chat/completions \
  -d '{"model":"model_name","stream":true,"messages":[{"role":"user","content":"<600k-token input>"}]}'
```

Control: the same request with `"stream": false` works as designed: the small tier returns a buffered 400, Switchyard classifies it and falls through, and the large tier serves it.

## Expected vs. actual

- **Expected:** the streaming request falls back to `long_ctx` exactly like the buffered one. At the moment the overflow event arrives, nothing has been forwarded to the client yet, so a retry would be invisible to it.
- **Actual:** the call is declared successful at response headers, the error event is passed through to the client, and `routing_fallbacks` stays at zero. The response is `200 OK`, `content-type: text/event-stream`, body:

```text
data: {"error":{"code":400,"message":"Input length (600013 tokens) exceeds the maximum allowed length (536826 tokens). Use a shorter input or enable --allow-auto-truncate.","object":"error","param":null,"type":"BAD_REQUEST"},"model":"short_ctx"}

data: [DONE]
```

The buffered control logs `model call failed; trying next candidate ... reason=context_window` and is served by `long_ctx`; the streaming case logs a normal successful call.

## Environment

- Switchyard version (or commit SHA): main @ `053a61e2`
- Python version: n/a (Rust server; no Python bindings in the path)
- OS / arch: macOS / arm64
- Install path: source build with Cargo
- Backend: other — OpenAI-compatible gateway in front of two SGLang deployments of the same model

## Additional context

The behavior is pinned by a test, so today it is by design:
- `crates/libsy-llm-client/src/client.rs:196` — a streamed "response is returned as soon as its successful headers arrive", so the fallback boundary ends at headers and `call_first_available` (`run.rs:154`) never sees a streaming overflow.
- `crates/libsy-llm-client/src/run.rs:707` — the test `streams_are_outside_the_candidate_fallback_boundary` asserts an in-stream failure does not try another model.

This is not gateway-specific: SGLang's own streaming tests assert scheduler rejections are delivered in-stream, so a direct SGLang target behaves the same. In a 900-request replay of recorded agentic coding traces, all 26 over-ceiling streaming requests errored to the client and none fell back; the same corpus buffered falls back every time.

**Suggested fix:** read the first stream event before declaring the call successful, since nothing has reached the client yet, so a retry is invisible to it. The decoder already turns upstream `{"error": ...}` frames into a typed `StreamError` (`switchyard-translation/src/codecs/openai_chat/stream.rs:60`); classifying its message with the existing overflow phrases would let the normal candidate loop fall through, and any other first event is reattached and streamed on unchanged. Errors after content starts keep today's semantics — the same boundary #390 drew for stream transport errors on the escalation router. (Success moves from headers to first event; engines send nothing before their admission check, so in practice the same instant.)

Happy to contribute the implementation with a regression test, filing the issue first since the change will exceed the 100-line contribution threshold.

Related: #388 / #390 (transport-error fallback, escalation router only), #424 (SSE terminal-event hygiene), #426 (overflow phrases), #431 (per-request fallback).

fparisio:
## What

Streaming calls now classify the first stream event before the call is declared successful. If that event is an in-band error whose message matches the existing overflow phrases (`is_context_overflow`), the call fails as `ContextWindowExceeded` and the ordered candidate loop falls through — same path, log line, and metrics as a buffered overflow 400. Any other first event (content, a non-overflow error, a transport failure) is reattached to the front of the stream and everything behaves as before. Errors after the first content event keep today's semantics: delivered output cannot be retried.

~30 LOC in `crates/libsy-llm-client/src/client.rs`, four regression tests, and a short addition to `docs/operations/context_window.md`.

## Why

Ordered candidate fallback works for buffered requests but never engages for streaming ones. When a request carries `"stream": true`, an engine that rejects it for context overflow answers `200 OK` + `text/event-stream` and delivers the rejection as the first SSE event:

```text
data: {"error":{"code":400,"message":"Input length (600013 tokens) exceeds the maximum allowed length (536826 tokens). Use a shorter input or enable --allow-auto-truncate.","object":"error","param":null,"type":"BAD_REQUEST"},"model":"short_ctx"}

data: [DONE]
```

Because a streaming call is declared successful as soon as response headers arrive, the error event is never classified, no fallback fires, and the client receives the overflow from a route whose purpose is to absorb it. Agent clients stream by default, so context-window fall-through never engages for exactly the traffic it targets. At the moment the first event arrives nothing has been forwarded to the caller, so retrying on the next candidate is invisible to it.

Closes #531

## How tested

- [x] `uv run pytest tests/` green (the Docker-compose e2e needs a local daemon; verified it fails identically on unpatched `main`, unrelated to this change)
- [x] `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check` all green
- [x] Manual smoke: replayed 900 streaming requests from recorded agentic coding traces through a two-tier route (admission ceilings 536,826 / 1,048,576 tokens, two SGLang deployments behind an OpenAI-compatible gateway, concurrency 4). Before the patch: 26 over-ceiling requests, all errored to the client, second tier idle. With the patch: 18 over-ceiling requests, all rerouted and served by the second tier (largest 600,029 tokens), zero overflow errors reaching clients; the largest request served by the first tier was 536,761 tokens, so the tiering boundary lands exactly on the admission limits.

New tests (in `run.rs`, driving a streaming request through `TranslatingLlmClient` and the candidate loop against a mock SSE server):

- `first_stream_event_overflow_falls_back_to_the_next_candidate` — replays the captured rejection frame verbatim; asserts the second candidate serves and its stream aggregates intact.
- `non_overflow_first_stream_events_are_delivered_unchanged` — a non-overflow first-event error streams through; no fallback.
- `overflow_after_content_stays_within_the_stream` — an overflow-phrased error after content does not reroute; surfaces during aggregation as today.
- `empty_streams_are_returned_without_fallback` — a stream ending before any event aggregates to an empty response; no fallback, no error.

The pre-existing boundary tests (`streams_are_outside_the_candidate_fallback_boundary`, `candidate_failures_follow_the_fallback_policy`) pass unchanged.

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. (n/a — no Python changes)
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (n/a — no new public symbols)
- [x] README / `--help` updated if customer-facing surface changed. (`docs/operations/context_window.md` now documents the streaming case)
- [x] Commits signed off (`Signed-off-by`) per the DCO.

## Notes for reviewers

- **Semantic shift, deliberate:** streaming success now means "first event received" rather than "headers received," so client TTFB moves by one upstream event. Engines send nothing before their admission check, so in practice it is the same instant.
- **Scope is first-event-only on purpose** — the boundary #390 drew for transport errors on the escalation router. Mid-stream errors (even overflow-phrased) and non-overflow first events keep exact current behavior; each case is pinned by a test.
- **A transport error as the first item is reattached, not converted** into a candidate failure. Kept minimal intentionally; easy follow-up if broader first-event handling is wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming requests now detect context-window errors before displaying output and can automatically fall back when appropriate.
  * Errors that occur after content has started streaming remain visible and are not retried.
  * Improved handling of empty streams and other initial streaming errors.

* **Documentation**
  * Added guidance on context-window overflow behavior during streaming requests.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -5,16 +5,18 @@
 //! request, call the configured backend over HTTP, decode the neutral response.
 
 use std::collections::{BTreeMap, HashMap};
+use std::future::ready;
 use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
-use futures_util::StreamExt;
+use futures_util::{StreamExt, stream};
 use http::StatusCode;
 use reqwest::RequestBuilder;
 use reqwest::header::{HeaderMap, RETRY_AFTER};
 use serde_json::{Map, Value};
 use switchyard_protocol::{
-    LlmRequest, LlmResponse, Metadata, ModelId, Request, Response, RoutedLlmClient,
+    LlmRequest, LlmResponse, LlmResponseChunk, LlmResponseStreamEvent, Metadata, ModelId, Request,
+    Response, RoutedLlmClient,
 };
 use switchyard_translation::{
     WireFormat, decode_aggregated_response, decode_request, decode_stream,
@@ -445,8 +447,23 @@ impl TranslatingLlmClient {
                         }
                     })
                 });
-                let chunks = decode_stream(bytes, wire_format)?;
-                LlmResponse::Stream(chunks)
+                let mut chunks = decode_stream(bytes, wire_format)?;
+                // Providers reject an over-ceiling streaming request with an in-band
+                // error event on an HTTP 200. Classify the first event before returning
+                // the stream: nothing has reached the caller yet, so an overflow can
+                // still fail the call and let routing try the next candidate.
+                match chunks.next().await {
+                    None => LlmResponse::Stream(stream::empty().boxed()),
+                    Some(first) => {
+                        if let Some(message) = first_event_overflow(&first, backend) {
+                            return Err(LlmClientError::ContextWindowExceeded {
+                                model: model_id.clone(),
+                                message,
+                            });
+                        }
+                        LlmResponse::Stream(stream::once(ready(first)).chain(chunks).boxed())
+                    }
+                }
             }
             EncodedResponse::Buffered { body, .. } => {
                 let body = serde_json::from_slice::<Value>(&body).map_err(|error| {
@@ -593,6 +610,25 @@ impl AttemptFailure {
             _ => false,
         }
     }
+}
+
+// The overflow message when a stream's first event is an in-band provider rejection
+// of the whole request, rather than the start of a response.
+fn first_event_overflow(
+    first: &Result<LlmResponseStreamEvent>,
+    backend: &Backend,
+) -> Option<String> {
+    first
+        .as_ref()
+        .ok()?
+        .normalized()
+        .iter()
+        .find_map(|chunk| match chunk {
+            LlmResponseChunk::StreamError { message } if backend.is_context_overflow(message) => {
+                Some(message.clone())
+            }
+            _ => None,
+        })
 }
 
 // Uses Retry-After when supplied, capped so an upstream cannot stall a request indefinitely.
--- a/crates/libsy-llm-client/src/run.rs
+++ b/crates/libsy-llm-client/src/run.rs
@@ -729,4 +729,136 @@ mod tests {
         assert_eq!(&*client.calls.lock(), &[ModelId::from("weak")]);
         Ok(())
     }
+
+    // Verbatim shape of an SGLang admission rejection on a streaming session:
+    // HTTP 200, and the overflow arrives as the first SSE event.
+    const OVERFLOW_SSE: &str = "data: {\"error\":{\"code\":400,\"message\":\"Input length (600013 tokens) exceeds the maximum allowed length (536826 tokens). Use a shorter input or enable --allow-auto-truncate.\",\"object\":\"error\",\"param\":null,\"type\":\"BAD_REQUEST\"},\"model\":\"weak\"}\n\ndata: [DONE]\n\n";
+
+    // A normal one-turn streamed answer, as the second candidate serves it.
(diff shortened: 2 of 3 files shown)


grahamking:
@fparisio This looks good, thanks!

Could you reduce to a single test? Most current models write far too many tests.

fparisio:
Thanks @grahamking! Agree, it's one test now, folded all tests as sections, covering edge cases that I think we should not lose. But, happy to trim even more if you prefer.

grahamking (approved):

PR #544, opened 2026-08-25
fix(llm-client): follow only same-origin redirects to keep credentials on origin

### LINKED ISSUE #543: [bug]: default upstream client follows cross-origin redirects while holding the deployment x-api-key
## symptom

the default upstream client follows http redirects, and the forwarded request
still carries the deployment `x-api-key`. so an anthropic `api_key_env` key sent
to a `base_url` that returns a cross-origin 307 or 308 is delivered to the
redirect target origin, not just the configured url. a gemini
`extra_headers.x-goog-api-key` behaves the same way. the openai bearer key does
not, because reqwest strips `authorization` across origins.

the `forward_auth` client already guards against this. `client.rs` builds it with
`redirect(Policy::none())` and the comment "a redirect could move provider-specific
headers to another origin. forwarded credentials are sent only to the configured
url." the default client (used for `api_key_env` and `extra_headers`) does not get
the same policy, so that invariant does not hold for deployment credentials.

## reproduction

local only, canary strings, no real keys. two listeners on different ports: the
origin returns a 307 with a `Location` to the sink, the sink records the headers
it receives.

routes.toml:

```toml
schema_version = 1
[llm_clients.anth]
format = "anthropic_messages"
base_url = "http://127.0.0.1:19420"   # origin, returns 307 to 127.0.0.1:19421
api_key_env = "SY_HUNT_ANTH_KEY"
[targets.cap]
id = "claude-hunt"
llm_client = "anth"
[routes.primary]
id = "claude-hunt"
type = "passthrough"
target = "cap"
```

```bash
export SY_HUNT_ANTH_KEY=canary-anthropic-key
switchyard-server --config routes.toml --host 127.0.0.1 -p 19422

curl -sS http://127.0.0.1:19422/v1/messages \
  -H 'anthropic-version: 2023-06-01' -H 'content-type: application/json' \
  -d '{"model":"claude-hunt","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
```

observed, 3/3:

- client sees http 200, no diagnostic.
- the sink (a different origin) received `x-api-key: canary-anthropic-key`.

control, same 307 shape, openai `api_key_env` (bearer): the origin hop has
`authorization: bearer ...`, the sink hop has no `authorization`. reqwest strips
it across origins.

gemini variant, `extra_headers.x-goog-api-key`: the sink hop has no
`authorization` but still has `x-goog-api-key`.

## expected vs. actual

- **expected:** a deployment credential reaches only the configured `base_url`
  origin, the same guarantee the `forward_auth` client already provides. a
  cross-origin redirect either is not followed or does not carry the credential.
- **actual:** the default client follows the cross-origin redirect and re-sends
  `x-api-key` / `x-goog-api-key` to the new origin. only the openai bearer is
  dropped, and only because reqwest treats `authorization` as sensitive.

## confirmed with live deployment keys

the same repro was run with live anthropic and gemini deployment keys. the 307
pair is all localhost, so no traffic reaches real providers, unredacted captures
are deleted, and only match or no-match is recorded. 5/5:

- anthropic `api_key_env`: the sink (a different origin) received `x-api-key`
  byte-identical to the live deployment key. client http 200.
- gemini `extra_headers.x-goog-api-key`: the sink received the live google key
  while the openai bearer on the same request was dropped. client http 200.
- control, openai `api_key_env`: the origin hop carried the live bearer, the sink
  hop had no `authorization` and no key.

note: on current builds, setting `x-api-key` through `extra_headers` is rejected
at config validation ("extra_headers cannot set x-api-key"), so that sub-vector
no longer applies. the `api_key_env` path and the gemini `x-goog-api-key` header
path are unaffected.

## root cause

- `crates/libsy-llm-client/src/client.rs`: the default `client` is
  `reqwest::Client::builder()` (follows up to 10 redirects). only
  `forward_auth_client` sets `redirect(Policy::none())`. `send_once` picks the
  default client for every non-`forward_auth` backend.
- `crates/libsy-llm-client/src/backend.rs`: `apply_auth` sets anthropic
  `x-api-key` as a plain header and gemini `x-goog-api-key` via `extra_headers`;
  openai uses `bearer_auth`. reqwest scrubs `authorization` on a cross-origin
  redirect but not arbitrary headers.

## suggested fix

either apply `redirect(Policy::none())` to the default client as well, or scrub
provider credential headers (`x-api-key`, `x-goog-api-key`, and configured
`extra_headers` secrets) before following a cross-origin redirect.

## environment

- switchyard version: `switchyard-server 0.2.0`, and current `main` source has the
  same client construction.
- reqwest: 0.13.4.
- inbound format: anthropic messages (also reproduced on openai chat for the
  gemini header).
- backend: local capture pair (anthropic-shaped and openai-shaped canned 200).

Atharva-Kanherkar:
## The bug

`TranslatingLlmClient::new` built the default upstream client with a plain
`reqwest::Client::builder()`, which follows redirects (up to ten). Only
`forward_auth_client` disabled them. `send_once` uses the default client for every
non-forward-auth backend, i.e. deployments configured via `api_key_env` /
`extra_headers`. reqwest strips `Authorization` on a cross-origin redirect but not
arbitrary headers, so a cross-origin 307/308 from the configured `base_url` re-sends
Anthropic `x-api-key` (and Gemini `x-goog-api-key`) to the redirect target origin.
OpenAI (bearer/Authorization) was only accidentally safe.

## The fix

Both upstream clients now share one redirect policy: a redirect is followed only
while it stays on the origin the request was first sent to. A cross-origin
redirect is refused and returned to the caller untouched, so deployment
credentials (`x-api-key`, `x-goog-api-key` via `extra_headers`) and forwarded
caller credentials never leave the configured origin — while same-origin
redirects (Azure-style LB 307s, path normalization) keep working with the
credential riding every hop. This supersedes the earlier all-redirects-refused
approach from the first push: refusing same-origin redirects broke legitimate
endpoints without adding any security.

## Review follow-ups addressed (52d0d9f8)

- Same-origin redirects are followed again; only cross-origin hops are refused.
- The sink assertion now fails on **any** contact with the redirect target;
  timeout/disconnect is the asserted success case (contact-without-key used to pass).
- Request heads are read to the header terminator, so TCP segmentation cannot
  split the headers the assertions match against.
- The refused redirect is pinned as an upstream HTTP 307 error.
- The cross-origin test exercises both credential transports on one request:
  `x-api-key` (`api_key_env`) and `x-goog-api-key` (`extra_headers`).
- New test proves same-origin redirects are followed with the credential
  present on every hop.
- CHANGELOG states the operator-facing behavior change explicitly: endpoints
  answering with a cross-origin redirect must be configured by their final origin.

## Testing

Test contract (locked before the code): deployment credentials reach only the
configured origin; same-origin redirects are followed, cross-origin targets are
never contacted.

- `cross_origin_redirect_never_carries_deployment_credentials`: hermetic
  two-`TcpListener` pair; the origin returns a cross-origin 307 while carrying
  both credential transports; the sink must receive nothing at all.
  Mutation-checked: the test fails if the policy wrongly follows cross-origin
  redirects.
- `same_origin_redirect_is_followed_within_the_configured_origin`: a 307 to the
  same origin is followed end-to-end; both hops carry `authorization`.
- `cargo test -p switchyard-llm-client` green (56 lib + 12 integration).
- `cargo fmt --check` and `cargo clippy --tests -D warnings` clean for the crate.

Closes #543

## Second review round addressed (716cbfeb)

- Dedicated test for the forwarded-auth client: a same-origin 307 is followed
  and the caller-provided `Authorization` rides both hops without leaving the
  configured origin. Its redirect policy was loosened from refuse-all in the
  first push, so this pins the new behavior explicitly.
- `read_request_head` now errors loudly on EOF-before-head-complete or buffer
  exhaustion instead of silently truncating oversized request heads.
- Both same-origin tests share one two-hop server harness; the policy helper
  is renamed to `same_origin_redirect_policy`; the CHANGELOG names both client
  paths (default and forwarded-auth) explicitly.
- Retitled: redirects are confined to the configured origin, not disabled.
  Gates re-run green: fmt --check, clippy --tests -D warnings, 57 lib + 12
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,11 +47,12 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - **Deployment credentials confined to the configured origin** — upstream
-  clients now follow only same-origin redirects, so a cross-origin 307/308
-  from the configured `base_url` is returned to the caller instead of being
-  followed, and `x-api-key`, `x-goog-api-key`, and forwarded caller
-  credentials can no longer reach another origin. Endpoints that answer with
-  a cross-origin redirect must be configured by their final origin. (#543)
+  clients, default and forwarded-auth alike, now follow only same-origin
+  redirects, so a cross-origin 307/308 from the configured `base_url` is
+  returned to the caller instead of being followed, and `x-api-key`,
+  `x-goog-api-key`, and forwarded caller credentials can no longer reach
+  another origin. Endpoints that answer with a cross-origin redirect must be
+  configured by their final origin. (#543)
 - **Reasoning order in mixed stream chunks** — the OpenAI Chat stream decoder
   emits reasoning deltas before content deltas from the same chunk, so
   interleaved reasoning is no longer reordered. (#387)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,11 +46,12 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
-- **Deployment credential leak on cross-origin redirect** — the default
-  upstream client no longer follows redirects, so a cross-origin 307/308 from
-  the configured `base_url` can no longer carry `x-api-key` or
-  `x-goog-api-key` to the redirect target; deployment credentials reach only
-  the configured origin, matching the forward-auth client. (#543)
+- **Deployment credentials confined to the configured origin** — upstream
+  clients now follow only same-origin redirects, so a cross-origin 307/308
+  from the configured `base_url` is returned to the caller instead of being
+  followed, and `x-api-key`, `x-goog-api-key`, and forwarded caller
+  credentials can no longer reach another origin. Endpoints that answer with
+  a cross-origin redirect must be configured by their final origin. (#543)
 - **Reasoning order in mixed stream chunks** — the OpenAI Chat stream decoder
   emits reasoning deltas before content deltas from the same chunk, so
   interleaved reasoning is no longer reordered. (#387)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,11 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **Deployment credential leak on cross-origin redirect** — the default
+  upstream client no longer follows redirects, so a cross-origin 307/308 from
+  the configured `base_url` can no longer carry `x-api-key` or
+  `x-goog-api-key` to the redirect target; deployment credentials reach only
+  the configured origin, matching the forward-auth client. (#543)
 - **Reasoning order in mixed stream chunks** — the OpenAI Chat stream decoder
   emits reasoning deltas before content deltas from the same chunk, so
   interleaved reasoning is no longer reordered. (#387)
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -1317,6 +1317,98 @@ mod tests {
         Ok(())
     }
 
+    // A cross-origin redirect must not carry the deployment credential to the
+    // redirect target. The default client serves non-forward-auth backends and
+    // refuses to follow redirects, so the sink origin never receives x-api-key.
+    #[tokio::test]
+    async fn default_client_does_not_follow_cross_origin_redirect_with_credential()
+    -> std::result::Result<(), Box<dyn Error + Sync + Send + 'static>> {
+        use std::sync::mpsc;
+        use std::time::Instant;
+
+        // The sink stands in for the redirect target on a different origin. It
+        // reports whether any request reached it carrying the credential.
+        let sink = std::net::TcpListener::bind("127.0.0.1:0")?;
+        sink.set_nonblocking(true)?;
+        let sink_addr = sink.local_addr()?;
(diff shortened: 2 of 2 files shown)


grahamking:
Thanks for the PR @Atharva-Kanherkar .

> a cross-origin 307/308 from the configured base_url re-sends Anthropic x-api-key (and Gemini x-goog-api-key) to the redirect target origin

This is the correct behavior. If a provider needs to change their URL, they would issue a 308 to the new one. The API key must follow.

Someone previously reported this as a security issue. That would only be a problem in a man-in-the-middle attack that was able to intercept and decrypt TLS, or if the provider themselves were compromised. Both of those attacks get the API key directly, no need for a redirect.

Atharva-Kanherkar:
> Thanks for the PR @Atharva-Kanherkar .
> 
> > a cross-origin 307/308 from the configured base_url re-sends Anthropic x-api-key (and Gemini x-goog-api-key) to the redirect target origin
> 
> This is the correct behavior. If a provider needs to change their URL, they would issue a 308 to the new one. The API key must follow.
> 
> Someone previously reported this as a security issue. That would only be a problem in a man-in-the-middle attack that was able to intercept and decrypt TLS, or if the provider themselves were compromised. Both of those attacks get the API key directly, no need for a redirect.

yeah you are right. seems okay to me. thanks for the oversight. I am using switchyard with claude code and other coding agents, and I thought this might bite. anyways thank you for clearing this out. thanks anyways! <3

PR #546, opened 2026-08-25
refactor(libsy): decide stage-router picker on score-as-probability

sabhatinas:
Switches the stage-router picker to think in probability instead of raw score.

Today it scores a turn as a signed value in `(-1, 1)` and checks `|score|` against a threshold. This maps that onto a `(0, 1)` probability with `p = (score + 1) / 2`: above `0.5 + t/2` goes capable, below `0.5 - t/2` goes efficient, in between is ambiguous. Same decisions as before, just relabeled — the one edge case is the exact boundary (`score == threshold`), which now reads ambiguous instead of resolving, but that basically never happens.

`probability` carries all the way out: renamed the field on `PickOutcome`, the exported metric (`...stage_router.score` -> `...stage_router.probability`, buckets now `0..1`), and its stats/Prometheus projection in `switchyard-server`. Hard-rule decisions (override, tests-passed) report a neutral `0.5` instead of the old `0.0` placeholder.

Also fixed a bug I found while tracing this: the classifier was reporting confidence as `score.abs()` instead of the confidence the picker already computed, which silently zeroed out confidence on override/tests-passed decisions that should've been maximally confident.

Dashboard heads up: `switchyard_stage_router_score` is renamed to `..._probability` and its buckets move from `[-1,1]` to `[0,1]`.

### DIFF
--- a/crates/libsy/src/algorithms/util/stage.rs
+++ b/crates/libsy/src/algorithms/util/stage.rs
@@ -624,7 +624,7 @@ impl Classifier<State> for StageClassifier {
                 tier,
                 source,
                 score,
-                ..
+                confidence,
             } => {
                 let target = self.targets.name(tier);
                 record_decision_source(state, source);
@@ -633,7 +633,10 @@ impl Classifier<State> for StageClassifier {
                 // is the only branch whose tier the signals actually chose — an
                 // ambiguous turn is decided further down the cascade.
                 self.apply_handoff_note(request, tier, source);
-                let conf = score.abs();
+                // Use the confidence pick_tier already computed (e.g. `1.0` for
+                // an Override) rather than re-deriving it from `score`, which is
+                // `0.0` on both hard-rule paths and would understate them.
+                let conf = confidence.unwrap_or(score.abs());
                 Ok((
                     Classification::Scores(vec![Score {
                         target: target.clone(),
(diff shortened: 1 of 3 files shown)


ayushag-nv (approved):

ayushag-nv:
@sabhatinas address coderabbit first and then merge

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -689,6 +709,95 @@ mod tests {
...
+        );
+        assert_eq!(
+            ScoreResult {
+                score: -t,
+                confidence: t
+            }
+            .probability(),
+            0.5 - t / 2.0
+        );
+    }
```
grahamking: No, come on. The entire function is `(self.score + 1.0) / 2.0`. Let's delete the whole test.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -689,6 +709,95 @@ mod tests {
         );
     }
 
+    #[test]
+    fn probability_remaps_the_score_brackets_onto_zero_point_five_plus_or_minus_half_threshold() {
```
grahamking: The test *name* is 3.5 times as long as the function being tested.

grahamking (changes requested):
That test.

crates/libsy/src/algorithms/util/stage.rs
ayushag-nv: actually we don't need this test

ayushag-nv (commented):
Is this change need to reflected in the docs as well ??

sabhatinas:
Dropped the trivial remap test and replaced the score-based sweep with a small test that exercises `pick_tier` directly at the exact boundary — 52f59ef.

grahamking (approved):

ayushag-nv (approved):
now looks good. Please add any doc changes in the routing algorithms if required

messiaen (approved):

PR #553, opened 2026-08-26
fix(translation): map anthropic metadata.user_id to Chat user

bgrins:
## What

Translates Anthropic `metadata.user_id` to Chat's `user` field in the openai_chat request codec instead of passing `metadata` through untouched. Anthropic metadata is a different shape than Chat metadata (a string map attached to stored completions), and some endpoints reject the unknown shape. Other metadata objects continue to pass through.

## Why

Anthropic clients attach `metadata.user_id` to requests, so pointing one at certain Chat upstreams through Switchyard fails on the first call:

```
curl -s https://opencode.ai/zen/v1/chat/completions \
    -H "Authorization: Bearer $OPENCODE_API_KEY" -H "Content-Type: application/json" \
    -d '{"model":"big-pickle","max_tokens":16,
         "metadata":{"user_id":"user_abc123"},
         "messages":[{"role":"user","content":"hi"}]}'
{"error":{"type":"server_error","message":"Error from provider (Console): Upstream request failed: [1210] Invalid API parameter, please check the documentation."}}
```

## How tested

- [x] `uv run pytest tests/` green
- [x] `cargo fmt --all --check` clean; `cargo clippy -p switchyard-translation --all-targets -- -D warnings` clean
- [x] `cargo test -p switchyard-translation` green — 145 tests
- [x] Manual smoke: the failing curl above, re-sent as the translated shape (`"user":"user_abc123"`, no `metadata`), returns a completion from `big-pickle`. Have also tested long Claude Agent SDK -> Switchyard -> opencode sessions on a wheel built from this branch (without the fix every session dies on its first request).

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. (Rust-only change)
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. (n/a)
- [x] Unit tests added for new components / bug fixes. (`anthropic_openai_compatible_extensions_are_preserved_for_openai_chat` now asserts the mapping)
- [x] README / `--help` updated if customer-facing surface changed. (n/a)

## Notes for reviewers

`user` is set with `or_insert`, so an explicit `user` already on the request wins.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved translation of chat request metadata.
  - When metadata contains only a user ID, it is now correctly represented in the request’s user field.
  - Other metadata continues to be preserved as metadata.

### DIFF
--- a/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
@@ -932,7 +932,6 @@ fn copy_openai_chat_request_extensions(
     extensions: &Map<String, Value>,
 ) {
     for field in [
-        "metadata",
         "parallel_tool_calls",
         "prompt_cache_key",
         "prompt_cache_retention",
@@ -947,6 +946,20 @@ fn copy_openai_chat_request_extensions(
             body.entry(field.to_string())
                 .or_insert_with(|| value.clone());
         }
+    }
+    match extensions.get("metadata") {
+        // Anthropic {"user_id": ...} metadata causes some endpoints to reject
+        // as an invalid parameter. Translate this to Chat's `user`
+        Some(Value::Object(map)) if !map.is_empty() && map.keys().all(|k| k == "user_id") => {
+            if let Some(user) = map.get("user_id") {
+                body.entry("user").or_insert_with(|| user.clone());
+            }
+        }
+        Some(value) => {
+            body.entry("metadata".to_string())
+                .or_insert_with(|| value.clone());
+        }
+        None => {}
     }
     if let Some(stop_sequences) = extensions.get("stop_sequences") {
         body.entry("stop").or_insert_with(|| stop_sequences.clone());
--- a/crates/switchyard-translation/tests/request_translation.rs
+++ b/crates/switchyard-translation/tests/request_translation.rs
@@ -575,7 +575,8 @@ fn anthropic_tool_without_name_is_dropped_before_openai_chat() -> TestResult {
     Ok(())
 }
 
-// Verifies OpenAI-compatible Anthropic extension fields are preserved.
+// Verifies Anthropic extension fields translate to their Chat equivalents:
+// metadata.user_id becomes Chat's `user` (some endpoints reject user_id).
 #[test]
 fn anthropic_openai_compatible_extensions_are_preserved_for_openai_chat() -> TestResult {
     let engine = TranslationEngine::default();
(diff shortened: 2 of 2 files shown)


crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
```diff
@@ -948,6 +947,20 @@ fn copy_openai_chat_request_extensions(
                 .or_insert_with(|| value.clone());
         }
     }
+    match extensions.get("metadata") {
```
afourniernv, line 950: I think this is a little risky because this extension map has already lost its source format. It could have come from Anthropic, Responses, or Chat, so metadata: {"user_id": ...} doesn’t necessarily mean Anthropic metadata.

afourniernv (changes requested):

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
bgrins: Do you think this should be done in a different place or have some other kind of conditional guard? It's a good point to at least fix the comments here, as the endpoint doesn't care what sort of metadata this is coming from (it just throws an error when you pass user_id).

bgrins:
Huh, the original error I was seeing is now gone, so it seems like the provider has just added support for passing user_id in metadata.

```
curl -s https://opencode.ai/zen/v1/chat/completions \
    -H "Authorization: Bearer $OPENCODE_API_KEY" -H "Content-Type: application/json" \
    -d '{"model":"big-pickle","max_tokens":16,
         "metadata":{"user_id":"user_abc123"},
         "messages":[{"role":"user","content":"hi"}]}'
{"id":"472ebc2b-...","object":"chat.completion","created":1788199123,"model":"big-pickle","choices":[{"index":0,"finish_reason":"length","message":{"role":"assistant","content":null,"reasoning_content":"Hmm, the user just said \"hi\" - a simple greeting. They","tool_calls":null}}],"usage":{"prompt_tokens":248,"completion_tokens":16,"total_tokens":264,"prompt_tokens_details":{"cached_tokens":192},"completion_tokens_details":{"reasoning_tokens":0}},"cost":"0"}
```

I don't know if there are others that still have the behavior from last week, but I don't think it's justified to land this anymore unless if there's evidence of requests failing in the wild.

crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
bgrins: Actually, going to close this PR for now given https://github.com/NVIDIA-NeMo/Switchyard/pull/553#issuecomment-5482408992. If you see this come up in practice feel free to reopen / update.

PR #560, opened 2026-08-26
fix(libsy): make tool signals readable from codex traffic

ryan-lempka:
**Background:** Stage routing reads an agent's tool calls and output to pick the capable or efficient tier. Those signals were calibrated against Claude Code. Codex reports its tools differently, and three pattern matches fire on text that is not an error.

Under Codex the tier only ever moves toward capable. Replaying a recorded session through `stage_router`, `efficient_first` went capable on the third request and never returned, `capable_first` never left capable.

**What:** All in `crates/libsy/src/algorithms/util/tool_signals.rs`.

- `command_of` decodes arguments sent as a JSON string, and accepts `command`, `cmd`, or an argv array. Responses sends a string, so the lookup always came back empty.
- `exec_command` joins the shell tools. Codex runs every shell action through it.
- `write_text(`, `writelines(`, `.write(` join the bash write patterns. Codex edits by piping a Python snippet to an interpreter.
- New `reports_no_failure` skips severity scoring when a tool result reports exit 0, or reports the command is still running. Opening a file containing "out of memory" scored as a critical failure, and a streaming chunk echoing a diff scored on whatever that diff contained. Neither is an error the agent hit.
- New `contains_failure_literal` stops `error:` matching the path `error::`. `libsy` has a module named `error`, so passing tests print `test error::tests::...` and a green `cargo test` was never seen as green.

**Why:** The first three make Codex traffic legible. Before them, zero of 39 real `exec_command` calls registered as an edit or write.

The last two are not Codex-specific and matter more. Any agent that reads or writes source code can trip the severity match, and any repo with an `error` module breaks pass detection. Both fail silently: 200s all the way, the cheap tier just stops being used.

Related to #264. Scorer confidence topped out at 0.46 against the 0.5 default, so the tests-passed shortcut was the only route down, and these bugs broke exactly that. Thresholds alone would not have fixed it.

Verified by replaying recorded Codex sessions through a live server: on a passing test run the tier now moves capable to efficient, where before it could not. No claim here about turns or cost, run-to-run variance across a handful of live Codex runs was larger than any difference between routes.

Five tests, each fails without its fix.

### DIFF
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -124,6 +124,9 @@ static BASH_WRITE_PATTERNS: &[&str] = &[
     "<<eof",
     "<<'eof'",
     "<< eof",
+    "write_text(",
+    "writelines(",
+    ".write(",
 ];
 
 static BASH_EDIT_PATTERNS: &[&str] = &[
@@ -161,6 +164,7 @@ static BASH_TOOL_NAMES: &[&str] = &[
     "shell",
     "local_shell_call",
     "terminal",
+    "exec_command", // codex
 ];
 
 // Prefer false negatives: tests_passed routes the picker to EFFICIENT, so a false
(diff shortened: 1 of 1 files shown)


ryan-lempka:
Splitting into two PRs: the harness-agnostic false-positive fixes, and the codex argument parsing.

Issue #563, opened 2026-08-27
[bug] Hardcoded input_modalities: ["text"] makes Codex drop attached images


mrPronin:
## Symptom

Every route's `GET /v1/models` entry declares `input_modalities: ["text"]`, hardcoded, regardless of the resolved target. Codex trusts that declaration and **drops an attached image before sending**, so routing a vision-capable model through Switchyard silently loses images.

## Reproduction

One passthrough route to any OpenAI-Responses target, then a Codex turn with an image:

```bash
switchyard-server --config routes.toml --host 127.0.0.1 --port 4100
codex exec -p <profile-pointing-at-switchyard> -i diagram.png -- 'Describe the image.'
```

`routes.toml`:

```toml
schema_version = 1
[llm_clients.upstream]
format = "openai_responses"
base_url = "http://127.0.0.1:4999/v1"
[targets.t]
id = "some-vision-model"
llm_client = "upstream"
[routes.r]
id = "r"
type = "passthrough"
target = "t"
```

## Expected vs. actual

- **Expected:** the `input_image` item reaches the upstream, as it does with no proxy in the path.
- **Actual:** Codex substitutes a placeholder before the request ever leaves the client:

```json
{"type": "input_text", "text": "image content omitted because you do not support image input"}
```

Measured at the wire with a stand-in upstream logging the request body:

| Run | Outbound body | Last user-message content |
|---|---|---|
| Codex → upstream, no proxy (control) | 759,745 B | `input_image` (543,102-char data URI) |
| Codex → Switchyard, `input_modalities: ["text"]` | **248,385 B** | 60-char placeholder |
| Codex → Switchyard, `input_modalities: ["text","image"]` | **739,155 B** | `input_image`, same as control |

The response is `200` and no diagnostic is emitted anywhere; the only other signal is a correspondingly smaller prompt-token count.

⚠ Note the failure is entirely client-side, so it cannot be seen by inspecting Switchyard's request path — the proxy never receives an image. The placeholder string is in the `codex` binary (6 occurrences in 0.150.1) and not in Switchyard.

## Cause

`crates/switchyard-server/src/lib.rs`, in `codex_model_entry_json`:

```rust
"input_modalities": ["text"],
```

The surrounding comment already anticipates this class of problem — *"Every field below is either derived from the route's declared capabilities or a required `ModelInfo` field the server has no better value for"*, with a TODO about sourcing capabilities from the backend. `input_modalities` is neither derived nor declarable.

[...]

PR #566, opened 2026-08-27
fix: encode Responses image and file content as wire-shaped parts

### LINKED ISSUE #564: [bug] Responses codec emits adjacently-tagged ImageSource/FileSource into image_url/file
## Symptom

The Responses codec serializes the adjacently-tagged `ImageSource` enum directly into `image_url`, emitting an object where the Responses API requires a bare URL or data-URI string. The same defect applies to `FileSource` in `input_file`. Any image or file translated **into** the Responses format is unreadable upstream.

## Cause

`crates/switchyard-translation/src/codecs/responses/buffered.rs` (in `encode_responses_content`):

```rust
ContentBlock::Image { source } => {
    blocks.push(json!({"type": "input_image", "image_url": source}));
}
...
ContentBlock::File { source } => {
    blocks.push(json!({"type": "input_file", "file": source}));
}
```

`ImageSource` and `FileSource` are adjacently tagged (`crates/protocol/src/llm.rs`):

```rust
#[serde(tag = "type", content = "data", rename_all = "snake_case")]
```

so the emitted item is

```json
{"type": "input_image",
 "image_url": {"type": "url", "data": {"url": "https://…", "detail": null}}}
```

## Expected vs. actual

- **Expected:** `"image_url": "https://…"` — a string, as the API requires and as the decoder reads back.
- **Actual:** a nested tagged object; upstream cannot read it.

⭐ The asymmetry is the clearest evidence this is unintended: all three codecs encode the same `ImageSource`, and only this one serializes it raw.

| Codec | Handling |
|---|---|
| `openai_chat/buffered.rs` | `openai_image_part(source)` — builds `{"url": …}` explicitly ✅ |
| `anthropic/buffered.rs` | `match source { … }` — destructures ✅ |
| `responses/buffered.rs` | serializes the enum inline ⛔ |

And within the *same* match block in the Responses codec, `Audio` and `Video` are destructured correctly — only `Image` and `File` are not.

## Why it survived the test suite

`crates/switchyard-translation/tests/lossless_roundtrip.rs` uses the **Chat** shape for the Responses fixture:

```json
{"type": "input_image", "image_url": {"url": "https://example.test/image.png", "detail": "high"}}
```

`decode_image_source` accepts both an object and a string, so the fixture round-trips while never pinning the shape the encoder must emit. Real clients send `"image_url": "data:image/png;base64,…"` as a string.

## Reproduction

Translate an Anthropic request carrying a base64 image to `openai_responses` and inspect `input[0].content`. The added test `anthropic_image_encodes_as_responses_input_image_string` fails on `main` and passes with the fix.

⚠ Note this is **not** reached on a same-format Responses→Responses route, because `encode_request` short-circuits to `exact_preserved_request` and replays the preserved body verbatim. It bites on cross-format routes.

## Environment

- Commit: `7f3b2fe9` (`main`); introduced in the initial commit `86020fab`, so present in `v0.2.0` too.

## Suggested fix

Destructure both enums into the Responses wire shape, mirroring `openai_image_part`, and record a lossy diagnostic instead of dropping an unmappable source silently. PR: see linked pull request.

mrPronin:
## What

Encodes Responses image and file content as the wire shapes the API expects, instead of serializing the adjacently-tagged `ImageSource` / `FileSource` enums inline.

## Why

`encode_responses_content` did:

```rust
ContentBlock::Image { source } => {
    blocks.push(json!({"type": "input_image", "image_url": source}));
}
```

Both enums are `#[serde(tag = "type", content = "data")]`, so that emits

```json
{"type": "input_image",
 "image_url": {"type": "url", "data": {"url": "…", "detail": null}}}
```

where the Responses API requires `image_url` to be a bare URL or data-URI string. Any image or file translated *into* the Responses format is unreadable upstream.

The asymmetry is the argument that this is unintended: all three codecs encode the same `ImageSource`, and only this one serialized it raw — `openai_chat` uses `openai_image_part`, `anthropic` destructures, and `Audio`/`Video` in this very match block are destructured correctly.

Closes #564

## How

- Adds `responses_image_part` / `responses_file_part`, mirroring `openai_chat::openai_image_part`, including the descent into an Anthropic `{"type": "image", "source": {…}}` block for a raw source.
- An unmappable image now records a **lossy diagnostic** instead of vanishing. The silent drop is what made this hard to find.
- `FileSource` keeps the nested `file` object shape, which is what `decode_file_source` reads back for all three variants.

## Why the suite missed it

`tests/lossless_roundtrip.rs` uses the **Chat** shape in the Responses fixture (`"image_url": {"url": …, "detail": "high"}`). `decode_image_source` accepts both forms, so the fixture round-trips while never pinning what the encoder must emit. Real clients send a plain string.

⚠ Also worth knowing for reviewers: this path is **not** reached on a same-format Responses→Responses route, because `encode_request` short-circuits to `exact_preserved_request`. It bites on cross-format routes, which is the other reason it went unnoticed.

## How tested

- [x] `cargo test --workspace` green (33 suites, 0 failures)
- [x] `cargo fmt --check` clean
- [x] Three added tests: a Responses `input_image` keeps `image_url` as a string; `detail` survives as a sibling key; an Anthropic base64 image encodes to a data-URI string. **The last one fails on `main`** — verified by reverting only the source change on this branch.
- [ ] `uv run ruff check .` / `mypy` / `pytest` — n/a, no Python touched

## Notes for reviewers

- ⚠ **DCO sign-off is not yet on the commit.** Say the word and I will `git rebase --signoff` and force-push; I did not want to add the attestation line without being asked.
- `FileSource` carries the identical tagging bug and is fixed the same way. I kept the nested `file` object rather than switching to the API's top-level `file_id`/`file_data`, because that is the shape the existing decoder and fixture assume — happy to change it if you would rather the encoder emit the top-level form.
- Discovered while investigating #563; the two are independent defects and neither fixes the other.

https://claude.ai/code/session_018yveJruskBHpt3EXehSuwo

## Summary by CodeRabbit

* **New Features**
  * Improved image and file handling in Responses requests.
  * Supports image URLs, base64 images, provider image formats, optional image details, and file IDs or data.
  * Converts supported content into the required image and file input formats.

* **Bug Fixes**
  * Preserves image details and correctly formats image URLs and base64 image data.
  * Provides a text fallback for unsupported image sources.

### DIFF
--- a/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
@@ -604,8 +604,20 @@ pub(crate) fn decode_file_source(block: &Map<String, Value>) -> FileSource {
         }
         return FileSource::Raw(Value::Object(file.clone()));
     }
+    // A Responses `input_file` carries the payload directly on the block instead
+    // of nesting it under `file`, so read that shape too. Chat blocks never reach
+    // here with these keys, having matched the nested branch above.
     if let Some(file_id) = block.get("file_id").and_then(Value::as_str) {
         return FileSource::FileId(file_id.to_string());
+    }
+    if let Some(file_data) = block.get("file_data").and_then(Value::as_str) {
+        return FileSource::FileData {
+            data: file_data.to_string(),
+            filename: block
+                .get("filename")
+                .and_then(Value::as_str)
+                .map(ToOwned::to_owned),
+        };
     }
     FileSource::Raw(Value::Object(block.clone()))
 }
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -18,9 +18,9 @@ use crate::diagnostic::TranslationDiagnostic;
 use crate::error::{Result, TranslationError};
 use crate::format::{FormatId, WireFormat};
 use crate::llm::{
-    AggLlmResponse, ContentBlock, InstructionBlock, LlmRequest, MediaSource, Message, OutputParams,
-    ProviderExtensions, ReasoningParams, ResponseOutput, Role, SamplingParams, StopReason,
-    ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
+    AggLlmResponse, ContentBlock, FileSource, ImageSource, InstructionBlock, LlmRequest,
+    MediaSource, Message, OutputParams, ProviderExtensions, ReasoningParams, ResponseOutput, Role,
+    SamplingParams, StopReason, ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
 };
 use crate::policy::{DeterministicIdPolicy, TranslationPolicy};
 use crate::util::{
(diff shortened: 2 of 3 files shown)


afourniernv (commented):
I think this overlaps almost entirely with #530, which is already under review. The extra piece I see here is decoding flat Responses file_data on the reverse path. Could we move that small case into another PR and close this?

afourniernv:
Thanks for taking the time to dig into this. #530 is already addressing the same Responses image and file encoding bug and is further along in review, so I’m closing this one to keep the work in one place. The flat Responses file_data decode is a useful additional case, but it’s separate from the encoder fix and can be handled as a focused follow-up. Appreciate the contribution.

PR #568, opened 2026-08-27
fix(libsy): stop reading source text as a runtime failure

ryan-lempka:
**Background:** Stage routing scores severity by pattern-matching the text a tool returns. Three matches fire on text that is not an error. Severity forces the capable tier ahead of every other rule, and a failure literal vetoes test-pass detection (e.g. exit code 0), so both block de-escalation.

- Reading a file containing "out of memory" scored as a critical failure.
- A streaming chunk echoing a diff scored on the diff, so writing a test fixture with "Traceback" looked like a crash.
- `libsy` has a module named `error`, so passing tests print `test error::tests::...`. Any repo with an `error` module cannot detect a green `cargo test`.

Both failures are silent: routing returns 200 and the cheap tier stops being used.

**What:** Both in `crates/libsy/src/algorithms/util/tool_signals.rs`.

- New `reports_no_failure` skips severity scoring when a tool result reports exit 0, or the command is still running.
- New `contains_failure_literal` stops `error:` matching the path `error::`.

**Why:** Status code should be trusted over text matching and we should not treat messages from source text reading as a runtime error.

### DIFF
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -601,7 +601,7 @@ fn detect_tests_passed(tool_texts: &[String], recent_window: usize) -> bool {
     tool_texts[start..].iter().any(|text| {
         let lower = text.to_lowercase();
         TEST_PASS_PHRASES.iter().any(|p| lower.contains(p))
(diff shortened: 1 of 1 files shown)


bhuvan002:
@ryan-lempka are you still working on this?

PR #569, opened 2026-08-27
fix(libsy): read tool signals from codex exec_command traffic

ryan-lempka:
**Background:** Tool signals are extracted from a tool call's name and its command. Codex reports both differently from Claude Code, so no tool use activity is classified as a read, write, or edit. This matters for routing algorithms in switchyard such as `stage` which rely on tool signals.

Sampled from 146 tool calls across 20 local codex-cli 0.149.1 sessions:

- Tool names used: `exec_command` (110), `write_stdin` (18), `update_plan` (18). No `bash`, `shell`, `terminal` or `local_shell_call`, so every shell action fell through to `Other`.
- Responses sends `arguments` as a JSON string rather than an object, so `arguments.get("command")` never resolved. This affects any Responses client.
- The command is under `cmd`.
- Codex rewrites whole files with `python3 - <<'PY' ... p.write_text(...)` in which case the write is only visible in the command text.

Each `exec_command` call carries the shell text it ran. Since Codex runs every shell action through that one tool, the text is the only way to tell a read from a write. When the arguments cannot be parsed it is unreachable, so nothing gets classified.

**What:** All in `crates/libsy/src/algorithms/util/tool_signals.rs`.

- `command_of` decodes `arguments` sent as a JSON string, and accepts `cmd` and `input` as well as `command`. New `command_text` handles an argv array as well as a string.
- `exec_command` joins the known shell tool names.
- `write_text(` and `.write(` join the bash write patterns, matching 18 and 3 commands in the sample, all of which modified a file.

Across the 110 sampled `exec_command` calls this moves classification from 100% `Other` to 41 reads, 21 writes, 48 other.

OpenAI ref: https://developers.openai.com/api/docs/guides/function-calling#handling-function-calls

## Summary by CodeRabbit

* **New Features**
  * Improved detection of file-writing commands in Python-style code, including `write_text` and `.write()` calls.
  * Added support for recognizing `exec_command` as a shell-like tool.
  * Enhanced command extraction from JSON arguments, including string and argument-array formats.

* **Bug Fixes**
  * Improved handling of heredoc-based file writes and varied command input formats.

### DIFF
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -836,44 +836,47 @@ mod tests {
         assert_eq!(sig.recent_edit_count, 1);
     }
 
-    fn exec_command(arguments: Value) -> Message {
+    fn exec_command(cmd: Value) -> Message {
         Message {
             role: Role::Assistant,
             content: vec![ContentBlock::ToolCall(ToolCall {
                 id: String::new(),
                 name: "exec_command".to_string(),
-                arguments,
+                arguments: cmd,
             })],
         }
     }
 
     #[test]
-    fn exec_command_arg_shapes() {
-        for arguments in [
-            json!({"command": "sed -i s/a/b/ src/lib.rs"}),
-            json!({"cmd": "sed -i s/a/b/ src/lib.rs"}),
-            json!({"cmd": ["bash", "-lc", "sed -i s/a/b/ src/lib.rs"]}),
-            json!(r#"{"cmd":"sed -i s/a/b/ src/lib.rs","workdir":"/x"}"#),
-        ] {
-            let request = with_messages(vec![exec_command(arguments.clone()), tr("ok")]);
-            let signal = ToolSignals::from_request(&request, None);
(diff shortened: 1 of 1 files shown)


crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -124,6 +124,8 @@ static BASH_WRITE_PATTERNS: &[&str] = &[
     "<<eof",
     "<<'eof'",
     "<< eof",
+    "write_text(",
+    ".write(",
```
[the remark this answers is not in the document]
ryan-lempka: Sounds like a bit of an edge case but easy to fix. Done.

crates/libsy/src/algorithms/util/tool_signals.rs
ayushag-nv, line 1: Do we need so many tests ? Can they be trimmed or merged ??

ryan-lempka: Merged two of the tests and drastically simplified others.

ayushag-nv (approved):

PR #573, opened 2026-08-28
fix(libsy): upstream sse error prop

messiaen:
## What

Adds LlmStreamError type to hold an upstream SSE error value or a LlmClientError.

Propagate SSE error events.

## Why

Closes 

SWITCH-1243

### DIFF
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -12,6 +12,7 @@ use futures::{Stream, StreamExt};
 use http::StatusCode;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use thiserror::Error;
 
 use crate::{
     LlmClientError,
@@ -23,6 +24,18 @@ use crate::{
 /// upstream already sent a success status line before failing, so there is no real
 /// code to propagate; 502 matches how a failed upstream call surfaces elsewhere.
 const MID_STREAM_UPSTREAM_STATUS: StatusCode = StatusCode::BAD_GATEWAY;
+
+/// Why a translated event stream stopped early.
+#[derive(Debug, Error)]
+pub enum LlmStreamError {
+    /// An error the upstream sse error
+    #[error("upstream stream error: {0}")]
+    Stream(Value),
+
+    /// The stream itself failed
+    #[error(transparent)]
+    Client(#[from] LlmClientError),
+}
 
 /// A boxed, `Send` stream of response events. Each item may fail independently mid-stream.
 pub type LlmResponseStream =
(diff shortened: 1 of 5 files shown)


crates/protocol/src/stream.rs
```diff
@@ -24,6 +25,18 @@ use crate::{
 /// code to propagate; 502 matches how a failed upstream call surfaces elsewhere.
 const MID_STREAM_UPSTREAM_STATUS: StatusCode = StatusCode::BAD_GATEWAY;
 
+/// Why a translated event stream stopped early.
+#[derive(Debug, Error)]
+pub enum LlmStreamError {
+    /// An error the upstream sse error
```
grahamking: Typo? Claude suggested:
> /// The upstream's own error event, already in the target wire format.

crates/switchyard-translation/src/helpers.rs
```diff
@@ -148,20 +144,29 @@ pub fn encode_stream_with_extensions(
...
+            }
+            // A codec that errored emits the error event last and nothing after it, so the
+            // final value is the terminal frame.
+            let terminal = state.errored.then(|| encoded.pop()).flatten();
+            for value in encoded {
                 yield value;
             }
             if state.errored {
-                return;
+                Err(terminal.map_or_else(
```
grahamking: Claude thinks this `map_or_else` can't trigger. At line 156 you set `terminal` if `state.errored` is true. Hence here, where `state.errored` is true, `terminal` will not be `None`.

The exception would be if `encode_response_stream_event` at line 149 returned an empty Vec.

crates/protocol/src/stream.rs
```diff
@@ -24,6 +25,18 @@ use crate::{
...
+    #[error("upstream stream error: {0}")]
+    Stream(Value),
+
+    /// The stream itself failed
+    #[error(transparent)]
+    Client(#[from] LlmClientError),
```
grahamking, line 37: Claude disapproves, seems reasonable:

> Naming nit — Stream vs Client reads ambiguously (both are "stream" errors). Upstream or InBand would be clearer.

grahamking:
Can you clean up the description? The template is still there.

grahamking (approved):

crates/protocol/src/stream.rs
messiaen: Typo fixed.

crates/protocol/src/stream.rs
messiaen: Changed name to `Upstream`

crates/switchyard-translation/src/helpers.rs
messiaen: @grahamking refactored to

```rust
            let terminal = encoded.pop();
            for value in encoded {
                yield value;
            }
            match (state.errored, terminal) {
                (false, Some(value)) => yield value,
                (true, Some(value)) => Err(LlmStreamError::Upstream(value))?,
                (true, None) => Err(LlmStreamError::Client(LlmClientError::ResponseTranslation(
                    "stream ended after an error without a terminal event".to_string(),
                )))?,
                _ => ()
            }
```

PR #578, opened 2026-08-30
feat(protocol): add typed request context

ilexpoon:
## Status and dependencies

**Draft.** This proposal is blocked on the request-context ownership decision in #583 and on the next breaking protocol release. It is independent of #579 and #586. I will not advance the implementation until the community agrees on the context boundary and public API shape.

## What

Adds a cloneable, typed, in-process context to protocol Metadata using http::Extensions.

The new field is documented and covered by clone and untrusted-header tests. It is not populated from headers, serialized, logged, or forwarded by the protocol crate.

## Why

Rust hosts embedding Switchyard need to carry trusted authorization and tenant context across request clones without reducing it to client-controlled string metadata.

## Notes for reviewers

Start with crates/protocol/src/metadata.rs. The field preserves runtime defaults but is source-incompatible for downstream exhaustive Metadata struct literals. It must ship in the next breaking switchyard-protocol release; callers should migrate to struct update syntax with Metadata::default() or initialize typed_context explicitly.

Validated with workspace Rust tests, strict clippy, rustdoc warnings-as-errors, Ruff, and all 115 Python tests including the Docker e2e.

## Summary by CodeRabbit

* **New Features**
  * Added support for storing cloneable, typed application context in metadata.
  * Typed context is preserved when metadata is cloned and remains available for local processing.

* **Documentation**
  * Added guidance and examples for inserting and retrieving typed context.
  * Clarified that typed context is process-local and is not serialized, logged, forwarded, or populated from HTTP headers.

### DIFF
--- a/crates/protocol/README.md
+++ b/crates/protocol/README.md
@@ -101,6 +101,11 @@ assert_eq!(request.llm_request.tools[0].name, "lookup_metric");
 
 ## Trusted in-process context
 
+This API is unreleased and must ship in the next breaking `switchyard-protocol` release. Adding a
+field to public `Metadata` is source-incompatible for downstream exhaustive struct literals. Such
+callers must add `..Metadata::default()` (or set `typed_context`) before upgrading. The `v0.2.0`
+setup dependency above does not contain this API.
+
 Hosts can attach cloneable, typed values to [`Metadata::typed_context`] without
 encoding trusted state into client-controlled headers or serialized protocol
 payloads:
--- a/crates/protocol/README.md
+++ b/crates/protocol/README.md
@@ -99,6 +99,35 @@ let request = Request {
 assert_eq!(request.llm_request.tools[0].name, "lookup_metric");
 ```
 
+## Trusted in-process context
+
+Hosts can attach cloneable, typed values to [`Metadata::typed_context`] without
+encoding trusted state into client-controlled headers or serialized protocol
+payloads:
+
+```rust
+use switchyard_protocol::Metadata;
+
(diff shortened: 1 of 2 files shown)


ilexpoon:
Addressed the review in 0bc851fb: the header-isolation test now uses and asserts a recognized session header, both behavioral tests document their contracts, and the README plus PR description explicitly schedule this source-incompatible public-field addition for the next breaking switchyard-protocol release with migration guidance. Revalidated strict clippy, 25 unit tests, 3 doctests, and rustdoc with warnings denied.

afourniernv:
The use case makes sense, but adding this to the released `Metadata` type is a bigger API decision. I opened #583 so the team can decide where it should live. Mind moving this back to draft for now? Thanks for putting the implementation together.

ilexpoon:
Thanks, agreed. I have moved this PR back to draft while we settle the ownership question in #583.

I verified that the trusted context needs to survive request cloning and routing and remain available to the final `RoutedLlmClient`, so a server-only context would not cover the embedding use case. However, that does not necessarily mean the protocol should expose `http::Extensions` directly.

I have documented the required propagation boundary in #583 and proposed an opaque protocol-owned typed context API for the next breaking protocol release. I will not advance this PR until that API direction is agreed.

ilexpoon:
I added a status note to the description. This remains draft pending the context ownership discussion in #583 and the next breaking protocol release. It is independent of #579 and #586.

messiaen:
Closing as it's still in draft after 3 weeks.
Feel free to reopen.

PR #579, opened 2026-08-30
feat(llm-client): configure HTTP client builders

ilexpoon:
## What

Adds TranslatingLlmClient::with_http_client_builders so an embedding host can configure the two existing provider HTTP clients through reqwest::ClientBuilder values.

The ordinary builder keeps the host redirect policy. The forwarded-credential builder always has redirects disabled so caller credentials cannot move to another origin.

## Why

The concrete caller is a private Rust gateway that needs host-controlled DNS, TLS, proxy, timeout, and connection-pool settings. It does not currently need cross-instance shared pools, so that broader API and claim have been removed.

## Status and review order

Draft while the narrowed builder-injection API is reviewed. This PR is independent of #578 and #586 and has no merge-order dependency on them.

## Compatibility

TranslatingLlmClient::new and its defaults remain unchanged. Model validation still occurs before fallible HTTP client construction. Standalone configuration loaders are unchanged.

## Validation

- cargo fmt --all -- --check
- cargo clippy --workspace --all-targets --locked -- -D warnings
- cargo test --workspace --locked
- RUSTDOCFLAGS="-D warnings" cargo doc -p switchyard-llm-client --no-deps --locked

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -128,6 +128,46 @@ impl TranslatingLlmClient {
     /// Builds a client over the given [`ModelConfig`]s, with a fresh shared HTTP
     /// client and the built-in translation codecs.
     pub fn new(model_configs: &[ModelConfig]) -> Result<Self> {
+        Self::new_with_http_client_factory(model_configs, || {
+            build_http_clients(reqwest::Client::builder(), reqwest::Client::builder())
+        })
+    }
+
+    /// Builds a client over `model_configs` using host-configured HTTP client builders.
+    ///
+    /// The ordinary builder retains its redirect policy. Redirects are always disabled on the
+    /// forwarded-credential builder so caller credentials cannot move to another origin.
+    /// Configuration validation is identical to [`Self::new`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`LlmClientError`] when model configuration is invalid or either builder cannot
+    /// construct its client.
+    pub fn with_http_client_builders(
+        model_configs: &[ModelConfig],
+        client_builder: reqwest::ClientBuilder,
+        forward_auth_client_builder: reqwest::ClientBuilder,
+    ) -> Result<Self> {
+        Self::new_with_http_client_factory(model_configs, || {
+            build_http_clients(client_builder, forward_auth_client_builder)
+        })
+    }
+
+    // Keeps configuration failures ahead of fallible default client construction.
+    fn new_with_http_client_factory(
+        model_configs: &[ModelConfig],
+        build_http_clients: impl FnOnce() -> Result<(reqwest::Client, reqwest::Client)>,
+    ) -> Result<Self> {
+        Self::validate_model_configs(model_configs)?;
+        let (client, forward_auth_client) = build_http_clients()?;
+        Ok(Self::from_validated_configs(
+            model_configs,
+            client,
+            forward_auth_client,
+        ))
+    }
+
+    fn validate_model_configs(model_configs: &[ModelConfig]) -> Result<()> {
         for config in model_configs {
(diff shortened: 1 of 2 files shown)


ilexpoon:
Addressed all review nits in 7cd85dbc by documenting HttpClients::new error behavior, the test helpers, and the injected-header/shared-pool/redirect-boundary contracts. No runtime behavior changed. Revalidated strict clippy, 69 tests, and rustdoc with warnings denied.

crates/libsy-llm-client/src/client.rs
```diff
@@ -100,6 +139,19 @@ impl TranslatingLlmClient {
     /// Builds a client over the given [`ModelConfig`]s, with a fresh shared HTTP
     /// client and the built-in translation codecs.
     pub fn new(model_configs: &[ModelConfig]) -> Result<Self> {
+        let http_clients =
```
afourniernv: Could we keep model validation before building the default HTTP clients? `new` previously returned configuration errors first, and changing that order seems unrelated to client injection.

afourniernv (commented):
What integration is this meant for? `Runner::load` and `Runner::from_toml` still build their own clients, and neither the server uses this constructor. I want to understand the caller before we add the public API.

crates/libsy-llm-client/src/client.rs
ilexpoon: Agreed. Fixed in e2097373. `TranslatingLlmClient::new` now validates all model configs before invoking the fallible default HTTP-client factory; `with_http_clients` uses the same validation and both constructors converge on one validated construction path. The regression test uses a factory that would panic if invoked and proves the configuration error wins first.

ilexpoon:
The intended caller is a programmatic embedded Rust gateway, not the standalone TOML loader. Its integration boundary is `host runtime builder -> host client factory -> HttpClients::new -> TranslatingLlmClient::with_http_clients -> host RoutedLlmClient -> Runner::new`. The host groups deployments by endpoint, credential identity/version, wire format, and network policy, then shares its DNS, TLS, proxy, timeout, redirect, and connection-pool policy across the translating clients in that group.

I updated the README to make that boundary explicit; `Runner::load` and `Runner::from_toml` intentionally retain their existing defaults.

The branch is now rebased onto current `main` at c597bfd, the `AuxiliaryOperation`/`HttpClients` export conflict is resolved, the constructor ordering regression is fixed, and the PR is independently mergeable at e2097373 with DCO green. Local exact-tree gates passed: rustfmt, strict package clippy, 73 package/integration tests, and rustdoc with warnings denied. The upstream GitHub Actions runs are currently awaiting maintainer approval for the fork workflow.

afourniernv:
@ilexpoon Thanks, that clears up the intended path. Is there a concrete Rust host that needs to share pools across multiple TranslatingLlmClient instances today? Since one client already supports multiple model configs, I want to make sure this is a current integration requirement before adding the reqwest-coupled public API.

afourniernv:
@ilexpoon like any project, repository, or current deployment that is driving this?

ilexpoon:
Yes — the concrete caller is a private Rust gateway under active development. I rechecked its current implementation, and the real need is host-configured HTTP client builders for DNS, TLS, proxy, and timeout policy. It does not currently need cross-instance shared pools, so that part of this PR was too broad. I have moved it to draft and will narrow it to builder injection, while keeping deployment policy downstream.

ilexpoon:
Updated. Our Rust gateway needs to configure DNS, TLS, proxy, and timeout policy for the provider clients. This PR now only injects the two existing client builders; I removed the shared-pool wrapper and claim. Redirect protection and validation order are covered, and the full Rust gates pass.

messiaen:
closing due to inactivity.

PR #580, opened 2026-08-30
feat(server): add composable ingress hooks

ilexpoon:
## What

Adds provider-neutral ingress hooks for pre-body preparation or rejection, logical route-model resolution, and final response presentation.

It also adds build_llm_router, a reduced router exposing only Chat Completions, Responses, and Messages. The existing full router keeps its current defaults and route set.

## Why

Rust hosts embedding Switchyard need to authenticate before buffering an untrusted body, attach trusted request context, map public models to internal routes, and keep operational endpoints on a separately protected surface without copying Switchyard handlers.

## Notes for reviewers

Start with IngressHooks and build_llm_router in crates/switchyard-server/src/lib.rs.

Tests cover rejection before invalid JSON extraction, public-to-internal model mapping, no upstream call on rejection, and exclusion of metrics from the reduced router. Existing protocol, streaming, cancellation, logging, and full-router regressions remain green.

Validated with workspace Rust tests, strict clippy, rustdoc warnings-as-errors, Ruff, and all 115 Python tests including the Docker e2e.

## Summary by CodeRabbit

* **New Features**
  * Added support for embedding LLM routes in Rust applications with customizable request authentication, metadata enrichment, model resolution, and response formatting.
  * Added a dedicated LLM router for Chat Completions, Responses, and Messages endpoints.
  * Standalone operation continues to include operational endpoints such as metrics.

* **Documentation**
  * Added guidance for integrating embedded LLM routes into Rust hosts.

### DIFF
--- a/crates/switchyard-server/README.md
+++ b/crates/switchyard-server/README.md
@@ -68,8 +68,9 @@ Rust hosts can install [`IngressHooks`] on [`ServerState`] and use [`build_llm_r
 only Chat Completions, Responses, and Messages. `prepare` runs before JSON body extraction, so a
 host can authenticate or reject without buffering an untrusted body. The async `resolve_model`
 runs after wire-format decoding and can enforce host policy before mapping a public model to an
-internal route. `present_response` can adapt the final error envelope. All methods have defaults
-that preserve the standalone server behavior.
+internal route. `present_response` receives the resolved request metadata and can adapt the final
+error envelope or add request-scoped headers. All methods have defaults that preserve the
+standalone server behavior.
 
 Server-generated failures retain public [`ApiError`] metadata in response extensions. A host
 presenter can inspect its status, message, type, and stable code without buffering or parsing the
@@ -113,6 +114,15 @@ impl IngressHooks for HostHooks {
         _wire_format: WireFormat,
     ) -> Result<ModelId, Response> {
         Ok(ModelId::from("internal/route"))
+    }
+
+    fn present_response(
+        &self,
+        response: Response,
+        _wire_format: WireFormat,
+        _request_metadata: &Metadata,
+    ) -> Response {
+        response
     }
 }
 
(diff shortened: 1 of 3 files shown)


ilexpoon:
Addressed the review in 4f6a8b80: the embedding docs now declare async-trait as a direct host dependency, resolve_model documents Err(Response) behavior, and both ingress tests state their ordering/router contracts. No runtime behavior changed. Revalidated strict clippy, 60 tests, and rustdoc with warnings denied.

afourniernv (changes requested):
Could you please catch branch up and resolve conflicts? Please also have proper DCO sign off. 

Is there any way to compress the PR to make it easier to review?

ilexpoon:
Thanks. I have moved this PR back to draft; it needs more than a conflict-only rebase.

I also need to correct my earlier validation statement: that validation was run against a combined local branch containing the typed-context patch, not against this PR exact head. I rechecked the exact head and confirmed that its tests reference `Metadata::typed_context` even though #578 is not present, so it is not independently buildable as submitted. Sorry for missing that distinction.

I will reduce the review scope rather than stack more fixes here: first the restricted LLM router, then request-side serving hooks after the context boundary in #583 is settled, and finally typed error/response-model presentation as focused follow-ups. Each change will be rebased onto current `main`, independently buildable, and use one matching author/sign-off identity.

ilexpoon:
The restricted router portion has now been split into #586 as a single focused commit with workspace tests, clippy, and rustdoc passing. I am keeping this PR in draft for the remaining ingress-hook work until the context ownership/API direction in #583 is agreed.

ilexpoon:
Closing this to simplify the review set. The restricted router moved into #586. The remaining hooks depend on the request-context design discussion, while our authentication, model-mapping, and response policy will stay downstream. If a generic hook is still needed after that discussion, I will bring it back as a small, independently testable PR. Thanks for the feedback.

PR #582, opened 2026-08-31
docs(readme): lead with value, algorithms, and benchmarks

itayzit:
## What

Restructures the README around the order a first-time reader needs, and adds the
Terminal-Bench 2.1 results as the value argument.

Section order is now **Maturity → What Is Switchyard → Why Switchyard →
Get Started → Routing Algorithms → Documentation → Community → License**.

- **Maturity is at the top**, carrying the per-component levels added in #552.
- **The two modes are runnable up front** — `cargo install switchyard-server` for
  the proxy, `cargo add --git …` for the embeddable library.
- **"Why Switchyard"** holds the accuracy-vs-cost results: a new theme-aware SVG
  chart plus a table twin, so the numbers are selectable and screen-reader
  accessible.
- **Get Started is a four-step path** ending in wiring a real coding agent
  (`ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL`), where most readers are headed.
  Step 2 writes `routes.toml` with a heredoc so the whole path is copy-paste.
- **The algorithm explainers are standardized.** All seven answer the same
  fields in the same order: Signal, Cost, Use it when, Benchmark, Route type.

## Why

The README is the main education surface for the project, and the previous one
buried the algorithms below the fold and never answered "what is different about
this router?" This puts that answer up top and backs it with numbers.

## Notes for reviewers

**Start with the rendered page, not the diff** — the diff is a near-total rewrite
and reads as noise. Two things only render on GitHub: the mermaid diagram in
*What Is Switchyard*, and the chart in *Why Switchyard*. The chart is a single
theme-aware SVG, so it is worth flipping your GitHub theme to check both.

Then the judgement calls:

- **Benchmark provenance lives in its own section at the bottom**, deliberately
  kept out of the value pitch: blog attribution, the note that those runs used
  NVIDIA-internal inference endpoints, and how to reproduce them. There is no
  pointer to it from *Why Switchyard* — say the word if you want one.
- **Only the escalation profile is checked in.** The staged and classifier
  configs behind two of the chart points are not in the repo, so the Get Started
  TOML is a stage router over the same model pair, *not* the benchmarked config.
  Publishing the other two profiles is a natural follow-up.
- **The advisor-gate row is not a placeholder** — it uses the TB 2.1 numbers
  already documented in `docs/routing_algorithms/advisor_gate_routing.md`.
- **The chart is embedded with markdown image syntax, not `<picture>`.**
  `pyproject.toml` uses this README as the PyPI long description, and PyPI's
  sanitizer drops `<picture>`/`<source>` entirely.
- **Rebased onto main** to pick up the new Maturity block (#552) and the
  `hierarchical` → `composite` rename (#548).

Docs-only; no Python or Rust source touched. Verified by hand: every link,
image, and in-page anchor resolves; the `routes.toml` heredoc parses as valid
TOML; the chart palette passes colorblind-separation and contrast checks against
both GitHub surfaces; the mermaid block parses.

Out of scope, worth a follow-up: `AGENTS.md` still documents the
`switchyard launch` CLI removed in #501, and the seven algorithm doc pages still
have seven different intro shapes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,24 +4,37 @@
 
 # Switchyard
 
-Switchyard is a Rust proxy and library for LLM traffic. It routes requests
-across providers, translates between OpenAI and Anthropic APIs, records
-operational metrics, and provides typed, composable routing algorithms.
+**Switchyard routes each LLM call to the cheapest model that can still do the job. Without changing a line of your agent.**
 
-**Why Switchyard?** Point a coding agent such as Claude Code or Codex at an
-open-source model. Switchyard translates between the OpenAI Chat, Anthropic
-Messages, and OpenAI Responses formats, so the agent keeps speaking its native
-API while the request is served by vLLM, NVIDIA NIM, Ollama, or any
-OpenAI-compatible endpoint. The same proxy can spread traffic across several
-models for A/B benchmarking, apply signal-driven stage routing, or run a custom
-algorithm you write yourself.
+It has three modes:
 
-## Features
+**1. A Rust proxy** — run it in front of the agent you already use:
 
-- **Protocol Translation**: convert between OpenAI Chat, Anthropic Messages, and OpenAI Responses formats
-- **Multi-Backend Routing**: random routing, LLM-as-classifier routing, signal-driven stage-router, or your own algorithm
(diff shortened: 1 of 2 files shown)


ayushag-nv (approved):
Thanks @itayzit

PR #584, opened 2026-09-01
fix(server): bind to localhost by default

aaronjmars:
## Summary by CodeRabbit

* **Security**
  * The server now binds to localhost by default, reducing unintended external exposure.
  * External access requires explicitly configuring a host address.


afourniernv:
Thanks for raising this. This changes the deployment behavior of `switchyard-server`, not just the bind address. It would also affect the checked-in systemd service and normal Docker port publishing.

I've opened #596 so we can decide what the default should be and account for those paths together. I'm going to close this PR for now.

PR #585, opened 2026-09-01
fix(translation): preserve Anthropic tool strictness

### LINKED ISSUE #577: switchyard drops anthropic tools[].strict when translating to openai chat
## summary

an anthropic `/v1/messages` request with `tools[].strict: true` loses that flag
when switchyard translates it to an openai chat backend. name, description, and
parameters survive, `strict` does not. the client still gets http 200 with no
error.

## repro

build switchyard main, tested at `27fc1ce9ff3846760337fe42ab09c28f5b01c807`,
and configure an `openai_chat` target backed by a local capture server. send
this complete Anthropic Messages request five times:

```bash
curl -sS http://127.0.0.1:9004/v1/messages \
  -d '{
    "model": "captured-model",
    "max_tokens": 64,
    "messages": [{"role": "user", "content": "Use the tool."}],
    "tools": [{
      "name": "strict_probe",
      "description": "Verify strictness survives translation.",
      "strict": true,
      "input_schema": {
        "type": "object",
        "properties": {"city": {"type": "string"}},
        "required": ["city"],
        "additionalProperties": false
      }
    }]
  }'
```

expected forwarded tool:

```json
{"type":"function","function":{"name":"strict_probe","strict":true}}
```

actual forwarded tool:

```json
{"type":"function","function":{"name":"strict_probe"}}
```

confirmed 5/5 against a keyless local capture backend. every client call
returned http 200 while `strict` was absent. the same tool sent as a native
openai chat request with `function.strict: true` keeps the field through the
same server and backend, so this is an ingress-side loss, not something openai
chat is unable to express.

the full keyless reproducer, five failure captures, and five control captures
are in the linked Kairo finding:
https://github.com/Atharva-Kanherkar/kairo/pull/22

## root cause

- `decode_anthropic_tools` in `crates/switchyard-translation/src/codecs/anthropic/buffered.rs`
  builds every tool with `strict: None` and never reads the `strict` field from
  the incoming request.
- `encode_openai_tools` in `codecs/openai_chat/buffered.rs` already writes
  `function.strict` when the value is present, so the loss is entirely on the
  decode side.
- `encode_responses_tools` in `codecs/responses/buffered.rs` forwards `strict`
  the same way, and reads from the same tool list `decode_anthropic_tools`
  builds. an anthropic request routed to an openai responses backend is likely
  affected too, but was not verified in this report.

## not a duplicate

issues #467, #452, and #462 are about `response_format` / `output_config.format`,
whole-response json-schema enforcement. this is `tools[].strict`, a separate
field on `ToolDefinition` for per-call argument enforcement. searched issues,
prs, discussions, and code for `strict` and `decode_anthropic_tools`, and found
nothing else covering this.

## impact

this is a silent loss: no diagnostic and http 200 on every run. an app relying
on `strict` to skip its own argument validation can receive schema-violating tool
arguments with no signal that enforcement was dropped.

cpakkamisaac-sae:
## What

Preserve top-level `tools[].strict` when decoding Anthropic Messages requests so explicit strictness reaches OpenAI Chat and Responses targets.

Add regression coverage for `true`, `false`, and omitted strictness across both translated formats.

## Why

The Anthropic decoder previously replaced every tool's strictness with an absent value, silently disabling schema enforcement after translation.

Closes #577.

## Notes for reviewers

Start with `decode_anthropic_tools`; the runtime change is one expression. The request-translation test exercises the complete decoder/encoder path for both destination formats.

Local validation:

- `cargo fmt --all --check`
- `cargo test --locked --offline -p switchyard-translation`
- `cargo test --workspace`
- `.venv/bin/ruff check .`
- `.venv/bin/mypy switchyard`
- Hermetic Python suite: 113 passed, 2 deselected
- Loopback proxy capture confirmed `function.strict: true` in the forwarded request

## Summary by CodeRabbit

- **Bug Fixes**
  - Anthropic tool definitions now correctly preserve the optional `strict` setting during translation.
  - Both enabled and disabled values are retained, while unspecified settings remain absent.

- **Tests**
  - Added coverage confirming `strict` values translate correctly to OpenAI Chat and Responses formats.

### DIFF
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -701,7 +701,7 @@ fn decode_anthropic_tools(value: Option<&Value>) -> Vec<ToolDefinition> {
                     .get("input_schema")
                     .cloned()
                     .unwrap_or_else(|| json!({})),
-                strict: None,
+                strict: tool.get("strict").and_then(Value::as_bool),
             })
         })
         .collect()
--- a/crates/switchyard-translation/tests/request_translation.rs
+++ b/crates/switchyard-translation/tests/request_translation.rs
@@ -654,6 +654,47 @@ fn anthropic_tool_without_name_is_dropped_before_openai_chat() -> TestResult {
     Ok(())
 }
 
+// Verifies Anthropic tool strictness survives translation into both OpenAI formats.
+#[test]
+fn anthropic_tool_strictness_is_preserved_for_openai_formats() -> TestResult {
+    let engine = TranslationEngine::default();
+    let body = json!({
+        "model": "claude-sonnet-4-20250514",
+        "messages": [{"role": "user", "content": "Use a tool."}],
+        "max_tokens": 100,
+        "tools": [
+            {"name": "strict_tool", "input_schema": {}, "strict": true},
+            {"name": "non_strict_tool", "input_schema": {}, "strict": false},
+            {"name": "unspecified_tool", "input_schema": {}}
+        ]
+    });
+
+    let chat = engine
+        .translate_request(
+            WireFormat::AnthropicMessages,
+            WireFormat::OpenAiChat,
+            &body,
+            &TranslationPolicy::default(),
+        )?
+        .body;
+    let responses = engine
+        .translate_request(
+            WireFormat::AnthropicMessages,
+            WireFormat::OpenAiResponses,
+            &body,
+            &TranslationPolicy::default(),
+        )?
+        .body;
+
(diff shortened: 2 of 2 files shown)


ting-hong-shieh (approved):
Verified at `32fe68a`. The change preserves `true` and `false` while leaving omitted strictness absent across OpenAI Chat and Responses. Format, workspace Clippy, and the full translation test suite pass. No blocking findings.

afourniernv (approved):
Tested this through Switchyard and against the live Anthropic API. This fixes the Anthropic to OpenAI direction. The reverse direction still drops `strict`, so I opened #599 to track that separately rather than hold this up.

PR #587, opened 2026-09-01
fix(llm-client): drop the upstream url from transport errors

ryan-lempka:


bhuvan002:
Closing as it's superceeded by #649

PR #590, opened 2026-09-01
fix(translation): decode flat Responses input_file payloads

mrPronin:
## What

`decode_file_source` now reads an `input_file` payload carried directly on the
content block, not only nested under a `file` object.

## Why

The Responses wire puts `file_data`/`filename` directly on the block, while
OpenAI Chat nests them under `file`. The nested shape decoded both `file_id`
and `file_data`, but the flat shape read only `file_id` — so a flat
`file_data` fell through to `FileSource::Raw`.

That is not a cosmetic loss. Chat's raw file encoder maps only Anthropic
`document` blocks and returns `None` for anything else, so the block is then
dropped: a Responses request carrying base64 file content loses the file
entirely en route to a Chat backend, silently and with no diagnostic.

The new test fails on `main` with

```
the file must survive into Chat, not be dropped as unmappable raw
```

and passes with this change.

## Notes for reviewers

Decode path only: one branch in `decode_file_source`, plus one test. A Chat
block cannot reach that branch, having matched the nested one above, so the
existing shapes decode unchanged.

Independent of #530, which fixes the Responses *encoder* for image and file
content and explicitly scoped this reverse-path case out. This applies to
`main` as-is and does not touch `responses/buffered.rs`, so the two do not
conflict and neither blocks the other.

Split out at reviewer request from #566, which was closed as overlapping #530.

- `cargo test --workspace`: 678 passed, 0 failed
- `cargo fmt --all --check`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- Commit carries the DCO sign-off

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with OpenAI Responses-style file inputs.
  * File data and filenames provided directly on file input blocks are now preserved during translation.
  * Existing nested file formats and raw fallback behavior remain supported.

* **Tests**
  * Added regression coverage to verify file content is translated correctly.

### DIFF
--- a/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/openai_chat/buffered.rs
@@ -588,6 +588,11 @@ pub(crate) fn decode_image_source(block: &Map<String, Value>) -> Option<ImageSou
 }
 
 /// Decodes OpenAI file block shapes into normalized file sources.
+///
+/// Accepts both the Chat shape, which nests the payload under a `file` object,
+/// and the Responses shape, which carries `file_id`/`file_data`/`filename`
+/// directly on the block. Returns `FileSource::Raw` when neither shape carries a
+/// field this can normalize.
 pub(crate) fn decode_file_source(block: &Map<String, Value>) -> FileSource {
     if let Some(file) = block.get("file").and_then(Value::as_object) {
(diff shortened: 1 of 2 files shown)


afourniernv (approved):
Tested this locally and through a live Responses-to-Chat request. The codec change is scoped correctly and the file survives as a Chat file part. LGTM.

PR #591, opened 2026-09-01
refactor(libsy): source advisor-gate guards from shared ToolSignals counts

eric-liu-nvidia:
## Summary

Followup committed in the #371 review thread: the advisor gate's request-side guards (`gate_stall_turns`, `gate_min_tool_results`) now read the shared `ToolSignals` extraction instead of bespoke counters — one definition of tool/turn counting across algorithms.

- `ToolSignals` gains two buckets: `tool_result_count` (per `ToolResult` block, including empty-content results, matching the gate's previous counting) and `assistant_turn_count` (`Role::Assistant` messages; documented in contrast to `turn_depth`, which counts all messages regardless of role).
- `AdvisorGate::route_inner` calls `ToolSignals::from_request` once and reads both counts; the bespoke `count_tool_results` / `assistant_turns` helpers in `advisor_gate/turn.rs` are deleted.
- **No behavior change**: the two new gate e2e tests were written first and passed against the old helpers; after the swap, both suites re-ran green with zero test modifications.

## Tests

- `ToolSignals` unit tests pin counting parity on both message shapes: one `ToolResult` per user message (OpenAI-chat normalized `role: tool`) and several blocks batched in one message (Anthropic) — plus empty-content results still counting, and the `assistant_turn_count` vs `turn_depth` contrast.
- Gate e2e: batched result blocks satisfy `gate_min_tool_results` per block; the stall checkpoint counts assistant turns, not messages.

Additive pub-struct fields only (no out-of-crate consumers). The trigger side (buffered-response inspection: `has_tool_use`, `visible_text`) is intentionally untouched and left for a separate followup.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved advisor gating accuracy when processing tool results and assistant responses.
  - Batched messages now count each tool result individually, including empty results.
  - Stall detection now counts assistant turns correctly rather than counting all conversation messages.
  - Threshold-based safeguards now respond more reliably to conversations with multiple tool calls or results.

- **Tests**
  - Added coverage for batched tool results, empty results, and varied conversation formats.

### DIFF
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -242,14 +242,11 @@ pub struct ToolSignals {
     pub pure_bash_streak: u32,
     /// At least one of the last three tool results matched a test-pass pattern.
     pub tests_passed: bool,
-    /// Total `ToolResult` blocks in the conversation, counted per block — an
-    /// Anthropic-style user message batching N results contributes N — and
-    /// including results with empty content, which the text-derived signals
-    /// skip. Guard input for the advisor gate's `gate_min_tool_results`.
+    /// Total `ToolResult` blocks, counted per block (a message batching N
+    /// results contributes N) and including empty-content results.
     pub tool_result_count: u32,
-    /// Messages with `Role::Assistant` — the advisor gate's stall clock.
-    /// Unlike [`ToolSignals::turn_depth`], which counts every message
-    /// regardless of role, this count is stable across wire formats.
+    /// Messages with `Role::Assistant`, unlike [`ToolSignals::turn_depth`],
+    /// which counts every message regardless of role.
     pub assistant_turn_count: u32,
     /// Message-count proxy for turn depth. Wire-format dependent (Anthropic batches
     /// tool results into fewer messages than OpenAI-chat), so gates keyed on it are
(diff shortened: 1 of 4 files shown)


crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -239,6 +242,15 @@ pub struct ToolSignals {
     pub pure_bash_streak: u32,
     /// At least one of the last three tool results matched a test-pass pattern.
     pub tests_passed: bool,
+    /// Total `ToolResult` blocks in the conversation, counted per block — an
```
nachiketb-nvidia: nit: can we have fewer comments?

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -803,6 +830,80 @@ mod tests {
         assert_eq!(sig.write_count, 1);
     }
 
+    #[test]
+    fn tool_result_count_counts_single_block_messages() {
```
nachiketb-nvidia: nit: do we need all these tests? should the ones in crates/libsy/src/algorithms/advisor_gate/tests.rs suffice?

crates/libsy/src/algorithms/util/tool_signals.rs
eric-liu-nvidia: Done

crates/libsy/src/algorithms/util/tool_signals.rs
eric-liu-nvidia: Done

nachiketb-nvidia (approved):

PR #593, opened 2026-09-01
feat(prefill-router): add libsy algorithm wrapper

nachiketb-nvidia:
## What
- Adds a libsy algorithm adapter inside the prefill-router crate.
- Exposes PrefillRouterConfig and PrefillRouterAlgo as part of the crate API.
- Makes PrefillRouterConfig the normal construction path: targets + checkpoint in, algorithm out.
- Keeps the existing PrefillRouter<F> scorer/binder unchanged.
- Keeps algorithm behavior coverage as an integration test under crates/prefill-router/tests.

## Why
This is step 3 of SWITCH-1274 / SWITCH-1280: bring prefill routing into libsy without putting prefill-router internals into switchyard-libsy itself. Since prefill-router is now the owner of this integration, the adapter is compiled directly instead of being hidden behind a Cargo feature.

## How
- PrefillRouterConfig::new(targets, checkpoint).build() creates the algorithm with current TransformersForward defaults.
- The algorithm routes on the latest text user message.
- User-turn affinity is the default behavior for this wrapper, so tool continuations reuse the previous decision instead of rerunning prefill inference.
- Prediction runs through spawn_blocking because the current forward path is synchronous embedded Python/PyTorch work.
- Tests use a doc-hidden fake-forward constructor from an integration test so they do not require a real checkpoint.

## Validation
- cargo test -p prefill-router
- cargo clippy -p prefill-router --all-targets -- -D warnings
- cargo check --workspace

Linear: https://linear.app/nvidia/issue/SWITCH-1280/make-prefill-router-a-libsy-algorithm-via-thin-wrapper

### DIFF
--- a/crates/prefill-router/src/algorithm.rs
+++ b/crates/prefill-router/src/algorithm.rs
@@ -1,0 +1,219 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! libsy adapter for checkpoint-backed prefill routing.
+
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+
+use async_trait::async_trait;
+use libsy::{
+    AffinityRouter, Algorithm, Classifier, Driver, Event, LibsyError, Processor, RoutingOutcome,
+};
+use switchyard_protocol::{ModelId, Request, Role};
+
+use crate::{
+    PrefillForward, PrefillRouter, PrefillRouterError, Result, TransformersForward,
+    TransformersForwardConfig,
+};
+
+const ALGORITHM_NAME: &str = "prefill_router";
+
+/// Configuration for a libsy prefill-router algorithm.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PrefillRouterConfig {
+    /// Ordered routing targets, matching the checkpoint head order.
+    pub targets: Vec<ModelId>,
+    /// Tensor-only router checkpoint path.
+    pub checkpoint: PathBuf,
+    /// Torch device override. Auto-detected when omitted.
+    pub device: Option<String>,
+    /// Optional Hugging Face cache directory.
+    pub cache_dir: Option<PathBuf>,
+    /// Maximum tokenized prompt length.
+    pub max_length: usize,
+    /// Maximum prompts per encoder forward pass.
+    pub batch_size: usize,
+}
+
+impl PrefillRouterConfig {
+    /// Creates a prefill-router config with sensible Transformers and routing defaults.
+    pub fn new(targets: Vec<ModelId>, checkpoint: impl Into<PathBuf>) -> Self {
+        let forward = TransformersForwardConfig::new(checkpoint);
+        Self {
+            targets,
+            checkpoint: forward.checkpoint,
+            device: forward.device,
+            cache_dir: forward.cache_dir,
+            max_length: forward.max_length,
+            batch_size: forward.batch_size,
+        }
+    }
+
+    /// Builds the libsy algorithm described by this config.
+    pub fn build(self) -> Result<PrefillRouterAlgo> {
+        let Self {
+            targets,
+            checkpoint,
+            device,
+            cache_dir,
+            max_length,
+            batch_size,
+        } = self;
+        let forward = TransformersForward::new(TransformersForwardConfig {
+            checkpoint,
+            device,
+            cache_dir,
+            max_length,
+            batch_size,
+        })?;
+        PrefillRouterAlgo::from_forward(targets, forward)
+    }
+}
+
+/// libsy [`Algorithm`] wrapper around a [`PrefillRouter`].
+///
+/// The adapter scores only the latest text user turn. With the default user-turn
+/// affinity, tool continuations reuse the prior decision instead of re-running
+/// prefill inference.
+pub struct PrefillRouterAlgo {
+    router: Arc<Mutex<PrefillRouter<AnyPrefillForward>>>,
+    targets: Vec<ModelId>,
+    default_target: ModelId,
+    affinity: Arc<AffinityRouter>,
+}
+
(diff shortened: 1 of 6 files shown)


crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+            .rev()
+            .filter(|message| message.role == Role::User)
+            .filter_map(|message| message.text_content("\n"))
+            .find(|text| !text.trim().is_empty())
+        {
+            Some(prompt) => {
+                let router = Arc::clone(&self.router);
+                let predictions = tokio::task::spawn_blocking(move || {
+                    router
+                        .lock()
```
grahamking: Here you start a new thread (spawn_blocking) and then block on a mutex. Lots of traffic could mean lots of idle threads.

Ideally the router would not need a lock. Does `predict` change the route itself?

If the lock is necessary change it to be a tokio mutex and acquire before spawning the thread.

crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+/// prefill inference.
+pub struct PrefillRouterAlgo {
+    router: Arc<Mutex<PrefillRouter<AnyPrefillForward>>>,
+    targets: Vec<ModelId>,
+    default_target: ModelId,
+    affinity: Arc<AffinityRouter>,
+}
+
+impl PrefillRouterAlgo {
+    fn from_forward(targets: Vec<ModelId>, forward: impl PrefillForward + 'static) -> Result<Self> {
```
grahamking: The only useful impl seems to be `TransformersForward`. Could we use that directly? Would simplify things.

If that makes testing difficult then I think the trait is OK, but it does add complexity.

crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+/// libsy [`Algorithm`] wrapper around a [`PrefillRouter`].
+///
+/// The adapter scores only the latest text user turn. With the default user-turn
+/// affinity, tool continuations reuse the prior decision instead of re-running
+/// prefill inference.
+pub struct PrefillRouterAlgo {
+    router: Arc<Mutex<PrefillRouter<AnyPrefillForward>>>,
+    targets: Vec<ModelId>,
+    default_target: ModelId,
+    affinity: Arc<AffinityRouter>,
```
grahamking: I've noticed AI loves to `Arc` all-the-things. Do you want to double check `router` and `affinity` and see if Arc is really necessary?

crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+        Ok(Self {
+            router: Arc::new(Mutex::new(router)),
+            targets,
+            default_target,
+            affinity: Arc::new(AffinityRouter::new().with_release_on_user_turn()),
+        })
+    }
+
+    #[doc(hidden)]
+    pub fn from_test_forward(
```
grahamking: `pub(crate)`

crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+            return Err(PrefillRouterError::InvalidConfiguration(
+                "at least one target is required".to_string(),
+            ));
+        };
+        let router = PrefillRouter::new(targets.clone(), AnyPrefillForward(Box::new(forward)))?;
+        Ok(Self {
+            router: Arc::new(Mutex::new(router)),
+            targets,
+            default_target,
+            affinity: Arc::new(AffinityRouter::new().with_release_on_user_turn()),
```
grahamking: I don't entirely understand this one, but Codex says:

> Affinity fails for requests without a session ID
> 
> AffinityRouter::new() has no fallback identity. Ordinary SDK requests may omit session metadata, so decisions are  never retained and every tool continuation reruns prefill inference. This contradicts the advertised “tool continuations reuse the previous decision” behavior. The test conceals this by always setting session_id. Add metadata-less continuation coverage and define the intended fallback behavior

crates/prefill-router/src/algorithm.rs
```diff
@@ -0,0 +1,219 @@
...
+
+use crate::{
+    PrefillForward, PrefillRouter, PrefillRouterError, Result, TransformersForward,
+    TransformersForwardConfig,
+};
+
+const ALGORITHM_NAME: &str = "prefill_router";
+
+/// Configuration for a libsy prefill-router algorithm.
+#[derive(Clone, Debug, Eq, PartialEq)]
```
grahamking: Do you need `Eq` and `PartialEq`?

nachiketb-nvidia: not yet no,

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: The trait is there because I wanted somethign working, and then maybe if we move to, say, a candle based forward, I can try both together. But yeah I can remove it if needed.

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: removed

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: basically saying if there's no session id, we're in trouble, added some simple hashing

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: done

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: yeah :sweat_smile: , its removed now

crates/prefill-router/src/algorithm.rs
nachiketb-nvidia: i moved the lock to before the spawn blocking, need this since the `predict` is running transformers and pyo3 in the back, so we still need the sync here.

grahamking (approved):

PR #600, opened 2026-09-02
fix(translation): preserve OpenAI tool strictness for Anthropic

### LINKED ISSUE #599: translation drops OpenAI tool strictness for Anthropic targets
## summary

an openai chat request with `function.strict: true` loses that field when switchyard translates it to an anthropic messages target.

## repro

send an openai chat request with a function tool that sets `strict: true` through an `anthropic_messages` target. the captured anthropic request contains the tool name, description, and input schema, but no `strict` field.

i also forwarded the captured request to the live anthropic messages api. it returned http 200, but the call was running without strict schema enforcement because the field had already been removed.

## root cause

the openai decoder stores the value in `ToolDefinition.strict`, but `encode_anthropic_tools` does not write it into the outbound anthropic tool definition.

## expected

preserve `true` and `false`, and leave the field out when it was not specified.

found while reviewing #585, which fixes the same loss in the opposite anthropic-to-openai direction.

afourniernv:
## What

Preserve `ToolDefinition.strict` when Switchyard encodes an Anthropic Messages tool definition.

The field is written for `true` and `false`, and left out when it was not set.

## Why

OpenAI Chat already decodes function strictness into the shared request type, but the Anthropic encoder was dropping it. That silently turned a strict tool into a normal tool before the upstream request was sent.

This is the reverse direction of #585.

Closes #599.

## Testing

- `cargo fmt --all --check`
- Live OpenAI Chat to Switchyard to Anthropic capture: the outbound tool contained `strict: true` and Anthropic returned HTTP 200 with the forced tool call.

No public API changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Tool definitions translated to Anthropic format now preserve explicitly configured strictness settings.
  - Unspecified strictness remains unchanged during translation.

- **Tests**
  - Added coverage for strict, non-strict, and unspecified tool configurations.

### DIFF
--- a/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/anthropic/buffered.rs
@@ -1044,11 +1044,15 @@ fn encode_anthropic_tools(tools: &[ToolDefinition]) -> Value {
         tools
             .iter()
             .map(|tool| {
-                json!({
+                let mut item = json!({
                     "name": tool.name,
                     "description": tool.description.clone().unwrap_or_default(),
                     "input_schema": tool.parameters,
-                })
+                });
+                if let Some(strict) = tool.strict {
+                    item["strict"] = Value::Bool(strict);
+                }
+                item
             })
             .collect(),
     )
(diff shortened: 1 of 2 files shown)


bhuvan002 (approved):
LGTM, thank you!

PR #605, opened 2026-09-02
refactor(libsy): decompose advisor_gate into signal/trigger/budget components

eric-liu-nvidia:
Decomposes the `advisor_gate` algorithm into libsy's standard building blocks — a signal processor (conversation counts from `Event::Request`, the generated turn's tool-use/text from `Event::ModelResponse`), a trigger classifier that runs after the executor call, and a reserve/refund review budget — leaving `AdvisorGate` as a thin orchestrator. Follow-up to the algorithm-structure review thread on #371.

This makes the trigger logic testable in isolation and lets future gate variants reuse the same components, with zero behavior change: the trigger stays fused to the buffered response, budget reservation stays atomic with the trigger firing, and the REDO and audit paths are untouched.

**Validation:** all existing advisor_gate unit tests and server e2e tests pass unmodified, plus new component-level tests; `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, rustdoc with `-D warnings`, `ruff`, and `pytest` are all green.

### DIFF
--- a/crates/libsy/src/algorithms/advisor_gate/budget.rs
+++ b/crates/libsy/src/algorithms/advisor_gate/budget.rs
@@ -2,15 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! The gate's mutable ledger: the per-scope review budget behind a
-//! reserve/refund interface, plus the per-conversation stall latch.
-//!
-//! Reservation is deliberately not split from the consult outcome: failed and
-//! unparseable consults refund their reservation (a pre-routing reservation
-//! would leak budget on advisor outages — whether to refund is only knowable
-//! after the consult), and reserving before the consult await keeps
-//! concurrent same-scope requests from overdrawing `max_reviews`. This is the
-//! gate's only mutable shared state, behind one lock, next to the code that
-//! refunds it.
+//! reserve/refund interface, plus the per-conversation stall latch — the
+//! gate's only shared mutable state, behind one lock.
 
 use std::collections::{HashMap, HashSet};
 use std::hash::{DefaultHasher, Hash, Hasher};
@@ -20,18 +13,14 @@ use switchyard_protocol::{Request, Role};
 
 use super::BENCH_SESSION_HEADER;
 
-/// Failed consults tolerated per scope before the gate stops consulting.
-/// Failures refund the review budget — a transient advisor error must not
-/// silently exhaust `max_reviews` with zero real reviews — so this separate
-/// cap is what bounds per-turn consult latency against a down advisor.
+/// Failed consults tolerated per scope before the gate stops consulting;
+/// failures refund the budget, so this cap is what bounds a down advisor.
 const MAX_FAILED_CONSULTS: u32 = 3;
 /// Bounds tracked budget scopes and stall keys; a scope dropped at the bound
 /// re-arms like a process restart (rare, harmless).
 const MAX_TRACKED_SCOPES: usize = 1_024;
 
-/// Review budget scope, in precedence order: the benchmark harness header
-/// (exact evaluation identity, sub-agents included), then the host-resolved
-/// session id, then one instance-wide scope for headerless clients.
+/// Review budget scope; resolution and precedence live in [`budget_scope`].
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub(super) enum ScopeKey {
     Instance,
@@ -47,17 +36,14 @@ struct ScopeState {
     exhaustion_logged: bool,
 }
 
-/// Shared mutable gate state; every access is a short critical section and
-/// the lock is never held across an await.
+/// Shared mutable gate state; locked briefly, never across an await.
 #[derive(Default)]
 struct GateState {
     scopes: HashMap<ScopeKey, ScopeState>,
     stall_fired: HashSet<u64>,
 }
 
-/// The review budget: at most `max_reviews` reserved reviews per scope, a
-/// failure cap on refunded consults, and the stall checkpoint's
-/// once-per-conversation latch.
+/// Per-scope review budget plus the stall checkpoint's per-conversation latch.
 pub(super) struct ReviewBudget {
     max_reviews: u32,
     state: Mutex<GateState>,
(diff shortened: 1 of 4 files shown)


crates/libsy/src/algorithms/advisor_gate/budget.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Latches the stall checkpoint for this conversation key.
+    pub(super) fn mark_stall_fired(&self, key: u64) {
+        let mut state = self.state.lock();
+        if state.stall_fired.len() >= MAX_TRACKED_SCOPES {
+            let drop = state.stall_fired.iter().next().copied();
+            if let Some(key) = drop {
+                state.stall_fired.remove(&key);
+            }
+        }
+        state.stall_fired.insert(key);
```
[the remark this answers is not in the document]
eric-liu-nvidia: Applied in 1086862c — `stall_already_fired`/`mark_stall_fired` are now a single atomic `try_mark_stall_fired` (check + insert in one critical section, same `MAX_TRACKED_SCOPES` eviction), and the caller uses its return value, so of two concurrent stalls on one conversation exactly one claims the checkpoint and the loser's turn passes through unreviewed. The latch is still only attempted when no terminal trigger fired, preserving the no-latch-on-simultaneous-trigger invariant.

Worth noting the blast radius was already bounded: `try_reserve` is atomic per scope, so this race could never overdraw `max_reviews` — at the default `max_reviews = 1` the losing stall simply failed the reservation; only `max_reviews >= 2` could spend an extra review on one checkpoint. All existing behavior tests pass unmodified.

crates/libsy/src/algorithms/advisor_gate/budget.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub(super) fn try_reserve(&self, scope: &ScopeKey) -> bool {
+        let mut state = self.state.lock();
+        if state.scopes.len() >= MAX_TRACKED_SCOPES && !state.scopes.contains_key(scope) {
+            let evict = state
+                .scopes
+                .keys()
+                .find(|key| **key != ScopeKey::Instance)
+                .cloned();
+            if let Some(key) = evict {
+                state.scopes.remove(&key);
```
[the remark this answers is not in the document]
eric-liu-nvidia: Verified against the current code, and keeping as-is: this is the pre-existing ledger behavior from #371 moved verbatim, and the trade-off is documented on `MAX_TRACKED_SCOPES` — a scope dropped at the 1,024-scope bound "re-arms like a process restart (rare, harmless)". Reaching it requires more than 1,024 concurrently tracked scopes *and* the arbitrarily evicted scope to have a consult in flight at that instant; the worst case is one extra advisor consult, never a lost or double-served client turn.

The proposed admission rule inverts the trade-off: rejecting reservations once tracked scopes are saturated silently skips reviews for *new* scopes under load, which is the wrong default for this gate. This PR is a behavior-preserving decomposition, so the ledger semantics stay bit-identical; a reservation-aware eviction policy is worth its own change if the bound ever becomes reachable in practice.

crates/libsy/src/algorithms/advisor_gate/signals.rs
ayushag-nv, line 1: @eric-liu-nvidia should we move this to common tool signal file ? or rename that. Seems like Turn Signal might be of use to other algos as well or might be extended to more richer signals.
what do you think ??

eric-liu-nvidia: main concern is no other algorithm can consume these signals today.

And this signal is more request side tool extraction. In the future, we can split this into two different categories. One response and one request.

crates/libsy/src/algorithms/advisor_gate/budget.rs
```diff
@@ -0,0 +1,239 @@
...
+//! The gate's mutable ledger: the per-scope review budget behind a
+//! reserve/refund interface, plus the per-conversation stall latch.
+//!
+//! Reservation is deliberately not split from the consult outcome: failed and
+//! unparseable consults refund their reservation (a pre-routing reservation
+//! would leak budget on advisor outages — whether to refund is only knowable
+//! after the consult), and reserving before the consult await keeps
+//! concurrent same-scope requests from overdrawing `max_reviews`. This is the
+//! gate's only mutable shared state, behind one lock, next to the code that
+//! refunds it.
```
ayushag-nv: @eric-liu-nvidia can you reduce these commented lines to a few may be. This and other files

crates/libsy/src/algorithms/advisor_gate/budget.rs
ayushag-nv, line 91: Could we avoid evicting an arbitrary scope here? If that scope has a review in flight, another request can recreate it and reserve again, so `max_reviews` is no longer enforced. A failed in-flight review can also reinsert the entry after eviction. Returning `false` for unseen scopes while the map is full seems safer; `session_final` can free entries normally.

ayushag-nv (approved):
looks good overall.

crates/libsy/src/algorithms/advisor_gate/budget.rs
eric-liu-nvidia: Done in ed520cad — module headers across all four files condensed (net −43 comment lines; this budget.rs header is 10 → 3 lines). Kept only what the code can't say itself: the reserve-before-the-consult-await concurrency rule and the refund-on-failed-consult contract.

crates/libsy/src/algorithms/advisor_gate/budget.rs
eric-liu-nvidia: Both failure modes you describe are real, and both are bounded and pre-existing — the ledger moved verbatim from #371: an evicted-mid-consult scope can re-arm (worst case one extra advisor consult; never a dropped or double-served client turn), and yes, a failed consult's refund can transiently resurrect the evicted entry since `refund_failure` has no capacity check.

The reason I'd rather not take return-false-for-new-scopes-when-full: at saturation it silently skips reviews for every *new* conversation, and silent degradation under load is the exact failure mode this gate is built to avoid (`fail_open = false` exists for the same reason) — a skipped review is invisible, an extra consult is just cost.

The fix that addresses both of your scenarios without that trade-off is **reservation-aware eviction**: track in-flight reservations per scope, make only reservation-free scopes evictable (deny admission only in the degenerate all-1,024-mid-consult case), and give `refund_failure` the same capacity rule so a refund can't resurrect past the bound. That needs an `in_flight` counter plus a success-settlement hook — real ledger semantics — and this PR is a behavior-preserving decomposition, so I've filed **SWITCH-1388** to do it as an immediate follow-up. If you'd rather it land in this PR, say the word and I'll add it here instead.

PR #606, opened 2026-09-02
feat(stage-router): support custom tool semantics

### LINKED ISSUE #333: [feature] Extensible stage router known tools
## Problem
The stage router currently derives progress and difficulty signals from a built-in vocabulary oriented primarily toward coding agents. It recognizes tools such as file reads, edits, writes, shell commands tests, and planning operations.

Agent harnesses in other domains often expose different tools. For example, Tau3 conversation tasks use tools such as:

  - KB_search and get_user_information_by_name for observation
  - log_verification and send_payment_request for mutation
  - send_message_to_user and start_conversation for protocol progress
  - end_conversation for completion

Because the stage router does not recognize these names, they are generally classified as unknown tool activity. The router therefore misses evidence that the agent is observing state, making changes, or progressing through a task.

  This has two practical impacts:

  1. With a capable-first policy, ambiguous turns disproportionately route to the expensive model.
  2. With an efficient-first policy and an LLM fallback, ambiguous turns invoke the fallback judge unnecessarily.

Any non-coding harness with domain-specific tool names—including customer-service, browser, database, and computer-use benchmarks—can lack the signals expected by the current stage-router vocabulary.

## Proposed solution

Allow users to extend the stage router’s known tool semantics through additive route configuration.

```
  [routes.router]
  id = "switchyard/router"
  type = "stage_router"
  capable_target = "capable"
  efficient_target = "efficient"
  picker = "efficient_first"
  confidence_threshold = 0.3

  [routes.router.tool_semantics]
  observe = [
    "KB_search",
    "get_user_information_by_name",
    "get_current_time",
    "get_credit_card_accounts_by_user",
  ]

  mutate = [
    "log_verification",
    "send_payment_request",
    "transfer_to_human_agents",
  ]

  progress = [
    "start_conversation",
    "send_message_to_user",
    "unlock_discoverable_agent_tool",
    "give_discoverable_user_tool",
  ]

  complete = [
    "end_conversation",
  ]
```

The categories would map into existing stage-router signals:
  - observe: count as read/investigation operations.
  - mutate: count as write/production operations.
  - progress: indicate forward protocol progress without pretending that a read or mutation occurred. This should suppress false stall signals.
  - complete: indicate terminal progress and remain separately identifiable for future completion-specific routing policy.

The extension should have the following properties:

  - Additive: configured names extend the built-in vocabulary rather than replacing it.
  - Case-insensitive exact-name matching by default.
  - Built-in semantics retain precedence if a configured name conflicts with a packaged tool.
  - Empty or absent configuration preserves current behavior.
  - The configuration is scoped to an individual route.

Library-mode construction could expose the same concept as a ToolSemantics value passed into the stage-router configuration:

### Rely exclusively on the LLM fallback judge

This works functionally but adds latency, cost, and another source of nondeterminism to every ambiguous turn. The fallback judge remains useful for genuinely ambiguous turns, but it should not compensate for tool vocabulary the user can describe deterministically.

 ### Attach semantic hints to tool definitions

Tool schemas could carry routing metadata such as observe, mutate, progress, or complete, allowing semantics to travel with the tool instead of being duplicated in route configuration. This is attractive for frameworks that control tool registration, but tool-schema metadata is not consistently preserved across providers, MCP adapters, and agent harnesses. Route-level configuration is more dependable at the Switchyard boundary and also works when upstream tool definitions cannot be modified.

  ### Improve unknown-tool handling heuristically

The router could treat successful use of a previously unknown tool as evidence of progress, while interpreting repeated calls to the same tool with similar arguments or results as evidence of stalling. This would generalize to new tools without explicit configuration and could be a better hueristic than treating all unknown tools identically. However, repetition is not always failure—pagination, polling, iterative search, and multi-record mutations may legitimately reuse a tool—so explicit semantics remain more predictable and auditable; improved unknown-tool heuristics could complement them as a fallback.

## Scope notes

  - This is an extension of the existing stage-router request-side signal extraction, rather than a new chain role.
  - The primary affected component is the existing request processor that extracts tool-result and progress signals before stage-router tier selection.
  - Server configuration parsing and validation would need to accept the new tool_semantics block.
  - Library-mode stage-router construction should accept the equivalent configuration.
  - It should not require changes to LLMBackend, ResponseProcessor, or ResponseTranslator.
  - It does not need to touch a public API listed in switchyard/__init__.py.__all__ unless ToolSemantics is intentionally promoted as a public library-mode type. A plain mapping accepted by the existing stage-
    router constructor could avoid adding a new public symbol.
  - Backward compatibility should be straightforward:
      - An absent or empty tool_semantics block preserves current behavior.
      - Existing built-in classifications retain precedence.
      - Existing route configuration remains valid.
      - Unknown tools continue to behave as they do today.

  - Configuration validation should reject a tool name appearing in multiple user-defined categories, or document deterministic precedence if overlap is allowed.

  ## Additional context

The issue arose while evaluating Switchyard as the model endpoint for the LangChain Deep Agents unified evaluation suite, particularly its Tau3 conversation tasks.

slopp:
Closes #333.

## Summary

This PR makes stage-router tool semantics extensible without coupling the router to any one agent framework or domain.

It is intentionally split into three commits:

1. `refactor(stage-router): generalize tool activity categories` replaces the internal coding-specific category enum with semantic activity (`observe`, `mutate`, `plan`, `unknown`) while preserving the existing built-in vocabulary and public counters.
2. `feat(stage-router): configure custom tool semantics` adds route-scoped exact-name mappings for `observe`, `mutate`, `plan`, and neutral `new` activity.
3. `docs(stage-router): document custom tool semantics` updates the canonical schema, server and routing guides, nested-route documentation, and Python typing facade.

`unknown` remains the fallback for unmatched tools. `new` represents forward activity that suppresses false spinning/exploring signals without otherwise biasing the capable/efficient score.
The issue's progress and terminal/complete tool examples intentionally both map to `new`: stage scoring needs to know that activity advanced, while terminal control flow remains the agent framework's responsibility.

## Configuration

```toml
[routes.stage.tool_semantics]
observe = ["KB_search", "get_customer_by_phone"]
mutate = ["send_payment_request", "update_inventory"]
plan = ["create_research_plan"]
new = ["start_conversation", "send_message_to_user"]
```

Mappings are:

- additive to the built-in coding-agent vocabulary;
- route-scoped;
- exact-name and ASCII case-insensitive;
- available in standalone and composite stage routes and the Python binding.

Configuration loading rejects empty names, duplicate names across categories, unknown category keys, and attempts to reclassify a built-in tool. Argument-aware wrapper tools, inferred semantics, and online/learned rules remain out of scope.

## Compatibility

With no `tool_semantics` section, routing behavior is unchanged. Existing write/edit/read/plan counters and built-in Bash command inference remain intact.

## Tests

- `cargo fmt --all --check`
- `cargo test --workspace`
- `uv run ruff check .`
- `uv run mypy switchyard`
- credential-scrubbed `uv run pytest tests/ -v -m "not integration"` — 117 passed, 2 deselected
- `cd docs && make publish` — strict MkDocs build passed

New coverage includes:

- category matching and validation in libsy;
- neutral `new` activity suppressing false stall dimensions without changing the score;
- standalone/composite TOML deserialization;
- Python binding acceptance and unknown-category rejection;
- an HTTP server integration test proving a configured domain mutation changes the selected tier.

## LangChain Deep Agents Unified Evals evidence

The evaluation uses the issue's concrete integration boundary rather than a synthetic-only test:

- LangChain Deep Agents Unified Evals Lite: 15 autonomous, 11 Tau3 conversation, and 10 context-retrieval tasks;
- the branch-built `switchyard-server` runs as a sidecar in each Harbor sandbox;
- efficient-first stage routing at threshold `0.3`, window `3`, with GPT-5.6 Sol / GLM 5.2 and Gemini 3.1 Flash Lite fallback;
- the Tau3 tool map classifies progress and terminal protocol tools as neutral `new`;
- pass@1, concurrency 3, with per-task routing and token telemetry;
- portable Bullseye build verified with a GLIBC 2.30 ceiling.

The exact tool mapping applied in that run was:

```toml
[routes.router.tool_semantics]
observe = [
  "ls",
  "grep",
  "glob",
  "KB_search",
  "get_user_information_by_name",
  "get_current_time",
  "get_credit_card_accounts_by_user",
  "get_details_by_id",
  "get_credit_card_transactions_by_user",
  "get_bills_for_customer",
  "get_customer_by_phone",
  "get_assistant_tool_schemas",
]
mutate = [
  "edit_file",
  "log_verification",
  "send_payment_request",
  "resume_line",
  "transfer_to_human_agents",
]
new = [
  "execute",
  "task",
  "configure_run",
  "start_conversation",
  "send_message_to_user",
  "unlock_discoverable_agent_tool",
  "call_discoverable_agent_tool",
  "give_discoverable_user_tool",
  "end_conversation",
]
```

`read_file` and `write_file` are deliberately absent because the built-in vocabulary already recognizes them. `execute` is neutral because its arguments may read, mutate, or run tests; argument-aware wrapper classification is a separate future feature.

### Results and scope

There are two complementary sources of external evidence:

| Run | Scope | Reward | Sol / GLM calls | Strong-call rate | Judge requests |
| --- | --- | ---: | ---: | ---: | ---: |
| Archived original router, no custom semantics | Full 36 tasks | 17/36 | 591 / 172 | 77.5% | 151 |
| This PR's native `tool_semantics` implementation | 8 autonomous tasks completed | 2/8 | Per-task traces collected | n/a for partial run | Per-task traces collected |

The branch-native replay exercised the exact TOML above against real Harbor
sandboxes. Eight autonomous tasks completed without harness exceptions before the
evaluation host was retired; the passing tasks were the SWE-smith OAuth1 repair and
the non-SWE OmniMath problem. This partial run is included as integration and
routing evidence, not as a new full-suite quality comparison.

The non-SWE OmniMath task passed with 4 Sol and 2 GLM calls in the PR replay. Its
agent used the configured neutral `task` wrapper, demonstrating that framework
protocol activity can be recognized without falsely counting it as repeated
observation or mutation. The archived no-PR OmniMath recovery trace also passed,
but used 19 Sol and 3 GLM calls. These are independent samples, so the difference
is descriptive rather than causal.

### Original-router counterexample: SWE-smith OAuth1

The closest archived no-PR comparison is the same
`swesmith-fix-oauth1-header-params` task under the same efficient-first + judge
policy. Both independent pass@1 rollouts passed, so this is routing evidence rather
than a claim that one rollout caused a quality improvement.

| Rollout | Sol calls | GLM calls | Behavior after the edit phase |
| --- | ---: | ---: | --- |
| Original stage router | 13 | 4 | Stayed on Sol through the remaining turns because `edit_file` was unknown |
| This PR + the mapping above | 9 | 7 | Returned turns 12–14 to GLM after two configured mutations and neutral `execute` activity |

In the PR rollout, the initial observation/exploration sequence moved from GLM to
Sol at turn 5 (deterministic confidence `0.462`). Sol handled repository search and
the two `edit_file` calls. Those configured mutations then produced enough
production evidence to de-escalate to GLM at turn 12 (confidence `0.321`), remain
there at turn 13 (`0.462`), and let the judge retain GLM at turn 14 after neutral
`execute` activity. A later test/error result escalated back to Sol at turn 15
(`0.462`), and the task passed. The archived original-router trace had the same
first four GLM turns, but stayed on Sol from turn 5 through completion; without
custom semantics it could not recognize the domain's edit boundary.

### Archived no-PR suite context

The full archived 36-task no-tool-semantics run provides context for the non-SWE tasks:
it scored 17/36 overall (autonomous 2/15, conversation 5/11, context 10/10) and
routed 591 calls to Sol versus 172 to GLM, a 77.5% strong-model call rate. Its only
non-SWE autonomous pass was OmniMath. Endpoint variance and different model samples
mean these comparisons describe observed routing behavior, not controlled quality
attribution.

## Summary by CodeRabbit

* **New Features**
  * Added configurable tool-activity semantics for stage and composite routing, including observe, mutate, plan, and new-activity categories.
  * Added Python support for supplying custom tool mappings, with validation for unknown categories.
  * Tool activity now contributes to routing decisions across broader agent workflows.

* **Bug Fixes**
  * Recent unclassified activity no longer incorrectly reports stalled behavior.

* **Documentation**
  * Expanded routing configuration guidance, matching rules, and supported tool-activity signals.

* **Tests**
  * Added coverage for custom mutation mappings, end-to-end routing, validation, and stall detection.

### DIFF
--- a/crates/libsy/src/algorithms/util/stage.rs
+++ b/crates/libsy/src/algorithms/util/stage.rs
@@ -257,9 +257,9 @@ impl ScoreResult {
 pub struct CodingAgentDimensions {
     /// Windowed max error severity in `[0, 1]`.
     pub severity: f64,
-    /// `1.0` when a deep turn has no reads, plans, writes, or edits (pure churn).
+    /// `1.0` when a deep turn has no recognized observation, mutation, plan, or new activity.
     pub spinning: f64,
-    /// `1.0` when a deep turn reads/plans but does not write or edit.
+    /// `1.0` when a deep turn observes/plans but does not mutate or show new activity.
     pub exploring: f64,
     /// Fraction of recent tool ops that produced code (writes + edits).
     pub production_intensity: f64,
@@ -302,10 +302,11 @@ pub fn dimensions_from_signal(signal: &ToolSignals) -> CodingAgentDimensions {
     let deep_enough = signal.turn_depth >= STALL_MIN_TURN_DEPTH;
     let no_production = signal.recent_write_count == 0 && signal.recent_edit_count == 0;
     let investigating = signal.recent_read_count >= 1 || signal.recent_todowrite_count >= 1;
+    let new_activity = signal.recent_new_count >= 1;
     // spinning vs exploring partition the "not producing" case by investigative
     // activity, so at most one fires — no double-counting on the production axis.
-    let spinning = deep_enough && no_production && !investigating;
-    let exploring = deep_enough && no_production && investigating;
+    let spinning = deep_enough && no_production && !investigating && !new_activity;
+    let exploring = deep_enough && no_production && investigating && !new_activity;
 
     CodingAgentDimensions {
         severity: f64::from(signal.severity),
@@ -744,6 +745,21 @@ mod tests {
             }
             other => panic!("expected consult-classifier, got {other:?}"),
         }
+    }
+
+    #[test]
+    fn new_activity_suppresses_stall_dimensions_without_biasing_the_score() {
+        let signal = ToolSignals {
+            turn_depth: STALL_MIN_TURN_DEPTH,
+            recent_new_count: 1,
+            ..Default::default()
+        };
+
+        let dimensions = dimensions_from_signal(&signal);
+
+        assert_eq!(dimensions.spinning, 0.0);
+        assert_eq!(dimensions.exploring, 0.0);
+        assert_eq!(score_signal(&signal).score, 0.0);
     }
 
     // ─── StageClassifier ─────────────────────────────────────────────────
--- a/crates/libsy/src/algorithms/util/stage.rs
+++ b/crates/libsy/src/algorithms/util/stage.rs
@@ -762,6 +762,42 @@ mod tests {
         assert_eq!(score_signal(&signal).score, 0.0);
     }
 
+    #[test]
+    fn old_new_activity_does_not_suppress_a_current_stall() {
+        let signal = ToolSignals {
+            turn_depth: STALL_MIN_TURN_DEPTH,
+            new_count: 1,
+            recent_new_count: 0,
+            ..Default::default()
+        };
+
+        let dimensions = dimensions_from_signal(&signal);
+
+        assert_eq!(dimensions.spinning, 1.0);
+        assert_eq!(dimensions.exploring, 0.0);
+        assert!(score_signal(&signal).score > 0.0);
+    }
+
+    #[test]
+    fn new_activity_does_not_dilute_existing_production() {
+        let baseline = ToolSignals {
+            turn_depth: STALL_MIN_TURN_DEPTH,
+            recent_write_count: 1,
+            ..Default::default()
+        };
+        let with_new = ToolSignals {
+            recent_new_count: 2,
+            new_count: 2,
+            ..baseline.clone()
+        };
+
+        assert_eq!(
+            dimensions_from_signal(&with_new).production_intensity,
+            dimensions_from_signal(&baseline).production_intensity
+        );
+        assert_eq!(score_signal(&with_new), score_signal(&baseline));
+    }
+
     // ─── StageClassifier ─────────────────────────────────────────────────
 
     /// Tiers named the way a deployment would name them.
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -432,7 +432,8 @@ fn classify_tool_call_with_semantics(
     command: Option<&str>,
     semantics: &ToolSemantics,
 ) -> ToolSemantic {
-    let lower = name.to_lowercase();
+    // Built-in names and Bash command inference take precedence over route-scoped mappings.
+    let lower = name.to_ascii_lowercase();
     if WRITE_TOOL_NAMES.contains(&lower.as_str()) {
         return ToolSemantic::Mutate(MutationKind::Write);
     }
@@ -1419,6 +1420,23 @@ mod tests {
         assert_eq!(signal.new_count, 1);
         assert_eq!(signal.recent_new_count, 1);
         assert_eq!(signal.pure_bash_streak, 0);
+    }
+
+    #[test]
+    fn configured_tool_semantics_only_fold_ascii_case() {
+        let semantics = ToolSemantics {
+            observe: vec!["kb_search".to_string()],
+            ..Default::default()
+        };
+
+        assert_eq!(
+            classify_tool_call_with_semantics("KB_SEARCH", None, &semantics),
+            ToolSemantic::Observe
+        );
+        assert_eq!(
+            classify_tool_call_with_semantics("KB_SEARCH", None, &semantics),
+            ToolSemantic::Unknown
+        );
     }
 
     #[test]
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -278,12 +278,18 @@ struct ObservedToolCall {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum ToolCategory {
+enum MutationKind {
     Write,
     Edit,
-    Read,
+}
+
+/// Domain-neutral meaning assigned to an observed tool call.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ToolSemantic {
+    Mutate(MutationKind),
+    Observe,
     Plan,
-    Other,
+    Unknown,
 }
 
 /// Request-side processor that extracts tool-result signals from each request
@@ -314,38 +320,38 @@ impl Processor<State> for ToolSignalProcessor {
     }
 }
 
-fn classify_tool_call(name: &str, command: Option<&str>) -> ToolCategory {
+fn classify_tool_call(name: &str, command: Option<&str>) -> ToolSemantic {
     let lower = name.to_lowercase();
     if WRITE_TOOL_NAMES.contains(&lower.as_str()) {
-        return ToolCategory::Write;
+        return ToolSemantic::Mutate(MutationKind::Write);
     }
     if EDIT_TOOL_NAMES.contains(&lower.as_str()) {
-        return ToolCategory::Edit;
+        return ToolSemantic::Mutate(MutationKind::Edit);
     }
     if READ_TOOL_NAMES.contains(&lower.as_str()) {
-        return ToolCategory::Read;
+        return ToolSemantic::Observe;
     }
     if PLAN_TOOL_NAMES.contains(&lower.as_str()) {
-        return ToolCategory::Plan;
+        return ToolSemantic::Plan;
     }
     if BASH_TOOL_NAMES.contains(&lower.as_str())
         && let Some(cmd) = command
     {
         // Write/edit redirection trumps read-like operands.
         if BASH_WRITE_PATTERNS.iter().any(|p| cmd.contains(p)) {
-            return ToolCategory::Write;
+            return ToolSemantic::Mutate(MutationKind::Write);
         }
         if cmd.contains("python") && PYTHON_WRITE_PATTERNS.iter().any(|p| cmd.contains(p)) {
-            return ToolCategory::Write;
+            return ToolSemantic::Mutate(MutationKind::Write);
         }
         if BASH_EDIT_PATTERNS.iter().any(|p| cmd.contains(p)) {
-            return ToolCategory::Edit;
+            return ToolSemantic::Mutate(MutationKind::Edit);
         }
         if BASH_READ_PATTERNS.iter().any(|p| cmd.contains(p)) {
-            return ToolCategory::Read;
+            return ToolSemantic::Observe;
         }
     }
-    ToolCategory::Other
+    ToolSemantic::Unknown
 }
 
 // ─── extraction entry point ───────────────────────────────────────────────────
@@ -474,7 +480,7 @@ fn build_signal(
     let no_error_streak = compute_no_error_streak(&tool_texts);
 
     // Single pass: cumulative + sliding-window counters together. Also tracks
-    // the trailing pure-bash streak (consecutive `Other`-category calls back
+    // the trailing pure-bash streak (consecutive `Unknown` calls back
     // from the end) — the build-pit proxy.
     let recent_start = tool_calls.len().saturating_sub(recent_window);
     let mut write_count = 0u32;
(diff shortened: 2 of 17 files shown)


crates/switchyard-py/src/libsy_bindings.rs
```diff
@@ -793,6 +794,7 @@ fn stage_router_algorithm(
     only_on_wrong_signal_escalation: bool,
     capable_system_prompt: Option<String>,
     efficient_system_prompt: Option<String>,
+    tool_semantics: Option<HashMap<String, Vec<String>>>,
```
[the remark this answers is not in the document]
slopp: Fixed in 0a9b1058 by narrowing the public Python facade annotation to `dict[str, Sequence[str]]`, matching the PyO3 `HashMap` extraction contract while preserving category validation.

crates/switchyard-server/README.md
```diff
@@ -115,8 +120,10 @@ fallback produced while the judge was unreachable. `message_hash_fallback` keys
 content rather than a session id, so unrelated callers sending identical text share one
 assignment.
 
-A `stage_router` route scores tool-result and agent-progress signals from recent turns to pick a
-tier per turn, without an extra classifier call on every turn. `capable_target`,
+A `stage_router` route scores tool-result and activity signals from recent turns to pick a
+tier per turn, without an extra classifier call on every turn. Domain-specific exact tool names
+can extend its built-in coding vocabulary through `tool_semantics.observe`,
+`tool_semantics.mutate`, `tool_semantics.plan`, and neutral `tool_semantics.new`. `capable_target`,
```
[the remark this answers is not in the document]
slopp: Fixed in 0a9b1058. The server README and both standalone/composite schema tables now state exact ASCII case-insensitive matching for every semantic list.

docs/routing_algorithms/composite_routing.md
```diff
@@ -33,6 +33,11 @@ classify_trigger = "user_turn"
 capable_target = "strong"
 efficient_target = "weak"
 confidence_threshold = 0.5
+
+[routes.switchyard.stage.tool_semantics]
+observe = ["lookup_customer"]
+mutate = ["update_inventory"]
+new = ["send_message"]
```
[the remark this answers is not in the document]
slopp: Fixed in 0a9b1058 by adding `plan = ["create_research_plan"]` to the composite example.

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -307,45 +412,65 @@ impl Default for ToolSignalProcessor {
...
+fn classify_tool_call(name: &str, command: Option<&str>) -> ToolSemantic {
+    classify_tool_call_with_semantics(name, command, &ToolSemantics::default())
+}
+
+fn classify_tool_call_with_semantics(
+    name: &str,
+    command: Option<&str>,
+    semantics: &ToolSemantics,
+) -> ToolSemantic {
     let lower = name.to_lowercase();
```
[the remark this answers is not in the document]
slopp: Fixed in 0a9b1058: custom lookup now uses `to_ascii_lowercase()`, with a Kelvin-sign regression test proving non-ASCII characters remain distinct. I also documented that built-ins and Bash inference take precedence.

slopp:
Addressed the four inline findings and the three maintainability suggestions in `0a9b1058`:

- use ASCII-only normalization, with a non-ASCII regression test;
- align the Python annotation with PyO3 dict extraction;
- document matching consistently and complete the composite example;
- document precedence and the HTTP integration contract;
- make the Python binding test exercise actual stage scoring.

For the linked-issue warning: progress and terminal/complete tool names intentionally map to neutral `new`. Both represent forward activity for stage scoring and suppress false stall signals without biasing the tier; terminal control flow remains the agent framework responsibility. `unknown` remains the fallback. The PR description now makes this design choice explicit.

Validation: workspace fmt, clippy with warnings denied, and tests pass; Ruff, mypy, 117 Python tests, and the strict docs build pass.

@coderabbitai review

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -302,10 +302,11 @@ pub fn dimensions_from_signal(signal: &ToolSignals) -> CodingAgentDimensions {
     let deep_enough = signal.turn_depth >= STALL_MIN_TURN_DEPTH;
     let no_production = signal.recent_write_count == 0 && signal.recent_edit_count == 0;
     let investigating = signal.recent_read_count >= 1 || signal.recent_todowrite_count >= 1;
+    let new_activity = signal.recent_new_count >= 1;
     // spinning vs exploring partition the "not producing" case by investigative
     // activity, so at most one fires — no double-counting on the production axis.
-    let spinning = deep_enough && no_production && !investigating;
-    let exploring = deep_enough && no_production && investigating;
+    let spinning = deep_enough && no_production && !investigating && !new_activity;
```
sabhatinas, line 308: This is a great idea. Suppressing stalls from new activity is critical when tool signals are missing on a new harness or workload.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -302,10 +302,11 @@ pub fn dimensions_from_signal(signal: &ToolSignals) -> CodingAgentDimensions {
...
+    let exploring = deep_enough && no_production && investigating && !new_activity;
```
sabhatinas, line 309: what happens when a new category is detected btw?

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -261,12 +343,25 @@ pub struct ToolSignals {
...
 
 impl ToolSignals {
-    /// Extracts tool and progress signals from `request`.
+    /// Extracts tool and activity signals from `request`.
     ///
     /// `window_size` limits recent counters to the newest tool results. `None`
     /// uses [`DEFAULT_RECENT_WINDOW`].
     pub fn from_request(request: &Request, window_size: Option<usize>) -> Self {
-        extract_tool_signals_with_window(request, window_size.unwrap_or(DEFAULT_RECENT_WINDOW))
+        Self::from_request_with_semantics(request, window_size, &ToolSemantics::default())
```
sabhatinas, line 351: thanks for keeping this backward compatible.

sabhatinas: we need to have unit tests to ensure categorising to tool semantics does not produce a different score.

docs/routing_algorithms/stage_router_routing.md
```diff
@@ -208,6 +207,35 @@ switchyard-server --config routes.toml --port 4000
...
 This is the recommended default: routing on tool signals alone, no classifier.
 
+### Optional: custom tool semantics
+
+Extend the built-in coding vocabulary when an agent uses domain-specific tool
+names. Mappings are route-scoped, additive, and matched by exact name without
+regard to ASCII case:
+
+```toml
+[routes.stage.tool_semantics]
```
sabhatinas, line 219: @slopp are these tool calls specific to LangChain? can we specify it in the doc for clarity? Maybe it helps some users to re-use directly.

crates/libsy/src/algorithms/util/stage.rs
slopp: A `new` match increments `new_count` / `recent_new_count`. Recent new activity suppresses the `spinning` and `exploring` stall dimensions, because the agent is demonstrably making forward progress, but it adds no positive or negative scoring weight of its own. Any other signals on the turn still apply normally. So in the otherwise-empty case the score stays neutral rather than falsely favoring the capable tier. This is covered by `new_activity_suppresses_stall_dimensions_without_biasing_the_score`.

crates/libsy/src/algorithms/util/tool_signals.rs
slopp: Agreed on preserving scoring semantics. The categories are aliases into the existing signal counters: a configured `observe`, `mutate`, or `plan` tool is scored the same way as the corresponding built-in semantic; it is intentionally different from leaving that tool unknown. `configured_tool_semantics_extend_the_builtin_vocabulary` verifies the mapped counters, and `new_activity_suppresses_stall_dimensions_without_biasing_the_score` directly verifies that the new-only case remains score-neutral. The Python binding and server integration tests also verify the configuration reaches routing. The default/empty mapping continues through the prior extraction behavior unchanged.

docs/routing_algorithms/stage_router_routing.md
slopp: These names are not LangChain-specific conventions. They are illustrative application/toolset names; some were exercised in a LangChain-based evaluation, but `tool_semantics` operates on the literal tool/function name in the request and is independent of the agent framework and supported wire format. A LangChain agent, another framework, or a custom client can reuse these entries if it emits the same names; otherwise it maps its own names into the four categories.

sabhatinas:
Rebased onto current `main` (`8b3c62b`) and added the compatibility/coverage fix in `96927d3`:

- Preserve the legacy Unicode-aware lookup for built-in tools while keeping custom semantics ASCII case-insensitive, so an absent `tool_semantics` section behaves exactly as before.
- Add coverage for built-in precedence, additive custom mappings, unknown fallback, neutral/recent `new` scoring, and invalid route configuration.

Validated with workspace fmt, Clippy (`-D warnings`), full Rust workspace tests, 117 non-integration Python tests, Ruff, mypy, and the strict docs build.

sabhatinas (approved):

docs/routing_algorithms/stage_router_routing.md
sabhatinas: is it worth adding your langchain TOML here? maybe just a few tool names?

crates/libsy/src/algorithms/util/tool_signals.rs
sabhatinas: Added a direct score-equivalence regression in 53f67e7: configured `observe`, `mutate`, and `plan` tools are compared against `Read`, `Write`, and `TodoWrite`, and the full `ScoreResult` must be exactly equal with a non-zero control score. I also replayed 28 archived Terminal-Bench 2.1 Anthropic requests against pre-PR main and this branch with empty semantics; scores, confidence bits, and every legacy signal field matched exactly (0 mismatches).

docs/routing_algorithms/stage_router_routing.md
sabhatinas: Per the follow-up, keeping this PR to tests only—no LangChain-specific documentation update. The existing example remains framework-independent.

PR #612, opened 2026-09-03
feat(relay): report answer fallback in routing marks

afourniernv:
> Builds on the routing-mark schemas merged in [#604](https://github.com/NVIDIA-NeMo/Switchyard/pull/604).

## Why

The routing decision currently records the model Switchyard selected, while the Relay LLM span records the model that eventually answered. If the selected model fails and Switchyard falls back, neither record explains the difference, and the failed answer attempt is not visible.

This makes the complete serving path visible without changing routing, retries, fallback, responses, or metrics.

## What changed

| Mark | Change |
| --- | --- |
| `switchyard.routing.decision` | Adds nullable `served_model` and `fallback_used` fields. |
| `switchyard.routing.llm_call` | Emits one mark for each answer candidate, including call order, model, outcome, and latency. |

## Trace shape

Before, fallback has to be inferred from the selected model and the outer response:

```text
agent-run
├── openai.chat_completions              response from model/strong
├── mark:switchyard.routing.requested    algorithm=random
├── mark:switchyard.routing.overhead
└── mark:switchyard.routing.decision     selected=model/weak
```

After, both answer attempts and the final serving path are explicit:

```text
agent-run
├── openai.chat_completions               response from model/strong
├── mark:switchyard.routing.requested     algorithm=random
├── mark:switchyard.routing.llm_call      answer #1, model/weak, error
├── mark:switchyard.routing.llm_call      answer #2, model/strong, ok
├── mark:switchyard.routing.overhead
└── mark:switchyard.routing.decision      model/weak → model/strong, fallback=true
```

With an agent scope, Relay parents the LLM span and Switchyard marks into one trace. The smoke harness used for these screenshots had no agent scope, so Phoenix displays them as separate roots.

### Before

The decision records `model/weak`, while the successful LLM span reports `model/strong`. There is no fallback field or failed-attempt mark.

Before: the decision mark records model/weak

Before: model/strong served the response

### After

The decision reports the selected and served models and `fallback_used = true`. Two answer-call marks show `model/weak` failing before `model/strong` succeeds.

After: the decision records selected model, served model, and fallback

<details>
<summary>Answer-candidate attempt details</summary>

| First candidate | Fallback candidate |
| --- | --- |
| The model/weak call failed | The model/strong call succeeded |

</details>

## Contract notes

- These are additive fields and values on the version-1 schemas introduced in #604.
- `served_model` and `fallback_used` are `null` when serving metadata is unavailable.
- For streaming answers, `outcome = ok` and latency cover creation of the upstream stream. Later stream failures continue through `switchyard.routing.error`.
- No provider response body or free-form provider error is added to telemetry.

## Testing

- Focused plugin tests, formatting, and Clippy pass; GitHub CI runs the full suite.
- Live smoke test: `model/weak` returned HTTP 503, `model/strong` succeeded, and the resulting marks were verified in Phoenix.
- A direct successful call reports the same selected and served model with `fallback_used = false`.

## Summary by CodeRabbit

- **Observability**
  - Improved telemetry for routing decisions and answer-model calls.
  - Added model selection, served model, fallback status, outcomes, and latency details.
  - Expanded streaming lifecycle and token coverage across call types.
  - Added separate failure reporting and versioned telemetry schemas.
- **Documentation**
  - Documented telemetry mark fields, request details, call information, overhead, decisions, errors, and nullable serving metadata.

### DIFF
--- a/crates/libsy-llm-client/src/run.rs
+++ b/crates/libsy-llm-client/src/run.rs
@@ -130,16 +130,14 @@ fn emit_routing_observations(
     let (Some(observer), Some(observations)) = (observer, observations) else {
         return;
     };
-    let mut answer_observation = None;
+    let mut answer_observed = false;
     for observation in observations.lock().drain(..) {
-        if answer_observation.is_none() && answered_model == Some(&observation.selected_model) {
-            answer_observation = Some(observation);
+        if !answer_observed && answered_model == Some(&observation.selected_model) {
+            answer_observed = true;
+            observer(RunObservation::AnswerCall(observation));
         } else {
             observer(RunObservation::LlmCall(observation));
         }
-    }
-    if let Some(observation) = answer_observation {
-        observer(RunObservation::AnswerCall(observation));
     }
 }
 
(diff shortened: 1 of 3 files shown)


bbednarski9:
[P2] call_index can report the wrong execution order. The new answer-mark path assigns indexes in RunObservation delivery order ([runtime.rs](https://github.com/NVIDIA-NeMo/Switchyard/blob/9713c0fc6a9fcb88e54c7e0450a7287ee776b686/crates/switchyard-nemo-relay-plugin/src/runtime.rs#L253-L284)). However, the client deliberately buffers the observation that became the answer and emits it after every other routing observation ([run.rs (line 124)](/Users/bbednarski/Projects/switchyard_integration/switchyard/crates/libsy-llm-client/src/run.rs:124)). For example, advisor_gate can call the executor first, consult a reviewer second, and return the executor’s response; telemetry will number the reviewer as call 1 and the executor as call 2. That contradicts the PR’s promise that call_index represents call order. The observation layer should preserve original sequence—likely by retaining an index or marking the matching observation in place—before this field becomes part of the version-1 schema.

afourniernv:
Good catch. Fixed in 4a7f5800 by classifying the answer observation in place, so it keeps its original call order. I added a focused test for the answer-then-judge case too.

bbednarski9 (approved):

bbednarski9:
RFR @grahamking  @nachiketb-nvidia

grahamking (approved):

PR #613, opened 2026-09-03
docs(readme): restructure around three integration paths

itayzit:
## What

Restructures the README around how people actually adopt Switchyard, in the
order they should reach for it: **use it inside a gateway you already run →
integrate it into a gateway you build → run the standalone proxy.**

- **`What is Switchyard`** opens with the benchmark chart, then three sections:
  *Use Switchyard* (NeMo Relay, LiteLLM), *Integrate Switchyard into your
  gateway or harness* (`pip install nemo-switchyard`, then the guide), and *Run
  Switchyard as a standalone proxy*. Each has a diagram; all three share one
  shape and differ only in what box Switchyard sits inside. The Switchyard node
  is NVIDIA green in every diagram.
- **`Maturity` → `Components`.** The warning callout is replaced by a table of
  stability, intended use, and adoption guidance per crate — beta / alpha / demo.
- **Get Started is three self-contained paths** — Relay, Library, Proxy — each
  stating where you start and what you finish with. Path 2 is now built around
  a **verified Python example**: the `run_stream` loop over `CallModel` / `Done`,
  executed against the real `switchyard.libsy` bindings before it went in.
- **Routing algorithm names follow the Slack thread**: Capability, Stage,
  Capability + Stage, Escalation, Advisor Gate, Sub-Agent-Aware, Custom, Random.
- **No links to `docs/getting_started.md`.** Chart labels read "Switchyard:"
  rather than "SY:"; a footnote states the cost basis.

## Why

The library is the primary way we want Switchyard adopted, and the README led
with the proxy. It also had no working example for the embed path in any
language, and every install path started mid-story.

## Notes for reviewers

**Read the rendered page, not the diff** — it is a rewrite. The mermaid diagrams
and the chart only render on GitHub. Flip your theme once: the green node is
verified to survive mermaid's dark theme, but worth one look.

Judgement calls worth a second opinion:

- **Path 2's Python loop is the only embed example in the repository.** I
  grepped every crate README and `docs/` page for `Step.CallModel` / `Step.Done`
  — zero hits. It should probably also live in `crates/libsy/README.md`; that is
  a follow-up, not this PR.
- **`docs/getting_started.md` is now unreferenced by the README but still
  exists.** Deleting it is an 11-file change that also removes
  `tests/getting_started/` and a CI workflow, so it is left for its own PR. Note
  that test asserts `## Server Path` before `## Library Path` — the opposite of
  the order adopted here.
- **The proxy diagram has no subgraph, deliberately.** The other two are boxed
  inside something you already run; the proxy is its own process.
- **The chart footnote says "ISP token cost"** as requested. Most readers will
  parse ISP as Internet Service Provider; spelling out "inference-provider" may
  be worth it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,45 @@
 
 ![Accuracy versus total cost on Terminal-Bench 2.1. Switchyard's staged, escalation, and classifier routes reach 71-76% accuracy for 13-30% less than the Opus 4.8 baseline, while single fixed models stay below 56%.](assets/benchmark-accuracy-vs-cost.svg)
 
+_\*Total cost based on average ISP token cost_
 
 ## What is Switchyard
 
-**1. An embeddable Rust library** `switchyard-libsy` picks the model for you. Drops into whatever gateway or agent runtime you already have.
+Switchyard picks which model serves each LLM call.
 
-```bash
-cargo add --git https://github.com/NVIDIA-NeMo/Switchyard.git --tag v0.2.0 \
-  switchyard-libsy switchyard-protocol
+### Use Switchyard
+
+Switchyard runs inside gateways you may already have.
+
+- **NeMo Relay** — a native plugin. Load a `routes.toml` into a Relay deployment
+  you already run. [Setup →](#path-1--load-the-nemo-relay-plugin)
+- **LiteLLM** — a routing plugin for LiteLLM's `Router` and proxy.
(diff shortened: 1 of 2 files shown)


ayushag-nv (approved):
Loooks good. I would like to see it a bit concise in the future with more redirection links but its okay for now

afourniernv (commented):
withdrawn

PR #618, opened 2026-09-04
feat(server): record route attribution in durable log

### LINKED ISSUE #617: [feature] Record route_id (and algorithm) in the durable routing log
## Problem

The durable routing log written by `--routing-log-file` records the **served model** and `tier`, but not the **route** the client asked for. Two routes that resolve to the same model are therefore indistinguishable in the log.

Concrete case on 0.2.0 (`switchyard-server`, Rust): a `passthrough` route to a local Nemotron model and a `stage_router` route whose `efficient_target` is the same model both log as

```
{"ts":"…","model":"nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4","tier":"", …}
```

Consequences for anyone using the log as an accounting source:

- **No escalation rate.** Escalations *are* visible (`tier: "strong"`), but the denominator — how many requests went *through the stage_router* — cannot be recovered, because its efficient-tier decisions look identical to the passthrough route's traffic.
- **No per-route / per-project attribution.** Semantic route ids (e.g. `copper.analyst`, `scg.generator`) are the whole reason for a router in front of shared hardware, and they are the one thing the log drops.

`/metrics` does carry `algorithm=` and `tier=` labels (`switchyard_decisions_total{algorithm="stage_router",selected_model=…}`), so the rate is computable there — but those counters reset on every server restart (observed 89 in `/v1/stats` vs 159 records in the durable log after one restart), so they cannot serve as the durable weekly source the log otherwise is.

Who hits it: any operator running several routes over a shared model pool who wants weekly usage or escalation figures from the durable log rather than from a scrape pipeline. Every routed request.

## Proposed solution

Add the route id to each routing-log record — the `id` from the matching `[routes.*]` entry, i.e. exactly the string the client sent as `model`:

```json
{"ts":"…","route_id":"copper.escalate","algorithm":"stage_router","model":"nemotron-3-super:120b-a12b","tier":"strong", …}
```

`route_id` is the essential field. `algorithm` is a cheap companion that mirrors what `/metrics` already labels and lets a log consumer group by router type without joining back to the TOML.

## Alternatives considered

- **Scrape `/metrics` to durable storage.** Works, but adds a collector to every deployment, loses the interval around each restart, and still yields no `route_id` — the metrics label the algorithm and selected model, not the route.
- **Set `session_id` / `task` per route from the client.** Both fields exist in the record, but they require every consumer to be modified and disciplined; the router already knows the route at the moment it writes the line.
- **Infer route from `(model, tier)`.** Fails exactly in the common case above: same model, default tier, two routes.

The proposal is preferable because the server has the route id in hand when it emits the record, and it is one additive key.

## Scope notes

- Surface: **server — the routing-log writer** (`--routing-log-file`). No routing-algorithm, protocol, codec, or launcher change.
- Public interface: the JSONL record gains new keys. `/v1/routing/session-stats` could optionally accept a `route_id` filter, but that is not required for the core request.
- Backward compatibility: additive. Existing consumers that read `model`/`tier`/tokens are unaffected; a new key is ignored by anything that does not look for it.

## Additional context

- Related: #276 fixed a case where fallback/affinity decisions logged **without a tier** — same log, adjacent gap. This request is about the route, not the tier.
- Observed while using the log as the source for a weekly per-backend usage report across a 3-host fleet (one `stage_router`, several `passthrough` routes, four backends). Happy to test a build.

cpakkamisaac-sae:
## What

- add `route_id` and `algorithm` to each durable routing-log record
- carry the resolved route identity through buffered, streaming, and classifier/judge log paths
- preserve compatibility when reading older JSONL records
- document the additive fields and cover shared-backend routes end to end

## Why

Two configured routes can serve the same backend model. Today their durable records can be identical, so operators cannot recover per-route usage or the denominator for an algorithm's routing decisions after metrics counters restart.

Closes #617.

## Test plan

- native server process with passthrough and stage-router routes sharing one mock backend
- `cargo +1.96.1 fmt --all --check`
- `cargo +1.96.1 clippy --locked --offline --workspace --all-targets -- -D warnings`
- `cargo +1.96.1 test --locked --offline --workspace`
- `uv run --offline ruff check .`
- `uv run --offline mypy switchyard`
- `uv run --offline pytest tests/ -q` (`117 passed, 2 subtests passed`)

## Summary by CodeRabbit

* **New Features**
  * Routing logs now include the resolved route ID and routing algorithm alongside the served model, tier, and token usage.
  * Requests using different routes that share the same backend model are now distinguishable in routing records.
  * Existing routing logs remain compatible when the new metadata is unavailable.
  * Streaming routing logs now preserve canonical route and algorithm metadata for verification.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -987,6 +987,7 @@ fn resolve_route(
     Ok((route, request))
 }
 
+/// Resolves and executes an LLM request, attaching route identity when durable logging is enabled.
 async fn handle_llm_request(
     state: ServerState,
     started: RequestStart,
@@ -1000,9 +1001,12 @@ async fn handle_llm_request(
         Ok(resolved) => resolved,
         Err(response) => return response,
     };
-    let route_id = request.llm_request.model.clone().unwrap_or_default();
-    let routing_log_context =
-        routing_log_context.map(|context| context.with_route(route_id, route.algorithm_name()));
+    let routing_log_context = routing_log_context.map(|context| {
+        context.with_route(
+            request.llm_request.model.as_deref().unwrap_or_default(),
+            route.algorithm_name(),
+        )
+    });
     // Only the Codex namespace mapping is needed downstream, not the whole request.
     let request_extensions = request.llm_request.extensions.clone();
     let observer = stats_observer(
@@ -1673,13 +1677,11 @@ mod tests {
     #[test]
     fn stats_observer_logs_judge_calls_to_the_routing_log() {
         let dir = tempfile::tempdir().expect("temp dir");
-        let log_path = dir.path().join("routing.jsonl");
-        let log = SharedRoutingLog::new(log_path.clone()).expect("routing log");
+        let log = SharedRoutingLog::new(dir.path().join("routing.jsonl")).expect("routing log");
         let mut headers = HeaderMap::new();
         headers.insert("proxy_x_session_id", "session-1".parse().expect("header"));
         let metadata = metadata_from_headers(headers);
-        let context = routing_log::RoutingLogContext::from_metadata(&metadata)
-            .with_route("switchyard/classifier", "llm_classifier");
+        let context = routing_log::RoutingLogContext::from_metadata(&metadata);
         let observer = stats_observer(StatsAccumulator::default(), Some((log.clone(), context)));
 
         let call = |model: &str, answer: bool| {
@@ -1711,18 +1713,6 @@ mod tests {
         assert_eq!(snapshot["models"]["judge-model"]["prompt_tokens"], 100);
         assert_eq!(snapshot["models"]["judge-model"]["completion_tokens"], 7);
         assert!(snapshot["models"].get("routed-model").is_none());
-
-        let record: Value = serde_json::from_str(
-            std::fs::read_to_string(log_path)
-                .expect("read log")
-                .lines()
-                .next()
-                .expect("judge record"),
-        )
-        .expect("valid record");
-        assert_eq!(record["route_id"], "switchyard/classifier");
-        assert_eq!(record["algorithm"], "llm_classifier");
-        assert_eq!(record["tier"], CLASSIFIER_TIER);
     }
 
     #[derive(Clone)]
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -1000,6 +1000,9 @@ async fn handle_llm_request(
         Ok(resolved) => resolved,
         Err(response) => return response,
     };
+    let route_id = request.llm_request.model.clone().unwrap_or_default();
+    let routing_log_context =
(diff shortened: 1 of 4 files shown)


crates/switchyard-server/src/routing_log.rs
```diff
@@ -98,6 +100,8 @@ pub(crate) fn snapshot(
 /// Request fields retained until terminal usage and routing are available.
 #[derive(Clone)]
 pub(crate) struct RoutingLogContext {
+    route_id: String,
+    algorithm: String,
```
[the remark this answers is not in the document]
cpakkamisaac-sae: Addressed in 84ef6139: the struct comment now states that the context holds request metadata until route resolution attaches the durable route identity. Formatting and the focused routing-log tests pass (2 unit tests and 4 HTTP integration tests).

cpakkamisaac-sae:
@coderabbitai review

crates/switchyard-server/src/lib.rs
```diff
@@ -1000,6 +1000,9 @@ async fn handle_llm_request(
         Ok(resolved) => resolved,
         Err(response) => return response,
     };
+    let route_id = request.llm_request.model.clone().unwrap_or_default();
```
afourniernv: Small thing: this clones the model on every request, even when the routing log is off. Could we move it inside `map` and pass `request.llm_request.model.as_deref().unwrap_or_default()` to `with_route`? Then we only allocate when logging is enabled.

afourniernv:
Could you take a quick pass at reducing the test LOC? The shared-model and streaming coverage makes sense. I think the new classifier assertions could go in the existing `advisor_route_routing_log_records_classifier_tier` test, which already covers that path end to end. That should let us drop the extra file read from the unit test.

crates/switchyard-server/src/lib.rs
cpakkamisaac-sae: @afourniernv Addressed in 6538ab72. Route attribution now borrows the requested model inside `routing_log_context.map(...)`, so the `String` allocation only occurs when durable routing logging is enabled. The focused routing-log tests, complete server suite, full workspace tests, formatting, and workspace Clippy all pass.

cpakkamisaac-sae:
@afourniernv Addressed the test-size request in 6538ab72. The classifier-tier `route_id` and `algorithm` assertions now live in the existing `advisor_route_routing_log_records_classifier_tier` HTTP test, and the unit test no longer clones the log path, manually attaches route identity, or rereads the JSONL file. The shared-model and streaming coverage remains intact. The follow-up is net -8 lines; focused tests, the full server and workspace suites, formatting, and workspace Clippy all pass.

afourniernv (approved):
Looks good. I checked the route attribution across shared-model, classifier, and streaming paths, and ran the wider CI. All green.

afourniernv:
Thanks for the contribution! This is a useful addition.

PR #619, opened 2026-09-04
fix(translation): Responses wire compatibility for strict backends (system/developer roles, list input)

RedEyeNinja-BKK:
Strict Responses backends (the chatgpt.com Codex endpoint) reject two shapes that the Responses codec currently emits:

1. **`system`-role input items** - `400 "System messages are not allowed"`; the backend requires the `developer` role.
2. **Top-level input STRING shorthand** - `400 "Input must be a list"` when `encode_responses_input` emits a bare string for a single-user-text-message request.

This series makes every Responses-encoded request wire-legal for strict backends:

- `role_to_responses` maps `Role::System` to `developer` (mirroring the Codex wire rules for chat -> Responses conversions). `developer` is a first-class role in the Responses API, so non-strict backends are unaffected.
- Exact-request passthrough replays a preserved body verbatim; `system`-role input items are normalized to `developer` at replay time (typed `{"type":"message","role":"system"}` items and untyped role-keyed items alike), so preserved same-format requests stay wire-legal too.
- `encode_responses_input` always emits canonical message-item lists. Message-item lists are valid for every Responses backend, so this is wire-shape hygiene rather than a behavior fork.

The inline-instruction decode architecture is unaffected: system/developer items decode to instruction blocks and re-encode into the `instructions` field; the role mapping additionally protects any system-role IR message reaching the generic encoder. Regression tests cover typed/untyped normalization, wire-level absence of `system` roles, the exact-replay and embed-preservation boundaries, and single-message list encoding.

DCO signed.

## Summary by CodeRabbit

- **Bug Fixes**
  - OpenAI Responses requests using `system` roles are now consistently translated to `developer` roles.
  - Preserved and replayed requests now produce canonical input arrays without unsupported `system` roles.
  - Improved compatibility for both structured and unstructured input items when processing role-based instructions.
- **Tests**
  - Added coverage for role normalization across direct translation and preserved-request replay scenarios.

### DIFF
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -1022,10 +1022,14 @@ fn encode_responses_input(
     policy: &TranslationPolicy,
     namespaces: Option<&Map<String, Value>>,
 ) -> Result<Value> {
-    // Strict Responses backends (chatgpt.com Codex endpoint) reject non-list
-    // input; callers of translated chat-completions cannot influence the shape.
-    // Always encode canonical message-item lists instead of the input-string
-    // shorthand.
+    if messages.len() == 1
+        && matches!(messages[0].role, Role::User)
+        && messages[0].content.len() == 1
+        && matches!(messages[0].content[0], ContentBlock::Text { .. })
+        && let ContentBlock::Text { text } = &messages[0].content[0]
+    {
+        return Ok(Value::String(text.clone()));
+    }
     let mut encoded = Vec::new();
     for message in messages {
         // Anthropic-signed thinking cannot be sent as Responses input.
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -128,8 +128,11 @@ impl FormatCodec for OpenAiResponsesCodec {
             // Strict Responses backends reject `system`-role input items ("System
             // messages are not allowed"); they require `developer`. The preserved
             // body replays verbatim, so normalize it here to keep exact-request
-            // passthrough wire-legal.
+            // passthrough wire-legal. A scalar string `input` is likewise
+            // converted to the canonical message-item list so preserved bodies
+            // keep the always-list wire shape the encode path guarantees.
             normalize_system_input_roles(&mut body);
+            normalize_input_to_message_list(&mut body);
             return Ok(EncodedRequest {
                 body,
                 diagnostics: Vec::new(),
@@ -1016,20 +1019,17 @@ fn encode_responses_text_format(response_format: &Value) -> Value {
 }
 
 // Encodes normalized messages into the Responses `input` shape.
+//
+// The result is ALWAYS a message-item list. Strict Responses backends
+// (chatgpt.com Codex endpoint) reject the scalar string form, and list input
+// is universally accepted by normal `/v1/responses` endpoints, so the
+// historical single-user-text scalar fast path is intentionally gone.
 fn encode_responses_input(
     messages: &[Message],
     diagnostics: &mut Vec<TranslationDiagnostic>,
     policy: &TranslationPolicy,
     namespaces: Option<&Map<String, Value>>,
 ) -> Result<Value> {
-    if messages.len() == 1
-        && matches!(messages[0].role, Role::User)
-        && messages[0].content.len() == 1
-        && matches!(messages[0].content[0], ContentBlock::Text { .. })
-        && let ContentBlock::Text { text } = &messages[0].content[0]
-    {
-        return Ok(Value::String(text.clone()));
-    }
     let mut encoded = Vec::new();
     for message in messages {
         // Anthropic-signed thinking cannot be sent as Responses input.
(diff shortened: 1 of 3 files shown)


crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -122,9 +122,14 @@ impl FormatCodec for OpenAiResponsesCodec {
...
     ) -> Result<EncodedRequest> {
-        if let Some(body) =
+        if let Some(mut body) =
             exact_preserved_request(&request.preservation, WireFormat::OpenAiResponses, _policy)
         {
+            // Strict Responses backends reject `system`-role input items ("System
+            // messages are not allowed"); they require `developer`. The preserved
+            // body replays verbatim, so normalize it here to keep exact-request
+            // passthrough wire-legal.
+            normalize_system_input_roles(&mut body);
```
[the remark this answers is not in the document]
RedEyeNinja-BKK: Fixed in af0d85bb (branch tip). The scalar fast path in `encode_responses_input` is removed - it now always returns a message-item list - and the preserved replay path converts scalar string `input` to the canonical single user message-item list, so the always-list guarantee holds on both paths (matching the production reference behavior that motivated this PR). Regression coverage: non-preserved scalar encode, preserved scalar replay, and byte-exact replay for already-canonical list input. Full `switchyard-translation` suite green at the final head, which also folds in a `cargo fmt` pass for the CI rustfmt gate.

crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -1011,20 +1019,17 @@ fn encode_responses_text_format(response_format: &Value) -> Value {
 }
 
 // Encodes normalized messages into the Responses `input` shape.
+//
+// The result is ALWAYS a message-item list. Strict Responses backends
```
afourniernv: Can we cut down on the large comment strings - it's a smell.

afourniernv:
I pulled on this a bit more. The compatibility issue looks real.

The normal Responses API accepts scalar input and system messages, but the ChatGPT Codex endpoint has been observed rejecting both. OpenAI’s Codex client sends list input and developer messages instead.

Since both endpoints use the Responses format, we need something else to tell them apart. I couldn’t find an earlier Switchyard issue or maintainer decision covering these Codex restrictions.

Where I’m uncomfortable is the boundary. #619 makes the Codex restrictions part of the shared Responses codec, even though those request shapes are valid on the normal OpenAI API.

#621 then adds `Backend::is_codex()` as public API and changes shared URL handling to identify the endpoint. This feels like a larger API and behavior decision than the original compatibility bug needs.

I don’t think we should merge this as-is. The outbound client looks like the more natural place to isolate backend-specific handling, but I don’t want to prescribe whether that should use the URL, an explicit backend type, or something else.

@grahamking could you weigh in on where you’d want this boundary before we ask for a rework?

bhuvan002:
Thank for for the contribution @RedEyeNinja-BKK. The team is discussing the design question raised by @afourniernv and we hope to get back with an answer next week!

crates/switchyard-translation/src/codecs/responses/buffered.rs
```diff
@@ -122,9 +122,17 @@ impl FormatCodec for OpenAiResponsesCodec {
         request: &LlmRequest,
         _policy: &TranslationPolicy,
     ) -> Result<EncodedRequest> {
-        if let Some(body) =
+        if let Some(mut body) =
             exact_preserved_request(&request.preservation, WireFormat::OpenAiResponses, _policy)
         {
+            // Strict Responses backends reject `system`-role input items ("System
+            // messages are not allowed"); they require `developer`. The preserved
+            // body replays verbatim, so normalize it here to keep exact-request
```
grahamking, line 130: We should not normalize the preserved body. We should forward it as-is. If it fails that's an issue between the client and the backend.

grahamking:
1. Every Responses backend should get this stricter Responses implementation - `developer` role and messages array. It belongs in translations, this PR is correct.

> Where I’m uncomfortable is the boundary. https://github.com/NVIDIA-NeMo/Switchyard/pull/619 makes the Codex restrictions part of the shared Responses codec, even though those request shapes are valid on the normal OpenAI API.

My Astra said something similar, so I'm disagreeing with the AIs here. The OpenAI API supports this for backwards compatibility. We're the client so we don't need to worry about that.

Thanks for closing #621 @afourniernv , I agree we don't want `is_codex` type detection, just clean protocol support.

2. However, I don't think we want to change the preserved client's JSON. Part of the promise is that we forward that unchanged. 

This means you can't use an older Responses client with the Codex endpoint. I can't think of anyone wanting to do that. Every potential Codex-endpoint client (Pi, OpenCode, Hermes, OpenClaw, etc) will have upgraded to stricter Responses long ago.

@RedEyeNinja-BKK Can you remove the preserved JSON normalization, and rebase?

afourniernv:
@RedEyeNinja-BKK friendly bump when you get a chance 🙂 Graham clarified the direction above: keep the canonical Responses encoding in translation, but leave preserved request JSON unchanged. Once that normalization is removed and the branch is rebased, I can take another look. Thanks!

RedEyeNinja-BKK:
> @RedEyeNinja-BKK friendly bump when you get a chance 🙂 Graham clarified the direction above: keep the canonical Responses encoding in translation, but leave preserved request JSON unchanged. Once that normalization is removed and the branch is rebased, I can take another look. Thanks!

Yes. I have received the pointer and your nudge. I'm just in the midst of aligning routes and once that is settled, I'll get my agent on it. Thank you for the pointer and the nudge; I'll update again soon.

RedEyeNinja-BKK:
Thanks both — done as directed.

* Rebased onto current `main` (`9cf6fadf`); PR is mergeable again. Final diff is 3 files, +120/−14.
* Removed preserved-request normalization completely. `exact_preserved_request(...)` is forwarded unchanged: preserved scalar `input` stays scalar and preserved `system` stays `system`.
* Canonical Responses encoding retains the stricter wire shape: generated input is always a message-item list, and generated `System`/`Developer` roles encode as `developer`.
* No backend/URL detection and no `is_codex`.
* Added regression coverage for both sides of that boundary, including mutation/negative controls.
* Validation: `cargo fmt` clean, clippy clean, workspace **939 passed / 0 failed**, `switchyard-translation` **225 passed / 0 failed**.
* Head: `1f9d50b6`.

The remaining GitHub Actions jobs are currently at the first-contributor approval gate rather than reporting PR-related failures.

Thanks for clarifying the preservation boundary.

PR #620, opened 2026-09-04
fix(llm-client): canonical reasoning representation + target reasoning-pin precedence

RedEyeNinja-BKK:
OpenAI-family targets pin their thinking policy via `extra_body` (for example a non-thinking lane carrying `reasoning = { effort = "none" }`, or `reasoning = { enabled = false }`). Because `merge_extra_body` only fills absent keys (or_insert semantics), a caller that supplies its own nested `reasoning` object silently displaces the target's policy: a non-thinking lane would run reasoning its route author disabled.

This series re-asserts the target's reasoning pin after merge (replacing any caller-supplied nested `reasoning` object and dropping the flat `reasoning_effort`), on OpenAI Chat and Responses backends. It also fixes the conflicting dual representation OpenRouter rejects ("reasoning_effort and reasoning.effort are both provided with conflicting values"): with no target pin, a body carrying both a nested `reasoning.effort` and a conflicting flat `reasoning_effort` now drops the flat field and keeps the nested object, which is the canonical form.

Regression tests cover: target NT-pin precedence (Chat and Responses legs), `enabled = false` pins, and dual-representation suppression with and without a target pin. Anthropic's separate thinking contract is explicitly out of scope.

DCO signed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of reasoning settings for OpenAI Chat and Responses requests.
  - Backend-provided reasoning defaults now take precedence when explicitly configured.
  - Conflicting reasoning options are resolved consistently, preventing duplicate or contradictory settings from being sent.
  - Reasoning behavior is now more predictable across supported OpenAI request types.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -256,8 +256,18 @@ impl TranslatingLlmClient {
         if matches!(backend, Backend::Anthropic(_)) {
             enable_anthropic_prompt_caching(&mut body);
         }
+        // Target-pinned reasoning policy is canonical on both OpenAI leg formats:
+        // when the target pins a nested `reasoning` policy (e.g. an NT lane's
+        // `reasoning = { effort = "none" }` or `enabled = false` pin via
+        // extra_body), it overrides any caller-supplied reasoning representation.
+        // merge_extra_body only fills absent keys (or_insert), so the target pin
+        // must be re-asserted here; without it, a caller's own `reasoning` object
+        // would silently displace the target's policy.
         if matches!(backend, Backend::OpenAiChat(_)) {
+            canonicalize_reasoning_effort(&mut body, backend.extra_body().get("reasoning"));
             ensure_openai_stream_usage(&mut body);
+        } else if matches!(backend, Backend::OpenAiResponses(_)) {
+            canonicalize_reasoning_effort(&mut body, backend.extra_body().get("reasoning"));
         }
         let streaming = endpoint.allows_streaming()
             && body.get("stream").and_then(Value::as_bool).unwrap_or(false);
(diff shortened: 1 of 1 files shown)


crates/libsy-llm-client/src/client.rs
```diff
@@ -829,6 +839,48 @@ fn merge_extra_body(body: &mut Value, extra_body: &BTreeMap<String, Value>) {
...
+        object.remove("reasoning_effort");
+        return;
+    }
+    // Rule 2: no target pin - avoid the conflicting dual representation.
+    if object
+        .get("reasoning")
+        .and_then(Value::as_object)
+        .and_then(|reasoning| reasoning.get("effort"))
+        .and_then(Value::as_str)
+        .is_some()
```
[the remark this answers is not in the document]
RedEyeNinja-BKK: Fixed in e1b7fa6c (branch tip d0d56b3c = e1b7fa6c + a `cargo fmt`-only commit): any object-valued `reasoning` is now treated as canonical in the no-pin path - the flat `reasoning_effort` is dropped regardless of whether the object carries a string `effort`. Regression test covers the `reasoning.enabled=false` + flat-effort case end-to-end through the OpenAI Chat wire path, alongside the existing no-pin string-effort and target-pin tests.

afourniernv:
This feels like a lot of test code for the size of the proposed change. Could we reduce the amount of repeated setup and keep the coverage more focused?

afourniernv (changes requested):
please try to shorten inline comments and test cases. This is a really large PR for the proposed work

crates/libsy-llm-client/src/client.rs
```diff
@@ -256,8 +256,18 @@ impl TranslatingLlmClient {
         if matches!(backend, Backend::Anthropic(_)) {
             enable_anthropic_prompt_caching(&mut body);
         }
+        // Target-pinned reasoning policy is canonical on both OpenAI leg formats:
+        // when the target pins a nested `reasoning` policy (e.g. an NT lane's
+        // `reasoning = { effort = "none" }` or `enabled = false` pin via
```
afourniernv: Can we compress inline comments down in this PR

afourniernv:
I think the client is the right layer for this, but this changes `extra_body.reasoning` from a default into an override. That can silently change reasoning level and discard other caller settings across every OpenAI-compatible backend. I do not think we should merge that behavior without explicitly deciding on the configuration contract first.

@nachiketb-nvidia could you take a look at the intended `extra_body` precedence here?

crates/libsy-llm-client/src/client.rs
```diff
@@ -2276,4 +2325,397 @@ mod tests {
...
+        assert!(
+            seen_body.get("reasoning_effort").is_none(),
+            "flat reasoning_effort must be dropped when any object-valued reasoning is present"
+        );
+        Ok(())
+    }
+
+    // OpenAI Responses backend: a target `reasoning = { effort = "none" }` pin
+    // must replace the caller's nested reasoning object on the Responses wire.
+    #[tokio::test]
```
afourniernv: can we make tests more concise / slim / maybe even just a singleton?

PR #627, opened 2026-09-04
fix(llm-client): validate configured headers at startup

### LINKED ISSUE #626: [bug]: --dry-run accepts configured HTTP headers that cannot be sent
## Symptom

`switchyard-server --dry-run` accepts configured HTTP headers that cannot be used to build an upstream request. The server also starts normally, but every request routed through the affected client returns HTTP 502 before reaching the upstream.

## Reproduction

No provider credentials or external network access are required. Start a loopback HTTP server on port 19450, then save this configuration as `routes.toml`:

```toml
schema_version = 1

[llm_clients.upstream]
format = "openai_chat"
base_url = "http://127.0.0.1:19450/v1"
extra_headers = { "bad header" = "value" }
max_retries = 0

[targets.upstream]
id = "upstream/model"
llm_client = "upstream"

[routes.default]
id = "switchyard/default"
type = "passthrough"
target = "upstream"
```

Run the deployment check:

```bash
cargo run --locked -p switchyard-server -- --config routes.toml --dry-run
```

It exits with status 0 and prints:

```text
server OK: switchyard/default
```

Start the same configuration:

```bash
cargo run --locked -p switchyard-server -- \
  --config routes.toml --host 127.0.0.1 --port 19451
```

Send a request:

```bash
curl -i http://127.0.0.1:19451/v1/chat/completions \
  -d '{"model":"switchyard/default","messages":[{"role":"user","content":"hello"}]}'
```

The response is:

```text
HTTP/1.1 502 Bad Gateway

{"error":{"message":"failed to build upstream request: builder error","type":"upstream_error","code":"upstream_configuration_error"}}
```

The loopback upstream receives no request. A valid control using
`extra_headers = { "x-audit-header" = "valid-control" }` reaches the same upstream and returns HTTP 200.

The same startup/runtime mismatch occurs when `api_key_env` resolves to a value containing an HTTP control character, such as a newline: `--dry-run` exits successfully, but the routed request returns the same HTTP 502 without reaching the upstream.

## Expected vs. actual

- **Expected:** Client construction and `--dry-run` reject header names, configured header values, and generated authentication header values that the HTTP client cannot send. The error should identify the affected client and field without exposing secrets.
- **Actual:** Validation checks conflicts with reserved headers, but header construction errors are deferred until the first routed request.

## Environment

- Switchyard commit SHA: `4eae4bf1e464bd778f1e68c4d5824bc195c3e692`
- Python version: N/A for the server path; Python standard library used only for the loopback capture server
- Install path: source build with `cargo build -p switchyard-server`
- Backend: offline loopback OpenAI-compatible server

## Additional context

The behavior was reproduced with the compiled server as a standalone process and real TCP requests. Invalid header-name and invalid API-key cases failed before the loopback server recorded a request; the valid-header control recorded the rewritten model and configured header.

Related work does not cover this validation gap:

- #398 added early validation for malformed upstream URLs after the same dry-run/runtime mismatch.
- #360 rejects reserved authentication and protocol names in `extra_headers`, but does not validate whether names and values are constructible.

A startup check should use the same header conversion semantics as request construction and remain entirely offline.

cpakkamisaac-sae:
## What

- validate configured header names and values when constructing the HTTP client
- validate provider authentication headers before any request can be routed
- return field-specific configuration errors without including configured values or credentials
- add deployment and process-level regression coverage, including valid-header controls

## Why

Closes #626.

An invalid `extra_headers` name or an API key containing an HTTP control character previously passed `switchyard-server --dry-run` and allowed the server to start. Every request through the affected route then failed with HTTP 502 before reaching the upstream.

Validation now uses the same header conversion rules as request construction, keeping the check offline while moving deterministic failures to deployment startup.

## Notes for reviewers

Start with `Backend::validate_configured_headers` in `crates/libsy-llm-client/src/backend.rs`. It validates both configured additional headers and the provider-specific authentication representation. Error messages name the model and field but never include header values or credentials.

The valid control retains extended-byte header values accepted by the underlying HTTP library, so this does not broaden validation beyond values that would fail request construction.

Validation completed locally:

- `cargo fmt --all --check`
- `cargo test --workspace`
- `ruff check .`
- `mypy switchyard`
- `pytest tests/ -v` — 117 passed, including the Docker integration tests
- standalone executable checks: invalid configuration exits 1 with and without `--dry-run`; invalid credential exits 1 without exposing its value; valid control exits 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for configured custom HTTP headers and API keys.
  * Invalid header names, values, or API keys now produce clearer configuration errors.
  * Prevented sensitive API keys from appearing in error output.
  * Dry-run checks now reliably reject configurations containing headers or credentials that cannot be encoded for HTTP requests.

* **Tests**
  * Added coverage for valid and invalid headers, API keys, and command-line dry-run behavior.

### DIFF
--- a/crates/libsy-llm-client/src/backend.rs
+++ b/crates/libsy-llm-client/src/backend.rs
@@ -44,7 +44,7 @@ pub struct HttpBackendConfig {
     /// Base URL of the provider API (e.g. `https://api.openai.com/v1`).
     pub base_url: String,
     /// API key for the provider, loaded by the caller. `None` sends no configured auth.
-    /// Client construction rejects values that cannot form the provider's auth header.
+    /// Client construction rejects active values that cannot form the provider's auth header.
     pub api_key: Option<String>,
     /// Whether this backend forwards the caller's provider credential instead.
     pub forward_auth: bool,
@@ -130,7 +130,7 @@ impl Backend {
             });
         }
 
-        let Some(api_key) = self.config().api_key.as_deref() else {
+        let Some(api_key) = self.configured_api_key() else {
             return Ok(());
         };
         let valid_api_key = match self {
@@ -167,6 +167,15 @@ impl Backend {
         }
     }
 
+    // Static credentials are unused when the caller's authorization is forwarded.
+    fn configured_api_key(&self) -> Option<&str> {
+        if self.is_forwarding_auth() {
+            None
+        } else {
+            self.config().api_key.as_deref()
+        }
+    }
+
     /// The fully resolved upstream URL for this backend's endpoint.
     ///
     /// Tolerates base URLs that already include the provider path (or a bare
@@ -186,11 +195,7 @@ impl Backend {
     /// `x-api-key: <key>` plus the required `anthropic-version` header. A backend
     /// with `forward_auth` uses the caller's provider credential instead.
     pub fn apply_auth(&self, mut builder: RequestBuilder) -> RequestBuilder {
-        let api_key = if self.is_forwarding_auth() {
-            None
-        } else {
-            self.config().api_key.as_deref()
-        };
+        let api_key = self.configured_api_key();
         match self {
             Backend::OpenAiChat(_) | Backend::OpenAiResponses(_) => {
                 if let Some(api_key) = api_key {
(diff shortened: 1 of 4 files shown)


afourniernv (changes requested):
Two things I’d like to address before this merges:

- Could validation share the configured-key selection with `apply_auth()`? `apply_auth()` ignores `api_key` when `forward_auth` is enabled, while this path still validates it. Direct `HttpBackendConfig` callers can construct both today, and keeping that decision in one place would prevent the two paths from drifting.

- The coverage makes sense, but the test setup feels pretty large for this change. Could we trim some of the repeated TOML between the runner and CLI cases, or move the low-level header cases into `switchyard-llm-client` and keep the CLI tests focused on startup failure and redaction?

cpakkamisaac-sae:
@afourniernv Addressed both review requests in 2caf3f6f:

- Startup validation and `apply_auth()` now share one configured-key selector. Direct backend configurations with `forward_auth = true` ignore an inert configured key in both paths; the behavior is covered for OpenAI and Anthropic backends.
- Low-level name/value, Unicode, API-key encoding, and redaction coverage now lives with the LLM-client backend. The duplicate runner cases were removed, and the CLI cases share one TOML/dry-run harness focused on startup rejection and credential redaction.

Focused backend, CLI, runner, and complete LLM-client tests pass, as do the full Rust workspace, formatting, and workspace Clippy with warnings denied.

afourniernv:
@cpakkamisaac-sae LGTM otherwise. Could you update the branch against current main and resolve the CLI test conflict? Once that is done and CI is green, I’ll merge this after the release code freeze ends.

cpakkamisaac-sae:
@afourniernv Rebased onto current `main` and resolved the CLI test conflict in c973a842. The branch is now 0 commits behind and GitHub reports it cleanly mergeable.

Local validation passes: full Rust workspace tests, workspace Clippy with warnings denied, prefill-router tests/Clippy, formatting, Ruff, mypy, and Python non-integration tests. The upstream CI, Getting Started, and Package Portability workflows are currently `action_required` with no jobs created, so they still need repository approval to run. Commitlint is also ending in `startup_failure` across current `main` and unrelated PRs, rather than from this branch.

afourniernv (approved):
LGTM, thank you for the contribution. We’ll merge this after 0.3 is cut today.

PR #628, opened 2026-09-04
docs(readme): make Relay and libsy quickstarts runnable

### LINKED ISSUE #624: docs(readme): make Relay and libsy quickstarts runnable
## Problem

The Relay and embedded-libsy quickstarts added in #613 do not run as written.

## What needs fixing

- The Relay path calls `package_bundle.py` without first building the plugin or passing its required `--library` and `--output` arguments. It also skips plugin enablement and Relay's trust requirements. Put one tested install flow in the plugin README and link to it from the root README.
- `pip install nemo-switchyard` currently installs 0.2.0, which does not export the documented `Step` or `LlmResponse` API. The Rust example pins the same release, where `run_stream` and its step variants also have the older shape. Either show the released API or clearly identify the version the example requires.
- `run_stream` and `LlmResponse.Agg` take normalized Switchyard request and response dictionaries, not OpenAI wire payloads. Update the wording and link to `examples/libsy.py` or the protocol types.
- The sample routing loop drops the remaining selected candidates and `outcome.response`. Use a complete helper that handles both cases, or keep the full example under `examples/` and link to it.
- The TOML reference currently points to `crates/switchyard-server/CONFIGURATION.md`, which is a contributor guide. It should point to `docs/reference/toml_schema.md`.

## Checks already run

- Built and packaged the native plugin, then validated and loaded it with NeMo Relay 0.8.1.
- Installed `nemo-switchyard` from PyPI in a clean environment and checked the documented imports.
- Ran raw OpenAI and normalized Switchyard request/response dictionaries through the current Python bindings.
- Ran the sample loop with Stage and Escalation to check fallback candidates and routing-produced responses.

Follow-up to #613.

itayzit:
Closes #624.

## What changed

- **Relay path**: the plugin README now has one install flow: `cargo build --release -p switchyard-nemo-relay-plugin`, `package_bundle.py --library ... --output ...`, then `nemo-relay plugins validate / add / enable / validate`. It explains the `enabled = false` manifest default, the artifact digest Relay verifies, and the optional signature policy. The root README Path 1 links to it instead of the incomplete `package_bundle.py` call.
- **Version**: Path 2 states that the `Step` / `LlmResponse` API is newer than `nemo-switchyard` 0.2.0 on PyPI and shows a source install. The Rust block depends on `main` instead of the `v0.2.0` tag, with a note to pin the tested `rev`.
- **Wording**: `run_stream` and `LlmResponse.Agg` are described as taking normalized Switchyard request and response dicts, linking to `switchyard-protocol` and `examples/libsy.py`.
- **Sample loop**: replaced with a `call_with_fallback` helper that walks `call.models` / `outcome.selected_model_ids` in order and returns `outcome.response` when routing already produced the answer.
- **TOML reference**: Path 1, Path 3, and the plugin README now link to `docs/reference/toml_schema.md` instead of the contributor guide.

## Checks

- Built the plugin in release mode and ran `package_bundle.py` with the documented arguments; the bundle contains the library, `config.schema.json`, licenses, and a manifest with the digest filled in. The `nemo-relay plugins` commands were not run locally (CLI not installed here); they follow Relay 0.8.1's discoverable-plugins guide.
- Ran the Path 2 Python blocks verbatim against `stage_router` with stub clients: efficient path served, and a failing efficient client fell back to capable.

## Summary by CodeRabbit

* **Documentation**
  * Updated setup instructions for the Relay plugin, including building, packaging, registration, configuration, validation, and signature policies.
  * Documented deployment paths, generated manifests, and dynamic-plugin configuration.
  * Added guidance for installing Python components from the repository and pinning Rust dependencies to a repository revision.
  * Updated examples to use normalized protocol requests and responses.
  * Refreshed links to the TOML schema and routing documentation.
  * Documented ordered model fallbacks and algorithm-generated responses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,19 +104,20 @@ step 1, stop when you reach the result named under the heading.
 ### Path 1 — Load the NeMo Relay Plugin
 
 You finish with an existing NeMo Relay deployment routing through Switchyard.
-Requires NeMo Relay `>=0.8.1,<0.9.0`.
+Requires NeMo Relay `>=0.8.1,<0.9.0` and a Rust toolchain to build the plugin.
 
-**1. Build the plugin bundle.**
-
-```bash
-python crates/switchyard-nemo-relay-plugin/scripts/package_bundle.py
-```
+**1. Build, package, and enable the plugin.** Follow the
+[install steps in the plugin README](crates/switchyard-nemo-relay-plugin/README.md#install).
+They build the shared library, package it into a bundle with a digest-bearing
+`relay-plugin.toml`, and register and enable it with `nemo-relay plugins`.
 
 **2. Write the Switchyard deployment** to `/etc/switchyard/routes.toml` — the
 same version-1 TOML the proxy uses. Copy the file from step 2 of Path 3 below.
 
-**3. Point Relay at the generated manifest.** Use exactly one deployment
-source: a path, as here, or the config nested under `switchyard_config`.
+**3. Point the plugin at the deployment.** Add a `config` table to the
+`[[plugins.dynamic]]` entry that `nemo-relay plugins add` wrote. Use exactly one
+deployment source: a path, as here, or the config nested under
+`switchyard_config`.
 
 ```toml
 [[plugins.dynamic]]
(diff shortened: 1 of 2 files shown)


afourniernv (changes requested):
I ran both quickstarts. The libsy example works and the docs build cleanly. I hit two issues in the Relay install flow:

- The macOS build produces a `.dylib`, but the packaging command points to the Linux `.so`.
- The final plugin validation runs before the required Switchyard config is added, so it fails. Could we configure the deployment before enabling and validating the plugin?

Once those are fixed, this looks good to me.

itayzit:
Ran the documented Relay flow end to end on macOS with the NeMo Relay 0.8.1 CLI binary from the GitHub release, using an isolated `XDG_CONFIG_HOME`:

1. `cargo build --release` + `package_bundle.py --library ...dylib --output ./plugins/switchyard`
2. `plugins validate <manifest>` and `plugins add --user` succeed. `plugins validate nvidia.switchyard` before configuring fails with a config-schema `oneOf` error, so the README now configures before enable/validate.
3. With the config but no policy override, `plugins validate` passes (`attestation_mode: integrity_only`) but gateway start fails with `dynamic plugin 'nvidia.switchyard' requires integrity.signature under host policy`. Adding `[plugins.policy.overrides."nvidia.switchyard"] attestation = "integrity_only"` to `plugins.toml` fixes it; the docs now list that override as required.
4. Gateway started, the plugin registered and activated, and `POST /v1/chat/completions` with `model = "switchyard"` returned 200 from a local mock upstream, served by the stage router's efficient target.

Not changed: pinning the source install and Cargo deps to a commit hash. The text tells readers to pin the `rev` they tested; a hash in the README would go stale on every release.

afourniernv (approved):
Retested the current head. The plugin builds and packages correctly on macOS, the linked libsy example runs, and both doc issues are fixed. Thanks for checking the Relay flow end to end.

Issue #601, opened 2026-09-02
[Proposal] Unified routing surface


ryan-lempka:
Proposal co-authored by @ayushag-nv 

**Background**

Today in Switchyard every routing algorithm ends by naming a model and every algorithm carries its own copy of the model catalog.

That suits a route running one strategy against a fixed pair of models. At wider scope it introduces these issues: the same two models are configured four different ways, strategies cannot be combined because each terminates on a model, and selection cannot weigh price or preferences because neither is available where the decision happens.

To successfully unify the existing algorithms there also needs to be a layer that makes a final decision when signals diverge. For instance stage tool signals say X, task classification says Y, and the advisor says Z all at the same time.

In addition, right now switchyard requires customers to categorize models (e.g. efficient/capable) which means they have to make a judgement call on what those definitions mean. Or use an existing configuration we provide that is "blessed". This proposal is a stepping stone (not the final state) towards being able to support an arbitrary pool of models without a need to categorize them explicitly and solves the immediate need of unifying the algorithms.

```mermaid
flowchart LR
  subgraph T["Today: every algorithm holds the catalog"]
    direction LR
    R1[Request] --> A1["Classifier<br/><i>knows sol, luna</i>"]
    R1 --> A2["Stage<br/><i>knows sol, luna</i>"]
    R1 --> A3["Advisor<br/><i>knows sol, luna</i>"]
    A1 --> M1[ModelId]
    A2 --> M1
    A3 --> M1
    M1 --> U1[Upstream]
  end

  subgraph P["Proposed: one place holds the catalog"]
    direction LR
    R2[Request] --> S["Signals<br/><i>task, tool loop, review</i>"]
    S -->|"needs + confidence"| O["Pool optimizer<br/><i>catalog, cost,<br/>budget, preferences</i>"]
    O --> M2[ModelId]
    M2 --> U2[Upstream]
  end
```

**Goals**

One vocabulary for declaring models. Strategies that combine. A deployment that works from a list of models. Less surface area in the algorithm layer.

**Non-Goals**

- Infrastructure signals in the pool optimizer
- Automatic discovery of model capability from providers
- A learned or online-optimizing policy

**Proposal**

Algorithms emit what a request needs. A pool optimizer decides which model provides it.

Three core changes:

**1. What an algorithm returns**

Today. The capability classifier computes an estimate, compares it to a threshold, and throws the estimate away:

```rust
[...]

PR #632, opened 2026-09-05
fix(escalation): require fresh consistent failure evidence

pst2154:
## What

- Normalize duplicated Terminus command serialization in the escalation judge's transcript without changing model-facing history.
- Require consecutive escalation verdicts to cite fresh evidence for the same bounded failure category before latching to the strong tier.
- Log the parsed category, freshness flag, and a bounded reason; update the packaged schema, prompt, tests, and routing documentation.

## Why

Agent harnesses can represent one command both as assistant JSON/text and as a structured tool call. The trajectory judge could count those same-turn representations as repeated failed attempts. More generally, boolean-only confirmations could combine stale evidence or unrelated failure modes into a false escalation.

The new confirmation rule keeps the existing one-way latch while requiring a coherent multi-turn failure pattern. This is a focused correctness improvement related to the signal-composition direction discussed in https://github.com/NVIDIA-NeMo/Switchyard/issues/601.

Live SWE-Atlas diagnostics with GLM 5.2 as the efficient tier and Opus 4.8 as the capable tier:

- Positive path, RF `task-697e7458be1623d850a88838`: two consecutive fresh `repetition` verdicts caused exactly one GLM-to-Opus transition; the run used 28 GLM and 36 Opus worker calls and received verifier reward 1.
- Recovery control on the same task: two `repetition` votes separated by a decline did not latch; the GLM-only run received verifier reward 1.
- Prior false-positive cases RF `task-69391d8d1ce51c407be1e531`, RF `task-694b4b99829f00e24fd11889`, and TW `task-6902ef3ab97fe23e2ad271f3` all remained GLM-only under the new rule. In the first case, positive votes changed from `repetition` to `false_progress`, so they did not combine into a latch.

These are routing diagnostics, not an aggregate accuracy claim: the positive-path task also passed in a separate direct-GLM rollout, and the false-positive controls still received reward 0. The evidence shows correct confirmation/latch behavior and avoided strong-tier calls on previously wasteful trajectories.

## Notes for reviewers

Start with `crates/libsy/src/algorithms/escalation.rs` for the confirmation state machine and `crates/libsy/src/algorithms/util/escalation.rs` for judge transcript normalization and typed verdicts.

Validation on Rust 1.96.1:

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test --workspace --all-features`
- `uv run ruff check .`
- `uv run mypy switchyard`
- wheel build/install through maturin
- `uv run pytest tests/ -v` (`117 passed`, `2 subtests passed`)

### DIFF
--- a/crates/libsy/src/algorithms/escalation.rs
+++ b/crates/libsy/src/algorithms/escalation.rs
@@ -12,7 +12,9 @@ use switchyard_protocol::{
 
 use super::util::classifier_contract::ClassifierContractConfig;
 use super::util::decisive;
-use super::util::escalation::{self, EscalationJudge, EscalationJudgeConfig, EscalationPolicy};
+use super::util::escalation::{
+    self, EscalationCategory, EscalationJudge, EscalationJudgeConfig, EscalationPolicy,
+};
 use super::util::llm_judge::JudgeClassifier;
 use crate::core::algorithm::Driver;
 use crate::core::classifier::{Classification, Classifier};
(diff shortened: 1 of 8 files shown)


pst2154:
Superseded by #637, which carries the current fresh-evidence implementation and paired SWE-Atlas report. The separate directional/model-order follow-up is #638.

PR #633, opened 2026-09-05
fix(server): omit absent request-log ids instead of logging them empty (#301)

### LINKED ISSUE #301: Session affinity keys on an empty session id, and is silent when it cannot key at all

chethanuk:
## What

- Drop `.unwrap_or("")` on `session_id`/`correlation_id` in `RequestLogContext::emit` and `emit_cancelled`, so an absent id omits the field instead of logging it as empty.
- Widen the test subscriber to capture event fields (not just the message) and add a table test over absent/empty/present id shapes.

## Why

The terminal request log renders `session_id=` for both "no session was sent" and "an empty session was sent" — an operator can't tell them apart. This is part 3 of #301; parts 1 and 2 landed in #308.

Closes NVIDIA-NeMo/Switchyard#301.

## Notes for reviewers

`requested_model`, `selected_model`, and `error` keep `unwrap_or("")` on purpose: `emit_cancelled` hardcodes `selected_model = ""`, so those three mean "absent implies empty" by design. The only non-test producer of `session_id` (`resolve_path`) never returns `Some("")` today, so the empty-string test row is a regression guard rather than a live bug — the user-visible change is that an absent session stops printing `session_id=`.

## Test plan

- `cargo test -p switchyard-server request_log` — new test fails today (`left: Some(""), right: None`), 4 passed after the fix
- `cargo fmt --all --check`
- `cargo test --workspace` — 586 passed, 0 failed, 1 ignored

`cargo clippy --workspace` fails on `crates/prefill-router` (`chunks_exact`), pre-existing and untouched here — reproduces on `origin/main` alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Request logs now omit session and correlation IDs when they are unavailable, instead of displaying empty values.
  - Logging consistently preserves the distinction between missing and empty identifiers for both completed and cancelled requests.

- **Documentation**
  - Added an unreleased changelog entry describing the updated request-log behavior.

### DIFF
--- a/crates/switchyard-server/src/lib.rs
+++ b/crates/switchyard-server/src/lib.rs
@@ -1132,8 +1132,8 @@ impl RequestLogContext {
                     requested_model = self.requested_model.as_deref().unwrap_or(""),
                     selected_model,
                     streaming = self.streaming,
-                    session_id = self.session_id.as_deref().unwrap_or(""),
-                    correlation_id = self.correlation_id.as_deref().unwrap_or(""),
+                    session_id = self.session_id.as_deref(),
+                    correlation_id = self.correlation_id.as_deref(),
                     handling_duration_ms = duration_ms,
                     error,
                     $message
@@ -1160,8 +1160,8 @@ impl RequestLogContext {
             requested_model = self.requested_model.as_deref().unwrap_or(""),
             selected_model = "",
             streaming = self.streaming,
-            session_id = self.session_id.as_deref().unwrap_or(""),
-            correlation_id = self.correlation_id.as_deref().unwrap_or(""),
+            session_id = self.session_id.as_deref(),
+            correlation_id = self.correlation_id.as_deref(),
             handling_duration_ms = self.started.elapsed().as_secs_f64() * 1_000.0,
             error = "client disconnected before a response was written",
             "LLM request cancelled"
@@ -1692,6 +1692,7 @@ mod tests {
     use switchyard_llm_client::LlmCallObservation;
     use tokio::io::{AsyncReadExt, AsyncWriteExt};
     use tokio::sync::{Notify, oneshot};
+    use tracing::field::{Field, Visit};
 
     use super::*;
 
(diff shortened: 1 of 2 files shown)


crates/switchyard-server/src/lib.rs
```diff
@@ -1908,9 +1909,28 @@ mod tests {
         let _ = tokio::time::timeout(Duration::from_secs(5), server).await;
     }
 
+    // One terminal request event: its level, its message, and every other field it
+    // recorded. A field the event never recorded is a missing key, not an empty value.
+    type CapturedEvent = (Level, String, BTreeMap<String, String>);
```
messiaen, line 1914: I would make this a struct instead of tuple.

messiaen:
@nachiketb-nvidia can you review this one?

afourniernv:
@chethanuk could you switch the capture type to a named struct as requested in the open review thread? @nachiketb-nvidia gentle bump for review once that is updated.

PR #634, opened 2026-09-05
feat(libsy): map graded classifier verdicts onto routing targets (#348)

### LINKED ISSUE #348: feat(libsy): map graded classifier verdicts onto routing targets
## Problem

#275 gave custom mode a `target_selector` policy: a JSON Pointer reads one field out of the verdict and its **value is used as a target label**. That works when the rubric asks "which target?" and the judge answers with a target name.

Rubrics more often ask for a *judgement on a scale that means something to the model* — difficulty, complexity, risk — and let the deployment decide what each grade routes to:

```json
{"recommended_tier": "complex"}
```

Today the only way to route that is to make the rubric answer `"strong"` instead. That is worse prompting: it asks the model to apply the deployment's routing policy rather than to assess the task, and it means every operating point (conservative / balanced / aggressive) needs its own rewritten rubric even though the *assessment* is identical and only the mapping differs.

`targets` cannot express the mapping either — it is the set of models the route may dispatch to, so two entries resolving to one model are rejected.

## Proposal

An optional `labels` map on the `target_selector` policy:

```toml
targets = ["weak", "strong"]
default_target = "strong"

[routes.assistant.policy]
type = "target_selector"
selector = "/recommended_tier"
labels = { simple = "weak", medium = "weak", complex = "strong", reasoning = "strong" }
```

Several verdict values may share a target. Omit `labels` and verdict values are target labels exactly as they are today — this is purely additive.

Validation at load time, because both failures are silent at runtime otherwise:

- every value in `labels` must name a configured target;
- the table must not be empty (every verdict would be unroutable, which is a config mistake rather than a policy).

A verdict value absent from the map is unroutable in the same way an unknown target label already is, so `default_target` decides.

## Why this is worth a key rather than a prompt

The prompt workaround is real and we have been running it — an appended line telling the judge to answer with the routing target rather than the grade. It costs a rubric edit per operating point, and it silently couples prompt text to routing config: change `targets` and the prompt is wrong with nothing to catch it. Moving the mapping into config is what makes one rubric serve several operating points.

Builds directly on #275; no change to the selector mechanism itself. Rebased onto current `main` and green.

chethanuk:
## What

Adds an optional `labels` map to the custom classifier's `target_selector` policy, mapping each verdict value onto a configured target name:

```toml
targets = ["weak", "strong"]
default_target = "strong"

[routes.assistant.policy]
type = "target_selector"
selector = "/recommended_tier"
labels = { simple = "weak", medium = "weak", complex = "strong", reasoning = "strong" }
```

Several verdicts may share a target. Omitting `labels` keeps today's behaviour exactly. Empty tables and values naming no configured target are rejected when the deployment loads. A verdict outside the map falls back to `default_target`, same as an unknown target label today, and now logs a `warn!` naming the value that missed.

`CustomClassifierPolicy::TargetSelector` gains a field, so external code matching that variant by its fields must be updated. Recorded under **Changed** in the changelog. `CustomClassifierPolicy::target_selector(selector)` is unchanged.

## Why

Closes #348.

A rubric usually asks for a judgement — difficulty, complexity, risk — not a model name, but a verdict like `complex` cannot route today. `build_custom` fills `target_map` and `resolved_targets` in one pass over `targets`, so one list does two jobs: the fallback chain, which must not repeat a ModelId, and the vocabulary the judge may answer with, which may. Making a grade sayable means adding it to `targets`, where the duplicate-ModelId rejection then fires — correctly for the chain, wrongly for the vocabulary.

## Notes for reviewers

Start at the `match policy` arm in `build_custom` (`crates/libsy/src/algorithms/llm_class.rs`) — that's the whole change. Everything else is the field, the serde plumbing, and the docs.

With `labels` set, `response_schema` must enumerate the **verdict** values rather than the target names; nothing validates that pairing today, and getting it wrong sends all traffic to `default_target`.

## Test plan

```
cargo test -p switchyard-libsy --lib algorithms::llm_class        ok. 24 passed; 0 failed
cargo test -p switchyard-runner --lib config::                    ok. 39 passed; 0 failed
cargo test -p switchyard-server --test server custom_classifier   ok.  2 passed; 0 failed
cargo fmt --all -- --check                                        clean
cargo test --workspace --no-fail-fast                              688 passed / 34 targets
```

`cargo clippy --workspace --all-targets -- -D warnings` fails only on `crates/prefill-router/src/transformers.rs`, untouched by this branch and identical on `main`.

## Summary by CodeRabbit

* **New Features**
  * Custom classifier routes can map verdict labels to configured targets.
  * Multiple verdict labels can share a target, with unmatched verdicts using the configured default target.
  * Existing behavior is preserved when no label mapping is provided.
  * Invalid or missing mappings produce clearer warnings and validation errors.

* **Documentation**
  * Added configuration guidance, examples, validation requirements, and fallback behavior for verdict-based routing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,13 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- **Graded custom-classifier verdicts** — the `target_selector` policy takes an
+  optional `labels` map from verdict value to configured target, so a rubric can
+  grade a request (`simple` / `complex`) instead of naming a model and several
+  grades may share one target. Omitting `labels` keeps today's behavior, but when
+  it is set it is the whole verdict vocabulary and a verdict outside it falls back
+  to `default_target`. `CustomClassifierPolicy::TargetSelector` gains a field, so
+  external code matching that variant by its fields must be updated. (#348)
 - **`Algorithm::route` returns `Result<RoutingOutcome>`** — instead of the
   bare final `Result`, so callers observe the full routing outcome (see #458
   for the design). (#459)
--- a/crates/libsy/src/algorithms/llm_class.rs
+++ b/crates/libsy/src/algorithms/llm_class.rs
@@ -381,6 +381,12 @@ pub enum CustomClassifierPolicy {
     TargetSelector {
         /// JSON Pointer evaluated against each schema-validated verdict.
         selector: String,
+        /// Maps each verdict value the judge may return onto a configured target label.
+        ///
+        /// `None` keeps the direct lookup, where the verdict is itself a target label.
+        /// When set, this map is the whole verdict vocabulary: several verdicts may share
+        /// one target, and a verdict outside the map falls back to `default_target`.
+        labels: Option<BTreeMap<String, String>>,
     },
 }
 
@@ -389,6 +395,7 @@ impl CustomClassifierPolicy {
     pub fn target_selector(selector: impl Into<String>) -> Self {
         Self::TargetSelector {
             selector: selector.into(),
+            labels: None,
         }
     }
 }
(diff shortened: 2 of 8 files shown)


bhuvan002:
Hi @chethanuk, thank you for the contribution! We recently merged #630 to main which makes it so that algorithms now work with `Category` instead of `ModelId` and adds runtime model groups, so I am closing this PR as that should cover the same cases.

PR #636, opened 2026-09-05
fix(libsy): preserve provider events when escalation serves a buffered reply

linj-glitch:
## Problem

The escalation router calls the efficient tier, buffers the reply so the trajectory judge can read the completed turn, and then serves that same buffered reply whenever the judge declines. It rebuilt the served stream with `AggLlmResponse::into_stream()`, which documents itself as lossy:

> This conversion is lossy: only text, reasoning, and tool-call content has a synthetic chunk representation. Refusals, tool results, media, files, unknown blocks, response extensions, and preservation metadata are omitted.

The loss happened one step earlier as well. `LlmResponse::into_agg()` consumed only each event's `normalized` chunks and dropped its `preservation` payload, so the original provider bodies were already discarded before the response was rebuilt.

The result is that every unlatched escalation turn reached the outbound codec as synthetic chunks with no preserved provider bodies, so the codec could not produce a faithful same-format response. Escalation is the only route type that buffers a reply and then serves it, so no other route is affected. Latched turns are also unaffected, because a confirmed session returns before the buffering path.

This is invisible in normal operation: the requests all succeed, and the degradation only shows up as reduced agent quality on the turns that were served from the buffer.

## Change

- `LlmResponse::into_agg_retaining_events(retain_events)` aggregates as before, and optionally returns the original stream events alongside the aggregate. `into_agg()` now delegates to it and keeps its existing behaviour.
- `replay_stream_events()` serves previously buffered events verbatim, so `preservation` survives.
- The escalation classifier retains events when the inbound request is streaming and replays the originals instead of re-synthesising them. Non-streaming requests keep returning the aggregate.

The judge still reads the aggregated reply, so verdict behaviour is unchanged.

## Testing

- New `retained_events_replay_with_preservation_intact` asserts both halves of the contract: replayed events keep the provider payload, and the `into_stream()` path still emits synthetic events without preservation.
- `cargo test -p switchyard-protocol -p switchyard-libsy --lib` passes: 306 tests, 0 failures.

Found while benchmarking the escalation router with Codex on DeepSWE-v1.1, where the buffered path is taken on every turn until a session latches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming responses that bypass escalation by preserving and replaying the original provider events.
  * Provider-specific payloads and event details are now retained more accurately during streamed responses.
  * Non-streaming responses and error handling remain unchanged.

### DIFF
--- a/crates/libsy/src/algorithms/escalation.rs
+++ b/crates/libsy/src/algorithms/escalation.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use switchyard_protocol::{
     AggLlmResponse, LlmClientError, LlmResponse, Message, ModelId, Request, Response, Role,
+    replay_stream_events,
 };
 
 use super::util::classifier_contract::ClassifierContractConfig;
@@ -116,8 +117,17 @@ impl Classifier<State> for EscalationClassifier {
         };
         // The call resolves when its stream handle arrives; transport can still fail while
         // buffering. Fall back only for that availability failure and keep other errors typed.
-        let agg = match efficient_response.llm_response.into_agg().await {
-            Ok(agg) => agg,
+        // Retain the provider events: the buffered reply is served downstream when the judge
+        // declines, and replaying the originals keeps `preservation` intact. Rebuilding from the
+        // aggregate instead would emit synthetic chunks and drop the provider payloads the
+        // outbound codec needs, degrading every unlatched turn.
+        let streaming = request.llm_request.stream;
+        let (agg, retained_events) = match efficient_response
+            .llm_response
+            .into_agg_retaining_events(streaming)
+            .await
+        {
+            Ok(pair) => pair,
             Err(LlmClientError::Transport { .. }) => {
                 return Ok((decisive(&self.capable), None));
             }
@@ -132,8 +142,8 @@ impl Classifier<State> for EscalationClassifier {
             .messages
             .push(assistant_message(&agg));
         let efficient_response = Response {
-            llm_response: if request.llm_request.stream {
-                LlmResponse::Stream(agg.into_stream())
+            llm_response: if streaming {
+                LlmResponse::Stream(replay_stream_events(retained_events))
             } else {
                 LlmResponse::Agg(agg)
             },
--- a/crates/protocol/src/stream.rs
+++ b/crates/protocol/src/stream.rs
@@ -126,6 +126,16 @@ impl From<LlmResponseChunk> for LlmResponseStreamEvent {
     }
 }
 
+/// Replays previously buffered events as a stream, preserving their provider payloads.
+///
+/// Pair with [`LlmResponse::into_agg_retaining_events`] when a caller had to buffer a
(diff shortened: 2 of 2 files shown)


linj-glitch:
Converting to draft: this change regresses in end-to-end testing and is not ready to merge.

The underlying defect is real — `into_agg()` discards each event's `preservation` payload, so escalation's buffered replies are rebuilt as synthetic chunks — but replaying the retained provider events verbatim is not a correct fix on its own.

Benchmarking the escalation router with Codex against a Kimi-K3 efficient tier, the same five tasks behave as follows:

- without this change: 5/5 evaluations complete, no infrastructure failures
- with this change: 3/5 fail outright, and the upstream rejects requests with

```
Kimi K3 tool messages need a resolvable tool name: carry `tool`/`name`,
or match a preceding assistant tool_call by order.
```

So replaying the original events changes what the client records for the assistant turn, and the tool-call/tool-result pairing in the client's next request no longer resolves upstream. The synthetic path, whatever else it loses, at least emits tool calls in a shape the client echoes back correctly.

Leaving this open as a draft because the preservation loss is worth fixing; the fix needs to keep the replayed items consistent with what the outbound codec and client expect, which this does not yet do. I will update once I have a version that holds up end to end.

linj-glitch:
Closing this. I found why the change regresses, and the approach is wrong.

Replaying the retained provider events serves the upstream body verbatim, which bypasses `TranslationEngine::encode_response_with_extensions` and therefore `codex_namespaces::restore_qualified_tool_names`. Codex tools are namespaced, the request codec flattens them to `<namespace>__<tool>` for OpenAI-compatible upstreams, and the response encoder restores the namespaced spelling using the mapping carried in the request's `ProviderExtensions`. Serving raw events skips that step, so the client records the flattened name, echoes it back, and the upstream can no longer resolve the tool call.

Reproduced by diffing the upstream request bodies with and without the change, on the same client session. The conversation the client replays differs:

```
with change:    message, reasoning, function_call(call_1), function_call_output(call_1)
without change: message, message,   function_call(call_1), function_call_output(call_1)
```

End to end against a Kimi-K3 efficient tier, the same five tasks go from 5/5 evaluations completing (unchanged) to 3/5 failing outright, with the upstream returning:

```
Kimi K3 tool messages need a resolvable tool name: carry `tool`/`name`,
or match a preceding assistant tool_call by order.
```

So the lossiness documented on `AggLlmResponse::into_stream` is partly load-bearing: rebuilding from the aggregate is what routes the served reply back through response encoding, including tool-name restoration.

The underlying observation still stands and may be worth addressing separately: `LlmResponse::into_agg` discards each event's `preservation` payload, so a caller that must buffer a reply and then serve it cannot produce a faithful same-format response. A correct fix would need to keep the provider payload *and* still apply response-encoding concerns like namespace restoration, rather than bypassing them. That is a design decision for maintainers who know the translation layer better than I do, so I am not going to guess at it here.

Worth noting the passthrough route serves the same provider events without this problem, so whatever passthrough does differently is probably the right model for a future fix.

PR #637, opened 2026-09-05
fix(escalation): require fresh consistent failure evidence

pst2154:
## Summary

- deduplicate equivalent Terminus raw-command and structured-tool serialization only in the escalation judge transcript
- require fresh, same-category failure evidence across distinct turns before switching
- preserve the maximum-one-switch and post-switch latch invariants
- add focused Rust coverage and a 40-task paired SWE-Atlas report

## Benchmark evidence

On 20 RF and 20 TW tasks, the patched router scored 15/40 versus 10/40 for the original router and reduced switches from 9 to 5. Under the documented cached-token-aware synthetic rates, estimated cost fell from $1.171/task to $0.889/task. Direct GLM also scored 15/40 at $0.501/task, so this PR claims a false-positive fix and improvement over the original escalation policy, not superiority over the best fixed model.

Full methodology, task-level observations, formulas, limitations, and validation evidence are in benchmark/SWE_ATLAS_ESCALATION_REPORT.md.

## Validation

- cargo fmt --all --check
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace --all-features
- uv run ruff check .
- uv run mypy switchyard
- uv run maturin develop --uv
- uv run pytest tests/ -v

All passed on Rust 1.96.1. Pytest reported 115 passed, 2 deselected, and 2 subtests passed.

### DIFF
--- a/benchmark/SWE_ATLAS_ESCALATION_REPORT.md
+++ b/benchmark/SWE_ATLAS_ESCALATION_REPORT.md
@@ -1,0 +1,81 @@
+# SWE-Atlas escalation-router findings
+
+## Summary
+
+The escalation router could interpret duplicate command serialization in a NeMo Gym Terminus
+trajectory as repeated failed work. A single assistant turn may contain the same command both as
+raw JSON text and as a structured tool call. The judge prompt previously treated a command shown
+two or more times as loop evidence without requiring those attempts to occur in distinct turns.
+
+This change makes escalation depend on fresh, consistent failure evidence across turns. It keeps
+the existing one-way switch and latch: a session can move from the efficient model to the capable
+model at most once and remains there afterward.
+
+## Changes
+
+- Normalize only the transcript sent to the escalation judge. When a raw Terminus command batch
+  has a one-to-one match with structured tool calls in the same assistant turn, the redundant raw
+  representation is omitted. The worker model still receives the original history.
+- Require the judge to classify positive evidence as `repetition`, `false_progress`, `drift`,
+  `desperation`, or `capability_gap`, and report whether the evidence is new.
+- Advance the confirmation streak only for fresh positive evidence in the same category. A
+  category change starts a new streak; a decline or stale evidence resets it. An unavailable judge
+  preserves the prior streak rather than converting an infrastructure failure into a routing
+  decision.
+- Clarify in the judge prompt that duplicate serialization within one turn and an immediate,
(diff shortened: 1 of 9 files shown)


linj-glitch:
I reviewed the latest rebased head. The fresh-evidence and same-category confirmation design makes sense, and the transcript-only Terminus deduplication is a good fit for the observed false positive. A few comments before merge:

1. The current head does not compile under cargo check -p switchyard-libsy. crates/libsy/src/algorithms/escalation.rs:170 calls JudgeClassifier::verdict without the new judge_models argument, and crates/libsy/src/algorithms/util/escalation.rs:382 calls collect_text without its new tool_commands argument. These look like omissions from the latest rebase; the validation recorded in the report appears to predate that head.

2. crates/libsy/src/algorithms/escalation.rs:177-182 logs the judge-generated reason at INFO. Bounding and stripping control characters prevents log injection but does not prevent the reason from echoing sensitive transcript content. I would keep category and new_evidence at INFO and either omit the reason or retain it only at DEBUG.

3. I read the benchmark as good evidence for fixing the concrete duplicate-serialization false positive, but not yet as evidence that escalation beats a fixed model generally. It is one stochastic trial per task; patched escalation ties direct GLM at 15/40 while remaining materially more expensive, and some switches occur very late. The report mostly states these limitations clearly.

Overall, the #637 direction is sound. The two build errors and logging level are the actionable items I found.

pst2154:
Codex here — I addressed the actionable review feedback and pushed commit `8fd14f2b` after rebasing onto current `main`.

- Passed the runtime `Judge` model list to `JudgeClassifier::verdict`.
- Updated `collect_text` callers for the new tool-command argument.
- Removed the judge-generated free-text `reason` from INFO logging; only the structured escalation fields remain.
- Updated the escalation tests to use the current runtime model map.
- Updated both server mock judges to emit the typed `category`/`new_evidence` verdict schema. The second fixture was found by the server state-ownership integration test during validation.
- Clarified the report with explicit post-rebase validation evidence. Its benchmark conclusion remains intentionally narrow: this supports the concrete false-positive fix, not a claim that routing generally beats the best fixed model.

Fresh validation used Rust 1.96.1 and Python 3.12: `cargo fmt --all --check`, workspace clippy with all targets/features, and workspace tests with all features all passed. That includes 323 libsy tests and all 56 server integration tests. No live provider calls were made.

PR #641, opened 2026-09-08
feat(server): record request origin in routing logs

### LINKED ISSUE #640: [feature] Record request origin in the durable routing log
## Problem

The per-request routing JSONL (`--routing-log-file`, `crates/switchyard-server/src/routing_log.rs`) records `task`, `trial_id`, and `session_id`, but nothing that identifies **who/what sent the request**. A gateway serving several harnesses (Claude Code, Codex CLI, custom agents, direct API callers) can't answer "which client produced this traffic?" from the durable log alone.

Downstream consumers tail the JSONL and can only show fields that exist in the record — origin cannot be recovered after the fact, because request headers aren't persisted anywhere durable.

## Proposed change

Add an optional `origin` field to `RoutingRecord`, extracted in `RoutingLogContext::from_metadata` exactly like the existing `task`/`trial_id` headers:

- Primary source: new `x-switchyard-origin` request header.
- Fallback (optional, mirrors the `LEGACY_SESSION_ID_HEADER` precedent): derive a coarse origin from `User-Agent` when the header is absent — e.g. `claude-code`, `codex-cli`, `openai-sdk`.

`RoutingRecord` already uses `#[serde(default)]`, so old and new records coexist on read (`snapshot()`) with no migration.

```rust
const ORIGIN_HEADER: &str = "x-switchyard-origin";

// in RoutingLogContext:
origin: Option<String>,

// in from_metadata:
origin: headers
    .and_then(|headers| nonempty_header(headers, ORIGIN_HEADER))
    .map(str::to_string),
```

## Why this is cheap

`RoutingLogContext::from_metadata` (routing_log.rs, commit `c665174`) already plucks `x-switchyard-intake-task` / `x-switchyard-trial-id` out of `metadata.http_headers` — origin is the same pattern one line over. The field rides into the record in `RoutingLog::append`; no new plumbing beyond that.

## Consumers

At least one known consumer: [switchyard-studio](https://github.com/suncanyon/switchyard-studio) tails this JSONL into SQLite for a routing-visibility console. It parses a field allowlist, so an additive field is ignored by existing readers and picked up with a small parse change on the consumer side.

We run the routing log enabled in production (`x-switchyard-session-id` already flows end-to-end), so we can validate the field live immediately after it lands, and are happy to test a prototype build.

ting-hong-shieh:
## What

Record the caller-supplied `x-switchyard-origin` header as optional `origin` in the durable routing JSONL. Buffered answers, streamed answers, and judge records retain the same request origin. Missing, empty, or non-text values serialize as `null`; older records remain readable.

## Why

A gateway serving multiple clients needs to identify their traffic from the durable log. For example, `x-switchyard-origin: codex-cli` now produces `"origin":"codex-cli"` in its routing records.

Closes #640.

## Notes for reviewers

The implementation follows the existing task/trial header extraction in `routing_log.rs`. This implements the explicit header only; the optional User-Agent inference is deferred.

Pre-PR adversarial review checked quoted/backslash values, missing and invalid headers, mixed old/new JSONL, context retention through stream completion, and origin propagation to both judge and answer records. The judge/answer assertion was added during review. No unresolved findings remain.

Validation on macOS with Rust 1.96.1 and Python 3.11:
- `cargo test -p switchyard-server routing_log`: 7 passed.
- `PYO3_PYTHON=.venv/bin/python cargo test --workspace`: 699 passed, 1 ignored (requires `PREFILL_ROUTER_HANDOFF_DIR`).
- `PYO3_PYTHON=.venv/bin/python cargo clippy --workspace --all-targets -- -D warnings`: passed.
- `cargo fmt --all --check` and `git diff --check`: passed.
- `uv run ruff check .` and `uv run mypy switchyard`: passed.
- `uv run pytest tests/ -m "not integration"` with provider keys unset: 115 passed, 2 deselected.

No live provider tests were run.

## Summary by CodeRabbit

* **New Features**
  * Routing logs now record valid values from the `x-switchyard-origin` request header.
  * Origin metadata is preserved for streaming requests and advisor routes.
  * Missing, empty, or invalid origin values are recorded as absent.
* **Documentation**
  * Added documentation describing origin recording, validation, and compatibility with older routing records.

### DIFF
--- a/crates/switchyard-server/README.md
+++ b/crates/switchyard-server/README.md
@@ -102,6 +102,11 @@ served model. The legacy `proxy_x_session_id` remains a fallback when no normali
 present. The endpoint returns `404` when the session has no records and is not registered when
 routing logging is disabled.
 
+Clients can send `x-switchyard-origin: codex-cli` (or another client label) to include an
+`origin` field in each routing record. Missing, empty, or non-text header values produce
+`"origin": null`. The value is supplied by the caller; it is not inferred from `User-Agent`.
+Older records without `origin` remain readable by the session stats endpoint.
+
 An `llm_classifier` route sends each task to `classifier_target` for a capability verdict, then
 routes to `weak_target` or `strong_target`. Beyond the three targets it accepts these keys; only
 `base_threshold` is required, and anything the judge cannot decide routes to `strong_target`:
--- a/crates/switchyard-server/src/routing_log.rs
+++ b/crates/switchyard-server/src/routing_log.rs
@@ -18,6 +18,7 @@ use crate::usage_metrics::token_usage;
 use crate::{ServerError, ServerResult};
 
 const LEGACY_SESSION_ID_HEADER: &str = "proxy_x_session_id";
+const ORIGIN_HEADER: &str = "x-switchyard-origin";
 const TASK_HEADER: &str = "x-switchyard-intake-task";
 const TRIAL_ID_HEADER: &str = "x-switchyard-trial-id";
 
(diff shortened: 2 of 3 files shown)


bhuvan002 (commented):
Thank you for the contribution, LGTM!

bhuvan002 (approved):

PR #645, opened 2026-09-08
fix(translation): replay responses reasoning as input history

sabhatinas:
## Linear
- [SWITCH-1416](https://linear.app/nvidia/issue/SWITCH-1416/fix-responses-reasoning-replay-after-stage-router-handoff-notes)

## Summary
- Preserve encrypted Responses reasoning items through decode/encode after exact replay is invalidated.
- Re-emit reasoning input history using `summary`/`encrypted_content` fields and omit non-empty `content` arrays.
- Add a regression test for Codex-style encrypted reasoning after target prompt mutation.

## Why
Benchmark failure with Switchyard stage-router profile with `openai_responses`, Luna as efficient, and Sol as capable. Luna handled the first Responses calls successfully. After tool signals triggered escalation, the stage router added a handoff note to the normalized request IR as plain user text. That intentionally cleared exact same-format replay so the note could reach Sol.

Once exact replay was cleared, the Responses codec reconstructed `input` from IR. The old encoder reconstructed prior reasoning history as a top-level `reasoning` input item with a non-empty `content` array using `reasoning_text`. Sol rejected that replay shape with HTTP 400: `Invalid input[2].content: array too long. Expected an array with maximum length 0, but got an array with length 1 instead.`

This fix keeps the handoff-note behavior unchanged and fixes the reconstruction path for prior encrypted reasoning.

## Validation
- `cargo fmt --all --check`
- `cargo test --workspace` (rerun outside sandbox because wiremock binds local ports)
- `uv run ruff check .`
- `uv run mypy switchyard`
- `uv run pytest tests/ -v --ignore=tests/e2e`

Note: full `uv run pytest tests/ -v` stops locally at the Docker e2e test because the Docker daemon is not running.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -246,6 +246,9 @@ impl TranslatingLlmClient {
         // which keeps the caller's original `model`; force the resolved model so
         // the upstream always sees the target id.
         set_json_model(&mut body, model);
+        if matches!(backend, Backend::OpenAiResponses(_)) {
+            sanitize_openai_responses_provider_body(&mut body);
+        }
         // Strip before `merge_extra_body` so a target can reinstate either field
         // deliberately via `extra_body`.
         if matches!(backend, Backend::Anthropic(_)) {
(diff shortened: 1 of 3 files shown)


linj-glitch:
This overlaps with one commit that was in #646 (replaying reasoning history to Responses in the summary_text shape with encrypted_content and id, no content array); I dropped that commit from #646 in favour of this PR. One coordination note for whichever lands second: #646's response-side encoders read the encrypted payload and item id through shared helpers, which I taught to accept the verbatim reasoning item you store as the detail, so a reply that is buffered (for a judge) and then re-streamed keeps its encrypted content under the original id with your decoder shape. The remaining overlap is a textual conflict in decode_responses_reasoning_item in buffered.rs. For what it is worth, this same 400 ("array too long, maximum length 0") reproduced in our Sol+Luna escalation runs once the request was no longer exact-replayable, so the fix is needed by every transforming route, not only the stage router.

linj-glitch (approved):
Reviewed the diff against the failure we both hit (Sol rejecting replayed reasoning with "input[N].content: array too long, maximum length 0"). The change is correct: input reasoning items now carry only id, summary and encrypted_content, the provider's content array is never replayed, and keeping the whole item as the detail preserves the id the encrypted payload was issued under, which OpenAI verifies. The updated round-trip test and the new prompt-mutation test cover both the plain and the encrypted case, and CI is green.

One non-blocking observation. When a reasoning block has neither text nor an encrypted detail, encode_responses_reasoning_input returns None and the block falls through to encode_responses_content, which treats a reasoning-only message as text and emits an assistant message item with an empty string as content. Codex always requests encrypted content so this does not arise there, but a client that replays an empty-summary reasoning item without encrypted content would now send an empty assistant message instead of nothing. Skipping the message when the only block was an omitted reasoning item would close that gap; fine as a follow-up.

For coordination with #646: its response-side helpers already accept the verbatim reasoning item you store as the detail, so a reply that is buffered for a judge and re-streamed keeps the payload under the original id with this decoder shape. The only expected overlap is a textual conflict in decode_responses_reasoning_item for whichever lands second.

PR #649, opened 2026-09-09
fix(llm-client): drop upstream urls from transport errors

### LINKED ISSUE #423: [bug]: transport 502 echoes base_url including ?key=

afourniernv:
## What

Removes the upstream request URL from transport and timeout errors before they leave `switchyard-llm-client`.

The same conversion now covers buffered requests and streaming body failures. The server's raw fallback proxy applies the same protection because it sends requests without going through `switchyard-llm-client`.

## Why

Fixes [SYGH-64](https://linear.app/nvidia/issue/SYGH-64/bug-transport-502-echoes-base-url-including-key) and closes [#423](https://github.com/NVIDIA-NeMo/Switchyard/issues/423).

A failed upstream request currently returns the full reqwest URL in the client-visible 502. If a deployment uses a Gemini-style `base_url` containing `?key=...`, that response returns the configured key to the caller.

I reproduced this on untouched `main` with a closed local port:

```text
error sending request for url (http://127.0.0.1:1/v1?key=SYGH64_CANARY/chat/completions)
```

After this change, the response keeps the existing status, type, and code without the URL:

```json
{
  "error": {
    "message": "error sending request",
    "type": "upstream_error",
    "code": "upstream_error"
  }
}
```

## Behavior

- `LlmClientError::Transport` and `LlmClientError::Timeout` remain typed errors with their existing sources.
- The attached reqwest error no longer carries the upstream URL.
- HTTP status codes, error codes, retries, and fallback behavior are unchanged.
- No public APIs or dependencies change.

This uses the `reqwest::Error::without_url()` approach from Ryan's draft [#587](https://github.com/NVIDIA-NeMo/Switchyard/pull/587). The original server-side investigation is in [#428](https://github.com/NVIDIA-NeMo/Switchyard/pull/428).

## How tested

- [x] Live local reproduction for OpenAI Chat, streaming Chat, OpenAI Responses, Anthropic Messages, and the raw fallback proxy. Every request returned 502 without the canary.
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/` — 117 passed
- [x] Commit carries the required DCO sign-off.

## Notes for reviewers

The production change is five added lines across the shared reqwest error conversion and the raw fallback path. The remaining additions are regression coverage for both client-visible 502 paths.

### DIFF
--- a/crates/libsy-llm-client/src/client.rs
+++ b/crates/libsy-llm-client/src/client.rs
@@ -464,17 +464,9 @@ impl TranslatingLlmClient {
                 // Adapt the reqwest body stream to plain bytes; the SSE-decode itself is
                 // transport-agnostic and lives in `switchyard-translation`.
                 let bytes = http_response.bytes_stream().map(|chunk| {
-                    chunk.map(|bytes| bytes.to_vec()).map_err(|error| {
-                        if error.is_timeout() {
-                            LlmClientError::Timeout {
-                                source: Box::new(error),
-                            }
-                        } else {
-                            LlmClientError::Transport {
-                                source: Box::new(error),
-                            }
-                        }
-                    })
+                    chunk
+                        .map(|bytes| bytes.to_vec())
+                        .map_err(convert_reqwest_error)
                 });
                 let mut chunks = decode_stream(bytes, wire_format)?;
                 // Providers reject an over-ceiling streaming request with an in-band
@@ -707,6 +699,7 @@ fn record_gen_ai_request(url: &str, model: &str, streaming: bool) {
 fn convert_reqwest_error(error: reqwest::Error) -> LlmClientError {
     // Reqwest labels truncated or otherwise unreadable response bodies as decode
     // errors, so distinguish them from serde JSON failures at the call site.
+    let error = error.without_url();
     if error.is_timeout() {
         LlmClientError::Timeout {
             source: Box::new(error),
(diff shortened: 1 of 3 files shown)


bhuvan002 (approved):
Thank you for the fix, approved!

PR #651, opened 2026-09-09
fix(libsy): avoid zero exit tool-signal matches

sabhatinas:
## Summary

- Correct Codex tool signals so clean shell exits stay clean and genuine failures escalate reliably.
- Add a configurable capable-tier recovery window (`capable_hold_turns`, default 2), released early by clean tests.
- Expose routing reasons and document the new configuration.

## Why

False failure signals caused unnecessary escalation, while immediate hand-back could interrupt recovery after a real failure. This makes routing more accurate and stable.

## Validation

- Format and clippy checks pass.
- Targeted tests pass: tool signals (63), stage router (28), runner config (6).
- DeepSWE v1.1 (`bmr_6450d921f3b24644904b3ea902`, unrestricted network): 80/113 strict (70.80%), with 15.55% capable-tier calls.

### DIFF
--- a/crates/libsy/src/algorithms/util/tool_signals.rs
+++ b/crates/libsy/src/algorithms/util/tool_signals.rs
@@ -83,17 +83,14 @@ static ERROR_PATTERNS: &[(&str, f32, &[&str])] = &[
         ],
     ),
     // SOFT: plain non-zero exit without a recognisable exception traceback.
-    (
-        "exit_nonzero",
-        SOFT,
-        &[
-            "exit code 1",
-            "exit code 2",
-            "exit status 1",
-            "returned non-zero",
-            "exited with code",
-        ],
-    ),
+    ("exit_nonzero", SOFT, &["returned non-zero"]),
+];
+
+static NONZERO_EXIT_PHRASES: &[&str] = &[
+    "exit code",
+    "exit status",
+    "exited with code",
+    "exited with status",
 ];
 
 static EDIT_TOOL_NAMES: &[&str] = &[
@@ -744,7 +741,39 @@ pub(crate) fn classify_text(text: &str) -> (f32, Vec<String>) {
             severity = severity.max(*sev);
         }
     }
+    if has_nonzero_exit_status(&lower) && !patterns.iter().any(|p| p == "exit_nonzero") {
+        patterns.push("exit_nonzero".to_string());
+        severity = severity.max(SOFT);
+    }
     (severity, patterns)
+}
+
+/// Detects `exit_nonzero` only when a supported exit phrase is followed by a
+/// nonzero decimal status.
+///
+/// Codex includes "Process exited with code 0" on clean tool results, so exit
+/// phrases must parse their numeric status instead of matching the phrase alone.
+fn has_nonzero_exit_status(lower: &str) -> bool {
+    NONZERO_EXIT_PHRASES
+        .iter()
+        .any(|phrase| phrase_followed_by_nonzero_integer(lower, phrase))
+}
+
+/// Matches common "exit code/status N" spellings after optional separators.
+fn phrase_followed_by_nonzero_integer(lower: &str, phrase: &str) -> bool {
+    let mut cursor = 0usize;
+    while let Some(rel) = lower[cursor..].find(phrase) {
+        let value_start = cursor + rel + phrase.len();
+        let rest = lower[value_start..].trim_start_matches(|c: char| {
+            c.is_ascii_whitespace() || matches!(c, ':' | '=' | '\'' | '"' | '`')
+        });
+        let digits: String = rest.chars().take_while(|c| c.is_ascii_digit()).collect();
+        if !digits.is_empty() && digits.chars().any(|d| d != '0') {
+            return true;
+        }
+        cursor = value_start;
+    }
+    false
 }
 
 fn compute_no_error_streak(tool_texts: &[String]) -> u32 {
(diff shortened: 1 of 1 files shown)


crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -559,15 +599,22 @@ impl Classifier<State> for StageClassifier {
...
-        let Some(signal) = tool_signals else {
+        let Some(signal) = state.tool_signals.clone() else {
             // No tool activity yet — nothing to score, so the signals have no
             // opinion, same as a below-threshold turn.
             return Ok((Self::abstain(state), None));
         };
 
-        let outcome = pick_tier(signal, self.mode, self.confidence_threshold);
-        record_score_metrics(signal, &outcome);
+        let clean_test_pass = signal.tests_passed && signal.no_error_streak > 0;
```
grahamking, line 619: From AI: `tests_passed` means any result in the recent window passed. `no_error_streak > 0` only means the latest result was clean. A sequence of passing tests → compiler failure → successful edit therefore clears the hold, without any tests passing after the failure. Check that the passing test occurred after the latest failure.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -549,6 +563,32 @@ impl StageClassifier {
             prompts::append_note(request, note);
         }
     }
+
+    fn consume_capable_hold(state: &mut State) -> bool {
```
grahamking: I'm not sure if we care about subagents (probably we do), but Astra said:

The new counter lives in state keyed only by `session_id`. Subagents sharing that session can consume the parent’s recovery turns or clear its hold with their own passing tests. My router-level test also showed a clean child being routed to the capable tier because its parent failed. Keep recovery state separate for each agent.

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -474,6 +513,143 @@ fn is_builtin_tool_name(lower: &str) -> bool {
...
 
+/// Split a shell line at common command separators. This intentionally avoids
+/// pretending to be a full shell parser; only the leading program and flags of
+/// each segment are inspected below.
+fn shell_segments(command: &str) -> impl Iterator<Item = &str> {
+    command
+        .split(['\n', ';', '|', '&'])
+        .map(str::trim)
+        .filter(|segment| !segment.is_empty())
+}
```
grahamking, line 558: I think you also need to look for quotes (single and double). Example: `rg 'foo|rm obsolete.rs'`

grahamking:
Just noting, this is specific to English ("exited with code") and to Linux or Mac (forward slashes, `cp`, etc). No action for now.

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -474,6 +513,143 @@ fn is_builtin_tool_name(lower: &str) -> bool {
...
+        .is_some_and(|word| word.contains('=') && !word.starts_with('='))
+    {
+        start += 1;
+    }
+
+    words[start..].to_vec()
+}
+
+fn program_name(word: &str) -> &str {
+    word.rsplit('/').next().unwrap_or(word)
```
grahamking: You can also use `std::path::Path` to get the program name.

crates/libsy/src/algorithms/util/tool_signals.rs
```diff
@@ -474,6 +513,143 @@ fn is_builtin_tool_name(lower: &str) -> bool {
...
+/// each segment are inspected below.
+fn shell_segments(command: &str) -> impl Iterator<Item = &str> {
+    command
+        .split(['\n', ';', '|', '&'])
+        .map(str::trim)
+        .filter(|segment| !segment.is_empty())
+}
+
+fn shell_words(segment: &str) -> Vec<&str> {
+    let words: Vec<&str> = segment.split_ascii_whitespace().collect();
```
grahamking: You should be able to avoid the `collect()`, which allocates a Vec on the heap. Without the `collect` you have an `Iterator`.

crates/libsy/src/algorithms/util/stage.rs
sabhatinas: Fixed in 841488b: capable-hold state is now keyed by agent ID. Added coverage proving a child cannot consume its parent hold.

crates/libsy/src/algorithms/util/tool_signals.rs
sabhatinas: Fixed in 841488b: shell_words now returns an iterator directly, avoiding the temporary Vec.

crates/libsy/src/algorithms/util/tool_signals.rs
sabhatinas: Fixed in 841488b: shell segmentation now ignores separators inside single and double quotes, with coverage for both.

crates/libsy/src/algorithms/util/stage.rs
sabhatinas: Fixed in 841488b: a pass now clears the hold only when it occurs after the latest recent failure. Added regression coverage for pass, failure, then clean edit.

crates/libsy/src/algorithms/util/tool_signals.rs
sabhatinas: Fixed in 841488b: command basenames now use Path::file_name().

grahamking (approved):

PR #653, opened 2026-09-09
fix(server): separate the Codex reasoning summary capability

eugenn:
## Problem

Codex reads route capabilities from `GET /v1/models`. Switchyard derives
reasoning summary support from the general reasoning capability. Some models
accept reasoning effort but reject `reasoning.summary`, which causes HTTP 400
responses.

Codex also renamed the capability, so publishing one name only leaves one
Codex version reading nothing and sending `reasoning.summary` anyway.

## Change

- Add an optional `reasoning_summaries` route capability. It defaults to
  `reasoning` when unset, so existing configurations are unaffected.
- Publish it under both Codex capability names: the legacy
  `supports_reasoning_summaries` and the current
  `supports_reasoning_summary_parameter`.
- Keep the benchmark harness model card in sync.
- Document and test a route that supports reasoning effort but not summaries.

## Configuration

```toml
reasoning = true
reasoning_summaries = false
```

## Compatibility

Codex 0.144.x reads `supports_reasoning_summaries` and gates its whole
reasoning control on it, so disabling summaries also stops reasoning effort.
Codex 0.145+ reads `supports_reasoning_summary_parameter`, keeps reasoning
effort, and omits only `reasoning.summary`.

## Validation

- `cargo fmt --all --check`
- `cargo test -p switchyard-runner`
- `uv run ruff check .`


eugenn:
@coderabbitai review

messiaen (changes requested):
Thanks for the PR!

The config is only required for reasoning models that reject reasoning.summary. But users must know:

What each model accepts.
What its provider accepts.
Which Codex version interprets the metadata.
Whether every target in a route has the same support.
Ideally that knowledge should live in Switchyard, not route configuration.
Can we discover capability from provider metadata?

PR #655, opened 2026-09-09
feat(libsy): populate bounded outcome evidence

nachiketb-nvidia:
## What

Populate `RoutingOutcome.metadata.evidence` with bounded, algorithm-produced JSON for built-in libsy routing decisions.

Closes [SWITCH-1418](https://linear.app/nvidia/issue/SWITCH-1418/populate-bounded-decision-evidence-for-built-in-algorithms).

## Why

Outcome metadata already identifies the run and algorithm, but a string cannot represent structured decision evidence without another encoding layer. Built-in algorithms also need to expose the small set of facts that actually determined a route without leaking prompts, responses, raw errors, or arbitrary internal state.

## How

- Change evidence from `Option<String>` to `Option<serde_json::Value>`.
- Keep evidence run-scoped in the private driver state; each `run_stream` invocation owns an independent slot.
- Record only decision-defining fields:
  - capability classifier score and threshold
  - stage source, plus confidence and threshold when dimension scoring decides
  - advisor and escalation verdicts
  - bounded fail-open reason codes
  - retained, fall-open, and subagent source labels
- Keep random, passthrough, noop, and successful opaque custom decisions empty because the algorithm name or selected model already describes them.
- Preserve evidence from the deciding component across fallback classifiers.

## What to review

- Whether each source and verdict describes the component that made the final decision.
- Whether the JSON remains bounded and contains no request-derived text.
- The private per-run evidence handoff and last-decision-wins behavior within one cascade.

## Validation

- `cargo clippy -p switchyard-libsy --all-targets -- -D warnings`
- `cargo test -p switchyard-libsy` (282 passed)
- `cargo check -p switchyard-llm-client -p switchyard-server -p switchyard-py`

One existing top-level driver test now covers structured evidence attachment and fallback precedence; no algorithm-specific test matrix was added.

## Summary by CodeRabbit

* **New Features**
  * Added structured decision evidence to outcomes, including routing sources, confidence, thresholds, verdicts, and fallback paths.
  * Outcomes now record whether decisions came from retained assignments, classifiers, subagents, escalation handling, or fail-open processing.
  * Added standardized reason details for consultation failures and unparseable responses.
  * Evidence is preserved across processing stages and included in completed outcome metadata.

* **Bug Fixes**
  * Improved visibility into escalation and fallback decisions, including why a review was refunded or deferred.

### DIFF
--- a/crates/libsy/src/algorithms/util/llm_judge.rs
+++ b/crates/libsy/src/algorithms/util/llm_judge.rs
@@ -237,11 +237,11 @@ where
         self
     }
 
-    /// Replaces run evidence only for judges that opted into structured evidence.
+    /// Adds fail-open evidence only for evidence-enabled judges and preserves an earlier decision.
     fn report_fail_open(&self, driver: &Driver, error: String, reason: &'static str) {
         report_fail_open(self.target.as_str(), error, reason);
         if self.evidence.is_some() {
-            driver.set_evidence(serde_json::json!({
+            driver.set_evidence_if_empty(serde_json::json!({
                 "source": "fail_open",
                 "reason_code": reason,
             }));
@@ -350,14 +350,20 @@ where
             });
         };
         let verdict = self.verdict(state, request, driver).await;
+        let classification = self.policy.to_classification(verdict.as_ref());
         if let Some(evidence) = self
             .evidence
             .and_then(|evidence| evidence(&self.policy, verdict.as_ref()))
         {
-            driver.set_evidence(evidence);
+            match &classification {
+                Classification::Scores(scores) if !scores.is_empty() => {
+                    driver.set_evidence(evidence);
+                }
+                _ => driver.set_evidence_if_empty(evidence),
+            }
         }
         // A judge consultation is a side call, never the turn's answer.
-        Ok((self.policy.to_classification(verdict.as_ref()), None))
+        Ok((classification, None))
     }
 }
 
--- a/crates/libsy/src/algorithms/util/llm_judge.rs
+++ b/crates/libsy/src/algorithms/util/llm_judge.rs
@@ -237,6 +237,7 @@ where
         self
     }
 
+    /// Replaces run evidence only for judges that opted into structured evidence.
     fn report_fail_open(&self, driver: &Driver, error: String, reason: &'static str) {
         report_fail_open(self.target.as_str(), error, reason);
         if self.evidence.is_some() {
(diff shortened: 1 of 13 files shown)


crates/libsy/src/algorithms/util/llm_judge.rs
```diff
@@ -220,6 +227,23 @@ where
...
+        self
+    }
+
+    fn report_fail_open(&self, driver: &Driver, error: String, reason: &'static str) {
+        report_fail_open(self.target.as_str(), error, reason);
+        if self.evidence.is_some() {
+            driver.set_evidence(serde_json::json!({
+                "source": "fail_open",
+                "reason_code": reason,
+            }));
```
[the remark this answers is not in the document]
nachiketb-nvidia: check now

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -629,6 +629,17 @@ impl Classifier<State> for StageClassifier {
                 // is the only branch whose tier the signals actually chose — an
                 // ambiguous turn is decided further down the cascade.
                 self.apply_handoff_note(request, tier, source);
+                if let Some(driver) = driver {
+                    let evidence = match (source, confidence) {
+                        (DecisionSource::Dimensions, Some(confidence)) => serde_json::json!({
+                            "source": source.as_str(),
+                            "confidence": confidence,
+                            "threshold": self.confidence_threshold,
```
sabhatinas, line 637: this is static, can we skipped

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -629,6 +629,17 @@ impl Classifier<State> for StageClassifier {
                 // is the only branch whose tier the signals actually chose — an
                 // ambiguous turn is decided further down the cascade.
                 self.apply_handoff_note(request, tier, source);
+                if let Some(driver) = driver {
+                    let evidence = match (source, confidence) {
+                        (DecisionSource::Dimensions, Some(confidence)) => serde_json::json!({
+                            "source": source.as_str(),
+                            "confidence": confidence,
```
sabhatinas, line 636: This is the prob score right? if yes, LGTM!

sabhatinas (approved):
Minimal change requested for stage, approving as it looks good overall

crates/libsy/src/algorithms/util/stage.rs
nachiketb-nvidia: its referenced per instantiation of stage router no? so not static, but fixed per instantiation, should still be useful

crates/libsy/src/algorithms/util/stage.rs
nachiketb-nvidia: yeah

PR #663, opened 2026-09-10
fix(translation): encode Responses image URLs

alam0rt:
Hey! Just writing here to show that a human is behind this PR. I required this patch (patching it using a Nix overlay) to avoid issues with mixing the different provider APIs via switchyard.

## What

Fix OpenAI Responses request encoding for normalized image content.

The buffered Responses codec previously serialized the internal tagged `ImageSource` enum directly into `input_image.image_url`. That produces an object such as `{ "type": "raw", "data": ... }`, while the Responses API requires `image_url` to be the scalar URL/data URL.

This change maps normalized image sources to the wire form Responses accepts:

- URL sources become scalar URLs.
- Base64 sources become `data:<media-type>;base64,<payload>` data URLs.
- Common raw Anthropic/OpenAI image shapes are recovered to their URL or data URL.
- Unmappable raw image shapes go through the existing lossy-translation diagnostic path.

Regression tests cover Anthropic URL and base64 image blocks translated into OpenAI Responses.

## Why

Image-bearing Anthropic requests routed through Switchyard to an OpenAI Responses backend currently emit a malformed `input_image` block. The route can then fail with an upstream schema error instead of preserving the image.

## How tested

TDD: added the two regression tests first and confirmed both failed before the codec change with the internal tagged `ImageSource` object in `image_url`; both pass after the fix.

- [ ] `uv run ruff check .` clean — not run; no Python files changed
- [ ] `uv run mypy switchyard` clean — not run; no Python files changed
- [ ] `uv run pytest tests/` green — not run; no Python tests changed
- [x] Manual smoke (describe what was run)

Rust checks run:

```text
cargo fmt --all -- --check
cargo clippy -p switchyard-translation --all-targets -- -D warnings
cargo test -p switchyard-translation
cargo test -p switchyard-translation --test request_translation
```

Results: formatting clean; clippy clean; translation suite green, including all 59 request-translation tests.

## Checklist

- [x] One class per file; filename = `snake_case` of the primary class. — N/A; Rust codec change only
- [x] New public symbols exported from `switchyard/__init__.py.__all__` if intended for downstream use. — N/A; no public Python symbols
- [x] README / `--help` updated if customer-facing surface changed. — N/A; bug fix only

## Notes for reviewers

This intentionally preserves the current scalar `image_url` output shape and does not add support for a Responses `detail` field. Unmappable raw image sources report lossy conversion through the existing policy hook rather than serializing a malformed internal enum.

## Summary by CodeRabbit

* **New Features**
  * Improved image handling in Responses translations, including URL, base64, and raw image sources.
  * Base64-encoded images are converted into compatible data URLs with media types.

* **Bug Fixes**
  * Image sources that cannot be converted now produce a diagnostic instead of failing direct serialization.

* **Tests**
  * Added coverage for translating Anthropic image blocks into OpenAI Responses image inputs.

### DIFF
--- a/crates/switchyard-translation/src/codecs/responses/buffered.rs
+++ b/crates/switchyard-translation/src/codecs/responses/buffered.rs
@@ -19,9 +19,9 @@ use crate::diagnostic::TranslationDiagnostic;
 use crate::error::{Result, TranslationError};
 use crate::format::{FormatId, WireFormat};
 use crate::llm::{
-    AggLlmResponse, ContentBlock, InstructionBlock, LlmRequest, MediaSource, Message, OutputParams,
-    ProviderExtensions, ReasoningParams, ResponseOutput, Role, SamplingParams, StopReason,
-    ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
+    AggLlmResponse, ContentBlock, ImageSource, InstructionBlock, LlmRequest, MediaSource, Message,
+    OutputParams, ProviderExtensions, ReasoningParams, ResponseOutput, Role, SamplingParams,
+    StopReason, ToolCall, ToolChoice, ToolDefinition, ToolResult, Usage,
 };
 use crate::policy::{DeterministicIdPolicy, TranslationPolicy};
 use crate::util::{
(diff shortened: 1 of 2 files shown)


messiaen:
I belive #530 addressed these issues.

Closing for now. re-open if this is still an issue after #530 was merge into main

PR #677, opened 2026-09-10
ci: lint GitHub Actions workflows

afourniernv:
## What

Adds `actionlint` to pre-commit and runs the same pinned hook in the existing CI lint job.

The first GitHub run also found two unquoted variables in `perf.yml`; those are quoted so the current workflow set starts clean.

## Why

Workflow YAML can parse successfully while still containing invalid GitHub expressions, contexts, matrix references, job dependencies, or embedded shell. This catches those errors before a workflow is merged.

## How tested

- [x] `uvx --from pre-commit==4.5.0 pre-commit validate-config`
- [x] `uvx --from pre-commit==4.5.0 pre-commit run actionlint --all-files` — all 9 workflows passed
- [x] GitHub Actions `Lint` job — actionlint, `uv sync`, and Ruff passed on Ubuntu 24.04
- [x] `git diff --check`
- [x] Commit carries the required DCO sign-off

## Notes for reviewers

This uses the same `actionlint` hook and version as NeMo Relay without copying its larger pre-commit CI setup. It adds no new GitHub `uses:` reference, public API, dependency lock change, packaging, or release behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
   # ruff is a hard gate — the codebase passes today, keep it that way.
   lint:
-    name: Lint (ruff)
+    name: Lint
     needs: changes
     if: needs.changes.outputs.full_ci == 'true'
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ayushag-nv (approved):
Looks good and relevant. Please rebase and merge

Issue #669, opened 2026-09-10
ci: evaluate a main-owned Rust dependency cache


afourniernv:
## Problem

Recent successful code CI runs spend a median of 3m30s in the Rust job, with a 2m48s to 3m56s range. Clippy takes a median 1m06s and cargo test takes 2m06s.

On https://github.com/NVIDIA-NeMo/Switchyard/actions/runs/34497080857, Rust took 3m35s while the longest Python job took 5m05s. A Rust cache should save runner time and return Rust results sooner, but it probably will not shorten the full required CI run yet.

Switchyard currently has no Rust cache. Its existing Actions caches are about 55 MB of uv data.

## Proposed trial

Add a SHA-pinned Swatinem/rust-cache step only to the Rust job in .github/workflows/ci.yml.

- Use the v2.9.1 SHA already used by NeMo Relay.
- Set cache-bin to false.
- Save only from main; pull requests can restore the cache written by main.
- Keep the action defaults. Do not add shared keys or cache workspace crates.
- Leave Python, docs, portability, perf, publishing, and release workflows alone.

This should be a four-line workflow change.

## Risks

Cache download time may cancel out the compile savings. New lockfile and toolchain generations also consume storage until GitHub evicts them. The default repository cache limit is 10 GB.

Fork pull requests can read the main cache, so it must not contain secrets. Main-only writes, cache-bin false, and a pinned action SHA keep the trust boundary narrow. A missing or bad cache must fall back to a normal Cargo build.

## Validation

1. Seed the cache from a main run.
2. Verify pull requests restore but do not save it.
3. Record cache size and restore/save time.
4. Compare at least 5 warm runs against the current 3m30s Rust median.
5. Keep it only if the Rust job improves enough to justify the storage and maintenance cost.

Local Cargo runs can show that warm artifacts help, but cache scope, archive size, and Actions timing have to be measured on GitHub.

PR #679, opened 2026-09-10
ci: trial main-owned Rust caching

afourniernv:
## What

Adds `Swatinem/rust-cache` to the existing Rust CI job.

Pull requests can restore the cache, but only `main` saves it. `cache-bin: false` keeps Cargo-installed binaries out. This follows the same pinned-action and main-writes pattern used by [NeMo Relay's Rust CI](https://github.com/NVIDIA/NeMo-Relay/blob/1ec3c63521b2323f939f1813c90d431121a74eaa/.github/workflows/ci_rust.yml#L116-L122), introduced as part of [NVIDIA/NeMo-Relay#46](https://github.com/NVIDIA/NeMo-Relay/pull/46).

## Why

Starts the trial described in [#669](https://github.com/NVIDIA-NeMo/Switchyard/issues/669).

Recent successful runs spend a median of 3m30s in the Rust job. Python currently takes longer, so I do not expect this to shorten the full workflow yet. It should reduce repeated Rust compilation and return Rust results sooner.

This is intentionally experimental. The first `main` run will seed the cache, then we can compare 5-10 warm runs and inspect cache size, restore time, and Rust job duration. If the savings are not meaningful or the cache churn is too high, we should revert it.

## Scope

- Only the core Rust job is cached.
- No custom cache keys or workspace-crate caching.
- Python, docs, portability, performance, publishing, and release workflows are unchanged.
- No public API, runtime dependency, package, or release behavior changes.

## How tested

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/ci.yml`
- `git diff --check`
- `cargo fmt --all --check`
- `cargo test --workspace`
- `uv run ruff check .`
- `uv run mypy switchyard`
- `env -u OPENROUTER_API_KEY -u NVIDIA_API_KEY -u OPENAI_API_KEY -u ANTHROPIC_API_KEY uv run pytest tests/ -v -m "not integration"` - 117 passed, 2 deselected

## Notes for reviewers

The production diff is four YAML lines. A cache miss falls back to the normal Cargo build.

## Summary by CodeRabbit

* **Chores**
  * Improved Rust CI caching to reduce unnecessary cache storage and limit cache updates to the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,10 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust components
         run: rustup component add rustfmt clippy
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        with:
+          cache-bin: false
+          save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: cargo fmt
         run: cargo fmt --all --check
       - name: cargo clippy


ayushag-nv:
@grahamking for review

nachiketb-nvidia (approved):

Issue #559, opened 2026-08-26
Add a reproducible NeMo Gym example for comparing routing strategies


cwing-nvidia:
## Motivation

Switchyard’s existing benchmark example shows how to evaluate routing with Harbor and coding-agent benchmarks. Switchyard also integrates with NeMo Gym, which supports a broader collection of benchmarks and agent harnesses, but the Switchyard repository does not currently include a runnable example of that workflow.

Adding one would help users:

- Evaluate routing strategies on benchmarks beyond Harbor’s coding-agent workloads.
- Compare a routed strategy with a fixed-model baseline using the same tasks.
- Reproduce the workflow directly from the Switchyard repository.
- Discover and validate the NeMo Gym integration without assembling commands from external documentation.

## Proposed solution

Add benchmark/nemo_gym/ with a small, executable, version-pinned example that:

- Defines a fixed-model baseline and one routed strategy.
- Runs both against the same benchmark and task subset.
- Writes each condition to a separate output directory.
- Pairs rollouts before comparing rewards.
- Compares model tokens, classifier tokens, latency, and routing statistics.
- Explains the generated provenance and statistics artifacts.
- Links to the [full NeMo Gym integration documentation](https://docs.nvidia.com/nemo/gym/main/model-server/switchyard/).

grahamking:
Approved! Are you able to submit something?

shashank3959:
@grahamking The [PR](https://github.com/NVIDIA-NeMo/Switchyard/pull/680) has been approved now. Whats the process to merge?

PR #595, opened 2026-09-02
feat(benchmark): add reproducible NeMo Gym routing comparison

afourniernv:
## What

Adds `benchmark/nemo_gym/`, a version-pinned workflow for comparing two Switchyard routing conditions on the same NeMo Gym MMLU-Redux tasks:

- `strong-only` always uses the strong target.
- `policy-model` uses an LLM classifier to select the efficient or strong target.

The runner builds the current checkout's `switchyard-server`, gives each condition a fresh proxy and output directory, and captures Gym rollouts alongside Switchyard statistics and metrics. The comparator pairs matching task/repeat rows before reporting reward, answer and classifier tokens, latency, model totals, and classifier fail-opens.

This adds no production dependency, API, or request-path change.

## Why

Gym already provides the `switchyard_model` adapter, but Switchyard did not have an executable repository example for using it to compare a routed strategy with a fixed-model baseline. This makes that workflow reproducible from a Switchyard checkout and keeps failed or mismatched rollouts visible in the comparison.

Closes #559

Linear: [SYGH-96](https://linear.app/nvidia/issue/SYGH-96/add-a-reproducible-nemo-gym-example-for-comparing-routing-strategies)

Builds on [NVIDIA-NeMo/Gym#2141](https://github.com/NVIDIA-NeMo/Gym/pull/2141), which added Gym's Switchyard model-server adapter.

## Notes for reviewers

Start with `benchmark/nemo_gym/README.md` for the workflow, then `run.sh` for process and artifact ownership, and `compare.py` for pairing and accounting semantics.

The example uses attached mode so it exercises the current Switchyard checkout rather than Gym's pinned released wheel. It starts a fresh proxy per condition because router state and `/v1/stats` are process-wide, and it snapshots statistics before shutdown because Gym's attached-mode snapshot is best-effort.

Validation:

- [x] `env -u OPENROUTER_API_KEY -u NVIDIA_API_KEY -u OPENAI_API_KEY -u ANTHROPIC_API_KEY uv run pytest tests/ -q -m "not integration"` — 121 passed, 2 deselected, 2 subtests passed
- [x] `uv run ruff check .`
- [x] `uv run mypy switchyard`
- [x] `bash -n benchmark/nemo_gym/run.sh`
- [x] `git diff --check`
- [x] `cargo run --release -p switchyard-server -- --config benchmark/nemo_gym/routes.toml --dry-run`
[...]

PR #680, opened 2026-09-10
feat(tutorial): add tutorial to evaluate Switchyard with NeMo Gym

shashank3959:
## What

Adds a NeMo Gym evaluation tutorial under `benchmark/nemo_gym/`:

- A tested Gym `v0.6.0` setup that runs the **current Switchyard checkout** through the existing LiteLLM integration.
- A fixed Nemotron 3 Ultra baseline and **seeded Random routing** between Ultra and Nemotron 3.5 Lightning.
- A single runner that prepares MMLU-Redux, manages the local LiteLLM proxy, evaluates both conditions with identical settings, and saves the comparison artifacts. Task count, repeats, and concurrency are configurable.
- An offline comparator that validates paired inputs and model-call evidence before reporting Gym-scored rewards, serving-model attribution, Gym-captured input/output tokens, latency, incomplete responses, captured failures, and unknown usage.

## Why

Provides a runnable example of:

```text
NeMo Gym → LiteLLM proxy → Switchyard library → upstream model
```

No Harbor, Docker, or separately managed Switchyard server is required.

The example demonstrates how to evaluate routing against a fixed baseline using paired, identical Gym inputs. The small Random-routing run validates the integration; it does not establish an improvement in model quality or cost.

Related to [#559](https://github.com/NVIDIA-NeMo/Switchyard/issues/559).

## Validation

- 18 focused tests passed with no skips.
- Hermetic tests covered paired-input and model-call evidence validation, Gym-scored incomplete responses, additional and failed calls with unknown usage, serving-model attribution, response normalization, and runner lifecycle behavior.
- A live run completed five tasks per condition, exercised both Nemotron 3 Ultra and Nemotron 3.5 Lightning, and completed all ten captured calls successfully.
- `uv run ruff check .`, `uv run mypy switchyard`, Bash syntax validation, and `git diff --check` passed.

## Notes for reviewers

- Start with the tutorial, then review the runner and the comparator’s pairing, completeness, model-attribution, and token-accounting checks.
- The tutorial-local LiteLLM callback records only response identity and the serving deployment while normalizing response fields required by Gym `v0.6.0`. Gym’s model-call capture supplies usage and error evidence. This PR does not modify Gym.
- Gym-scored incomplete responses and recovered errors remain in the comparison with warnings. Missing rollouts, terminal Gym failures, mismatched inputs, and incomplete model-call captures block comparison.
- Retries inside Gym’s model adapter, LiteLLM, or the upstream provider may not appear as separate captured calls.
- Random routing makes no classifier calls, so there is no classifier-token overhead in this example.

### DIFF
--- a/benchmark/nemo_gym/README.md
+++ b/benchmark/nemo_gym/README.md
@@ -1,0 +1,250 @@
+# Evaluate Switchyard routing with NeMo Gym
+
+[NeMo Gym](https://github.com/NVIDIA-NeMo/Gym) is a library for evaluating models and agents using tasks with verifiable outcomes.
+This tutorial uses its five included multiple-choice examples to compare a fixed model with Switchyard routing.
+
+![Gym evaluation with a hosted Switchyard model server](architecture.svg)
+
+## What changes between the runs?
+
+Both routes are defined in [routes.toml](routes.toml):
+
+| Route | Behavior |
+|---|---|
+| `fixed` | Always use Nemotron 3 Super. |
+| `routed` | Ask GPT-OSS 20B to classify the task, then use GPT-OSS 20B or Nemotron 3 Super. |
+
+The dataset, agent, verifier, temperature, and answer-token limit stay the same.
+The classifier is an extra model call: its tokens count even when the router selects GPT-OSS 20B.
+Super uses `enable_thinking=false`; GPT-OSS 20B uses `reasoning_effort=low`, including for
+classification. These per-model settings stay unchanged between conditions. This is not a
+benchmark of either model's maximum reasoning capability.
+
+## 1. Install the pinned Gym checkout
+
+You need Git, [uv](https://docs.astral.sh/uv/), two Bash terminals, and an NVIDIA API key for
+the public endpoint `https://integrate.api.nvidia.com/v1`. Use these exact model IDs:
+`openai/gpt-oss-20b` and `nvidia/nemotron-3-super-120b-a12b`.
+Inference may consume credits. The classifier requires strict JSON Schema responses from
+its selected model.
+
+**Smoke-tested:** A one-question paired run completed on this endpoint with correct answers
+in both conditions, GPT-OSS 20B serving the routed answer, and no reported model or classifier
+errors. The full five-question comparison below has not yet been validated for this pair.
+
+Run from the **Switchyard repository root**:
+
+```bash
+WORK="$PWD/scratch/nemo-gym-tutorial"
+mkdir -p "$WORK" &&
+git clone https://github.com/NVIDIA-NeMo/Gym.git "$WORK/Gym" &&
+git -C "$WORK/Gym" checkout 3a26c35fa90c243427378569511f7b06f503e0fd &&
+uv tool run --from uv==0.11.29 uv venv --python 3.13.14 "$WORK/.venv" &&
+uv tool run --from uv==0.11.29 uv pip install \
+  --python "$WORK/.venv/bin/python" uv==0.11.29 -e "$WORK/Gym"
+```
+
+The pinned uv can download Python 3.13.14 even if your existing uv is older. It is also installed
+inside the tutorial environment for Gym's component setup; your global uv is left unchanged.
+
+This keeps the checkout and environment under Switchyard's ignored `scratch/` directory,
+without changing another Gym checkout. Use a fresh directory for this one-time setup.
+The editable install lets Gym's component environments use the same pinned source.
+
+Gym installs **`nemo-switchyard==0.2.0`** into its model-server environment and hosts the native
+proxy in-process. You do **not** need Docker or a separately running `switchyard-server`.
+The Switchyard checkout contains this tutorial; its current `main` is **not** the proxy being
+executed. Do not copy newer routing options into this version-pinned example.
+
+## 2. Start the fixed condition — Terminal 1
+
+Keep this terminal at the Switchyard repository root. Set your API key in the terminal,
+replacing the placeholder with your key:
+
+```bash
+export NVIDIA_API_KEY='<your-api-key>'
+```
+
+Start Gym's resources, agent, and model servers. Run only one Gym environment at a time.
+The first start also installs their dependencies.
+
+```bash
+EXAMPLE="$PWD/benchmark/nemo_gym"
+WORK="$PWD/scratch/nemo-gym-tutorial"
+source "$WORK/.venv/bin/activate"
+RUN_DIR="$EXAMPLE/results/first-run"
+OUT="$RUN_DIR/fixed"
+mkdir -p "$RUN_DIR"
+
+mkdir "$OUT" &&
+git -C "$WORK/Gym" rev-parse HEAD > "$OUT/gym-commit.txt" &&
+gym env start --resources-server mcqa --model-type switchyard_model --model fixed \
+  "++policy_model.responses_api_models.switchyard_model.deployment=$EXAMPLE/routes.toml" \
+  ++policy_model.responses_api_models.switchyard_model.switchyard_base_url=null \
+  "++policy_model.responses_api_models.switchyard_model.condition_dir=$OUT" \
+  ++mcqa_simple_agent.responses_api_agents.simple_agent.max_steps=1 \
+  ++observability_enabled=true \
+  "++model_call_capture_dir=$OUT/model-calls" \
+  "++nemo_gym_log_dir=$OUT/server-logs" \
+  "hydra.run.dir=$OUT/hydra-start"
+```
+
+Wait for **`All 3 / 3 servers ready!`**. Leave this terminal running.
+`mkdir "$OUT"` deliberately refuses to reuse an existing condition directory.
+For a new comparison, change `RUN_DIR` to the same fresh path in **both terminals**.
+
+## 3. Run the five questions — Terminal 2
+
+Open another Bash terminal at the **same Switchyard repository root**:
+
+```bash
+EXAMPLE="$PWD/benchmark/nemo_gym"
+WORK="$PWD/scratch/nemo-gym-tutorial"
+source "$WORK/.venv/bin/activate"
+RUN_DIR="$EXAMPLE/results/first-run"
+OUT="$RUN_DIR/fixed"
+
+gym eval run --no-serve --agent mcqa_simple_agent \
+  --input "$WORK/Gym/resources_servers/mcqa/data/example.jsonl" \
+  --output "$OUT/rollouts.jsonl" \
+  --limit 5 --num-repeats 1 --concurrency 1 \
+  --temperature 0 --max-output-tokens 4096 \
+  ++route_failures_to_sidecar=true \
+  ++observability_enabled=true \
+  "++model_call_capture_dir=$OUT/model-calls" \
+  "hydra.run.dir=$OUT/hydra-eval"
+```
+
+These questions are included with Gym. The MCQA resources server checks the answer letter
+against the expected answer; it does not call an LLM judge. A wrong answer earns zero reward.
+An infrastructure failure is a different outcome, recorded separately.
+
+The two-terminal flow is intentional: Gym's one-command evaluation mode does not accept
+`--split example`. `--no-serve` collects against the servers you already started.
+
+**After collection finishes, press Ctrl-C in Terminal 1 and wait for shutdown to finish.**
+Gym writes `switchyard-stats.json` during shutdown, before stopping its hosted proxy.
+Do not compare the runs before that file has been written. If Gym reports that a worker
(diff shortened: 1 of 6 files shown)


benchmark/nemo_gym/architecture.svg
itayzit, line 1: this is a good diagram, but i don't think we need to give the 'fixed' example. maybe we should say: decide between models using a prompt or heuristics.

Image

benchmark/nemo_gym/compare.py
```diff
@@ -0,0 +1,265 @@
...
+import argparse
+import json
+import math
+import sys
+from collections import Counter
+from pathlib import Path
+from statistics import mean
+from typing import Any, cast
+
+GYM_COMMIT = "3a26c35fa90c243427378569511f7b06f503e0fd"
```
itayzit, line 17: do we have to pin it to a specific commit?

afourniernv: ++ agree

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+The pinned uv can download Python 3.13.14 even if your existing uv is older. It is also installed
+inside the tutorial environment for Gym's component setup; your global uv is left unchanged.
+
+This keeps the checkout and environment under Switchyard's ignored `scratch/` directory,
+without changing another Gym checkout. Use a fresh directory for this one-time setup.
+The editable install lets Gym's component environments use the same pinned source.
+
+Gym installs **`nemo-switchyard==0.2.0`** into its model-server environment and hosts the native
```
afourniernv, line 54: Could we use the current Switchyard checkout here instead of the 0.2.0 wheel? Lin’s Gym work already covers the hosted released-wheel path. Since this example lives in Switchyard, I think it should tell us whether the checkout we’re looking at still works. The adapter supports that through attached mode; this is the main piece I’d keep from #595.

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Run from the **Switchyard repository root**:
+
+```bash
+WORK="$PWD/scratch/nemo-gym-tutorial"
+mkdir -p "$WORK" &&
+git clone https://github.com/NVIDIA-NeMo/Gym.git "$WORK/Gym" &&
+git -C "$WORK/Gym" checkout 3a26c35fa90c243427378569511f7b06f503e0fd &&
+uv tool run --from uv==0.11.29 uv venv --python 3.13.14 "$WORK/.venv" &&
+uv tool run --from uv==0.11.29 uv pip install \
```
afourniernv, line 43: This pins the Gym commit, Python, and uv, but `uv pip install -e` still resolves dependencies without using Gym’s lockfile. Could we use Gym’s frozen environment here so the dependency set is pinned too?

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+Open another Bash terminal at the **same Switchyard repository root**:
+
+```bash
+EXAMPLE="$PWD/benchmark/nemo_gym"
+WORK="$PWD/scratch/nemo-gym-tutorial"
+source "$WORK/.venv/bin/activate"
+RUN_DIR="$EXAMPLE/results/first-run"
+OUT="$RUN_DIR/fixed"
+
+gym eval run --no-serve --agent mcqa_simple_agent \
```
afourniernv, line 107: Could we make this one runner over a real Gym benchmark, maybe MMLU-Redux, with configurable limits and repeats? The five bundled questions are useful as a smoke test, but they are not much of a routing comparison, and the repeated two-terminal/Ctrl-C flow is easy to get wrong. Lin’s adapter already supports any Gym benchmark.

tests/test_nemo_gym_compare.py
```diff
@@ -0,0 +1,393 @@
...
+        "Mean rollout latency (ms)": ["200", "200"],
+        "Mean routing overhead (ms)": ["5", "5"],
+        "Classifier requests": ["0", "2"],
+    }
+    for metric, values in expected.items():
+        line = next(line for line in output.out.splitlines() if line.startswith(metric))
+        assert line[len(metric) :].split() == values
+
+
+@pytest.mark.parametrize(
```
afourniernv, line 171: Could we trim this test surface down quite a bit? This file is 393 lines and collects 77 cases, mostly from this malformed-field table and the error-count matrix below. For this tutorial I think we need one successful paired comparison, a couple of incomplete or mismatched-run cases, one bad capture case, and maybe the Bash syntax check. Testing every malformed JSON leaf and executing commands scraped from the README feels like a lot to maintain for the behavior we’re adding.

afourniernv (changes requested):
I think we can use this PR as the one implementation. The incomplete-run checks and walkthrough are good. Before merge, I’d like to make it a smaller Switchyard-owned benchmark: run the current checkout, automate both conditions, use a real Gym workload, and cut the test matrix down to the behavior we need. I left the concrete asks inline.

benchmark/nemo_gym/compare.py
```diff
@@ -0,0 +1,265 @@
...
+    require(
+        condition["nemo_switchyard_version"] == SWITCHYARD_VERSION,
+        f"{route}: unexpected Switchyard version",
+    )
+    require(run["gym_commit"] == GYM_COMMIT, f"{route}: unexpected Gym commit")
+    stats = snapshot["stats"]
+    classifier = stats["classifier"]
+    model_errors = number(stats["total_errors"], "model errors")
+    classifier_errors = number(classifier["total_errors"], "classifier errors")
+    require(
```
afourniernv, line 136: Do we want to reject every run with an internal error? A classifier or target can fail and Switchyard can still return a successful rollout through fail-open or fallback. That seems like useful routing behavior to report. The exact request-count checks below also reject a successful fallback. I’d keep incomplete Gym runs blocking, but report recovered errors and fallbacks instead of stopping the comparison.

afourniernv (commented):
One more general note: while moving this to one runner, can we aim for a simple setup, one command, and a result? The current README is effectively the runner, and the comparator and tests revalidate a lot of malformed Gym and Switchyard artifacts. I’d keep the checks that make the comparison trustworthy, but take a hard pass at the LOC so the scripts are easy for someone else to read and maintain.

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+source "$WORK/.venv/bin/activate"
+RUN_DIR="$EXAMPLE/results/first-run"
+OUT="$RUN_DIR/fixed"
+mkdir -p "$RUN_DIR"
+
+mkdir "$OUT" &&
+git -C "$WORK/Gym" rev-parse HEAD > "$OUT/gym-commit.txt" &&
+gym env start --resources-server mcqa --model-type switchyard_model --model fixed \
+  "++policy_model.responses_api_models.switchyard_model.deployment=$EXAMPLE/routes.toml" \
+  ++policy_model.responses_api_models.switchyard_model.switchyard_base_url=null \
```
itayzit: why null?

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+EXAMPLE="$PWD/benchmark/nemo_gym"
+WORK="$PWD/scratch/nemo-gym-tutorial"
+source "$WORK/.venv/bin/activate"
+RUN_DIR="$EXAMPLE/results/first-run"
+OUT="$RUN_DIR/fixed"
+mkdir -p "$RUN_DIR"
+
+mkdir "$OUT" &&
+git -C "$WORK/Gym" rev-parse HEAD > "$OUT/gym-commit.txt" &&
+gym env start --resources-server mcqa --model-type switchyard_model --model fixed \
```
itayzit: now that i think of it, we can leave the 'fixed'

benchmark/nemo_gym/README.md
```diff
@@ -0,0 +1,250 @@
...
+  "++policy_model.responses_api_models.switchyard_model.deployment=$EXAMPLE/routes.toml" \
+  ++policy_model.responses_api_models.switchyard_model.switchyard_base_url=null \
+  "++policy_model.responses_api_models.switchyard_model.condition_dir=$OUT" \
+  ++mcqa_simple_agent.responses_api_agents.simple_agent.max_steps=1 \
+  ++observability_enabled=true \
+  "++model_call_capture_dir=$OUT/model-calls" \
+  "++nemo_gym_log_dir=$OUT/server-logs" \
+  "hydra.run.dir=$OUT/hydra-start"
```
itayzit: are there any lines here that are redundant that we can delete?

benchmark/nemo_gym/compare.py
shashank3959: Removed the hard-coded commit requirement from the comparator. It now checks that the fixed and routed runs used the same Gym revision, rather than requiring one specific SHA.

The setup uses the tested v0.6.0 release tag to give users a known-working starting point. Other compatible revisions aren’t rejected by the comparator.

benchmark/nemo_gym/README.md
shashank3959: Thanks, Alex. Updated the PR around these asks:

Replaced the released-wheel setup with the existing Switchyard–LiteLLM integration, which builds Switchyard from this checkout. This takes a different path from the suggested attached mode: Gym calls LiteLLM, with Switchyard running as a library inside the proxy, not as a separate server.

benchmark/nemo_gym/README.md
shashank3959: Updated the Gym setup to use `uv sync --frozen --no-dev` against its checked-in `uv.lock`, replacing `uv pip install -e`. This installs the recorded dependency versions rather than resolving a fresh set. Gym and LiteLLM remain in separate environments.

benchmark/nemo_gym/README.md
shashank3959: Replaced the manual two-terminal flow with one runner over MMLU-Redux 2.0.

It prepares the dataset, manages the local LiteLLM proxy, evaluates fixed and routed conditions on the same inputs, and prints the comparison.

tests/test_nemo_gym_compare.py
shashank3959: Trimmed this further to six behavioral cases: successful pairing and metrics, incomplete runs, mismatched inputs, one bad capture, truncation, and recovered-error accounting.

benchmark/nemo_gym/compare.py
shashank3959: Updated so recovered errors and extra requests don’t automatically fail the comparison. Completed runs now report the recorded errors and token usage with a warning; incomplete Gym runs still block comparison.

Retries or fallbacks handled internally by LiteLLM or the upstream service may not appear separately in our logs, so we only report what we can observe.

benchmark/nemo_gym/README.md
shashank3959: `null` told Gym to start and host the Switchyard proxy itself rather than connect to an existing one. That setup has since been replaced: the runner now starts LiteLLM, and Gym connects to its explicit URL through `litellm_model`. The `null` override is no longer needed and has been removed.

benchmark/nemo_gym/README.md
shashank3959: Replaced the repeated fixed/routed command blocks with one command after setup:

`bash benchmark/nemo_gym/run.sh`

The runner defines the shared options once, runs both conditions, handles proxy startup and shutdown, and prints the comparison. Users no longer need to repeat the path setup or manage separate terminals.

afourniernv (approved):
Looks good. I ran this end to end on the PR head and against current main. Keeping #559 open for the native Gym path and classifier statistics.

itayzit (approved):
lgtm!

tests/test_nemo_gym_run.py
```diff
@@ -0,0 +1,325 @@
...
+printf 'gym_start:%s:%s:%s\\n' "$route" "$$" "$*" >> "$EVENT_LOG"
+run_dir="$(dirname "$output")"
+mkdir -p "$run_dir"
+printf '{"route": "%s"}\\n' "$route" > "$output"
+printf 'gym_done:%s\\n' "$route" >> "$EVENT_LOG"
+if [[ "$route" == "fixed" && "${FAKE_GYM_BEHAVIOR:-}" == "exit7" ]]; then exit 7; fi
+if [[ "$route" == "fixed" && "${FAKE_GYM_BEHAVIOR:-}" == "sidecar" ]]; then
+    printf '{"failure": true}\\n' > "$run_dir/rollouts_failures.jsonl"
+fi
+""",
```
grahamking, line 94: Nit for the future, these shell scripts should be in their own file. That allows validating them with something like `shellcheck`, and makes IDE shell script tooling work.

PR #681, opened 2026-09-11
feat(libsy): add an up-front capability gate to the escalation route

linj-glitch:
After merge, an escalation route can carry `escalation.gate = { base_threshold = 0.4 }`. On the first request of a session the route calls the judge with the packaged capability-classifier prompt, reads its p_solve forecast for the task, and latches the session to the strong tier when the forecast is below the threshold, before the weak tier is called; otherwise the request proceeds and the trajectory judge takes over as today. `threshold_step` and `prompt` behave as in capability mode, unusable verdicts and failed gate calls fall open to the weak tier, and the runner needs no change because `escalation` already deserializes into `EscalationJudgeConfig`.

The reason is cost. A hand-off mid-task makes the strong tier redo the work (90 to 120 strong calls per escalated session on DeepSWE regardless of when the latch fired), so the cheapest moment to escalate a task that was never going to fit the weak tier is turn one, and a task-level rule in the judge prompt cannot set that split: an absolute wording latched 30 of 30 DEV sessions and a relative wording 28 to 29 of 30, while capability mode's numeric threshold sent 14 of 30 to the strong tier in the same setting. The gate reuses capability mode's threshold policy, prompt and schema unchanged.

Tests: 354 pass across libsy and the runner including five new ones (latch before the efficient call, pass then trajectory judge, once per session, fall-open on an unusable verdict, threshold validation); clippy is clean; the server dry-run accepts the new table. Kept as a draft until the DEV benchmark on the Luna and Kimi pairs shows a selective split and a run-total saving against the strong tier alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   the weak tier is called. Sets the turn-one split with a number instead of a
   task-level rule in the judge prompt, which latched almost every multi-file
   task in practice. The trajectory judge takes over afterwards.
+  `gate.classifier_target` lets the once-per-session forecast use a stronger
+  model than the per-turn trajectory judge.
 - **Per-target `reasoning_effort`** — a target can force the reasoning effort
   of every request it serves, replacing the caller's value (`reasoning.effort`
   on the Responses wire, `reasoning_effort` on Chat Completions), so a strong
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- **Escalation up-front capability gate** — `escalation.gate = { base_threshold }`
+  judges the first request of a session with the packaged capability forecaster
+  and latches to the strong tier when `p_solve` is below the threshold, before
+  the weak tier is called. Sets the turn-one split with a number instead of a
+  task-level rule in the judge prompt, which latched almost every multi-file
+  task in practice. The trajectory judge takes over afterwards.
 - **Per-target `reasoning_effort`** — a target can force the reasoning effort
   of every request it serves, replacing the caller's value (`reasoning.effort`
   on the Responses wire, `reasoning_effort` on Chat Completions), so a strong
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   judge for one of eight named rungs instead of a probability; rungs map to
   band midpoints so thresholds keep working. Models rank tasks more reliably
   than they number them, and a stated numeric cutoff invites answers just
-  under it.
+  under it. `escalation.gate.min_confidence` writes the gate's threshold as a
+  rung, so an ordinal configuration carries no numbers at all.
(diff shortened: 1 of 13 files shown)


linj-glitch:
First benchmark results for the gate, DeepSWE-v1.1 DEV30 subset, Codex 0.149.1, run total = agent tokens of every member plus judge and gate calls in relative units (uncached input 1, cached 0.1, output 8 per million tokens, all at GPT-5.6 rates). Sol max alone scores 26 to 28 of 30 for 54 units.

Kimi-K3 cheap tier, Sol max strong tier, gate at base_threshold 0.50 through a Sol-medium target, per-turn judge (stuck bar only) on DeepSeek flash: 26 of 30 for about 46 units. The gate latched 10 sessions on turn one with forecasts between 0.18 and 0.48 and kept the sessions forecast at 0.52 to 0.72; Kimi solved 11 of the 12 it kept to the end; the stuck bar latched 8 late. Without the gate, every prompt-level task rule latched 19 to 30 of 30 sessions on turn one and the route cost 52 to 66 units.

Luna cheap tier at medium effort, gate at 0.60: the gate half worked (14 of the 15 tasks it sent to Sol solved, forecasts spread 0.22 to 0.78 with no cluster at the threshold) but Luna solved 0 to 1 of the tasks it kept, so that pair lands at 12 to 16 of 30 for 41 to 47 units. A 64-task confirmation of the Kimi configuration is running; I will post it here before asking for review.

linj-glitch:
64-task confirmation of the Kimi configuration (gate 0.50 through Sol-medium, DeepSeek stuck-only per-turn judge, generic route): 38 solved of 58 completed members, 65.5 percent; six members failed before the harness ran and left no artifacts. Same 64 tasks, other arms: Kimi-K3 alone 40 and 35 (58.6 percent), Sol max alone 41 to 46 (68 percent). Run total 1.30 units per member, about 84 for the set, against about 116 for Sol max alone and 56 for Kimi alone. The gate latched 16 of 58 sessions on turn one, the stuck bar latched 14 more late, and Kimi kept 28 and solved 16 of them.

So on both the 30-task and the 64-task sets the gate puts the route at the lower edge of Sol max's score band, four to five tasks above the cheap tier alone, at 16 to 28 percent below Sol's total. Without the gate, every prompt-level task rule latched 19 to 30 of 30 sessions on turn one and cost as much as or more than Sol alone.

linj-glitch:
Follow-up on the DeepSWE numbers, for anyone weighing this feature on them. Matching each gate verdict to its task and checking it against the cheap tier's own direct results shows that on this benchmark the turn-one forecast does not rank the cheap tier's failures: on 58 tasks the Sol-medium forecaster averaged 0.568 on tasks Kimi solves and 0.570 on tasks it fails (AUC 0.48), and Kimi forecasting its own chance did no better (AUC 0.53 as a probability; two adjacent rungs and no escalation on the ordinal ladder). The sessions the gate sent to Sol were solved at Sol's rate and the kept ones at Kimi's, which is what a random split gives. A random route sending 30 percent of sessions to Sol with no judge scored 21 of 30 for 55 units against the gated route's 26 for 46; the score gap is within noise, the cost gap comes from the stuck-run rescue capping looping cheap-tier sessions.

So on DeepSWE the gate behaves as a budget dial plus a stuck rescue rather than as a capability judgement. The feature is unchanged by this: where a forecaster does rank a pair's failures, the same threshold becomes selective, and the ordinal ladder and rung-name threshold exist for that case. I have added the analysis to the benchmark report and will keep the PR a draft until there is a pair on which the forecast ranks.

PR #684, opened 2026-09-11
feat(relay): report routing outcome evidence

afourniernv:
## What

- Forward existing libsy `OutcomeMetadata` through the LLM client's run observer.
- Add `outcome_id` and bounded, nested `evidence` to the existing Relay routing decision and route-execution error marks.
- Document the new fields in the Relay plugin's version-1 mark contract.

## Why

#647, #655, and #658 attach an outcome ID and bounded decision evidence to `RoutingOutcome` and the native `libsy.run` span. The Relay plugin does not export Switchyard's native spans, and the LLM client currently consumes that metadata before the plugin builds its marks.

As a result, Relay and Phoenix can show the selected and served models from #612, but not why the route was chosen. If every answer candidate fails, the routing outcome is also missing from the terminal error mark.

This passes the existing metadata through Relay's existing observation callback. It does not change routing, retries, fallback, responses, metrics, configuration, or mark names.

## How

- Add `RunObservation::Outcome(OutcomeMetadata)` and emit it once routing succeeds, before answer candidates run.
- Move the outcome ID and supported evidence fields into `switchyard.routing.decision` on success or `switchyard.routing.error` when all answer candidates fail.
- Keep evidence under an `evidence` object. Only known string fields (`source`, `verdict`, `trigger`, and `reason_code`) and numeric fields (`score`, `confidence`, and `threshold`) are retained.
- Ignore the new observation in `switchyard-server`, where the native `libsy.run` span already records the same metadata.

## Live Phoenix result

Before, the decision mark showed which model was selected and served, but not why:

Before: routing decision without outcome evidence

After, a scored Stage decision includes the outcome ID and its bounded evidence:

After: scored routing decision with source, confidence, and threshold

When both answer candidates fail, the terminal error mark retains the route's outcome ID and evidence:

After: terminal routing error retaining outcome evidence

## Notes for reviewers

Start with `RunObservation::Outcome` in `crates/libsy-llm-client/src/observation.rs` and its emission in `run.rs`. The Relay projection is in `crates/switchyard-nemo-relay-plugin/src/runtime.rs`.

`RunObservation` gains one public enum variant, so an external exhaustive match against current `main` must handle it. This is source-breaking for git consumers tracking `main`, but not a published API break: the latest released `switchyard-llm-client` is 0.2.0 and does not contain `RunObservation`. No existing function signatures change, and there are no new public methods or structs.

The route-execution regression makes both Stage answer candidates return HTTP 503 with retries disabled, then checks that the error mark retains the outcome ID and evidence.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace`
- `git diff --check`
- Live Relay to Switchyard to Phoenix check with both answer candidates failing

## Summary by CodeRabbit

* **New Features**
  * Added structured routing outcome metadata to run observations.
  * Routing telemetry now includes outcome identifiers and supported evidence fields for decisions and execution errors.
  * Evidence fields are filtered to documented string and numeric values.

* **Documentation**
  * Updated integration and plugin documentation to describe routing outcome metadata, evidence, and error fields.

* **Bug Fixes**
  * Preserved existing statistics behavior while supporting the new routing outcome observation.

### DIFF
--- a/crates/libsy-llm-client/src/observation.rs
+++ b/crates/libsy-llm-client/src/observation.rs
@@ -6,6 +6,7 @@
 use std::sync::Arc;
 use std::time::Duration;
 
+use switchyard_libsy::OutcomeMetadata;
 use switchyard_protocol::{ModelId, Usage};
 
 /// One completed model call observed while serving an algorithm run.
@@ -24,6 +25,8 @@ pub struct LlmCallObservation {
 /// One request-scoped observation emitted by the algorithm runner.
 #[derive(Clone, Debug)]
 pub enum RunObservation {
+    /// Metadata attached to the completed routing outcome.
+    Outcome(OutcomeMetadata),
     /// A completed model call requested by the algorithm for routing work.
     LlmCall(LlmCallObservation),
     /// A completed terminal model call made from the routing outcome.
--- a/crates/libsy-llm-client/src/run.rs
+++ b/crates/libsy-llm-client/src/run.rs
@@ -33,7 +33,8 @@ use crate::{metrics, observability};
 /// Run one request to completion, serving every offloaded model call with `client`.
 ///
 /// Returns the model selected by the algorithm and the final [`Response`]. `observer`, when
-/// present, receives each completed routing or answer call and the routing overhead.
+/// present, receives metadata for the routing outcome, each completed routing or answer call,
+/// and the routing overhead.
 ///
 /// `clients` resolves each offloaded call to the client for the target the algorithm
 /// selected — an algorithm may route among targets served by different providers, so this is
@@ -71,6 +72,11 @@ pub async fn run(
     metrics::record_routing_overhead(&algorithm_name, overhead);
 
     let selected_model_id = outcome.selected_model_id()?.clone();
+    if let Some(observer) = &observer
+        && let Some(metadata) = outcome.metadata
+    {
+        observer(RunObservation::Outcome(metadata));
+    }
     let (result, answer_duration) = if let Some(response) = outcome.response {
         (Ok(response), None)
     } else {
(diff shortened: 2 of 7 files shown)


bbednarski9:
streaming terminal errors still lose these fields. The PR passes Map::new() when returned_events creates the later switchyard.routing.error mark for an in-stream failure ([runtime change](https://github.com/NVIDIA-NeMo/Switchyard/pull/684/files)). At that point Route::execute already succeeded and the outcome observation has been consumed, so the streaming error mark cannot contain outcome_id/evidence. The new regression only tests the buffered “both candidates return 503” case.

bbednarski9:
I think we should try to get this into RC 2 for 0.3.0 because its an API change to public RunObservation - adding an enum variant breaks exhaustive matches for direct libsy-llm-client consumers.

bbednarski9 (changes requested):
just come comment from me and existing CR feedback

bbednarski9:
1. Evidence is allowlisted, but not actually bounded.
[evidence_for_mark](https://github.com/NVIDIA-NeMo/Switchyard/pull/684/files) restricts keys and JSON types, but it accepts arbitrary-length strings for source, verdict, trigger, and reason_code. OutcomeMetadata::new accepts caller-provided JSON, so a custom algorithm could emit a very large string—or sensitive text under an allowed key—and Relay would export it.
That conflicts with the PR description’s “bounded evidence” claim and weakens the apparent PII safeguard in the test. I’d request either:
- explicit length limits plus truncation/rejection, or
- a contract that these are controlled enum-like values, enforced at the producer side and documented clearly.

bbednarski9:
2. The documented outcome_id looks mandatory, but code makes it conditional.
The plugin emits it only for Some(outcome.metadata), while RoutingOutcome permits metadata to be absent. Built-in routes may always supply it now, but custom/external algorithms need not. The Relay README and integration docs should say outcome_id is optional, or the underlying outcome contract must guarantee metadata for every successful routed outcome.

afourniernv:
Thanks @bbednarski9 — fixed these in 698482bd:

- late stream failures now keep outcome_id and evidence
- evidence strings over 64 bytes are dropped
- the docs now say outcome_id is optional

Also covered the related CodeRabbit comments and tests. Focused tests, clippy, rustfmt, and the docs build pass locally. Agreed this should land before RC2 because of the RunObservation change.

bbednarski9 (approved):

Issue #565, opened 2026-08-27
[bug] base_instructions stub replaces Codex's own system prompt on every routed session


mrPronin:
## Symptom

Switchyard's Codex model card advertises `base_instructions: "You are Codex, a coding agent."`, and **Codex adopts it in place of its own system prompt**. Every routed session therefore runs with a six-word system prompt instead of the full Codex instructions, which changes agent behaviour on every turn and quietly invalidates any A/B comparison made *through* Switchyard.

## Cause

`crates/switchyard-server/src/lib.rs`, in `codex_model_entry_json`:

```rust
// Required `ModelInfo` string. Unlike the launcher, the server cannot read
// Codex's bundled prompt, so it sends a minimal stub.
"base_instructions": "You are Codex, a coding agent.",
```

The comment is accurate about the constraint but the consequence looks unintended: this is not a fallback Codex ignores, it is a value Codex **prefers over its own**.

## Evidence

Codex records the provenance of the instructions it used. Routed through Switchyard:

```json
"base_instructions": {
  "text": "You are Codex, a coding agent.",
  "provenance": {"type": "model", "model": "sy-solo"}
}
```

The same client on the same task with no proxy records its real multi-paragraph prompt (*"You are a coding agent running in the Codex CLI, a terminal-based coding assistant…"*).

Corroboration that the string travels over the wire rather than being a client-side default: it is **absent from the `codex` 0.150.1 binary** and present in the Switchyard binary.

## Expected vs. actual

- **Expected:** a proxy that routes requests does not replace the client's system prompt.
- **Actual:** it does, on every route, silently.

## Impact

Any benchmark or routing evaluation that compares "through Switchyard" against "direct" is measuring two different agents, not two different routes. That is a measurement-integrity problem rather than a crash, which is why it is easy to miss.

Two neighbouring fields on the same entry have the same shape of problem: `default_reasoning_level` is forced to `"xhigh"` for any reasoning-capable route, and `truncation_policy` to `{"mode": "tokens", "limit": 10000}`, regardless of client or target.

## Suggested direction

Omit `base_instructions` if `ModelInfo` permits it, so Codex keeps its own; otherwise make it configurable per route with "leave the client's prompt alone" as the default. Happy to send a PR if you can say which of those you would accept.

## Environment

- Commit: `7f3b2fe9` (`main`)
- Client: `codex-cli` 0.150.1

mrPronin:
[...]

PR #686, opened 2026-09-12
fix(server): allow per-route Codex base instructions

ting-hong-shieh:
## What

Add an optional per-route `base_instructions` setting to the Codex model catalog. The text is served verbatim; empty or whitespace-only values are rejected. Routes without the setting retain the existing placeholder, with a warning listing their IDs when the discovery router is built. The warning covers standalone and embedded hosts and does not log prompt text.

## Why

Codex adopts the catalog's prompt, so Switchyard's six-word placeholder replaces the agent's original instructions. Omitting the field can invalidate the entire catalog, including vision metadata.

Addresses #565. This follows the per-route direction investigated in #565 and implemented in #631, which was closed with a request to split concerns. This PR contains only the prompt configuration work.

## Notes for reviewers

The setting travels through runner configuration and `Route` into `/v1/models`; it does not rewrite request messages. `ModelInfo` keeps its existing public fields, `ModelCapabilities` remains `Copy`, and existing constructor and accessor signatures are unchanged. A new read-only `Runner::routes()` iterator lets the server obtain the prompt without changing those APIs.

Existing deployments remain loadable, but operators must supply the same resolved prompt as their direct session to obtain prompt parity. This does not automatically restore the original prompt for unconfigured routes or change reasoning/truncation defaults.

Validation on macOS:

- Rust workspace: 750 passed, 1 ignored; prefill-router runner tests passed.
- Workspace and prefill-router clippy with `-D warnings`, rustfmt, ruff, mypy, and strict MkDocs build passed.
- Python tests with the rebuilt native extension: 116 passed, 2 integration tests deselected, 2 subtests passed.
- Regression coverage includes distinct route prompts, whitespace/Unicode preservation, blank rejection, default and vision metadata, and warning scope/content. An external Rust compile probe verified that existing `ModelInfo` construction and exhaustive destructuring still compile.
- Codex CLI 0.152.1, using a local mock Responses backend and a one-pixel image:

  | Session | Base instructions | Image reached upstream |
  |---|---|---|
  | Direct, bundled `gpt-5.6-sol` metadata | 15,685 bytes | Yes |
  | Configured Switchyard route | Same 15,685 bytes, byte-for-byte | Yes |
  | Unconfigured route | 30-byte placeholder | Yes |

  Tested both the server's actual catalog JSON loaded through `model_catalog_json` and HTTP discovery with a dummy command-auth provider. No external model was called. HTTP discovery required a warmup: Codex 0.152.1 initially reused the preceding provider's catalog cache and selected fallback metadata for the new route; after refresh, the configured prompt matched. This separate client cache behavior is not fixed here.

## Summary by CodeRabbit

* **New Features**
  * Added optional route-level base instructions for Codex model discovery.
  * Model listings now use each route’s configured instructions, with a default fallback when omitted.
  * Added access to configured routes and their settings.

* **Bug Fixes**
  * Blank or whitespace-only base instructions are rejected.
  * Servers warn when routes lack base instructions.

* **Documentation**
  * Documented configuration, validation, fallback, and warning behavior.

### DIFF
--- a/crates/switchyard-runner/src/config.rs
+++ b/crates/switchyard-runner/src/config.rs
@@ -69,6 +69,7 @@ struct RouteConfig {
     tool_calling: Option<bool>,
     reasoning: Option<bool>,
     vision: Option<bool>,
+    base_instructions: Option<String>,
     algorithm: AlgorithmSpec,
 }
 
@@ -88,6 +89,7 @@ impl<'de> Deserialize<'de> for RouteConfig {
         let tool_calling = take_optional(&mut table, "tool_calling")?;
         let reasoning = take_optional(&mut table, "reasoning")?;
         let vision = take_optional(&mut table, "vision")?;
+        let base_instructions = take_optional(&mut table, "base_instructions")?;
         let algorithm = AlgorithmSpec::deserialize(toml::Value::Table(table))
             .map_err(serde::de::Error::custom)?;
         Ok(Self {
@@ -96,6 +98,7 @@ impl<'de> Deserialize<'de> for RouteConfig {
             tool_calling,
             reasoning,
             vision,
+            base_instructions,
             algorithm,
         })
     }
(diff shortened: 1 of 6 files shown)


messiaen:
seems to be OBE after #693

PR #687, opened 2026-09-13
feat(libsy): cap the windowed classifier judge payload

ardada2468:
## What

Add a route-level `judge_char_budget` that caps the characters sent to the classifier judge per call. It applies in all three modes. The default is 18000, which the escalation judge used as a hard-coded constant until now.

With `recent_turn_window`, the window ends at the first message that does not fit, so the newest turns survive. Without a window, an oversized task statement is clipped and marked with `...[trimmed]`. Tool JSON is never clipped. Budgets below 256 are rejected so the routing instruction always fits.

## Why

Addresses #279, part 2. Part 1 landed in #520.

`recent_turn_window` counts turns, and turns vary a lot in size. Four turns can be a few hundred characters, or tens of thousands if one carries a large tool result. So judge cost and latency depended on the request, not the config. The escalation judge had a fixed cap. The classifier judges had none.

This is the judge budget half of #631, split out as requested there. The Codex half was resolved by #693.

## Notes for reviewers

- Window selection is one newest-to-oldest pass in `window_start`. It tracks the running size and tool pairing together, so cost is one visit per message. When the budget cuts between a call and its result, the result is dropped rather than sent orphaned.
- Reasoning counts as 0 characters. #610 strips it before the judge is called, so counting it would drop turns the judge could have kept.
- Characters, not tokens: a token count needs a tokenizer per judge model, and the existing `window_message_chars` and `transcript_max_chars` settings count characters too. The docs note roughly four characters per token.
- `escalation.judge_char_budget` is rejected as an unknown field, matching how `max_output_tokens` works. The route-level key is the only spelling.
- `truncate_middle` moved from `escalation.rs` to `util.rs` so both judges share it.

Validation: `cargo test --workspace`, `cargo clippy --workspace --all-targets -D warnings`, `cargo fmt --check`, `ruff check`, `mypy`.


crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -351,6 +461,21 @@ const fn default_judge_max_output_tokens() -> u64 {
...
+    DEFAULT_JUDGE_CHAR_BUDGET
+}
+
+/// A zero budget would clip every message to the trim marker, leaving the judge a payload
+/// it cannot route. Rejecting it at construction beats serving empty verdicts.
+fn validate_judge_char_budget(budget: usize) -> Result<()> {
+    if budget == 0 {
+        return Err(LibsyError::AlgorithmError {
+            message: "judge_char_budget must be at least 1".to_string(),
+        });
```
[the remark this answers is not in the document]
ardada2468: Good catch. Budgets now must be at least 256, so the routing instruction always fits, with a boundary test in cc8fbe9d.

docs/reference/toml_schema.md
```diff
@@ -202,6 +202,7 @@ Capability mode classifies before serving. See
 | `classify_trigger` | No | `every_request` | When the judge runs. `every_request` judges every request, tool continuations included. `user_turn` judges each new user message and retains that target across intervening tool calls only when requests carry a session ID; without a session ID, it behaves like `every_request`. `new_session` judges once and reuses that target for the session. |
 | `message_hash_fallback` | No | `false` | Keys affinity on the first user message. Requires `classify_trigger = "new_session"`. |
 | `recent_turn_window` | No | unset | When unset, the judge sees the opening task and latest user follow-up, when present. When set, it also sees trailing turns. |
+| `judge_char_budget` | No | `18000` | Most characters a windowed judge payload may use. The window narrows from the oldest turn until it fits, so one large tool result cannot decide judge cost and latency. Ignored without `recent_turn_window`. |
```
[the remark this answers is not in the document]
ardada2468: Agreed. All three rows now say the budget must be at least `256`, which is the new minimum from cc8fbe9d.

switchyard_rust/libsy.py
```diff
@@ -75,6 +75,7 @@ def __init__(
             session_affinity: bool = False,
             message_hash_fallback: bool = False,
             recent_turn_window: int | None = None,
+            judge_char_budget: int = 18_000,
```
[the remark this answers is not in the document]
ardada2468: Partly agree. These docstrings list constraints rather than every parameter, so both now say `judge_char_budget` caps the windowed payload, must be at least 256, and is ignored without `recent_turn_window`.

crates/libsy/src/algorithms/util.rs
```diff
@@ -29,3 +29,37 @@ pub(crate) fn decisive(target: &ModelId) -> Classification {
 
 /// Default completion budget for internal classifier and escalation judge calls.
 pub(crate) const DEFAULT_JUDGE_MAX_OUTPUT_TOKENS: u64 = 4_096;
+
+/// Default character budget for a judge payload, shared by the escalation judge and the
+/// windowed classifier judges so one turn carrying a large tool result cannot decide how
+/// much a judge call costs.
+pub(crate) const DEFAULT_JUDGE_CHAR_BUDGET: usize = 18_000;
```
messiaen, line 36: More natural to have a token limit instead of char limit.

crates/libsy/src/algorithms/util.rs
```diff
@@ -29,3 +29,37 @@ pub(crate) fn decisive(target: &ModelId) -> Classification {
...
+
+/// Separator marking where [`truncate_middle`] dropped a message's interior.
+pub(crate) const TRIM_MARKER: &str = " ...[trimmed] ";
```
messiaen, line 52: Is this marker model specific at all?

crates/switchyard-runner/src/algorithm.rs
```diff
@@ -245,6 +247,11 @@ pub struct LlmClassifierRouteConfig {
     /// How many trailing turns the judge sees. Unset shows it the opening task
     /// and the latest user follow-up only.
     pub recent_turn_window: Option<usize>,
+    /// Most characters a windowed judge payload may use. The window narrows from the
+    /// oldest turn until it fits, so one large tool result cannot decide the judge's
+    /// cost. Ignored without `recent_turn_window`.
+    #[serde(default = "default_judge_char_budget")]
+    pub judge_char_budget: usize,
```
messiaen, line 254: This should effect the escalation judge also, no? Currently it looks not to.

crates/libsy/src/algorithms/util/escalation.rs
messiaen: This should be replaced with the new max chars config, right?

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -160,17 +160,119 @@ fn task_messages(messages: &[Message]) -> Vec<Message> {
...
+/// Whole turns are dropped from the oldest end rather than clipping individual messages,
+/// because [`trim_messages`] is what keeps tool calls paired with their results; removing
+/// messages by hand would hand the judge a result whose call was never introduced.
+fn window_within_budget(messages: &[Message], window: usize, budget: usize) -> Vec<Message> {
+    let mut window = window;
+    let mut kept = trim_messages(messages, window);
+    while window > 0 && payload_chars(&kept) > budget {
+        window -= 1;
+        kept = trim_messages(messages, window);
+    }
```
messiaen: I think there is an edge case here if a single message / turn does not fit in the budget.

Or more generally this could loop quite a few times if window is large, and the bulk of the chars are early in the conversation.

messiaen (changes requested):

crates/libsy/src/algorithms/util.rs
ardada2468: Kept characters. A token count needs a tokenizer per judge model, and the existing `window_message_chars` and `transcript_max_chars` settings count characters too. The docs now note roughly four characters per token.

crates/libsy/src/algorithms/util.rs
ardada2468: No. It is plain text the escalation judge already used; moving it to `util.rs` just lets both judges share it.

crates/switchyard-runner/src/algorithm.rs
ardada2468: Right, it did not. As of 92052f4f the route-level key applies in all three modes and replaces the old `MAX_REQUEST_CHARS` constant in escalation mode. A runner test checks the value reaches the escalation judge.

crates/libsy/src/algorithms/util/escalation.rs
ardada2468: Yes, done in 92052f4f. `MAX_REQUEST_CHARS` is gone and `summarize_for_judge` reads `config.judge_char_budget`.

crates/libsy/src/algorithms/llm_class.rs
ardada2468: Both fixed in 92052f4f. Window selection is now one newest-to-oldest pass that tracks size and tool pairing together, so cost is one visit per message. A message that does not fit on its own drops with the older turns; newer turns that fit stay, and the anchors are clipped instead of dropped.

ardada2468:
Superseded by #791, which carries the same change rebased onto current main with the review feedback applied.

PR #688, opened 2026-09-14
feat(relay-plugin): support nemo relay 0.8 through 0.x

bbednarski9:
## Summary

- declare NeMo Relay compatibility as `>=0.8.0, <1.0.0` across build metadata, bundle manifest, and user-facing documentation
- refresh Cargo.lock to resolve `nemo-relay-plugin` and `nemo-relay-types` 0.8.4
- assert the packaged manifest retains the compatibility range

## Validation

- `cargo test --locked -p switchyard-nemo-relay-plugin --no-fail-fast`
- `pytest tests/relay_plugin/test_package_bundle.py -v`
- `make publish`
- clean compile against `nemo-relay-plugin 0.8.0`

## Prerelease note

Standard semver prerelease matching means this range admits stable 0.9.x releases but does not itself admit `0.9.0-rc.1`. Explicit prerelease support needs a separate RC-inclusive constraint.


ibankolehame:
@bbednarski9 ,  I really appreciate the quick turnaround. I went through the diff, and the compatibility range now looks consistent across the dependency, manifest, and documentation. The packaging test is a nice addition as well.

One quick question, since the new range includes stable Relay 0.9, are you planning to include a compile or plug in load check against 0.9 in the release validation? I saw the current validation covers 0.8.0 and the lock file uses 0.8.4. If 0.9 is not available yet, that could be a follow up release check.

The note about release candidates makes sense. Everything else looks good to me, and this resolves the earlier inconsistency , thank you!

bbednarski9:
Hey @ibankolehame thanks for the call-out. Now is a good time to update the docs with instructions on the new NeMo-Relay-Plugins repo for distributing trusted external dynamic plugins (like switchyard). See docs update in [4cc3100](https://github.com/NVIDIA-NeMo/Switchyard/pull/688/commits/4cc3100af50b5e6cfc172968841fd5d201e6b8c4)

ibankolehame:
Thanks so much, @bbednarski9 . I reviewed the update in 4cc3100, and the new bundle instructions are really helpful. This gives us a much clearer deployment path for AA. This addresses my concern. I really appreciate you taking the time to update both the compatibility range and the documentation. It looks perfect from my side!

grahamking (approved):

PR #690, opened 2026-09-14
fix: Don't forward HTTP headers in forward_auth=false

### LINKED ISSUE #410: [bug]: proxy forwards client x-* headers including x-goog-api-key to the upstream
## symptom
switchyard-server 0.2.0 copies inbound non-reserved headers onto the upstream http call. a client `x-goog-api-key` or any other `x-*` secret (except the reserved `authorization` / `x-api-key` / `cookie` set) is sent to whatever backend the admin configured.

## reproduction
capture rig (no provider keys):

```bash
python3 tools/capture_headers.py 9999 cap.jsonl canned-openai.json
switchyard-server --config switchyard-capture.toml --host 127.0.0.1 --port 9000
  -H 'x-goog-api-key: CANARY_X_GOOG_API_KEY' \
  -H 'x-custom-internal-secret: CANARY_X_CUSTOM_SECRET' \
  -H 'authorization: Bearer CANARY_AUTHORIZATION' \
  -d '{"model":"captured-model","messages":[{"role":"user","content":"ping"}]}'
```

the mock's recorded request includes `x-goog-api-key: CANARY_X_GOOG_API_KEY` and `x-custom-internal-secret: CANARY_X_CUSTOM_SECRET`. `authorization` on the upstream is the deployment value `Bearer x`, not the client bearer. same headers also appear when the inbound path is `post /v1/messages`. 5/5.

live gemini (openai-compat base url, `api_key_env = GEMINI_API_KEY`): the same fake `x-goog-api-key` still returns http 200 because google prefers the configured bearer. the mock is what shows the header still went out.

## expected vs. actual
- expected: secret-bearing client headers such as `x-goog-api-key` are dropped unless the admin opts into header forwarding, matching how `authorization` / `x-api-key` / `cookie` are already reserved.
- actual: those headers are attached to the upstream request by default. litellm only does this when `forward_client_headers_to_llm_api` is true (off by default).

## environment
- os / arch: macos arm64
- install path: source build `tools/switchyard/target/release/switchyard-server`
- inbound format (chat completions / anthropic messages / responses): chat completions and anthropic messages
- backend (openai / anthropic / nvidia inference hub / other): openai-compatible mock, plus live gemini 2.5 flash via `https://generativelanguage.googleapis.com/v1beta/openai`

## additional context
this is a confidentiality leak (client secrets leave the proxy), not an auth bypass of the deployment key. no real provider keys appeared in error bodies.

wire captures and replay tests: https://github.com/Atharva-Kanherkar/kairo/pull/3

grahamking:
. In `forward_auth = false` mode:

Only forward HTTP header `x-request-id`. We are going to re-build the
request with our own auth, so nothing need to survive apart from that

. In `forward_auth = true` mode:

Strip the hop-by-hop headers but forward everything else.

Also strip auth from HTTP error messages.

Most of the change is a new test.

Closes: https://github.com/NVIDIA-NeMo/Switchyard/issues/410

Reviewed-by: Claude:Sonnet 5 medium

## Summary by CodeRabbit

- **New Features**
  - Added configurable authentication header forwarding for supported providers.
  - Forwarded application headers are filtered to exclude transport-related and provider-owned headers.
  - Provider-specific credential and beta-header handling is supported.

- **Security**
  - Forwarded credentials are redacted from upstream error responses.
  - Without authentication forwarding enabled, only the request ID is forwarded.

- **Documentation**
  - Updated header-forwarding guidance and configuration details.


crates/libsy-llm-client/src/client.rs
```diff
@@ -721,21 +721,29 @@ fn convert_reqwest_error(error: reqwest::Error) -> LlmClientError {
...
+            !is_non_forwardable_header(name.as_str(), headers)
+                && !backend.is_provider_owned_header(name.as_str())
+        } else {
+            is_allowed_metadata_header(name.as_str())
+        };
+        if !is_allowed {
             continue;
         }
-        builder = builder.header(name, value);
+        forwarded.append(name, value.clone());
```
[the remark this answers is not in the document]
grahamking: No. The whole point is to forward the Codex / Claude auth _through_ Switchyard. This allows us for example to route between Luna and Sol automatically to optimize costs.

crates/libsy-llm-client/src/client.rs
```diff
@@ -29,28 +29,27 @@ use crate::error::{LlmClientError, Result};
...
+const NON_FORWARDABLE_HEADERS: &[&str] = &[
     "host",
     "content-length",
     "connection",
-    "authorization",
-    "proxy-authorization",
+    "proxy-connection",
+    "keep-alive",
     "proxy-authenticate",
-    "cookie",
```
ryan-lempka, line 42: So we are dropping cookie/set-cookie from `NON_FORWARDABLE_HEADERS`? Just confirming that was intentional

grahamking: It is.

- In `forward_auth = true` we need the cookies, they are usually essential for auth.
- In `forward_auth = false` it's not in the list because we drop _all_ headers, except a one-item whitelist.

ryan-lempka (approved):
LGTM

PR #692, opened 2026-09-14
fix: Session affinity stores Category for fallback

grahamking:
Original bug report: With classify_trigger = "new_session", a custom classifier retains the selected primary model but loses the model group associated with that decision. If the retained primary model fails during a later request in the same session, Switchyard tries a target from models.any before the remaining fallback in the originally selected group.

To fix this we store the original `Category`, so that we can find other models in that category.


ayushag-nv (approved):
Nice addition

PR #694, opened 2026-09-14
fix(protocol): keep Codex compaction requests on the parent route

elyasmnvidian:
```http
x-codex-turn-metadata: {"session_id":"codex-session","thread_id":"compact-child","parent_thread_id":"codex-root","thread_source":"subagent","subagent_kind":"compact"}
```

When a Codex compaction request includes these child-thread fields, Switchyard sends it to the subagent worker and skips the parent route's classifier. Compaction summarizes the parent conversation and should use that conversation's route.

The recorded request selected the worker model:

```text
x-model-router-selected-model: openai/openai/gpt-5.6-sol
```

`parse_sub_agent` now checks the subagent task kind before using the child-thread fields. `collab_spawn` and `review` use subagent routing; `compact`, `memory_consolidation`, and unknown kinds use the parent route.

When no kind is supplied, the existing child-thread checks still apply. An explicit `x-switchyard-is-subagent: false` still disables subagent routing.

## Before and after

The live check sent four requests to a Composite route: a route that uses a classifier to choose between its capable and efficient models. The route also had a separate subagent worker.

| Request metadata | Before | After |
| --- | --- | --- |
| `x-openai-subagent: compact` | Parent route | Parent route |
| Parent thread ID without a subagent task kind or source | Parent route | Parent route |
| Structured `compact` header above | Subagent worker; no parent classifier call | Parent route; one parent classifier call |
| Same header with `subagent_kind: review` | Subagent worker | Subagent worker |

After the fix, the structured compaction request selected `nvidia/nvidia/nemotron-3.5-super-text-preview`, the parent's efficient model. The total parent classifier count increased from two to three. The review request still selected `openai/openai/gpt-5.6-sol`, the worker model. All four requests returned `PONG`.

The server regression test checks eight metadata cases through Chat Completions, then repeats the compaction and review cases through Anthropic Messages and Responses. It checks each selected model and the final total of seven parent classifier calls. The test fails with the old code, which selects `model/strong` for structured compaction, and passes with the fix, which selects `model/weak` through the parent classifier.

<details>
<summary>Request and configuration used for the live check</summary>

Send each metadata header to `POST /v1/chat/completions` with this body:

```json
{"model":"qa/composite-subagent-maintenance","messages":[{"role":"user","content":"Reply with exactly PONG."}],"max_tokens":16}
```

```toml
schema_version = 1

[llm_clients.gateway]
format = "openai_chat"
base_url = "https://gateway.example/v1"
api_key_env = "GATEWAY_API_KEY"
max_retries = 0

[targets.sol]
id = "openai/openai/gpt-5.6-sol"
llm_client = "gateway"

[targets.weak]
id = "nvidia/nvidia/nemotron-3.5-super-text-preview"
llm_client = "gateway"

[routes.composite]
id = "qa/composite-subagent-maintenance"
type = "composite"

[routes.composite.classifier]
target = "sol"
base_threshold = 0.01
classify_trigger = "user_turn"

[routes.composite.stage]
capable_target = "sol"
efficient_target = "weak"
confidence_threshold = 1.0
recent_turn_window = 5

[routes.composite.subagents]
type = "passthrough"
target = "sol"
```

The gateway address is replaced with an example address. Model selections came from the response headers; classifier counts came from `GET /v1/stats`.

Run the regression test with:

```bash
cargo test -p switchyard-server --test server codex_maintenance_uses_the_parent_classifier_across_apis
```

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved delegated-work classification by honoring explicit opt-out flags and recognizing only supported delegation types.
  - Refined handling of child lineage and unspecified delegation metadata for more accurate routing.
  - Standardized maintenance behavior across Chat Completions, Anthropic Messages, and Responses APIs.
  - Ensured explicit subagent settings take precedence over inferred metadata.
  - Improved consistency of target selection and related request statistics across supported APIs.


crates/switchyard-server/tests/server.rs
```diff
@@ -2749,6 +2749,106 @@ target = "shared"
     Ok(())
 }
 
+#[tokio::test]
+async fn codex_maintenance_uses_the_parent_classifier_across_apis() -> TestResult {
```
ayushag-nv, line 2894: @elyasmnvidian do we need such a big test ? any possibility of cutting it down ?? Does it even deserve a test ???

elyasmnvidian: Shortened the setup and reduced routing requests from 21 to 12. Compaction and review still run through Chat Completions, Anthropic Messages, and Responses; the other metadata cases run once through Chat Completions. I removed repeated model and stats checks while keeping the parent classifier count. The shorter test fails when the old code sends structured compaction to the subagent worker and passes with the fix.

ayushag-nv (approved):

PR #695, opened 2026-09-14
fix(libsy): classify subagent requests after tool results

elyasmnvidian:
```json
{"model":"agent","messages":[
  {"role":"user","content":"Review this parser carefully."},
  {"role":"assistant","content":null,"tool_calls":[{"id":"call_1","type":"function","function":{"name":"read_file","arguments":"{\"path\":\"parser.py\"}"}}]},
  {"role":"tool","tool_call_id":"call_1","content":"def parse(value): return value"}
],"max_tokens":32,"stream":false}
```

When a subagent sends this request after a tool returns, Switchyard skips the classifier and uses `default_target`, even with `classify_trigger = "every_request"`. In the live check, the classifier selected the capable model for the opening request, but the request ending with a tool result selected the default worker:

```text
x-model-router-selected-model: openai/openai/gpt-5.6-luna
```

Switchyard converts a Chat Completions tool message into a user message containing a tool result. `delegated_prompt_request` previously looked only at the newest user message, so it found no task text and skipped classification.

`delegated_prompt_request` now searches user messages from newest to oldest for nonblank text. For the request above, it sends `Review this parser carefully.` to the classifier. It does not use the tool output as the task. Requests with no user text still skip classification.

## Before and after

The live check sent three requests from the same Claude Code subagent. The classifier was configured to select `capable` every time.

| Request | Before | After |
| --- | --- | --- |
| Opening user task | Capable model | Capable model |
| Same task followed by a tool result | Default worker; classifier skipped | Capable model; classifier called |
| Tool result followed by new user text | Capable model | Capable model |

The total classifier count increased from two to three. All requests returned HTTP 200 with no classifier or answer errors. With `every_request`, requests ending with a tool result now use the classifier when the conversation contains user task text. `new_session` still reuses the subagent's initial decision.

The server regression test covers the opening task, a tool result, a whitespace-only user message, new task text, and a request with no user task text. Each case runs through Chat Completions, Anthropic Messages, and Responses. The test checks the classifier's exact instructions and latest nonblank task text. It also checks that the selected model receives the original system instructions, user context, task text, tool output, and tool call ID. Requests with no user task text use the default target without a classifier call.

The test fails with the old code, which selects `model/weak` after a tool result, and passes with the fix, which calls the classifier and selects `model/strong`.

<details>
<summary>Request headers and configuration used for the live check</summary>

Send the JSON above to `POST /v1/chat/completions` with:

```http
x-claude-code-session-id: review-session
x-claude-code-agent-id: review-child
```

```toml
schema_version = 1

[llm_clients.gateway]
format = "openai_responses"
base_url = "https://gateway.example/v1"
api_key_env = "GATEWAY_API_KEY"
max_retries = 0

[targets.worker]
id = "openai/openai/gpt-5.6-luna"
llm_client = "gateway"
extra_body = { reasoning = { effort = "low" } }

[targets.reviewer]
id = "openai/openai/gpt-5.6-sol"
llm_client = "gateway"
extra_body = { reasoning = { effort = "low" } }

[routes.agent]
id = "agent"
type = "passthrough"
target = "worker"

[routes.agent.subagents]
type = "llm_classifier"
mode = "custom"
models = { judge = ["reviewer"], capable = ["reviewer"], efficient = ["worker"], any = ["worker", "reviewer"] }
default_target = "efficient"
classify_trigger = "every_request"
max_output_tokens = 32
prompt = 'Return exactly the JSON object {"target":"capable"}.'
response_schema = '''
{
  "type": "object",
  "properties": {"target": {"type": "string", "enum": ["capable", "efficient"]}},
  "required": ["target"],
  "additionalProperties": false
}
'''
policy = { type = "target_selector", selector = "/target" }
```

The gateway address is replaced with an example address. Model selections came from the response headers; classifier counts came from `GET /v1/stats`.

Run the regression test with:

```bash
cargo test -p switchyard-server --test server subagent_tool_continuations_are_classified_on_every_request_across_apis
```

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved subagent continuation handling when conversations include tool-only messages.
  - The latest available textual task prompt is now used for classification, improving routing accuracy.
  - Requests without a textual task prompt now fall back to the efficient model.

- **Tests**
  - Added coverage across Chat Completions, Anthropic Messages, and Responses APIs to verify continuation classification, context preservation, model routing, and classifier statistics.


crates/switchyard-server/tests/server.rs
```diff
@@ -2749,6 +2749,162 @@ target = "shared"
     Ok(())
 }
 
+#[tokio::test]
+async fn subagent_tool_continuations_are_classified_on_every_request_across_apis() -> TestResult {
```
ayushag-nv, line 2754: do we need such a big test ??

elyasmnvidian: Simplified the setup and request sequence, and removed repeated response-model and stats checks. The test still checks the exact classifier input and confirms that the selected model receives the original instructions, context, task, and tool results. It covers requests with no task text and the blank user message suggested in the other review across Chat Completions, Anthropic Messages, and Responses. The shorter test fails when the old code skips classification after a tool result and passes with the fix.

elyasmnvidian:
Added the blank-message case from [the review](https://github.com/NVIDIA-NeMo/Switchyard/pull/695#pullrequestreview-5203578032). The test appends a whitespace-only user message and checks that the classifier still receives the earlier task. It passes through Chat Completions, Anthropic Messages, and Responses. A short comment explains why tool results and blank user messages must not replace the classifier's task text.

ayushag-nv (approved):
Looks good

PR #696, opened 2026-09-14
fix(libsy): warn when Claude Code predates sub-agent identity headers

linj-glitch:
## What

Sub-agent routing detects Claude Code children only through the `x-claude-code-agent-id` header. Claude Code started sending that header in 2.1.139. Older builds, such as the 2.1.121 QA used, send only the session id, so their sub-agent requests look like the parent's and silently route through the parent route.

This PR:
- adds a `subagent_identity_unsupported` flag to `switchyard_protocol::Metadata`, set during header normalization when the calling harness build is known not to send child identity (today: Claude Code before 2.1.139, read from its `User-Agent`);
- makes `SubagentRouter` log one warning per sub-agent route, without naming versions, when such a request reaches a route with a `subagents` table;
- documents the version floor in the sub-agent routing guide and the changelog.

Routing behavior is unchanged. Below 2.1.139 the parent and its children send identical headers and the body carries no stable child marker, so a heuristic fallback would risk misrouting parent turns.

## Why

Fixes SWITCH-1437 (NvBug 6771746). QA ran Claude Code 2.1.121 against v0.3.0-rc.1 with a Composite route and a `subagents` passthrough. The child target recorded zero calls with no error or warning. The header floor comes from the Claude Code 2.1.139 release notes: "API requests from subagents now carry x-claude-code-agent-id / x-claude-code-parent-agent-id headers."

## Notes for reviewers

Harness and version knowledge lives in `crates/protocol/src/metadata.rs` (`claude_lacks_child_identity`), next to the other Claude Code header handling. `libsy` only reads the resulting bool in `SubagentRouter::warn_if_subagent_identity_unsupported`. The Claude Code `User-Agent` has the form `claude-cli/<version> (...)`. One test in each crate: the protocol test covers the version floor and prerelease handling, the libsy test covers the once-only warning on the parent path. `cargo clippy --workspace --all-targets -- -D warnings` is clean.


crates/libsy/src/algorithms/subagent.rs
```diff
@@ -184,6 +241,81 @@ mod tests {
...
+            assert_eq!(selected, "parent");
+        }
+        assert!(router.outdated_claude_code_warned.load(Ordering::Relaxed));
+
+        let (selected, _) = test_drive_with_models(
+            configured(Arc::new(ScriptedClassifier {
+                calls: AtomicUsize::new(0),
+            }))?,
+            claude_code("claude-cli/2.1.139 (external, cli)"),
+            models,
```
ayushag-nv: @linj-glitch Can you please trim these tests down. I don't expect these many tests to be added. Usually a single test should be sufficient for this small fix.

ayushag-nv (changes requested):

crates/libsy/src/algorithms/subagent.rs
```diff
@@ -40,10 +41,32 @@ impl SubagentRouterConfig {
...
+    let mut parts = version.split('.').map(|part| part.parse::<u64>().ok());
+    Some([parts.next()??, parts.next()??, parts.next()??])
+}
+
+/// The Claude Code version behind `metadata` when it predates child identity headers.
+fn outdated_claude_code(metadata: Option<&Metadata>) -> Option<[u64; 3]> {
+    let version = claude_code_version(metadata?.user_agent.as_deref()?)?;
+    (version < MIN_CLAUDE_CODE_VERSION).then_some(version)
+}
+
```
ayushag-nv: Do we need to warn about outdated claude ? I feel backward compatibility is also important. So we should be supporting both here.

ayushag-nv (dismissed):

crates/libsy/src/algorithms/subagent.rs
linj-glitch: Done. The three tests are folded into one, `outdated_claude_code_warns_once_and_routes_through_the_parent`, in 4c6fc31a. It checks the version floor, that an old build still routes to the parent, and that the warning fires once.

crates/libsy/src/algorithms/subagent.rs
linj-glitch: Below 2.1.139 a child request is identical to a parent request on the wire: same session header, same body metadata, same User-Agent. The only differences are the system prompt and the first user message, and any heuristic on those misroutes the parent after compaction or on side-queries, which is worse than routing the child through the parent. Codex has carried lineage for a long time, so this only affects Claude Code.

I'd propose we set the supported Claude Code floor to 2.1.139 (released May 11; the benchmark already pins 2.1.211) and keep this warning so the gap is visible instead of silent.

crates/libsy/src/algorithms/subagent.rs
```diff
@@ -79,7 +102,37 @@ impl SubagentRouter {
             .with_classifier(Arc::new(SubagentGate::new(config.classifier)))
             .with_classifier(Arc::new(DefaultCategoryClassifier(config.default_target)));
 
-        Ok(Self { parent, subagent })
+        Ok(Self {
+            parent,
+            subagent,
+            outdated_claude_code_warned: AtomicBool::new(false),
```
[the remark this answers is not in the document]
linj-glitch: The intended scope is one warning per sub-agent route, not per process: each route has its own parent target and its own operators may want to know which route is affected. This matches how the affinity router's unkeyed-request warning is scoped today. The docs and changelog say per route; I fixed the PR body, which wrongly said per process.

crates/protocol/src/metadata.rs
```diff
@@ -186,6 +189,9 @@ pub struct Metadata {
     pub session_final: Option<bool>,
     /// External trace/request id for joining with the host's telemetry.
     pub correlation_id: Option<String>,
+    /// The caller's `User-Agent`, when present. Coding harnesses name themselves and
+    /// their version here, such as `claude-cli/2.1.139 (external, cli)`.
+    pub user_agent: Option<String>,
```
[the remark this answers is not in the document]
linj-glitch: Documented in addcb196: the changelog entry now states that `switchyard_protocol::Metadata` gains a public `user_agent` field and that downstream full struct literals must add it or use `..Metadata::default()`.

crates/libsy/src/algorithms/subagent.rs
```diff
@@ -40,10 +41,32 @@ impl SubagentRouterConfig {
...
 
+/// First Claude Code release whose sub-agents send `x-claude-code-agent-id`.
+const MIN_CLAUDE_CODE_VERSION: [u64; 3] = [2, 1, 139];
+
+/// Parses the version out of a Claude Code `User-Agent` such as
+/// `claude-cli/2.1.139 (external, cli)`. Returns `None` for any other client.
+fn claude_code_version(user_agent: &str) -> Option<[u64; 3]> {
+    let version = user_agent
+        .strip_prefix("claude-cli/")?
+        .split([' ', '-', '+'])
```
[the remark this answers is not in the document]
linj-glitch: Fixed in 54cc4742: the parser now keeps a prerelease flag, and a prerelease of the floor itself (`2.1.139-beta.1`) is treated as outdated. Build metadata after `+` is still ignored. Covered by a new assertion in the existing test.

crates/libsy/src/algorithms/subagent.rs
```diff
@@ -79,7 +110,37 @@ impl SubagentRouter {
...
+        let Some([major, minor, patch]) = outdated_claude_code(request.metadata.as_ref()) else {
+            return;
+        };
+        if self
+            .outdated_claude_code_warned
+            .swap(true, Ordering::Relaxed)
+        {
+            return;
+        }
+        let [min_major, min_minor, min_patch] = MIN_CLAUDE_CODE_VERSION;
```
ryan-lempka: @linj-glitch I don't think libsy should be aware of agent versions. This feels like something that belongs in translation / normalization. Perhaps in `crates/protocol/src/metadata.rs`. And then SubagentRouter can read a bool like: `subagent_identity_unsupported`.

I would also say we don't need to call out specific versions. We could just detect when no recognized child entity and warn harness upgrade may be required.

linj-glitch: Done in 5a1c5626. Version knowledge now lives in `crates/protocol/src/metadata.rs` next to the other Claude Code header handling: `claude_lacks_child_identity` reads the `User-Agent` during normalization and sets a new `Metadata::subagent_identity_unsupported` bool. `SubagentRouter` only reads that bool and warns once per route that the harness does not send sub-agent identity and an upgrade may be required, with no version named. The `user_agent` field is gone. One test per crate.

ryan-lempka (approved):

PR #697, opened 2026-09-14
fix(translation): report tool_use when a streamed Responses turn ends in a function call

linj-glitch:
## What

The Responses stream decoder reported every `response.completed` as a plain completion (`MessageStop { reason: None }`). A streamed `function_call` therefore reached Anthropic clients as `stop_reason: "end_turn"` and Chat clients as `finish_reason: "stop"`, while the buffered decoder correctly reported tool use for the same output.

The decoder now reports `tool_use` when the completed response's output holds a `function_call` or `custom_tool_call` item, or when it already decoded tool deltas earlier in the stream (for providers that end with a bare `response.completed`). Text and reasoning streams still report no reason, so their `end_turn` / `stop` mapping is unchanged.

## Why

Fixes SWITCH-1438 (NvBug 6771918). Stop-reason-driven tool loops, including the official `@anthropic-ai/sdk` TypeScript tool runner with `stream: true`, treat `end_turn` as final. They return the unfinished tool-use turn without running the tool and without any API error. QA's A/B probe shows the same runner completes the loop in buffered mode.

## Notes for reviewers

Start with the `response.completed` arm in `crates/switchyard-translation/src/codecs/responses/stream.rs`. The new `decoded_tool_call` flag on `StreamTranslationState` is set in the three places that decode tool items. The three new tests in `tests/stream_translation.rs` use QA's fixture and fail on the unpatched decoder: Anthropic `tool_use`, Chat `tool_calls`, and the bare-completed fallback. The reason carries the Anthropic spelling, as the `response.incomplete` arm already does with `max_tokens`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected streaming Responses tool-call handling so function and custom tool calls now report the appropriate tool-use completion reason instead of appearing as ordinary completions.
  * Preserved text-only stream behavior, which continues to complete without a tool-use reason.
  * Improved stop-reason translation for streamed tool calls across Anthropic and OpenAI Chat formats.


crates/switchyard-translation/tests/stream_translation.rs
```diff
@@ -1285,6 +1285,153 @@ fn responses_buffered_and_streamed_outputs_match() -> TestResult {
...
+    }
+    out.extend(engine.finish_stream(&mut state, target)?);
+    Ok(out)
+}
+
+// A streamed Responses function call must end the Anthropic message with `tool_use`, the same
+// stop reason the buffered decoder reports. Anthropic's tool runners dispatch on it; `end_turn`
+// makes them return the unfinished tool-use turn without running the tool.
+#[test]
+fn responses_function_call_stream_stops_with_tool_use_for_anthropic() -> TestResult {
```
ayushag-nv: @linj-glitch Looks like too many tests, is it psossible to fold the tests ??

linj-glitch: Done in 6fdcdb4b: the three tests are folded into one, `responses_function_call_stream_ends_with_tool_use_on_every_wire`. It covers the Anthropic `tool_use` stop reason, the Chat `tool_calls` finish reason, the bare-completed fallback, and the text-only control.

crates/switchyard-translation/tests/stream_translation.rs
```diff
@@ -1285,6 +1285,153 @@ fn responses_buffered_and_streamed_outputs_match() -> TestResult {
...
+#[test]
+fn responses_bare_completed_reports_tool_use_only_after_tool_deltas() -> TestResult {
+    let bare_completed = json!({"type": "response.completed", "response": {}});
+
+    let mut state =
+        StreamTranslationState::new(WireFormat::OpenAiResponses, WireFormat::OpenAiResponses);
+    let mut last = Vec::new();
+    for event in responses_function_call_stream()
+        .iter()
+        .take(5)
```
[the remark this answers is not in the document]
linj-glitch: Fixed in 6fdcdb4b: the fallback case now feeds only `response.created`, one `response.function_call_arguments.delta`, and the bare `response.completed`, with no output-item events, so it exercises the argument-delta path alone.

eric-liu-nvidia (approved):

PR #702, opened 2026-09-15
fix(client): enforce response deadlines and stop routing on errors

elyasmnvidian:
```toml
schema_version = 1
[llm_clients.http]
format = "openai_chat"
base_url = "http://127.0.0.1:60904/v1"
timeout_ms = 100
max_retries = 1
[targets]
stalled = { id = "stalled", llm_client = "http" }
weak = { id = "weak", llm_client = "http" }
strong = { id = "strong", llm_client = "http" }
[routes.classifier]
id = "classifier"
type = "llm_classifier"
classifier_target = "stalled"
weak_target = "weak"
strong_target = "strong"
base_threshold = 0.5
```

The local test provider accepts `stalled` requests but sends no response for 3 seconds. At `a2aec8a`, the classifier test returns HTTP 200 with `strong` / `PONG` after the judge times out. It should return HTTP 504 without calling another model. A stream that sends one event and then stalls also outlives the configured deadline.

The HTTP client now keeps one deadline across all attempts, retry delays, and the complete response, including stream reads. After retries, the HTTP driver stops routing on client errors; a timeout never tries another candidate. Streams used during routing are collected before the algorithm continues, with provider events preserved for replay.

**Compatibility:** HTTP client errors stop routing even when `timeout_ms` is unset or an advisor has `fail_open = true`. Invalid judge verdicts and non-timeout completion fallback keep their existing behavior. Unset deadlines remain unbounded; `0` is rejected. Runtime changes stay in `libsy-llm-client` and `switchyard-runner`; `libsy`, the protocol, and Python bindings are unchanged.

## Live results

This request to `/v1/chat/completions` now returns HTTP 504, with only `stalled` called:

```json
{"model":"classifier","stream":false,"max_tokens":32,"messages":[{"role":"user","content":"Ping"}]}
```

```json
{"error":{"message":"response did not finish within 100 ms, retries included","type":"upstream_error","code":"upstream_timeout"}}
```

All 12 live cases passed across Chat Completions, Messages, and Responses:

- Six stalled-judge or stalled internal-stream requests returned 504 without calling another model.
- Three final-answer streams sent an error event and ended without a success marker after sending HTTP 200 headers.
- Three healthy requests returned 200 with model `weak`, content `PONG`, and usage of 3 input tokens and 1 output token.
- Attempt counters matched all 15 captured upstream calls: 12 with HTTP 200 and 3 that timed out before a response. Expiry after streaming begins does not add another attempt.

## Testing

```bash
cargo test -p switchyard-server --test client_deadline --locked
```

`client_deadline_stops_routing_and_counts_attempts` checks HTTP 504 responses on Chat Completions, Messages, Responses, and `/v1/decision`. It also checks successful final-answer streams and timeout errors without success markers on the three streaming APIs. Retry delays, delayed bodies, routing stream errors, escalation-stream timeouts, and candidate timeouts are each exercised once. Captured upstream calls and attempt counters check that routing stops and stream timeouts do not add attempts.

The retry case combines a 300 ms first attempt and a 450 ms response body under one 600 ms deadline. Focused client and config tests cover expiry before sending, delayed stream consumption, and timeout validation. Existing client tests cover retry exhaustion and fallback after non-timeout errors. The stalled-classifier regression fails at `a2aec8a` with HTTP 200 instead of the expected 504.


crates/libsy/src/algorithms/util/llm_judge.rs
```diff
@@ -271,12 +291,22 @@ where
...
-            )
-            .await
+        let deadline = self.judge.timeout();
+        let call = driver.call_model(
+            self.judge.build_request(state, request),
+            judge_models.to_vec(),
+        );
+        // Dropping the call abandons the host's in-flight request; the host tolerates the
+        // orphaned reply and `drive` cancels it once the run finishes.
+        let Ok(response) = tokio::time::timeout(deadline, call).await else {
```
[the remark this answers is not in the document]
elyasmnvidian: `JudgeClassifier::verdict` now keeps `driver.call_model()` and `response.llm_response.into_agg()` inside the same `tokio::time::timeout`, so the deadline also covers stream reading. `configured_judge_deadline_covers_response_body_and_retries` sends a stalled judge stream through the configured HTTP client and checks that routing returns the fallback model before the caller's deadline. The test fails on the earlier implementation and passes with this change.

crates/libsy/src/algorithms/util/llm_judge.rs
```diff
@@ -271,26 +289,47 @@ where
...
+                        driver,
+                        safe_client_error(error),
+                        client_error_reason(error),
+                    );
+                })
+                .ok()
+        };
+        // On timeout, stop waiting for the judge. `serve` ignores a late
+        // `ResponseDropped`, and `drive` cancels pending calls when the run ends.
+        let Ok(aggregate) = tokio::time::timeout(deadline, call).await else {
```
ayushag-nv: Could we handle late replies in the Python bindings too? With timeout_ms=50 and a judge replying after 150 ms, call.respond() raises `LibsyError: driver response promise was dropped`; call.fail() does the same. The Python routing loop exits before consuming the fallback. The Rust HTTP client handles this, but the bindings need the same handling and a regression test.

crates/libsy/src/algorithms/llm_class.rs
```diff
@@ -24,6 +23,7 @@ use super::util::llm_judge::{
     SerdeDecoder, StructuredJudge,
 };
 use super::util::target_selector::TargetSelectorPolicy;
+use super::util::{DEFAULT_JUDGE_MAX_OUTPUT_TOKENS, DEFAULT_JUDGE_TIMEOUT_MS};
```
nachiketb-nvidia: libsy should remain strictly client unaware, fallbacks, timeouts, retries; libsy shouldn't have this faculty.

Lets isolate it by being able to send the "error" back to the algorithm rather than keeping track of timeout in the library.

nachiketb-nvidia (changes requested):

crates/libsy/src/algorithms/llm_class.rs
elyasmnvidian: Moved deadline enforcement into `ClientRouter` in `libsy-llm-client`. `libsy` now marks the call with `Category::Judge` and receives a response or a typed timeout error; the algorithm keeps its existing fallback decision.

`judge_timeout_ms` is a common route setting under `[routes.<name>]`, including for stage routers, composites, and subagents. It defaults to `10000` milliseconds, with a minimum of `1`. One deadline covers candidate attempts, retries and backoff, and reading the complete response. It applies only to judge calls, even when the same model also writes the answer.

`cargo test -p switchyard-server --test server judge_timeout` passes. The tests cover stalled streams, buffered responses, retry delays, and a model whose answer takes longer than its judge deadline. The Python docs also explain that custom hosts enforce their own deadline and return `TimeoutError` to the algorithm.

crates/libsy/src/algorithms/util/llm_judge.rs
elyasmnvidian: I reproduced both cases: with the original `timeout_ms=50` setup, calling `respond()` or `fail()` at 150 ms raised `LibsyError: driver response promise was dropped` and stopped the host loop. Both methods now ignore `DriverError::ResponseDropped` when routing has stopped waiting. Invalid argument types and completing the same call twice still raise errors.

Deadline enforcement has also moved out of `libsy`. For a custom Python host, `call.category == "judge"` identifies the call to time out. The host enforces the deadline and reports it with `call.fail(TimeoutError(...))`, or raises `TimeoutError` from the response stream. The algorithm then uses its existing fallback.

`tests/test_libsy_judge_deadline.py` covers both late replies after cancellation and duplicate completions. The late-reply test fails with the old extension and passes with the rebuilt extension. All five cases pass, including timeouts from buffered responses, awaited stream reads, and a synchronous `__anext__()` failure.

crates/libsy/src/algorithms/util/llm_judge.rs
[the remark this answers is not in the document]
elyasmnvidian: The deadline still covers `.into_agg()`, but its owner has changed since my earlier reply: `ClientRouter` in `libsy-llm-client` now runs the timer. One deadline covers candidate attempts, retries and backoff, and reading the complete response, including a stream that stalls after its first event. The client returns a typed timeout error to `libsy`, which keeps the algorithm's fallback decision.

`configured_judge_deadline_covers_response_body_and_retries` passes, including stalled-stream requests through Chat Completions, Anthropic Messages, and Responses. The three Python classifier constructor docs now explain that custom Python hosts enforce their deadlines and report `TimeoutError` so routing can continue without a verdict.

nachiketb-nvidia (changes requested):
Can we do this:

- for any model call inside run_stream (which should be in switchyard-runner and libsy-llm-client), we do a specific numebr of retries for a specific target until a number timeout (specified in the config). Then if there is still no response, we return an error, do not "continue" routing
- the changes should technically be isolated to libsy-llm-client and/or switchyard-runner

elyasmnvidian:
Done in a2aec8a. The change is now only in `libsy-llm-client` and `switchyard-runner`; `libsy` is untouched.

- `[llm_clients.<name>] timeout_ms` sits next to `max_retries`. `TranslatingLlmClient::send_encoded` runs the attempt loop for one target under that deadline: connect, each retry and its delay, and the response (whole body when buffered, first event when streamed).
- When the deadline passes the client returns `LlmClientError::Timeout` to the caller and does nothing else. The algorithm decides what an error means: a judge error already fell open in `libsy` before this PR, and an answer error already mapped to 504 in the server.
- No default. Unset keeps today's unbounded wait; `0` is rejected at load.

The judge-only route key, `Category::Judge`, `call_model_for_category`, and the Python `ModelCall.category` are gone. Live run against the gateway and the test list are in the PR body.

crates/libsy/src/algorithms/llm_class.rs
elyasmnvidian: Implemented in 3130d69. Calls inside `run_stream` retry the same target up to `max_retries` within `timeout_ms`. The deadline covers retry delays and the complete response, including streams. If the call fails, the HTTP driver returns the error and stops routing without trying another model.

Runtime changes are limited to `libsy-llm-client` and `switchyard-runner`; `libsy`, the protocol, and Python bindings are unchanged. The PR description includes the regression tests and live results.

crates/libsy/src/algorithms/util/llm_judge.rs
elyasmnvidian: The deadline stays in the Rust HTTP client. `libsy`, the protocol, and Python bindings are unchanged; the earlier `ModelCall.category` and `ResponseDropped` tolerance are not part of this PR. A late reply after cancellation is distinct from a duplicate completion.

Correction: `call.fail(TimeoutError(...))` maps to `LlmClientError::Ffi` and `reason=client_error`, not `reason=timeout`. A Python host driving `libsy` directly still enforces its own deadline.

crates/libsy/src/algorithms/util/llm_judge.rs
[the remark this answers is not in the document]
elyasmnvidian: `timeout_ms` now covers the complete response, including reads after the first stream event. The HTTP driver collects streams used inside `run_stream` before returning them to the algorithm, so a timeout, stream error, or decode error stops routing. Successful streams retain their provider events for replay. The regression test covers a weak-model stream that stalls after its first event: the server returns 504 without calling the judge or strong target.

elyasmnvidian:
> * the changes should technically be isolated to libsy-llm-client and/or switchyard-runner

done, PTAL @nachiketb-nvidia

crates/switchyard-server/tests/client_deadline.rs
```diff
@@ -0,0 +1,556 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Run deadline cases in one test because they share process-wide attempt counters.
+
+use std::collections::BTreeMap;
```
nachiketb-nvidia: do we need so many tests? its like 75% of the MR

ayushag-nv (approved):
lgtm

nachiketb-nvidia (approved):

PR #703, opened 2026-09-15
fix(translation): return HTTP 502 when Responses generation fails

elyasmnvidian:
```json
{
  "id": "resp_failed",
  "object": "response",
  "model": "failed",
  "status": "failed",
  "error": {
    "code": "server_error",
    "message": "deterministic upstream failure"
  },
  "output": [],
  "usage": null
}
```

## Request flow

An application sends a request to Switchyard. Switchyard forwards it to a model provider using the OpenAI Responses API, then returns the answer in the format the application requested: Chat Completions, Anthropic Messages, or Responses.

This PR covers a **non-streaming response**: the provider sends one JSON document with the generation status, output, and token usage. Switchyard reads the whole document before converting it; the code calls this a **buffered response**. A streaming response sends events as the model generates its answer and uses separate handling.

## Problem

The JSON above is a response **from the model provider to Switchyard**. The provider sends HTTP 200, but the document says `status: "failed"` and contains no generated answer (`output: []`). HTTP 200 alone does not mean that generation succeeded.

Switchyard converts this failed response into an empty successful answer for applications using Chat Completions or Anthropic Messages. Applications using Responses receive `status: "failed"` in the JSON but still get HTTP 200. Switchyard also counts the failed model call as a success and does not try another model.

The result is an application receiving an empty answer without an HTTP error:

```text
/v1/chat/completions: HTTP 200, content="", finish_reason="stop"
/v1/messages:         HTTP 200, text="", stop_reason="end_turn"
/v1/responses:        HTTP 200, status="failed"
```

## Fix

`OpenAiResponsesCodec::decode_response` now checks the provider's generation status before converting the answer. It treats `status: "failed"` as a provider failure, which the HTTP client reports as HTTP 502 (Bad Gateway).

Switchyard counts the failed model call as an error and tries another model when the route supports fallback. If this failure reaches the application, Switchyard returns HTTP 502 on Chat Completions, Anthropic Messages, and Responses.

Applications using an OpenAI API receive the provider's message and nonempty string error code. Applications using Anthropic Messages receive `type: "api_error"` and the provider's message. The shared HTTP error mapper also preserves codes from ordinary upstream HTTP errors and uses `upstream_error` when the code is missing, empty, or not a string.

The HTTP client uses `redact_forwarded_headers` to remove echoed caller credentials before returning or logging the error. A `failed` status still counts as a provider failure when the error details are missing or malformed.

## Verification

The local provider returns the JSON above. Switchyard uses this configuration:

```toml
schema_version = 1

[llm_clients.local]
format = "openai_responses"
base_url = "http://127.0.0.1:28081/v1"
max_retries = 0

[targets.failed]
id = "failed"
llm_client = "local"

[routes.failed]
id = "demo/failed"
type = "passthrough"
target = "failed"
```

```text
/v1/chat/completions: HTTP 502, code="server_error"
/v1/messages:         HTTP 502, type="api_error"
/v1/responses:        HTTP 502, code="server_error"
```

Every response includes `message: "deterministic upstream failure"`. The same three requests with a completed provider response return HTTP 200 and `PONG`, with 12 input tokens and 6 output tokens each.

Requests to a real Responses provider using `openai/openai/gpt-5.6-sol` also return HTTP 200 and `PONG` through all three endpoints, with 12 input tokens and 6 output tokens each. `/v1/stats` reports nine requests and three errors across the failed, completed, and real-provider cases.

**Compatibility:** When this failure reaches an application using Responses, the application now receives HTTP 502 instead of an HTTP-200 body with `status: "failed"`. Successful and incomplete Responses keep their existing behavior.

## Tests

```bash
cargo test -p switchyard-server --test server responses
cargo test -p switchyard-translation --test response_translation failed_responses_return_upstream_failure_with_provider_message
```

The HTTP tests check failed Responses, error counts, and fallback on all three public APIs. Chat Completions tests also cover missing or malformed error details, invalid or empty codes, and credential redaction. Translation tests check `UpstreamFailure` with the provider's message and retain coverage for incomplete responses.


crates/switchyard-server/tests/server.rs
```diff
@@ -803,6 +841,132 @@ async fn stats_accumulates_buffered_success_error_and_shared_routes() -> TestRes
...
+    assert_eq!(
+        response.json()?["error"]["message"],
+        "provider rejected [REDACTED]"
+    );
+    assert_eq!(response.json()?["error"]["code"], "[REDACTED]");
+    Ok(())
+}
+
+#[tokio::test]
+async fn failed_responses_try_the_next_candidate_across_endpoints() -> TestResult {
```
nachiketb-nvidia: Can we compress to one overarching test? if not its okay

ayushag-nv (approved):

PR #705, opened 2026-09-15
fix(python): avoid a runtime typing_extensions dependency

nachiketb-nvidia:
## What

- Load `typing_extensions.Self` only while type checking.
- Make the existing slim-install and release-wheel smoke checks import the native `Server` entry point.

## Why

A clean default install has no `typing_extensions` runtime dependency, but `switchyard_rust.server` imported it unconditionally. This made the supported server entry point fail before the native module could load.

## Validation

- Clean isolated install and `Server` import on Python 3.10 with `typing_extensions` absent.
- Clean isolated install and `Server` import on Python 3.14 with `typing_extensions` absent.
- `uv run ruff check switchyard_rust/server.py`
- `uv run mypy switchyard_rust/server.py`
- Repository pre-commit hooks.

Fixes [SWITCH-1425](https://linear.app/nvidia/issue/SWITCH-1425).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime loading for the server component by avoiding an unnecessary typing-related dependency during execution.
* **Tests**
  * Expanded installation and wheel smoke tests to verify that the server component can be imported successfully across supported Python versions.


ayushag-nv (approved):
Lgtm

PR #706, opened 2026-09-15
fix(packaging): include prefill-router in the Python sdist

nachiketb-nvidia:
## What

- Include the optional `prefill-router` crate and the pinned Rust toolchain in Python source distributions.
- Smoke install the generated source distribution before it can be uploaded.

## Why

The generated archive contained `switchyard-runner`, but omitted its local `prefill-router` dependency. Cargo therefore failed during package metadata generation. The archive also omitted the repository's pinned toolchain, which could leave clean builds below the declared Rust version.

## How

Maturin now includes only the prefill crate files needed to resolve and build the workspace dependency. The existing source-distribution job installs its generated archive in an isolated Python 3.10 environment and imports the native extension.

## Validation

- Built `nemo_switchyard-0.2.0.tar.gz` with Maturin.
- Confirmed the archive contains the prefill manifest, build script, Rust sources, embedded Python source, and `rust-toolchain.toml`.
- Installed that archive outside the checkout in an isolated Python 3.10 environment.
- Imported `switchyard_rust._switchyard_rust` from the installed archive.
- `actionlint` passed for the release workflow.

Fixes [SWITCH-1427](https://linear.app/nvidia/issue/SWITCH-1427).

## Summary by CodeRabbit

- **Chores**
  - Source distributions now include all required build configuration and source files, improving reliability when building from source.

- **Tests**
  - Added an automated smoke test that installs the source distribution in an isolated Python environment and verifies the compiled extension can be imported successfully.


.github/workflows/publish.yml
ayushag-nv, line 1: Can you verify if it works once

nachiketb-nvidia: Verified locally: the sdist builds, installs in an isolated Python 3.10 environment from `/tmp`, and imports `switchyard_rust._switchyard_rust` successfully.

ayushag-nv (approved):

PR #708, opened 2026-09-15
fix(libsy): escalate on context overflow while buffering

nachiketb-nvidia:
## What

- Fall back to the capable model when the efficient response stream reports a context-window error while it is being buffered.
- Cover the Python first-event failure boundary with one regression test.
- Extend the existing Rust overflow test to cover both direct and streamed errors.

## Why

Escalation already handled a context-window error returned directly by the efficient model call. A streamed response resolves that call before its events are read, so the same typed error could arrive later during aggregation and escape as `LibsyError`. Python hosts then received no final routing outcome.

## How

The aggregation error match now handles `ContextWindowExceeded` exactly like the existing direct-call path: it skips the judge, records `context_window` fallback evidence, and returns the capable-first routing decision without a buffered response. Other stream errors keep their existing behavior.

## Validation

- Reproduced the failure before the change with the public Python `run_stream` API.
- `cargo test --locked -p switchyard-libsy algorithms::escalation`: 7 passed.
- Focused Python first-event overflow regression: 1 passed.
- `cargo clippy --locked -p switchyard-libsy --all-targets -- -D warnings`.
- `uv run ruff check tests/test_libsy_minimal_bindings.py`.
- Repository pre-commit hooks passed.
- No live provider calls were made.

Fixes [SWITCH-1447](https://linear.app/nvidia/issue/SWITCH-1447).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model escalation when the efficient model reaches its context limit.
  - Requests now reliably fall back to the capable model whether the limit is detected immediately or while processing a streamed response.
  - Fallbacks correctly record the context-limit reason and preserve model selection details, providing more consistent results for both streaming and non-streaming requests.

- **Tests**
  - Added coverage for context-limit fallback behavior across streaming and non-streaming scenarios.


nachiketb-nvidia:
Validated the fix locally:

- Focused Rust escalation suite: **7 passed**
- Python regression: **1 passed**
- Exact SWITCH-1447 public Python reproducer: **5/5 repetitions passed**

Each reproduction returned:

```text
DONE ['strong', 'weak']
CALLS [['weak']]
```

The first-event context overflow now selects the strong model without calling the judge.

ayushag-nv (approved):
lgtm

PR #711, opened 2026-09-15
fix(stage-router): warn when capable-first cannot offload via scorer

### LINKED ISSUE #264: [Bug] stage_router: capable_first can't offload at the default confidence_threshold=0.5

ting-hong-shieh:
## What

Warn when a `capable_first` stage router uses a confidence threshold that prevents the scorer from selecting the efficient tier. For example, `0.5` exceeds the efficient confidence ceiling of approximately `0.462117`.

The warning runs when `StageClassifier` is constructed, covering the runner and Python bindings without logging on each request. It computes the ceiling through the existing scorer and uses the picker's closed probability band, including the exact boundary. Routing behavior is unchanged: hard de-escalation and an optional LLM classifier can still select efficient.

The documentation explains this limit and scopes the `0.5` recommendation to `efficient_first`.

## Why

Closes #264.

The current TOML schema requires an explicit threshold, but `capable_first` with `0.5` still silently disables scorer-driven offloading. This implements the issue's warning option without changing scoring weights or introducing new public APIs.

## Notes for reviewers

Start with `StageClassifier::new` in `crates/libsy/src/algorithms/util/stage.rs`. One regression test checks warning output and picker behavior at `0.45`, the exact ceiling, and `0.5`, plus the hard de-escalation escape path.

Validation:

- `cargo test --workspace`: 757 passed, 1 ignored before the test-only mutex cleanup; `cargo test -p switchyard-libsy`: 289 passed after it.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check`: passed.
- `uv run ruff check .`: passed; `uv run pytest tests/`: 116 passed, 2 skipped.
- Strict MkDocs build: passed.
- A temporary adversarial test checked 56,250 signal combinations against the efficient ceiling and 65 adjacent floating-point thresholds for agreement between the warning and picker. It passed and was removed from the final diff.

Validation used the project's Python 3.14 interpreter for PyO3 and allowed local ports for mock HTTP servers. No live model benchmark was run; `0.45` is documented as below the ceiling, not as a calibrated recommendation.

## Summary by CodeRabbit

- **New Features**
  - Added warnings in `capable_first` mode when production scoring cannot reach the configured confidence threshold for efficient selection.
  - Preserved hard de-escalation behavior and classifier exceptions for ambiguous cases.

- **Documentation**
  - Clarified confidence threshold comparisons, routing behavior, scorer limits, warning conditions, and threshold calibration guidance.
  - Specified that the recommended `0.5` threshold applies to `efficient_first`.


ayushag-nv:
++ @sabhatinas for review

afourniernv:
Small wording nit: direct `switchyard.libsy.stage_router()` reaches this code, but that path doesn’t install a tracing subscriber, so Python-only users won’t see the warning. The Python server path does. Could we narrow the “covers Python bindings” line? I don’t think this PR needs any additional Python logging work.

PR #712, opened 2026-09-15
fix(translation): number Chat tool calls within tool_calls instead of by content index

linj-glitch:
## What

The OpenAI Chat stream encoder copied the neutral tool-call index straight into `tool_calls[].index`. That index comes from the source stream: Anthropic's content-block index or Responses' output index, both of which count every block or item. When text streams ahead of the first tool call, the sole tool call arrives as index 1.

The encoder now assigns each tool call a position within the Chat `tool_calls` array in order of first appearance, tracked per source index on the stream state, and emits that instead. Chat-to-Chat streams are unaffected since their indexes already count tool calls.

## Why

Fixes SWITCH-1446 (NvBug SWYD-157). The official OpenAI Python SDK uses `tool_calls[].index` as a subscript while accumulating a stream. With one tool call tagged index 1 it indexes a one-element list and raises `IndexError: list index out of range`. QA reproduced this with a fixture and live against NVIDIA Inference Hub with `azure/anthropic/claude-opus-5`, which emitted text before its first tool call. The HTTP response is a clean 200, so the client crashes with no server-side signal.

## Notes for reviewers

Start with the `ToolCallDelta` arm in `crates/switchyard-translation/src/codecs/openai_chat/stream.rs`. The new `chat_tool_index` on `StreamToolState` mirrors how the Anthropic encoder already assigns `content_index`. One test covers both source formats: an Anthropic stream with text then two tool calls must emit indexes 0 and 1, and a Responses stream with a message item then a function call must emit 0. The test fails on main.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed OpenAI Chat tool-call indexing when text or other content appears before tool calls.
  - Tool calls now receive contiguous positions based on their order, preventing indexing errors for SDK clients.
  - Preserved consistent indexing across streamed updates.

- **Tests**
  - Added coverage for tool-call indexing across supported streaming formats.


crates/switchyard-translation/src/codecs/openai_chat/stream.rs
```diff
@@ -238,13 +238,29 @@ fn encode_openai_chat_stream(
...
-            id,
-            name,
-            arguments_delta,
-        )],
+        } => {
+            // The source index counts every content block (Anthropic) or output item
+            // (Responses), so text ahead of the first tool call shifts it. Chat clients use
+            // the index as a subscript into `tool_calls`, so number calls in that array
+            // instead, in order of first appearance.
+            let tool = state.tool_states.entry(index).or_default();
```
sabhatinas, line 246: we should use [tool_signal.rs](http://tool_signal.rs/) for parsing tool calls in all algos

linj-glitch:
Local end-to-end smoke run, cc @nachiketb-nvidia.

Setup is QA's black-box repro from SWITCH-1446: a local Anthropic Messages SSE fixture, a real `switchyard-server` on a loopback port with a passthrough route, and the official OpenAI Python SDK (2.34.0) streaming `/v1/chat/completions` through it. Two binaries from the same tree: main's translation sources vs this branch.

| Case | main | this PR |
|---|---|---|
| text then tool, raw `tool_calls[].index` | `[1, 1]` | `[0, 0]` |
| text then tool, OpenAI SDK | `IndexError: list index out of range` | PASS, 1 tool call, `finish_reason=tool_calls` |
| tool first (control), raw index | `[0, 0]` | `[0, 0]` |
| tool first (control), OpenAI SDK | PASS | PASS |

cc @sabhatinas on the `tool_signals.rs` suggestion: that module walks the request's conversation history to extract tool-result signals for the Stage Router and advisor gate. This PR is on the response side, in how the Chat stream encoder numbers the tool-call deltas it emits, and does no tool-call parsing, so I don't think it applies here. Happy to be corrected if you had a different spot in mind.

ayushag-nv (approved):

crates/switchyard-translation/src/codecs/openai_chat/stream.rs
nachiketb-nvidia: its translation not algos

linj-glitch: Thanks for looking. I think `tool_signals.rs` is on a different axis from this change. It lives in libsy and walks the request's conversation history to extract tool-result signals (error patterns, streaks, tests-passed) for the Stage Router and advisor gate. This PR is in the translation crate's Chat stream encoder, on the response side: it only renumbers the `index` field of the tool-call deltas we emit to the client so they count positions in `tool_calls` rather than source content blocks. No tool calls are parsed here, and the routing algorithms never see this path. If you meant that algorithms should share one tool-call parser, I agree that is worth doing, but it would be a libsy change separate from this fix.

sabhatinas: I see. I think tool parsing utility should be in a common crate. this seems to be duplicating code.

PR #714, opened 2026-09-15
fix(llm-client): Retry stream disconnects before doing fallback

grahamking:
Move the first stream-event check inside the retry check, before fallback. Transport failures and timeouts now retry before candidate fallback runs. Preserve no-replay behavior after streaming starts.

Add one regression test covering retry recovery and exhaustion.

Sonnet 5's review:
> Verdict: solid. The core change moves stream-decode of the first event inside `send_once`'s retry boundary via a new `prepare_response_stream` helper, and classifies pre-first-event `Transport/Timeout` errors as retryable rather than letting them fall straight through to candidate fallback. Logic checks out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streaming responses when the initial event encounters a transport or timeout error.
  * Streaming requests now retry appropriately when failures occur before the first event.
  * Context overflow responses are reported with a specific typed error.
  * Preserved valid initial stream events while continuing to process the remaining response.


ayushag-nv (approved):
nice change

PR #720, opened 2026-09-16
fix(runner): Remove API keys from URLs

grahamking:
Gemini, surprisingly, allows putting the API key in the URL. The `/v1/decision` API returns that URL, because that's where you need to send send your request to run inference on the selected model. Ergo that can leak the key.

So now we sanitize that URL. QA caught this.

Users of the decision API are expected to bring their own authentication. And as an aside, no-one should put secrets in URL parameters, ever. HTTP has a standard header for that.

Similar to https://github.com/NVIDIA-NeMo/Switchyard/issues/423 , which is the same Gemini key leak but for error messages.

Reviewed-by: Claude:Sonnet 5 medium

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive `key` and `api_key` query parameters are no longer exposed in decision response URLs.
  * Other query parameters, including `api_version`, remain unchanged.


ayushag-nv (approved):
Looks good for removing API keys 
Codex flagged this where user can expose password
```
https://user:password@example.test/v1?key=secret
```

PR #721, opened 2026-09-16
fix(llm-client): keep Responses follow-ups on the provider holding their history

elyasmnvidian:
Clients using the Responses API can use `previous_response_id` or `conversation` to leave earlier messages, tool calls, and tool results with the provider, then send new input plus an ID. This saves the caller from maintaining and retransmitting the complete conversation. The caller chooses this behavior; Switchyard receives the ID rather than replacing the caller's history with one. [Responses conversation state](https://developers.openai.com/api/docs/guides/conversation-state#openai-apis-for-conversation-state)

That ID refers to history stored at the original provider. It does not contain the history or offer a supported way to identify the provider from its value; applications should treat provider IDs as opaque. [API compatibility guidance](https://developers.openai.com/api/reference/overview#backwards-compatibility)

For example, suppose the strong target is [GPT-6 Astra](https://developers.openai.com/api/docs/models/gpt-6-astra) through OpenAI Responses, and the weak target is Claude Haiku through Anthropic Messages.

1. A complex question goes to Astra. OpenAI saves its answer and returns `resp_example`.
2. The caller asks, “Summarize that in one sentence,” sending only this new input and `previous_response_id: resp_example`.
3. Routing that short follow-up to Haiku would omit the analysis it needs to summarize. Anthropic cannot use the OpenAI response ID to retrieve that history; its Messages API requires earlier turns in the request. [Anthropic conversation history](https://platform.claude.com/docs/en/build-with-claude/working-with-messages#multiple-conversational-turns)

This change records each observed saved ID and the exact configured model that served it. A follow-up with a known ID returns to that model without another routing decision. In this example, the follow-up stays on Astra. If that request fails, Switchyard returns the error instead of trying a different provider. Fresh requests and requests containing full history without an ID use normal selection. Tracking applies only when answer models use separate configured HTTP clients; models sharing one client create no entries, even with a separate classifier client.

The local map preserves existing requests and provider IDs without storing transcripts or fetching history before each follow-up. It is a **best-effort compatibility fix**. Requiring full history would change callers and require complete message/tool replay; wrapping IDs would require rewriting and decoding responses, streams, and ID-based operations; a persistent shared registry would support restarts and replicas but add a storage dependency. [Caller-managed conversation option](https://developers.openai.com/api/docs/guides/migrate-to-responses#3-update-multi-turn-conversations)

This adds memory for **up to 65,536 distinct IDs per route, per process, over its lifetime**. This counts saved IDs, not tokens or concurrent requests. Records are never evicted or overwritten with another model. If a non-streaming reply adds an ID beyond the cap, Switchyard returns HTTP `503 response_state_limit_exceeded`; a conflicting owner returns `409 response_state_conflict`. A stream emits a framed error and stops after headers have been sent. Switchyard logs these failures; for `/v1/responses`, they increment `/v1/stats` `total_errors`. Other routes continue independently.

The provider has already returned a response or stream event when these checks run, so it may already have done work. Retrying does not free entries. Existing IDs remain usable only when the reply adds no new IDs; a stored follow-up normally needs a new response ID and can also fail at capacity. The chosen cap has not been justified against production workloads. Restarting loses all mappings, and replicas do not share them. Unknown IDs still use normal routing and can reach the wrong provider.

Maps allocate as they fill; requests sharing a route share its map. Each entry copies an ID and model name, with no transcript storage. An isolated probe at capacity, using 64-byte IDs and 32-byte model names, retained **12,713,992 bytes (about 12.1 MiB)** of requested heap. Longer strings cost more. This excludes allocator overhead, temporary allocations, fixed struct fields, and the rest of the server. Escalation also temporarily holds the original stream events and complete answer while the classifier checks the weaker model's reply. Keeping those events preserves provider IDs if that reply is returned.

The release server passed **two batches of 64 parallel requests with 1,000,000 input tokens each**, plus one warm-up request. All 129 ID-based follow-ups returned to the correct provider without a classifier call. Local HTTP stubs checked the complete payload and waited until every request in each batch had arrived; half of each 64-request batch used streaming. Sampled server RSS peaked at **3.03 and 3.23 GiB**, then measured about **605 and 740 MiB** after the batches. Replies were short and fixed. These checks establish this local workload, not unlimited concurrency or a hosted model's context capacity. Full methods, limits, and sources are below.

<details>
<summary>Configuration, requests, and before-and-after results</summary>

The comparison used `416ad0cb4` before the fix and `4aeaceb96` after it. The local classifier returns a fixed choice: provider B when the input contains `ROUTE_B`, and provider A otherwise. Each answer provider accepts only its own saved IDs. The follow-up input would therefore select B unless Switchyard uses its recorded owner.

```toml
schema_version = 1
[llm_clients.judge]
format = "openai_responses"
base_url = "http://127.0.0.1:58244/silo/judge"
max_retries = 0
[llm_clients.a]
format = "openai_responses"
base_url = "http://127.0.0.1:58244/silo/A"
max_retries = 0
[llm_clients.b]
format = "openai_responses"
base_url = "http://127.0.0.1:58244/silo/B"
max_retries = 0
[targets.judge]
id = "model/judge"
llm_client = "judge"
[targets.a]
id = "model/provider-a"
llm_client = "a"
[targets.b]
id = "model/provider-b"
llm_client = "b"
[routes.dynamic]
id = "route/dynamic"
type = "llm_classifier"
classifier_target = "judge"
weak_target = "a"
strong_target = "b"
base_threshold = 0.5
classify_trigger = "every_request"
```

The run sent these five JSON bodies to `POST /v1/responses`, in order. The first response returned `resp_A_2`; `conv_A_example` identifies a conversation on provider A.

```jsonl
{"model":"route/dynamic","input":"ROUTE_A: remember mango"}
{"model":"route/dynamic","input":"ROUTE_B: recall mango","previous_response_id":"resp_A_2"}
{"model":"route/dynamic","input":"ROUTE_B: a new task"}
{"model":"route/dynamic","input":"ROUTE_A: remember mango","conversation":"conv_A_example","store":false}
{"model":"route/dynamic","input":"ROUTE_B: recall mango","conversation":{"id":"conv_A_example"},"store":false}
```

Both follow-ups returned this error before the fix:

```json
{"error":{"message":"state is not present in provider B","type":"upstream_error","code":"state_not_found"}}
```

| Request | Before | After |
|---|---|---|
| First request | 200, `model/provider-a`, one classifier call | Same |
| Follow-up using `resp_A_2` | 400, `state_not_found`, one classifier call | 200, `model/provider-a`, no classifier call |
| New request | 200, `model/provider-b`, one classifier call | Same |
| First request using `conv_A_example` | 200, `model/provider-a`, one classifier call | Same |
| Follow-up using `conv_A_example` | 400, `state_not_found`, one classifier call | 200, `model/provider-a`, no classifier call |

Every successful reply reported `{"input_tokens":1,"output_tokens":1,"total_tokens":2}`. These counts come from the local services' fixed responses.

One server regression covers buffered and streamed follow-ups, `/v1/decision`, shared answer clients, ownership conflicts, and provider failures with shared setup and explicit TOML routes. Client tests cover storage settings, conversation-ID forms, response formats, and capacity. In the five-request run above, the classifier made five calls before the fix and three after it, with zero classifier errors in both runs.

```bash
cargo test -p switchyard-server --test server responses_continuations_preserve_state_ownership
cargo test -p switchyard-llm-client --lib responses_state_tracking_uses_upstream_storage_and_format
cargo test -p switchyard-llm-client --lib state_owners
```

Only Responses IDs are recorded. Chat Completions and Anthropic Messages response IDs are excluded. The provider's reported `store` value takes precedence over the request value; `false` skips response-ID tracking while conversation IDs remain tracked.

</details>

<details>
<summary>Reproduce the ID map's heap-allocation measurement</summary>

Measured on September 16, 2026, with Rust 1.96.1 on aarch64 macOS, compiled with `--edition 2024 -O`. `StateOwners` below is copied from `4aeaceb96`; the stand-in `ModelId(String)` has the same string storage and cloning behavior. The counter measures requested allocation sizes retained after insertion, subtracting the empty-map baseline. Repeated runs returned the same counts. This counts bytes rather than timing operations.

At 65,536 records with 64-byte IDs and 32-byte model names, the map retained 12,713,992 requested heap bytes. This excludes allocator metadata and size rounding, temporary allocations during growth, fixed struct fields, and whole-server RSS. The probe also checks that another ID returns the capacity error without removing an existing record.

Save this as `memory_probe_bounded.rs`:

```rust
use std::alloc::{GlobalAlloc, Layout, System};
use std::collections::HashMap;
use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};

static LIVE: AtomicUsize = AtomicUsize::new(0);
struct Counter;
unsafe impl GlobalAlloc for Counter {
    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
        let p = unsafe { System.alloc(layout) };
        if !p.is_null() { LIVE.fetch_add(layout.size(), Relaxed); }
        p
    }
    unsafe fn dealloc(&self, p: *mut u8, layout: Layout) {
        LIVE.fetch_sub(layout.size(), Relaxed);
        unsafe { System.dealloc(p, layout) }
    }
    unsafe fn realloc(&self, p: *mut u8, layout: Layout, size: usize) -> *mut u8 {
        let next = unsafe { System.realloc(p, layout, size) };
        if !next.is_null() {
            LIVE.fetch_sub(layout.size(), Relaxed);
            LIVE.fetch_add(size, Relaxed);
        }
        next
    }
}
#[global_allocator]
static ALLOCATOR: Counter = Counter;
#[derive(Clone, PartialEq)]
struct ModelId(String);
#[derive(Debug)]
enum LlmClientError {
    ResponseStateLimitExceeded { limit: usize },
    ResponseStateConflict,
}
#[derive(Default)]
struct StateOwners {
    by_id: HashMap<String, ModelId>,
}

const MAX_STATE_OWNERS: usize = 65_536;

impl StateOwners {
    fn owner(&self, id: &str) -> Option<&ModelId> {
        self.by_id.get(id)
    }

    fn remember(
        &mut self,
        response_id: Option<&str>,
        conversation_id: Option<&str>,
        model: &ModelId,
    ) -> std::result::Result<(), LlmClientError> {
        let ids = [
            response_id,
            conversation_id.filter(|id| Some(*id) != response_id),
        ];
        let mut new_ids = 0;
        for id in ids.into_iter().flatten() {
            match self.by_id.get(id) {
                Some(owner) if owner != model => return Err(LlmClientError::ResponseStateConflict),
                None => new_ids += 1,
                _ => {}
            }
        }
        if self.by_id.len() + new_ids > MAX_STATE_OWNERS {
            return Err(LlmClientError::ResponseStateLimitExceeded {
                limit: MAX_STATE_OWNERS,
            });
        }
        for id in ids.into_iter().flatten() {
            if !self.by_id.contains_key(id) {
                self.by_id.insert(id.to_owned(), model.clone());
            }
        }
        Ok(())
    }
}

fn main() {
    println!("id_bytes,model_bytes,records,retained_heap_bytes");
    for id_bytes in [16usize, 64, 256] {
        for count in [0usize, 1024, 65536] {
            let model = ModelId("m".repeat(32));
            let before = LIVE.load(Relaxed);
            let mut owners = StateOwners::default();
            for n in 0..count {
                let id = format!("{:0width$x}", n, width = id_bytes);
                owners.remember(Some(&id), None, &model).expect("within capacity");
            }
            let delta = LIVE.load(Relaxed) - before;
            assert_eq!(owners.by_id.len(), count);
            if count > 0 {
                let id = format!("{:0width$x}", 0, width = id_bytes);
                assert_eq!(owners.owner(&id).unwrap().0.len(), 32);
            }
            println!("{id_bytes},32,{count},{delta}");
            if count == 65_536 {
                assert!(matches!(
                    owners.remember(Some("over_limit"), None, &model),
                    Err(LlmClientError::ResponseStateLimitExceeded { limit: 65_536 })
                ));
            }
            drop(owners);
            assert_eq!(LIVE.load(Relaxed), before);
        }
    }
}
```

```bash
rustup run 1.96.1 rustc --version
rustup run 1.96.1 rustc --edition 2024 -O memory_probe_bounded.rs -o memory_probe_bounded
./memory_probe_bounded
```

The full-map row for 64-byte IDs is:

```csv
id_bytes,model_bytes,records,retained_heap_bytes
64,32,65536,12713992
```

</details>

<details>
<summary>Reproduce the parallel million-token, capacity, and request-limit checks</summary>

Run on September 16, 2026: `4aeaceb96`, Rust 1.96.1 release build, Python 3.13.9, macOS 26.6.2 on arm64 with 36 GiB RAM. Python dependencies were `aiohttp==3.14.3`, `tiktoken==0.14.0`, and `psutil==7.2.2`.

The script starts the real Switchyard server and local provider stubs. Each large input contains exactly 1,000,000 tokens under `o200k_base` (5,999,977 text bytes, about 6 MB per HTTP request). This is the stated tokenizer's count, not a claim about every model's native tokenizer. The provider checks a SHA-256 digest of every complete input. An arrival barrier holds replies until all requests in a batch are active at the provider, verifying overlap. Half of each 64-request batch streams its answer. All answers are short fixed strings, and follow-ups use `store: false`.

The script runs one warm-up request and two 64-request batches in the same server process. It samples that server's RSS every 20 ms; this excludes the Python caller/provider process. The values below are observations from these batches, not a latency benchmark or a long-term memory-retention test.

| Batch | Completed initial requests | Successful ID-only follow-ups | Peak simultaneous provider calls | RSS before | Sampled peak RSS | RSS after |
|---|---:|---:|---:|---:|---:|---:|
| Warm-up | 1 | 1 | 1 | 13,058,048 B | 69,976,064 B | 69,976,064 B |
| First parallel batch | 64 | 64 | 64 | 69,976,064 B | 3,258,417,152 B | 634,339,328 B |
| Second parallel batch | 64 | 64 | 64 | 634,339,328 B | 3,469,393,920 B | 775,733,248 B |

The same process then exercised the limits:

- With room for 16 IDs, 32 concurrent replies each supplied two new IDs. Eight requests succeeded and 24 returned `503 response_state_limit_exceeded`; existing records remained intact.
- Another buffered overflow returned 503 and increased `/v1/stats` `total_errors` from 24 to 25. A streamed overflow returned one error event after HTTP 200 headers.
- Reporting an existing ID from another model returned `409 response_state_conflict` without changing its owner. The oldest saved ID still routed correctly with `store: false`, and the server stayed running.
- A provider's `400 context_length_exceeded` reached the caller. The known continuation did not retry another provider.
- A request containing 33 MiB of text (34,603,067 request bytes) returned HTTP 413 under the existing 32 MiB body limit, before any provider call.

These limits are independent: a million-token input does not consume a million ID records, while enough short stored turns eventually fill the ID map. The test does not exercise a hosted model, long generated answers, arbitrary concurrency, or every request mix.

Save the following source as `scale_probe.py`. It writes its config, logs, and JSON results to a temporary directory unless `--output-dir` is supplied. Run these commands from the repository root:

```bash
rustup run 1.96.1 cargo build --release -p switchyard-server
uv run --python 3.13.9 \
  --with aiohttp==3.14.3 --with tiktoken==0.14.0 --with psutil==7.2.2 \
  python scale_probe.py target/release/switchyard-server --concurrency 64
```

```python
import argparse
import asyncio
import hashlib
import importlib.metadata
import json
import os
import platform
from pathlib import Path
import socket
import subprocess
import tempfile
import time

import aiohttp
from aiohttp import web
import psutil
import tiktoken

LIMIT = 65_536


def free_port():
    with socket.socket() as sock:
        sock.bind(('127.0.0.1', 0))
        return sock.getsockname()[1]


def strings(value):
    if isinstance(value, str):
        yield value
    elif isinstance(value, list):
        for item in value:
            yield from strings(item)
    elif isinstance(value, dict):
        for item in value.values():
            yield from strings(item)


async def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('binary')
    parser.add_argument('--concurrency', type=int, default=32)
    parser.add_argument('--skip-saturation', action='store_true')
    parser.add_argument('--output-dir', type=Path)
    args = parser.parse_args()
    work = args.output_dir or Path(tempfile.mkdtemp(prefix='responses-scale-'))
    work.mkdir(parents=True, exist_ok=True)
    print(json.dumps({'environment': platform.platform(), 'ram_bytes': psutil.virtual_memory().total, 'output_dir': str(work)}), flush=True)
    print(json.dumps({'dependencies': {name: importlib.metadata.version(name) for name in ['aiohttp', 'tiktoken', 'psutil']}}), flush=True)
    encoder = tiktoken.get_encoding('o200k_base')
    prefix, suffix = 'ROUTE_A: BEGIN ', ' END'
    repeats = 1_000_000
    text = prefix + ' alpha' * repeats + suffix
    repeats -= len(encoder.encode(text)) - 1_000_000
    text = prefix + ' alpha' * repeats + suffix
    tokens = len(encoder.encode(text))
    assert tokens == 1_000_000
    digest = hashlib.sha256(text.encode()).hexdigest()
    state = {'calls': 0, 'judge': 0, 'large': 0, 'active': 0, 'peak_active': 0,
             'owners': {}, 'tracked': 0, 'first_id': None, 'large_goal': 1,
             'large_arrivals': 0, 'gate': asyncio.Event()}

    async def provider(request):
        body = await request.json()
        silo = request.match_info['silo']
        state['calls'] += 1
        identity = f'resp_{silo}_{state["calls"]}'
        if silo == 'judge':
            state['judge'] += 1
            strong = any('ROUTE_B' in part for part in strings(body.get('input')))
            answer = json.dumps({'crux': 'request choice', 'primary_rule': 'LIM-1' if strong else 'SUP-1',
                                 'capability_boundary': 'unsupported' if strong else 'supported',
                                 'p_solve': 0.1 if strong else 0.9})
        else:
            previous = body.get('previous_response_id')
            if previous and state['owners'].get(previous) != silo:
                return web.json_response({'error': {'code': 'state_not_found', 'message': 'wrong provider'}}, status=400)
            content = body.get('input')
            if isinstance(content, str) and len(content) > 1_000_000:
                assert hashlib.sha256(content.encode()).hexdigest() == digest
                state['large'] += 1
                state['active'] += 1
                state['peak_active'] = max(state['peak_active'], state['active'])
                state['large_arrivals'] += 1
                if state['large_arrivals'] >= state['large_goal']:
                    state['gate'].set()
                await asyncio.wait_for(state['gate'].wait(), timeout=120)
                await asyncio.sleep(0.1)
                state['active'] -= 1
            if content == 'ROUTE_B: FORCE_CONTEXT_OVERFLOW':
                return web.json_response({'error': {'code': 'context_length_exceeded',
                                         'message': 'input exceeds model context window'}}, status=400)
            answer = f'served-by:{silo}'
            if body.get('store', True):
                state['owners'][identity] = silo
                if state['first_id'] is None:
                    state['first_id'] = identity
            state['tracked'] += int(body.get('store', True))
            if body.get('conversation'):
                state['tracked'] += 1
        if body.get('input') == 'ROUTE_B: CONFLICT':
            identity = state['first_id']
        input_tokens = tokens if silo != 'judge' and isinstance(body.get('input'), str) and len(body['input']) > 1_000_000 else 1
        response = {'id': identity, 'object': 'response', 'created_at': 1,
                    'model': body['model'], 'status': 'completed',
                    'store': body.get('store', True), 'conversation': body.get('conversation'),
                    'output': [{'id': 'msg_1', 'type': 'message', 'role': 'assistant', 'status': 'completed',
                                'content': [{'type': 'output_text', 'text': answer, 'annotations': []}]}],
                    'usage': {'input_tokens': input_tokens, 'output_tokens': 1, 'total_tokens': input_tokens + 1}}
        if body.get('stream'):
            frames = [json.dumps({'type': event, 'response': response})
                      for event in ('response.created', 'response.completed')]
            return web.Response(text=''.join(f'event: {json.loads(frame)["type"]}\ndata: {frame}\n\n' for frame in frames),
                                content_type='text/event-stream')
        return web.json_response(response)

    app = web.Application(client_max_size=40 * 1024 * 1024)
    app.router.add_post('/silo/{silo}/responses', provider)
    runner = web.AppRunner(app)
    await runner.setup()
    upstream_port, server_port = free_port(), free_port()
    await web.TCPSite(runner, '127.0.0.1', upstream_port).start()
    config = ['schema_version = 1']
    for name, silo in [('judge', 'judge'), ('a', 'A'), ('b', 'B')]:
        config.append(f'[llm_clients.{name}]\nformat = "openai_responses"\nbase_url = "http://127.0.0.1:{upstream_port}/silo/{silo}"\nmax_retries = 0')
    for name, model in [('judge', 'judge'), ('a', 'provider-a'), ('b', 'provider-b')]:
        config.append(f'[targets.{name}]\nid = "model/{model}"\nllm_client = "{name}"')
    config.append('[routes.dynamic]\nid = "route/dynamic"\ntype = "llm_classifier"\nclassifier_target = "judge"\nweak_target = "a"\nstrong_target = "b"\nbase_threshold = 0.5\nclassify_trigger = "every_request"\ncontext_window = 1048576')
    config_path = work / 'scale.toml'
    config_path.write_text('\n'.join(config) + '\n')
    log = (work / 'scale-server.log').open('w')
    process = subprocess.Popen([args.binary, '--config', str(config_path), '--host', '127.0.0.1', '--port', str(server_port)],
                               stdout=log, stderr=subprocess.STDOUT, env=dict(os.environ, RUST_LOG='error'))
    observed = psutil.Process(process.pid)
    samples = []
    async def memory_samples():
        while process.poll() is None:
            samples.append(observed.memory_info().rss)
            await asyncio.sleep(0.02)
    monitor = asyncio.create_task(memory_samples())
    base = f'http://127.0.0.1:{server_port}'
    try:
        async with aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=180), connector=aiohttp.TCPConnector(limit=256)) as client:
            for _ in range(200):
                try:
                    async with client.get(base + '/health') as response:
                        if response.status == 200:
                            break
                except aiohttp.ClientError:
                    pass
                assert process.poll() is None, 'server exited at startup'
                await asyncio.sleep(0.1)
            async def call(fields):
                payload = json.dumps({'model': 'route/dynamic', 'max_output_tokens': 16, **fields}, separators=(',', ':')).encode()
                async with client.post(base + '/v1/responses', data=payload, headers={'Content-Type': 'application/json'}) as response:
                    raw = await response.text()
                    if 'text/event-stream' in response.headers.get('Content-Type', ''):
                        data = [json.loads(line[5:]) for line in raw.splitlines() if line.startswith('data:')]
                    else:
                        data = json.loads(raw)
                    return response.status, data, len(payload)
            results = []
            for concurrency in [1, args.concurrency, args.concurrency]:
                state.update(large_goal=concurrency, large_arrivals=0, gate=asyncio.Event(), peak_active=0)
                samples.clear()
                before = observed.memory_info().rss
                async def large_turn(index):
                    start = time.monotonic()
                    status, answer, size = await call({'input': text, 'stream': index % 2 == 1})
                    assert status == 200, (status, answer)
                    if isinstance(answer, list):
                        assert all(item.get('type') != 'error' for item in answer), answer
                        answer = next(item['response'] for item in answer if item.get('type') == 'response.completed')
                    assert answer['model'] == 'model/provider-a', answer
                    return answer['id'], time.monotonic() - start, size
                batch = await asyncio.gather(*(large_turn(i) for i in range(concurrency)))
                judge_before = state['judge']
                followups = await asyncio.gather(*(call({'input': 'ROUTE_B: recall', 'previous_response_id': identity, 'store': False}) for identity, _, _ in batch))
                assert all(status == 200 and answer['model'] == 'model/provider-a' for status, answer, _ in followups)
                assert state['judge'] == judge_before
                await asyncio.sleep(0.2)
                record = {'case': 'million_token_input', 'concurrency': concurrency, 'tokens': tokens,
                          'encoding': 'o200k_base', 'text_bytes': len(text.encode()), 'request_bytes': batch[0][2],
                          'completed': len(batch), 'successful_followups': len(followups), 'upstream_peak_active': state['peak_active'],
                          'rss_before_bytes': before, 'rss_peak_bytes': max(samples), 'rss_after_bytes': observed.memory_info().rss}
                assert record['upstream_peak_active'] == concurrency, record
                results.append(record)
                print(json.dumps(record), flush=True)
            if not args.skip_saturation:
                remaining = LIMIT - 16 - state['tracked']
                if remaining % 2:
                    status, _, _ = await call({'input': 'ROUTE_A: fill'})
                    assert status == 200
                    remaining -= 1
                queue = asyncio.Queue()
                for index in range(remaining // 2):
                    queue.put_nowait(index)
                done = 0
                async def fill():
                    nonlocal done
                    while not queue.empty():
                        index = queue.get_nowait()
                        status, body, _ = await call({'input': 'ROUTE_A: fill', 'conversation': f'conv_fill_{index}'})
                        assert status == 200, (status, body)
                        done += 1
                        if done % 4096 == 0:
                            print(json.dumps({'case': 'filling', 'requests': done, 'target': remaining // 2}), flush=True)
                await asyncio.gather(*(fill() for _ in range(args.concurrency)))
                assert state['tracked'] == LIMIT - 16, state['tracked']
                boundary = await asyncio.gather(*(call({'input': 'ROUTE_A: boundary', 'conversation': f'conv_boundary_{i}'}) for i in range(32)))
                statuses = [status for status, _, _ in boundary]
                assert statuses.count(200) == 8 and statuses.count(503) == 24, statuses
                for status, body, _ in boundary:
                    if status == 503:
                        assert body['error']['code'] == 'response_state_limit_exceeded', body
                async with client.get(base + '/v1/stats') as response:
                    before_stats = await response.json()
                status, overflow, _ = await call({'input': 'ROUTE_A: buffered at capacity'})
                assert status == 503 and overflow['error']['code'] == 'response_state_limit_exceeded', (status, overflow)
                async with client.get(base + '/v1/stats') as response:
                    after_stats = await response.json()
                assert after_stats['total_errors'] == before_stats['total_errors'] + 1, (before_stats, after_stats)
                print(json.dumps({'case': 'capacity_stats', 'errors_before': before_stats['total_errors'],
                                  'errors_after': after_stats['total_errors']}), flush=True)
                status, conflict, _ = await call({'input': 'ROUTE_B: CONFLICT'})
                assert status == 409 and conflict['error']['code'] == 'response_state_conflict', (status, conflict)
                status, events, _ = await call({'input': 'ROUTE_A: stream at capacity', 'stream': True})
                assert status == 200 and len(events) == 1 and events[0]['type'] == 'error', (status, events)
                assert '65536' in events[0]['error']['message']
                judge_before = state['judge']
                status, answer, _ = await call({'input': 'ROUTE_B: recall oldest', 'previous_response_id': state['first_id'], 'store': False})
                assert status == 200 and answer['model'] == 'model/provider-a', (status, answer)
                assert state['judge'] == judge_before
                record = {'case': 'capacity', 'limit': LIMIT, 'boundary_requests': 32, 'accepted': 8, 'http_503': 24,
                          'stream_error_events': len(events), 'oldest_id_still_works': True, 'conflict_http_409': True, 'server_alive': process.poll() is None}
                results.append(record)
                print(json.dumps(record), flush=True)
            status, answer, _ = await call({'input': 'ROUTE_B: FORCE_CONTEXT_OVERFLOW', 'previous_response_id': state['first_id'], 'store': False})
            assert status == 400 and answer['error']['code'] == 'context_length_exceeded', (status, answer)
            print(json.dumps({'case': 'upstream_context_limit', 'status': status, 'code': answer['error']['code']}), flush=True)
            calls_before = state['calls']
            status, answer, request_bytes = await call({'input': 'x' * (33 * 1024 * 1024)})
            assert status == 413 and state['calls'] == calls_before, (status, answer)
            print(json.dumps({'case': 'request_body_limit', 'status': status, 'request_bytes': request_bytes, 'upstream_calls': 0}), flush=True)
            (work / 'scale-results.json').write_text(json.dumps(results, indent=2) + '\n')
    finally:
        process.terminate()
        await asyncio.to_thread(process.wait, 10)
        await monitor
        log.close()
        await runner.cleanup()


asyncio.run(main())
```

</details>


crates/libsy-llm-client/src/run.rs
```diff
@@ -414,6 +498,63 @@ impl ClientRouter {
...
 
+    /// The model holding the provider state `request` continues, when this router remembers
+    /// one and still serves it.
+    fn stored_state_owner(&self, request: &Request) -> Option<StateOwner> {
+        let owners = self.inner.state_owners.as_ref()?;
+        let fields = &request.llm_request.extensions.fields;
+        let (trigger, id) = match fields.get("previous_response_id").and_then(Value::as_str) {
+            Some(id) => ("previous_response_id", id),
+            None => ("conversation", conversation_id(fields)?),
+        };
```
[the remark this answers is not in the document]
elyasmnvidian: I kept `previous_response_id` as the sole lookup when it is present. The [Responses API reference](https://developers.openai.com/api/reference/cli/resources/responses/methods/create) says `previous_response_id` cannot be used with `conversation`, so the proposed fallback would apply only to an invalid request. Requests with `conversation` alone still use its recorded model.

crates/switchyard-server/tests/server.rs
```diff
@@ -10,7 +10,7 @@ use std::io::Write;
 use std::sync::Arc;
 
 use axum::body::{Body, Bytes};
-use axum::extract::{DefaultBodyLimit, State};
+use axum::extract::{DefaultBodyLimit, Path, State};
```
grahamking, line 13: Can you reduce the new tests in here to only one? That should reduce the PR a lot.

grahamking: Tips from my agent for your agent, to simplify:

 - Keep end-to-end coverage for buffered replies, streaming, and /v1/decision.                                                                                                                                                                                
 - Move storage flags, format exclusions, and cache behavior into focused client tests. The nested server-test matrix and string-based TOML replacements are cumbersome.                                                                                      
 - Keep the shared-answer-client check: excluding a separate classifier client is necessary behavior, not needless complexity.                                                                                                                                
 - Small cleanups: use debug! instead of per-continuation info! at run.rs:355; remove the redundant model lookup at run.rs:538—the router’s model map is immutable.

elyasmnvidian: I reduced the server coverage to one regression with shared setup and explicit TOML routes. It checks buffered and streamed follow-ups, `/v1/decision`, shared-client routing, conflicts, and provider failures. Storage settings, conversation ID forms, and format cases now live in client tests. Both targeted tests pass: `responses_continuations_preserve_state_ownership` and `responses_state_tracking_uses_upstream_storage_and_format`. The continuation log is now at debug level, and I removed the extra model lookup because the router's model mapping is immutable.

grahamking (approved):

PR #727, opened 2026-09-16
fix(translation): preserve native Anthropic request controls

grahamking:
Preserve:
- `inference_geo` - this is the one QA found
- `service_tier`
- `stop_sequences`
- `metadata`
- `cache_control`
- `container`
- `speed`
- `diagnostics`

Preserved when a target system prompt forces Anthropic request reconstruction. Limit this preservation to Anthropic input. Prefer native `stop_sequences` over the OpenAI `stop` fallback. Skip automatic cache-marker insertion when top-level cache_control is present to avoid conflicting TTLs.

Previously we dropped these.

Add one regression test and extend an existing cache test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved caller-specified Anthropic prompt caching settings instead of overwriting them.
  - Anthropic-specific request options, including metadata, service tier, inference region, stop sequences, and container settings, are now retained when translating compatible requests.
  - Improved handling of stop-sequence fallbacks across Anthropic and OpenAI request formats.
  - Prevented internal provenance markers from appearing in outgoing requests.


ayushag-nv:
found one gap: explicit thinking settings still disappear when a target system prompt forces reconstruction.

ayushag-nv (approved):
discussed offline with @grahamking for some more fields that get dropped. tickets will be created for internal tracking

PR #729, opened 2026-09-16
fix(translation): preserve Responses built-in tool history

afourniernv:
## What

Preserve the six OpenAI Responses built-in call/output history items when a target
`system_prompt` forces request reconstruction:

- `apply_patch_call` / `apply_patch_call_output`
- `shell_call` / `shell_call_output`
- `computer_call` / `computer_call_output`

Responses replay keeps the original top-level item and its fields. Invalid nested items are
rejected, as are translations to Chat or Anthropic that would otherwise turn the item into text.

## Why

[SWITCH-1482](https://linear.app/nvidia/issue/SWITCH-1482) describes the prompted path converting
these items into user `input_text`, which breaks the call/output relationship.

This PR covers caller-returned history re-encoded after request mutation. Responses provider
output translated to another public API is tracked separately in
[SWITCH-1471](https://linear.app/nvidia/issue/SWITCH-1471).

## Notes for reviewers

The main path is in the Responses buffered request codec. The shared predicate is crate-private,
and there are no public API changes. The one raw JSON clone is needed because request encoding
borrows the IR and returns an owned body.

Validation:

- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`
- Strict loopback server matrix for all six item types
- Live OpenAI Responses continuation: a real `apply_patch_call` and its output completed through
  the prompted Switchyard route; both provider calls returned HTTP 200

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of built-in tool history in Responses-format requests.
  - Preserved supported top-level tool call and output items during request re-encoding.
  - Added validation to reject built-in tool items nested within message content.
  - Prevented unsupported conversion of built-in tool history into OpenAI Chat or Anthropic text formats.
  - Other unrecognized content blocks continue to produce diagnostics and text fallbacks where supported.

- **Tests**
  - Added regression coverage for tool-history preservation, ordering, and validation scenarios.


grahamking:
@afourniernv Can you do the Code Rabbit? My local review found it too - Astra says:

>  responses/buffered.rs:484–486 https://github.com/NVIDIA-NeMo/Switchyard/blob/26ed5a9ebfd7405f7aec8ce7d7ab3af18ff7fb98/crates/switchyard-translation/src/codecs/responses/buffered.rs#L484-L486 checks only array-shaped content. An object such as         
 "content": {"type": "shell_call", ...} bypasses validation and becomes JSON text.
> 
>  I reproduced this: after applying a target prompt, Responses, Chat, and Anthropic all accept the item as user text. This contradicts the PR’s promised rejection of invalid nested items. CodeRabbit’s existing finding is still valid.

grahamking (approved):

PR #730, opened 2026-09-16
fix(translation): preserve URL citations between Chat and Responses

grahamking:
A response that did a web search can include an `annotation` field with
a URL citation that looks like this:
```
{"type":"url_citation","url_citation":{"url":"https://linkerd.io/docs/reference/load-balancing/","title":"Load Balancing | Linkerd","start_index":0,"end_index":0,"content":"Linkerd uses a sophisticated .."}}
```

In fact I think it's the only type of `annotation` that currently exists. See here for more detail: https://openrouter.ai/docs/api_reference/responses/web-search#response-with-annotations

Now we support them, carrying them through normalized responses. We preserve URLs, titles, and character offsets in both buffered translation directions.

Adds one regression test covering both directions and merged Unicode text.

## Summary by CodeRabbit

* **New Features**
  * Added support for preserving URL citations in assistant responses, including citation titles, links, and highlighted text ranges.
  * URL citations are now translated between supported OpenAI response formats and included when responses are encoded.

* **Bug Fixes**
  * Improved citation offset handling across combined messages, content sections, and multibyte text.
  * Invalid citation annotations are safely ignored.


crates/libsy/src/algorithms/advisor_gate/signals.rs
```diff
@@ -89,6 +89,7 @@ mod tests {
     fn tool_call_agg() -> AggLlmResponse {
         AggLlmResponse {
             outputs: vec![ResponseOutput {
+                url_citations: Vec::new(),
```
ayushag-nv, line 92: is this only relevant for advisor gate ? is ResponseOutput not constructed in any other algo ??

grahamking: It's only in `noop` because that doesn't use `Fallthrough`. Here it's the advisor gate tests.

ayushag-nv: got it. This view is bit messy , didn't realize these are tests

ayushag-nv (approved):
looks good. Some good old index and offset stuff.

crates/libsy/src/algorithms/advisor_gate/signals.rs
grahamking: Yup, happens to me a lot. It's very difficult to get context on these little snippets.

PR #736, opened 2026-09-17
docs(relay): clarify caller and target format translation

bbednarski9:
## What

Correct the native-plugin README and Relay integration guide so both describe caller-to-target format translation consistently.

- Remove the README's blanket requirement that the target use the caller's wire format and its instruction to create separate targets and routes per format.
- Explain that, with server-owned credentials, one route using an `openai_chat` target can serve Chat Completions, Responses, and Anthropic Messages callers, with buffered or streaming output in the caller's format.
- Preserve the `forward_auth` credential-family restriction, explicitly grouping Chat Completions and Responses as OpenAI-compatible and Messages as Anthropic.
- Link the separate Relay credential-forwarding limitation, and label the tracker as NVIDIA internal.
- Retain the provider-specific field compatibility caveat and cross-link the two guides.

## Why

The README told users that each route's target must use the caller's wire format, while the integration guide explicitly allowed different formats. This could lead users to create redundant routes or conclude that a Chat upstream cannot serve Responses or Anthropic callers.

The contradiction was confirmed on upstream main at `7038342cb92b5952601e58993d5063ce74aca2fe`. Current code supports the integration guide's translation model:

- [`call_rewrite_model`](https://github.com/NVIDIA-NeMo/Switchyard/blob/7038342cb92b5952601e58993d5063ce74aca2fe/crates/libsy-llm-client/src/client.rs#L499) resolves the target's configured backend when request metadata does not pin a format, then encodes the request for that backend.
- [`check_caller_format`](https://github.com/NVIDIA-NeMo/Switchyard/blob/7038342cb92b5952601e58993d5063ce74aca2fe/crates/switchyard-runner/src/route.rs#L180) enforces credential-family compatibility for forwarded authentication. It does not impose a general same-wire-format requirement.
- Existing [plugin tests](https://github.com/NVIDIA-NeMo/Switchyard/blob/7038342cb92b5952601e58993d5063ce74aca2fe/crates/switchyard-nemo-relay-plugin/src/runtime.rs#L900) exercise Responses callers against Chat and Anthropic targets. The same file covers buffered and streamed Responses tool calls through a Chat target.

The supplied issue report also records six successful live requests on Switchyard v0.3.0-rc.2 with Relay 0.8.1: each of the three caller formats, buffered and streamed, using one Chat target with server-owned credentials. Those results are reporter-provided evidence, not a live-provider rerun performed for this PR. They make no claim about forwarded credentials.

## Notes for reviewers

Start with the Request handling section of the native-plugin README, then compare the matching paragraphs in the integration guide. This change affects only those two Markdown files. Runtime routing, translation, and authentication behavior are unchanged.

Validation completed:

- `make -C docs publish`: passed, including `mkdocs build --strict`.
- `git diff --check`: passed.
- Checked that the format/auth paragraphs match across both guides and that the built integration page contains the README cross-link and NVBug link.
- Verified the documented behavior against current-main implementation and existing test source.

Rust tests and live-provider requests were not run for this documentation-only change. The internal NVBug's current status was not independently verified; the link preserves the separate forwarding caveat requested in the report.

Suggested runtime acceptance check when repeating the report: use one `openai_chat` target and one passthrough route with server-owned credentials, exercise all three caller formats in buffered and streamed modes, and check decoded content and successful protocol termination. Remove the trailing semicolon from the report's `base_url` TOML assignment before using that reproduction.

## Summary by CodeRabbit

- **New Features**
  - Added support for translating requests and responses between OpenAI Chat, OpenAI Responses, and Anthropic Messages formats.
  - Translation supports both buffered and streaming responses.
  - Server-managed credentials can use a single target format for callers using different supported formats.

- **Documentation**
  - Documented credential compatibility requirements when forwarding authentication.
  - Clarified format-dependent provider field support and noted a known Relay limitation.


afourniernv (approved):
Matches the documented cross-format behavior and the NVBug acceptance criteria. Looks good.

PR #744, opened 2026-09-17
feat(stage-router): hand capable-first execution to efficient

sabhatinas:
## What

Fix `capable_first` so that, after the first observed edit or write, undecided turns route to the efficient tier.

Hard overrides and decisive recovery signals still select the capable tier. `efficient_first` is unchanged. The transition is reported as `decision_source = "mutation"`.

## Why

At `confidence_threshold = 0.5`, the production dimension cannot independently cross the threshold, so `capable_first` could remain on the capable tier for the full task.

This restores the intended capable-first transition while preserving signal-driven recovery.

## Validation

`cargo test -p switchyard-libsy` — 312 passed.

## Summary by CodeRabbit

* **Behavior Changes**
  * Capable-first mode now uses the efficient tier after edit or write activity when no other decision has been made.
  * Decision details now identify mutation-triggered tier selection.


grahamking:
From Astra:  @sabhatinas 
 
1. Text and comments can be mistaken for actual edits.                                                                                                                                                                                                       
    In tool_signals.rs:555–601 (https://github.com/NVIDIA-NeMo/Switchyard/blob/3252940a/crates/libsy/src/algorithms/util/tool_signals.rs#L555-L601), the shell-command search also matches code inside strings. The patch-call detector skips strings but not 
    comments. I reproduced both:                                                                                                                                                                                                                              
     - Printing a string containing tools.exec_command({cmd:"gofmt -w src/main.go"}) counts as an edit.                                                                                                                                                       
     - A comment containing tools.apply_patch(patch) counts as an edit.                                                                                                                                                                                       
                                                                                                                                                                                                                                                              
    Neither changes a file. With this PR, either can incorrectly switch subsequent undecided requests to the efficient model. Detection must distinguish actual calls from quoted or commented text.                                                          
                                                                                                                                                                                                                                                              
 2. Valid commands are missed when cmd is not the first field.                                                                                                                                                                                                
    The pattern at tool_signals.rs:209–214 (https://github.com/NVIDIA-NeMo/Switchyard/blob/3252940a/crates/libsy/src/algorithms/util/tool_signals.rs#L209-L214) requires cmd immediately after {. I reproduced a missed edit with:                            
    ```javascript                                                                                                                                                                                                                                             
      tools.exec_command({workdir:"/app", cmd:"gofmt -w src/main.go"})                                                                                                                                                                                        
    ```                                                                                                                                                                                                                                                       
    Field order should not determine whether the router notices a file change.

crates/libsy/src/algorithms/util/stage.rs
```diff
@@ -103,7 +103,7 @@ impl Tier {
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize)]
```
grahamking, line 103: See PR comment

sabhatinas: will do! The comments are valid, the string patterns we match have possible leaks, do you know if Relay event streams have normalized tool responses ? Would it get easier if we used their schema to normalize the IR with ATOF format?

PR #750, opened 2026-09-17
fix(server): reject inputs disabled by route capabilities

elyasmnvidian:
When a route declared `vision = false`, `reasoning = false`, or `tool_calling = false`, the server still forwarded those inputs to the provider. It now returns HTTP 400 before dispatch. OpenAI errors use `unsupported_capability`; Anthropic errors use `invalid_request_error`.

The shared route check covers Chat Completions, Responses, and Anthropic Messages. It reads decoded content and preserved provider JSON: forwarding can retain fields that decoding omits. This includes built-in `tool_choice`, raw reasoning controls, `file_id` images, screenshots, Codex `additional_tools`, and images inside tool results. Rejecting these requests lets callers correct them without Switchyard silently removing their content.

Explicit `true` and unset declarations preserve existing forwarding. Codex's `models` list stays empty to preserve its instructions; Codex still uses its own settings and does not automatically adapt to route declarations. Callers must remove unsupported inputs or select a compatible route. The checks apply to registered server routes, including decision and auxiliary requests. Transparent forwarding through `fallback_client` and direct library calls remain unchanged.

Validation scans decoded blocks and relevant raw JSON nodes, using a temporary vector of borrowed JSON references when content checks apply. It retains no state between requests. The existing 32 MiB request limit and JSON parser nesting limit still apply.

## Example and evidence

A local run used an OpenAI Responses stub and this route declaration:

```toml
[routes.custom]
id = "switchyard/text-only"
type = "passthrough"
target = "local"
vision = false
reasoning = false
tool_calling = false
```

This request to `POST /v1/responses` returned HTTP 400 without calling the stub:

```json
{"model":"switchyard/text-only","input":"hello","reasoning":{"effort":"high"}}
```

```json
{"error":{"message":"route switchyard/text-only declares reasoning = false; remove the unsupported input or select another route","type":"invalid_request_error","code":"unsupported_capability"}}
```

The same run tested image, reasoning, and tool-definition requests separately under both `switchyard/text-only` and `gpt-5.6-sol`, each with the declarations above. All six returned 400 with zero stub calls. Plain text returned 200/`PONG`, made one stub call, and preserved caller instructions.

The HTTP regression first failed on the unchanged server: an image request with `vision = false` reached the stub and returned 200. It now passes across all three APIs and confirms that explicit `true` and unset declarations still forward images, reasoning, tools, and instructions:

```bash
cargo test -p switchyard-server --test server disabled_route_capabilities_reject_requests_before_calling_upstream -- --nocapture
```

Focused regressions also cover preserved provider fields. The checks use local stubs; they do not establish Codex CLI adaptation or external model behavior.

## Summary by CodeRabbit

- **New Features**
  - Requests now validate route capabilities before dispatch.
  - Routes explicitly disabling vision, reasoning, or tool calling return a clear HTTP 400 error instead of forwarding unsupported requests.
  - Vision validation includes images nested in tool results.
  - Model discovery now reports applicable capability values, including unset capabilities as `null`.

- **Documentation**
  - Updated route and model discovery documentation to describe capability restrictions, error responses, and behavior for enabled or unspecified capabilities.


crates/switchyard-server/src/capabilities.rs
```diff
@@ -0,0 +1,265 @@
...
+                || kind.ends_with("_call_output")
+                || kind.ends_with("_tool_use")
+                || kind.ends_with("_tool_result"))
+        {
+            return Some("tool_calling");
+        }
+        pending.extend(value.get("content"));
+        if kind.ends_with("_call_output") {
+            pending.extend(value.get("output"));
+        }
```
grahamking, line 110: Astra says this misses `mcp_approval_response` - "Tool-approval replies bypass the restriction. Add explicit checks for tool-approval request and response items.".

crates/switchyard-server/src/capabilities.rs
```diff
@@ -0,0 +1,265 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+    use switchyard_translation::{WireFormat, decode_request, encode_request};
+
+    #[test]
+    fn rejects_tool_controls_and_history_preserved_outside_normalized_tools()
+    -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
elyasmnvidian: Added intent comments to the four capability tests and the cross-API test. Also documented `unsupported_capability` and `unsupported_content`, including preserved JSON and images nested in tool results.

crates/switchyard-server/tests/server.rs
```diff
@@ -3324,6 +3324,146 @@ target = "shared"
...
+            "{endpoint}: {capability}"
+        );
+        let error = response.json()?;
+        assert_eq!(error["error"]["type"], "invalid_request_error");
+        assert!(
+            error["error"]["message"]
+                .as_str()
+                .unwrap_or_default()
+                .contains(&format!("{capability} = false"))
+        );
```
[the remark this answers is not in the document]
elyasmnvidian: Added the top-level `type: "error"` assertion for Anthropic while keeping the nested error checks. The test also checks OpenAI’s `unsupported_capability` code and passes across all three APIs.

crates/switchyard-server/src/capabilities.rs
elyasmnvidian: Added explicit checks for `mcp_approval_request` and `mcp_approval_response`, with approval and denial cases that omit `tools`. The router returned 200 before the fix; it now returns 400 with no upstream calls. Enabled and unset routes still forward the input unchanged. All 82 server tests pass.

AGENTS.md
```diff
@@ -25,6 +25,9 @@ Write for a high-school level in short, simple sentences. Avoid jargon, analogie
 - Avoid abstractions, configurability, and defensive code for hypothetical needs.
 - If the implementation grows unexpectedly large, reconsider and simplify it.
 - For bugs, reproduce the failure before fixing it; for refactors, establish a behavioral baseline first.
+- For protocol restrictions, check supported input variants and preserved provider JSON as well as decoded content. Tool history includes approval requests and replies; suffix matching alone does not cover every variant.
+- Test a rejected input without another forbidden field that could hide a missing check. Assert API-specific error fields and zero upstream calls. Preserve explicit `true` and unset behavior when they differ.
```
grahamking: Can you remove these two? Not sure what they are trying to say, sounds specific to some work you were doing maybe.

AGENTS.md
```diff
@@ -52,5 +55,5 @@ One step, one reviewed, one-line commit.
   (`type(scope): summary`). No body, no `Co-Authored-By` trailer.
 - Pull request titles use the same Conventional Commits form.
 - Use `git commit -s` so every commit carries the required DCO sign-off.
-- Never commit unprompted. Show the diff, get approval, then commit.
+- Commit reviewed changes within the requested scope without asking for separate approval.
```
grahamking: Yes, but we should have discussed it. That might surprise people.

elyasmnvidian: I will revert this whole file

AGENTS.md
elyasmnvidian: I will revert this whole file

grahamking (approved):

PR #758, opened 2026-09-17
fix(runner): reject conflicting same-model clients within a route

grahamking:
Route execution keys clients by model ID, so targets sharing an ID can silently use the wrong endpoint and request settings.

Reject these conflicts at startup. Preserve support across separate routes, extend regression coverage, and update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Routes now reject configurations where the same model ID is assigned to different LLM clients within a single route.
  * Duplicate model IDs remain supported when used across separate routes.

* **Documentation**
  * Updated configuration guidance to clarify model ID and LLM client requirements for completion, judge, and classifier targets.


crates/switchyard-runner/src/config.rs
```diff
@@ -348,12 +348,25 @@ impl DeploymentConfig {
...
+            routing_answer_target,
+        } = self.build_route_target_prompts(route_name, route)?;
         let mut by_model = HashMap::new();
+        let mut targets_by_model: HashMap<&str, (&str, &TargetConfig)> = HashMap::new();
         let mut caller_auth = None;
         for name in route.callable_target_names() {
             let target = self.targets.get(name).ok_or_else(|| {
                 RunnerError::configuration(format!("route references unknown target {name}"))
             })?;
+            if let Some((first_name, first)) = targets_by_model.insert(&target.id, (name, target))
```
ayushag-nv, line 362: Do we want to keep doing this in long term ..rejecting based on same model id by diff clients ? In case we bring some latency routing kinda thing, in that case serving same model with diff clients can be helpful.

grahamking: I'm not sure, but I think so. I picturing it working like this:

1. Model: Switchyard is the first-tier router, it would choose say `z-ai/glm-5.3`, and it's done.
2. Provider: A second-tier router will choose the best provider for GLM 5.3, like what Open Router does. This is based on a combination of availability, price, latency, and throughput.
3. Backend: At the provider, a third-tier router like Dynamo's kv-router chooses the best specific backend.

But I feel that is all out of scope for Switchyard. It just chooses the model.

ayushag-nv: sure, this make sense. If in case we ever bring back latency aware stuff, then it can make sense.

ayushag-nv (approved):

PR #772, opened 2026-09-18
fix(metrics): record streamed calls by their final outcome

nachiketb-nvidia:
## What

Defer routed-call metrics until a streaming answer finishes or is dropped. Stream and decode errors count as failures, and an abandoned open stream cannot count as a success. Record each outcome once and remove the server's duplicate Prometheus error increment.

## Why

Fixes [SWITCH-1441](https://linear.app/nvidia/issue/SWITCH-1441). An HTTP 200 stream was counted as successful before a later in-band error arrived, contradicting `/v1/stats`.

## Notes for reviewers

85 added production lines. Extends one existing HTTP regression with three assertions; no new test functions. Stream events and buffered responses remain unchanged. Metrics include the time spent consuming the stream.

The local HTTP regression failed on unchanged main because a failed stream incremented `switchyard_requests_total`. It now verifies no success or successful-call latency is recorded, total errors increase, and the per-model error is counted exactly once. The reported bug no longer reproduces through the server.

Validation: targeted server `stream` tests (14 passed), buffered stats control (1 passed), client observability `stream` tests (3 passed, including cancellation), and classifier accounting control (1 passed). Formatting, workspace clippy, and diff checks passed. Only local mock providers; no full suite or live provider calls.

Landing note: this independent main-based PR overlaps #773 in `run.rs`. The combined resolution calls the stream-outcome observer at each completion candidate and counts algorithm-supplied answers once. That resolution passed both regressions plus classifier-accounting and streaming-success controls in a temporary worktree.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved usage and performance metrics for streamed LLM responses.
  - Streaming requests are now recorded as successful only after normal completion.
  - Stream interruptions, decoding errors, and abandoned streams are correctly counted as failures.
  - Prevented failed streaming requests from incorrectly increasing request and latency metrics.
  - Corrected error reporting to avoid duplicate error counts.


ayushag-nv (approved):
looks solid.

PR #776, opened 2026-09-18
fix(translation): omit text blocks for absent Chat response content

grahamking:
Preserve missing or null visible content as no text block in buffered Chat responses. This removes redundant empty Anthropic text blocks alongside reasoning while retaining explicit empty strings.

Reviewed-by: Pi:Grok 4.6 medium

## Summary by CodeRabbit

- **Bug Fixes**
  - Responses with omitted or `null` content now preserve an empty content array instead of producing an empty text block.
  - Reasoning responses now correctly distinguish between absent, empty, and non-empty visible text content.
  - Response settings continue to translate `length` into `max_tokens`.

- **Tests**
  - Added regression coverage for reasoning responses and content-presence handling.


ayushag-nv (approved):
neat

PR #782, opened 2026-09-18
fix(translation): Translate allowed_tools between codecs, reject when not possible

grahamking:
Before this change, a Responses tool_choice of `{"type": "allowed_tools", ...}` fell into the `Value::Object(_) => None` arm of `decode_responses_tool_choice`. `request.tool_choice` became `None`, and every translated request silently dropped the restriction. That's a security bug: a request constraining the model to safe_lookup would arrive at the target allowing delete_all_records too.

Now we preserve Responses allowed_tools during decoding, translating them between codecs where possible and rejecting otherwise. Includes test.

Reviewed-by: Pi:GLM 5.3 Flash high


crates/switchyard-translation/src/codecs/anthropic/buffered.rs
```diff
@@ -200,6 +200,14 @@ impl FormatCodec for AnthropicMessagesCodec {
         }
         let mut diagnostics = Vec::new();
         validate_request_capabilities(request, &mut diagnostics, policy)?;
+        if let Some(ToolChoice::Raw(choice)) = &request.tool_choice
+            && choice.get("type").and_then(Value::as_str) == Some("allowed_tools")
+        {
+            // A tool restriction must never be weakened, even under a lossy policy.
+            return Err(TranslationError::LossyConversion(
```
ayushag-nv: why can't it be translated ?? How does Anthropic format handle this ??

ayushag-nv (approved):
Looks good and important fix. Just one question left

crates/switchyard-translation/src/codecs/anthropic/buffered.rs
grahamking: It can be, very good point! Done.

PR #783, opened 2026-09-18
fix(translation): encode Anthropic PDFs as OpenAI data URLs

nachiketb-nvidia:
## What

Encode Anthropic base64 documents as `data:<media_type>;base64,<data>` in the shared OpenAI file encoder. Both Chat Completions and Responses retain the original bytes and filename.

## Why

Fixes [SWITCH-1484](https://linear.app/nvidia/issue/SWITCH-1484). Anthropic PDF requests currently send bare base64 as OpenAI `file_data`, which causes HTTP 400 from the upstream provider.

## Notes for reviewers

Eight added production lines. Extends the existing media regression to both OpenAI formats and both preservation modes, checks round trips, and corrects two existing tool-result expectations. No new test functions.

Validation:
- The extended regression failed on main for both target formats and passes with this fix.
- Native-server reproduction with a complete PDF and a strict local mock: HTTP 400 before, HTTP 200 after for both targets. The mock checks the data URL, unchanged PDF bytes, and filename.
- Seven targeted media/file tests passed (`request_translation` filters `media`, `multimodal`, and exact tests `responses_flat_file_data_survives_into_chat` and `openai_chat_image_and_file_parts_translate_to_valid_responses_input`).
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets --locked -- -D warnings` passed.

No full test suite or live provider calls were run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Base64-encoded documents now retain their media type in generated data URLs, improving compatibility with downstream services.
  * PDF translations now use complete `data:application/pdf;base64,...` URLs.
* **Tests**
  * Expanded media translation coverage across supported targets and normalization policies.
  * Added validation for round-trip restoration and equivalent file representations.


nachiketb-nvidia:
Validated SWITCH-1484 against baseline `b5a9a7480` and this MR (`63f564b27`). The missing-prefix defect no longer reproduces in the targeted regression or the native-server reproduction.

| Anthropic PDF request | Baseline | Fixed |
| --- | --- | --- |
| Chat upstream | Bare base64; HTTP 400 | PDF data URL; HTTP 200 |
| Responses upstream | Bare base64; HTTP 400 | PDF data URL; HTTP 200 |

The end-to-end check used a complete PDF with a valid cross-reference table and a local upstream that rejects an incorrect data URL, altered bytes, or an altered filename. It exercised the public `/v1/messages` endpoint and both target formats. This confirms the outbound request defect is fixed; it is not a live-provider retest.

The existing regression was extended, with no new test functions. It failed on baseline for both target formats and now passes for both default and normalized translation, including round trips. Related image, text, file-ID, native OpenAI file-data, and tool-result controls also pass.

Commands (7 targeted tests total):
```sh
cargo test -p switchyard-translation --locked --test request_translation media
cargo test -p switchyard-translation --locked --test request_translation multimodal
cargo test -p switchyard-translation --locked --test request_translation responses_flat_file_data_survives_into_chat -- --exact
cargo test -p switchyard-translation --locked --test request_translation openai_chat_image_and_file_parts_translate_to_valid_responses_input -- --exact
cargo fmt --all --check
cargo clippy --workspace --all-targets --locked -- -D warnings
```
All passed. No full suite or live provider calls were run.

grahamking (approved):

PR #788, opened 2026-09-18
fix(translation): reject unsupported native Responses output

nachiketb-nvidia:
## What

Reject translation of native Responses `apply_patch_call`, `shell_call`, `computer_call`, and `image_generation_call` output to Chat or Anthropic. Buffered requests return an unsupported-translation error; streams emit a terminal error instead of a normal successful stop.

Keep the native buffered items in the existing opaque content representation so Responses re-encoding can retain their payload and identity. Check retained stream events before they are discarded, including output found only in completed or incomplete snapshots.

## Why

Fixes [SWITCH-1471](https://linear.app/nvidia/issue/SWITCH-1471). These provider outputs are currently skipped, leaving clients with an empty successful response and no action or image. The ticket explicitly accepts an unsupported-translation error. #729 addressed request history; this fixes provider output.

## Notes for reviewers

88 added production lines and one regression test. The checks live in the translation crate and use existing content and error types. Responses passthrough and ordinary function/custom-tool translation are covered by controls.

Validation:
- The new regression failed against main and passes with the fix. It covers both preservation policies, separate buffered decode/encode, both stream entry points, partial output before an error, final-snapshot-only output, and exact Responses replay.
- Native-server checks with a local Responses mock: all 24 affected Chat/Anthropic cases changed from silent loss to explicit errors; all 12 Responses controls preserved the original output items.
- Thirteen targeted tests, `cargo fmt --all --check`, and `cargo clippy --workspace --all-targets --locked -- -D warnings` passed.

No full test suite or live provider calls were run. No returned tools were executed.

## Summary by CodeRabbit

* **New Features**
  * Native OpenAI Responses outputs—including patch, shell, computer, and image-generation calls—are preserved when using the same format.
  * Preserved outputs are retained in buffered and streaming responses alongside other content.

* **Bug Fixes**
  * Unsupported native outputs now produce clear translation errors instead of being incorrectly converted.
  * Invalid response and streaming data is rejected consistently, preventing incomplete or misleading target responses.


nachiketb-nvidia:
Validated SWITCH-1471 against unmodified main `62494ea70` and this MR at `edf23d465`. The reported silent-loss bug no longer reproduces in the targeted regression or local native-server reproduction: unsupported output now produces an explicit translation error.

The HTTP reproduction used a real Switchyard server and a deterministic loopback Responses provider. It covered `apply_patch_call`, `shell_call`, `computer_call`, and `image_generation_call`, each through Chat, Anthropic, and Responses, in buffered, full-stream, and final-snapshot-only modes: 36 cases per build.

| Cases | Before | After |
| --- | --- | --- |
| 8 buffered Chat/Anthropic cases | HTTP 200; action/image omitted and normal stop | HTTP 500 with explicit unsupported-translation error |
| 16 streamed Chat/Anthropic cases | HTTP 200; action/image omitted and normal stop | One terminal SSE error; no normal finish reason or success terminator |
| 12 Responses controls | Original output preserved | Original output preserved, including IDs, operations, and image bytes |

Streaming HTTP headers can already be committed before the unsupported item arrives, so those responses remain HTTP 200 and signal failure through the terminal SSE error. The check asserts that no successful completion follows it.

The single new test also covers preservation disabled, separate buffered decode/encode, both stream entry points, output arriving after partial text, incomplete snapshots, and exact same-format replay. It failed before the production fix and now passes. Existing controls confirm normal function/custom-tool output, response preservation, terminal snapshot recovery, and #729's built-in request history.

All 13 targeted tests passed:
```sh
cargo test -p switchyard-translation --locked --test response_translation responses_native_output_is_preserved_or_rejected -- --exact
cargo test -p switchyard-translation --locked --test request_translation responses_builtin_tool_history
cargo test -p switchyard-translation --locked --test stream_translation preserved_
cargo test -p switchyard-translation --locked --test stream_translation responses_terminal_snapshots_recover_missing_output_once -- --exact
cargo test -p switchyard-translation --locked --test stream_translation responses_function_call_stream_ends_with_tool_use_on_every_wire -- --exact
cargo test -p switchyard-translation --locked --test stream_translation responses_decode_emits_tool_arguments_once -- --exact
cargo test -p switchyard-translation --locked --test response_translation responses_custom_tool_call_output_round_trips_with_request_extensions -- --exact
cargo test -p switchyard-translation --locked --test response_translation responses_reasoning_and_message_preserve_the_final_answer -- --exact
cargo test -p switchyard-translation --locked --test lossless_roundtrip embedded_preservation_roundtrips_responses_for_every_source_target_pair_exactly -- --exact
cargo test -p switchyard-translation --locked --test lossless_roundtrip embedded_preservation_survives_every_distinct_three_format_response_cycle -- --exact
```
`cargo fmt --all --check`, workspace clippy with `--all-targets --locked -- -D warnings`, and `git diff --check` passed. No full test suite or live provider calls were run, and no returned tool action was executed.

ayushag-nv (approved):

PR #797, opened 2026-09-20
fix(translation): honor Chat streaming usage opt-in

ajcasagrande:
## What

Makes generated OpenAI Chat streaming usage opt-in through `stream_options.include_usage`. Requested usage now follows the finish chunk as one `choices: []` frame at clean EOF. The relay test confirms hidden client output still emits answer-token telemetry.

## Why

OpenAI Chat callers that omit usage opt-in must not receive generated usage frames. This keeps provider raw replay unchanged while preserving clean-EOF ordering and telemetry behavior.

## Notes for reviewers

- RED to GREEN contract tests cover omitted, false, non-boolean, true, missing backend usage, post-stop errors, and raw replay.
- Full workspace tests passed. Focused relay tests, relay crate tests, scoped Clippy, formatting, and diff checks passed.
- Fresh Graham review approved with zero findings.

## Summary by CodeRabbit

- **New Features**
  - OpenAI Chat streaming responses can now include a separate final usage-only chunk when explicitly requested.
  - Usage information is emitted after normal completion and remains available even when no backend usage data is returned.

- **Bug Fixes**
  - Prevented usage data from being embedded in terminal message chunks.
  - Usage is no longer emitted for errored streams or when usage inclusion is not requested.
  - Ensured usage is emitted at most once and stream replay reaches the correct final state.


ajcasagrande:
closing for now, while product alignment is conferred

PR #799, opened 2026-09-20
fix(translation): enforce reasoning source boundaries

ajcasagrande:
## What

- Binds Anthropic signed and redacted thinking replay to internal provenance trusted only from Anthropic decoding.
- Prevents Chat reasoning details from forging private Anthropic replay, including under Reject policy.
- Preserves valid Anthropic same-format and Responses encrypted replay.
- Handles foreign reasoning effort safely and reports static lossy diagnostics for unreplayable private metadata.

## Why

Caller-controlled Chat fields must not grant access to provider-private reasoning transport. This keeps opaque reasoning boundaries explicit while retaining supported round trips.

## Notes for reviewers

- Added strict RED-to-GREEN Chat spoof tests for signed and redacted markers, including Reject-policy behavior.
- Full workspace tests passed; scoped Clippy, cargo fmt check, and git diff check passed.
- Fresh Graham v3 review approved with zero findings.

## Summary by CodeRabbit

* **New Features**
  * Reasoning effort settings are now normalized across supported providers and formats.
  * Native reasoning data can be preserved and replayed when compatible with the target format.
  * Reasoning provenance is tracked to support safer cross-format translation.

* **Bug Fixes**
  * Prevented signed or redacted reasoning from being incorrectly converted, exposed, or replayed without valid provenance.
  * Unsupported or invalid reasoning-effort values are safely omitted with appropriate diagnostics.
  * Improved filtering to prevent reasoning data from leaking between formats.


harshitwandhare:
I have an open PR against the same leak, #792, so I would rather flag the overlap than leave two PRs sitting on the same file.

Measured today, same input through `AnthropicMessages -> OpenAiResponses` (`thinking: {"type":"enabled","budget_tokens":10000}`), three heads:

| head | emitted `reasoning` | diagnostics | under `LossyConversionPolicy::Reject` |
| --- | --- | --- | --- |
| `main` `bfcd023c` | `{"type":"enabled","budget_tokens":10000}` | `[]` | Ok |
| this PR `4af42218` | absent | `[]` | Ok |
| #792 `a367c6da` | absent | 1 x `lossy_conversion` | Err |

`raw_by_format` is the more general fix and it does remove the leak. At `4af42218` the body comes back

```json
{"model":"claude-opus-5","input":"hi","max_output_tokens":32000}
```

with no `reasoning` key and no `budget_tokens`, because the Anthropic decoder's raw object is keyed to its own format and `raw_for(WireFormat::OpenAiResponses)` returns `None`. #792 gets there by allowlisting `effort`, `summary` and `generate_summary` in the Responses encoder, which is narrower, and I am happy for it to give way to this.

What your branch does not yet do is report the loss. I applied only #792's three tests on top of `4af42218` and ran them: the leak assertions pass, and the two diagnostics assertions fail, `diagnostics.len()` 0 against 1 and `result.is_err()` false. So a caller who has explicitly asked to be told about information loss still gets a request that quietly lost `budget_tokens`.

Two ways to close that, and I do not mind which:

1. Take the two tests and the `push_lossy` call into this PR. I can post them here or open a PR against your branch.
2. Land this, and I cut #792 down to the diagnostic alone on top of it.

I do not think both should land as they stand. Reproduction for either: check out `4af42218`, append the three tests from #792's diff to `crates/switchyard-translation/tests/request_translation.rs`, and run `cargo test -p switchyard-translation --test request_translation`.

ajcasagrande:
closing for now, while product alignment is conferred

PR #801, opened 2026-09-20
fix(soak): accept nullable error fields in valid responses

enwaiax:
## Summary

Fix false failures in the native soak client when a valid payload contains `"error": null`.

Both buffered and SSE validation currently reject any top-level `error` field. A nullable error field is not an error. Reject only non-null values, while preserving rejection of explicit SSE `event: error` and JSON `type: error`.

Only `crates/switchyard-soak/src/client.rs` changes. No server, routing, workload, token-limit, retry, or error-rate-policy changes.

## Regression evidence

- Before the fix, an HTTP 200 Responses body with `{"output":[],"status":"completed","error":null}` fails through the real `send_request` path with `invalid_response`.
- A stream with `error:null` also falsely fails with `stream_error`.
- The focused tests fail before the corresponding fixes and pass afterward. Negative controls still reject non-null errors and explicit SSE error events/types.
- Real-provider A/B, using the same pinned server image, byte-identical routing configuration, one-minute workload, concurrency 1, and output-token budget 512:
  - Baseline: 52 requests, 44 successful, 8 `invalid_response` failures; exit 1.
  - Fixed: 47 requests, 47 successful, no failures; exit 0.
  - Both runs completed and cleaned up. Only the soak binary changed. These are separate runs, not continuous-soak evidence.

## Verification

- `cargo fmt --all --check`
- `cargo test -p switchyard-soak --locked`: 14 unit and 4 integration tests passed.
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`: 853 passed, 1 ignored (requires `PREFILL_ROUTER_HANDOFF_DIR`).
- `cargo build --release -p switchyard-soak --locked`
- Independent bounded review passed; the reviewer separately ran all 5 client unit tests, formatting, and whitespace checks.

The tested source tree matches this publication commit exactly. Local checks are not a claim that all feature-specific or Python/package CI gates ran.

## Unchanged scope

`response.incomplete` remains an accepted SSE transport terminal. Buffered validation still checks minimal shape rather than complete, nonempty answers. This fix does not establish response completeness, routing escalation coverage, or long-duration stability. Earlier failed runs remain failed evidence.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response validation to accept missing or null error fields.
  * Correctly identifies actual errors in buffered responses and streaming payloads.
  * Added support for handling named and typed stream error events.
  * Recognizes the `response.incomplete` event as a valid terminal response event.


enwaiax:
Fresh integration verification at published product head `9009f2e3f2a03398203e58021c47109078d2f2c4` is complete.

Built both native server and soak from that exact commit, then ran a five-minute real-provider Stage-configured smoke through Docker Compose: concurrency 1, short-interactive, output-token budget 512, retries disabled.

- **260/260 successful requests**, zero failures: Chat 88, Messages 88, Responses 84.
- Native process exit 0 and orchestration PASS; original failed runs retained.
- Health, metrics, and process checks: 60 each, zero failures. Four invalid-request canaries passed.
- RSS remained 14 MiB; no restart. Prometheus range export and Grafana datasource query passed.
- Source identity, native config hash, actual container arguments, retained log hashes, credential isolation, and teardown verified; all four containers stopped.

This complements the same-condition baseline/fixed one-minute A/B in the description. It is not a response-completeness claim: existing acceptance of incomplete terminals and buffered minimal-shape validation are unchanged. It does not prove Stage escalation or long-duration stability.

All currently reported GitHub checks pass at this PR head. No merge was performed.

PR #803, opened 2026-09-20
fix(llm-client): store canonical history under the conversation id

### LINKED ISSUE #802: [bug] Responses conversation continuation drops materialized cross-format history
## Symptom

A Responses client continuing a cross-format Chat conversation by `conversation` loses all prior turns. Continuing by `previous_response_id` keeps them.

## Reproduction

Pinned at `bfcd023c` (current main, 2026-09-20), passthrough route from OpenAI Responses to an OpenAI Chat backend (`format = "openai_chat"`) with a deterministic capture upstream. Seed then recall:

```json
{"model":"switchyard/chat","input":"SEED","conversation":"conv_1"}
{"model":"switchyard/chat","input":"RECALL","conversation":"conv_1"}
```

The recall reaches the Chat backend with only `RECALL`. Control: same route, same seed, replace `conversation` with the returned `previous_response_id`. The backend then receives the seed user turn, seed assistant answer, and recall.

Result: history lost 5/5 with `conversation`, preserved 5/5 with `previous_response_id`.

## Expected vs. actual

- **Expected:** both continuation mechanisms carry the same materialized history. PR #781 promises canonical history for cross-format stored Responses continuations.
- **Actual:** the conversation follow-up forwards only the new turn. The server returns HTTP 200 with a plausible but context-free answer.

## Root cause

`crates/libsy-llm-client/src/run.rs` at this commit. `remember_canonical_response` (line 839) calls:

```rust
.remember(response_id, None, model, Some(history))   // line 863
```

The conversation id is hardcoded to `None`. `remember_response` (line 828) passes the real `conversation`, and native Responses recording passes both ids (line 718). The later conversation lookup misses, so routing silently continues without history.

## Environment

- Switchyard version (or commit SHA): `bfcd023cb6791efd8ac59d32afb5095c74fef1ff`
- Python version: N/A
- OS / arch: macOS arm64
- Install path: source build
- Inbound format: Responses
- Backend: OpenAI Chat Completions (local deterministic capture)

## Additional context

- Gap in #781 (merged 2026-09-18), which materializes cross-format state for `previous_response_id` only. PR #721 documents `conversation` as an equivalent supported continuation.
- Suggested fix: pass the request conversation id into `remember_canonical_response` and store the materialized history under both the response id and the conversation id, with buffered and streamed cross-format conversation tests.

cc @sabhatinas

Atharva-Kanherkar:
## What

`remember_canonical_response` recorded cross-format materialized history under the response id only, hardcoding `None` for the conversation id. This PR passes the request conversation id through on both the buffered and streamed paths, and stores materialized conversation records with latest-wins semantics so multi-turn chains keep the full history. Three regression tests cover the behavior.

## Why

A Responses client continuing a cross-format conversation (Chat or Anthropic backend) by `conversation` silently lost all prior turns. The backend received only the new turn and answered without context while the server returned HTTP 200. The equivalent `previous_response_id` continuation works, so the two supported selectors disagreed. Fixes https://github.com/NVIDIA-NeMo/Switchyard/issues/802.

## Notes for reviewers

- Start at `crates/libsy-llm-client/src/run.rs`. `remember_canonical_response` gains a `conversation` parameter; both call sites in `remember_state_owner` pass `conversation.as_deref()`.
- `StateOwners::remember` now overwrites the conversation entry when the history is materialized, and keeps the existing first-writer-wins behavior for provider-owned state.
- Tests: `materialized_conversation_state_keeps_latest_history` pins the refresh semantics. `responses_conversation_continuation_materializes_for_anthropic` runs three turns by conversation id through a mock upstream and asserts the second and third upstream requests carry the full chain. `streamed_cross_format_conversation_state_is_recorded_after_completion` covers the stream completion path.
- Checked with rustc 1.96.1: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (855 passed, 0 failed). The three new tests fail with the fix reverted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conversation continuity across Responses, Chat, and Anthropic interactions.
  - Buffered and streamed responses now consistently retain and refresh canonical conversation history.
  - Conversation identifiers are preserved across aggregate and streamed responses.
  - Existing provider-managed Responses state remains preserved when no canonical history is available.
- **Tests**
  - Added coverage for conversation continuations, streamed history, and state preservation across backends.


crates/libsy-llm-client/src/run.rs
```diff
@@ -841,6 +852,7 @@ impl ClientRouter {
         response: &AggLlmResponse,
         model: &ModelId,
         store: bool,
+        conversation: Option<&str>,
```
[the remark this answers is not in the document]
Atharva-Kanherkar: Thanks, but I left this as is on purpose. This early return predates the PR. The test `cross_format_response_with_store_false_is_not_retained` pins that a canonical response with `store: false` is not retained, and `store: false` means the client asked for no persistence. Recording materialized history under the conversation id in that case would retain a turn the client opted out of, unlike the provider-owned path where the entry is only a routing pin (history is `None`). If maintainers want `store: false` to skip only the response id here too, that is a separate semantic decision worth a quick confirmation, and I can add it as a follow-up with a test.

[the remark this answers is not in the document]
Atharva-Kanherkar: Correction to my earlier reply: store:false disables response storage, not conversation history. The existing no-retention test has no conversation, so it does not establish the behavior I claimed. The conversation-state documentation distinguishes these: https://platform.openai.com/docs/guides/conversation-state#data-retention-for-model-responses . Fixed in d620444e with buffered and streamed regression coverage, including absent response IDs. Conversation tool outputs now retain their type, and chained histories reuse parent segments. All 855 workspace tests pass (1 ignored), with formatting and workspace Clippy clean.

afourniernv:
This bug seems pretty consequential. 

The bug happens when:

- A client calls /v1/responses.
- It supplies a non-null conversation ID.
- The selected target uses openai_chat or anthropic_messages.
- The next turn reuses that conversation ID and sends only the new input.

I repro'd on my end 

@elyasmnvidian @grahamking

afourniernv:
One more bit of context after testing this: I think this is a narrow bug rather than something affecting most Responses traffic.

The SDK does not send `conversation` automatically. An application has to create a Conversation and explicitly pass its ID. Simpler multi-turn flows generally use `previous_response_id` or resend history themselves. OpenAI documents all three approaches here: https://developers.openai.com/api/docs/guides/conversation-state

So this is probably an edge case in terms of usage. It is still a supported API path, and within that path the failure is deterministic: every cross-format follow-up loses the earlier conversation.

crates/libsy-llm-client/src/run.rs
```diff
@@ -1490,6 +1538,136 @@ mod tests {
         Ok(())
     }
 
+    // Every prior turn must reach the backend on each conversation continuation.
+    #[tokio::test]
+    async fn responses_conversation_continuation_materializes_for_anthropic() -> Result<()> {
```
afourniernv: Could we drop this test? The conversation cases in `responses_stored_tool_continuation_materializes_for_anthropic` already run three turns against Anthropic and verify the earlier tool call, result, and answer are replayed. Adding the seed-message assertion there should cover this without a second 125-line fixture.

crates/libsy-llm-client/src/run.rs
```diff
@@ -1567,6 +1745,104 @@ mod tests {
         Ok(())
     }
 
+    // Conversation history is recorded only once the stream completes.
+    #[tokio::test]
+    async fn streamed_cross_format_conversation_state_is_recorded_after_completion() -> Result<()> {
```
afourniernv: Could we fold this into `streamed_cross_format_state_is_recorded_only_after_completion` above? If that seed also carries a conversation ID, we can check both the existing response-ID lookup and the conversation lookup after the stream completes without keeping a second copy of the stream setup.

crates/libsy-llm-client/src/run.rs
Atharva-Kanherkar: Done in 671ffe9e. Dropped it and added the seed assertion to the tool continuation test.

crates/libsy-llm-client/src/run.rs
Atharva-Kanherkar: Done in 671ffe9e. Folded it in. The seed now carries a conversation id, and the test checks both lookups after completion, including `store: false` and a missing response id.

Atharva-Kanherkar:
@afourniernv thanks for reproducing it. Agreed, it is narrow since `conversation` is opt in, but deterministic on that path. 
I fixed what you said about tests, and made the commit. It is narrow, but it is a bug!

afourniernv:
Thanks for cleaning up the tests. I repro’d the fix at `671ffe9` and it works.

One thing I think we need to sort out before merging: `ClientRouter` is shared by callers on a route, but the history here is keyed only by the `conversation` ID. I changed the seed and follow-up to use different `session_id` values with the same conversation ID, and the second request still picked up the first session’s history.

Maybe that is intended, but I do not see it called out anywhere. Can you think through the boundary here and propose a fix? I do not think we can merge until this is resolved.

Atharva-Kanherkar:
> One thing I think we need to sort out before merging: `ClientRouter` is shared by callers on a route, but the history here is keyed only by the `conversation` ID. I changed the seed and follow-up to use different `session_id` values with the same conversation ID, and the second request still picked up the first session’s history.

I actually dont understand this right now, I'll have to look at the code once. Ill get back to you.

Atharva-Kanherkar:
@afourniernv  looked at it properly now. You're righ, the state map is one per `ClientRouter` and keyed only by the ID, so anyone on the route who sends the same conversation gets that history. Worth noting that `previous_response_id` already works this way on main. **The difference is that response IDs are generated upstream and hard to guess**. 
Conversation IDs on Chat/Anthropic targets are just whatever string the client picks, so collisions are a lot more likely. 
Also, `session_id `is a client-set header, not auth, so it can keep cooperating clients apart but it isn't a security boundary. 

I feel when `session_id` is present, key conversation history by (`session_id`, `conversation_id`), and fall back to the global key when it isn't. I'd add your case as a test and a short doc note on the boundary. Does that work for you, or would you rather scope it some other way? I remember you not liking large tests earlier.

PR #806, opened 2026-09-21
feat(server): forward x-litellm-response-cost upstream header

### LINKED ISSUE #805: [feature] Support forwarding x-litellm-response-cost for consistent cost tracking
## Problem

We use LiteLLM's `x-litellm-response-cost` response header to track per-request inference costs. The header is present when requesting `openai/openai/gpt-5.6-luna`, but missing when requesting `switchyard/openai/gpt-5.6-luna`.

## Proposed solution

#480 and #571 added upstream response-header forwarding, but the allowlist documented in #571 does not include `x-litellm-response-cost`.

## Scope notes

Expose `x-litellm-response-cost` consistently across models, including Switchyard routes, whenever cost information is available.

## Additional context

nothing

deepujain:
## What

Adds `x-litellm-response-cost` to the forwarded upstream response headers, so LiteLLM's per-request cost header reaches Switchyard clients the same way it does on direct LiteLLM routes. One regression test covers the allowlist.

## Why

Fixes https://github.com/NVIDIA-NeMo/Switchyard/issues/805. The header was dropped by the allowlist gate, so cost tracking broke on `switchyard/...` routes while direct `openai/openai/...` routes carried it.

## Notes for reviewers

Start at `FORWARDED_UPSTREAM_HEADERS` in `crates/switchyard-server/src/lib.rs`. The header is a plain numeric cost, not a credential, so it is safe to expose downstream. Validated with rustc 1.96.1: `cargo fmt --all -- --check`, `cargo clippy -p switchyard-server --all-targets` (plus `--features prefill-router`), `cargo test -p switchyard-server` (56 passed), `cargo clippy --workspace --all-targets`, `cargo test --workspace` (all suites green), and `cargo test -p switchyard-runner --features prefill-router` (green). The new test fails with the allowlist entry removed.

## Summary by CodeRabbit

* **New Features**
  * Clients can now receive upstream response cost information through the `x-litellm-response-cost` header.

* **Bug Fixes**
  * Confirmed that approved response headers are forwarded correctly while sensitive authorization headers remain blocked.


afourniernv:
I checked this against LiteLLM’s docs. `x-litellm-response-cost` is their documented response contract, so exact allowlisting here looks like the right boundary and is safer than forwarding every `x-litellm-*` header. One caveat: LiteLLM says the value can be unreliable for streaming, so this forwards the header when present rather than guaranteeing streaming cost accuracy.

[Cost tracking docs](https://github.com/BerriAI/litellm-docs/blob/main/docs/proxy/cost_tracking.md) · [Architecture](https://github.com/BerriAI/litellm/blob/main/ARCHITECTURE.md) · [Streaming limitation](https://github.com/BerriAI/litellm/issues/30816)

deepujain:
Agreed, thanks for checking against their docs. The PR forwards the header verbatim only when LiteLLM sends it and makes no claim about streaming cost accuracy; no behavior change is needed on that front. I can add a note to the PR description capturing the streaming caveat if that would be useful.

afourniernv (approved):

afourniernv:
LGTM. Approved. We’ll merge this once the 0.3 release code freeze ends. Thank you for the contribution!

